PR #6133, opened 2014-10-06
WIP: Multimedia improvments

LaurentGoderre:
/cc @crichard


LaurentGoderre:
@crichard I think this is ready!!! :D. I will merge right after v4.0.11 comes out!

PR #6410, opened 2015-01-04
Twitter Plugin: Added test suite (Fixes part of #6070)

melodyma:
Added test suite to Twitter plugin that tests initialization of plugin.
This PR is related to issue #6070.

### DIFF
--- a/src/plugins/twitter/test.js
+++ b/src/plugins/twitter/test.js
@@ -1,0 +1,52 @@
+/*
+ * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)
+ * @title Twitter Plugin Unit Tests
+ * @overview Test the favicon plugin behaviour
+ * @license wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
+ * @author @patheard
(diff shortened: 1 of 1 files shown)


src/plugins/twitter/test.js
```diff
@@ -0,0 +1,51 @@
...
+    $body = $document.find( "body" );
+
+  before(function( done ) {
+    // Trigger plugin init
+    $elm = $( "<div class='wb-twitter'><a class='twitter-timeline' href='https://twitter.com/search?q=%23WxT' data-widget-id='329066756620566528'>Tweets about '#WxT'</a></div>" )
+      .appendTo( $body )
+      .trigger( "wb-init.wb-twitter");
+
+    // Give the content time to load
+    setTimeout( done, 100 );
```
LaurentGoderre: Instead of using a timeout, you should listen for the ready event to make sure you are not a) timing out intermittently and b) not delaying the testing

Add the following code instead

```javascript
$document.on( "wb-twitter-ready.wb-twitter", ".wb-twitter", function() {
	done();
});
```

LaurentGoderre:
BTW, Thanks so much for this!

PR #6414, opened 2015-01-05
Theming: Added the ability for themes to customize WET variables

LaurentGoderre:


LaurentGoderre:
When this gets merged, all themes will need to be modified accordingly

src/_variables.scss
```diff
@@ -14,115 +10,96 @@ $clrMedium: #ccc;
...
+// Calendar
+$cal-text-head: #fff !default;
+$cal-text-body: #000 !default;
+$cal-dark: #333 !default;
+$cal-medium: #555 !default;
+$cal-active: #ccc !default;
+$cal-days-link: #111 !default;
+
+//Multimedia Player
+$mm-progress-fg-color: #176ca7 !default;
```
LaurentGoderre: @melodyma notice that we need to add `!default` for each variable to allow the variable to be customizable by themes

nschonni: Not :100: sure that libsass properly supports `!default` right now, but I haven't been following the 3.x dev that close

src/base/_bootstrap.scss
```diff
@@ -0,0 +1,13 @@
+// Hack for Bootstrap 3.1.1, since it doesn't convert the "fadein" LESS method.
+// Pending fix release in https://github.com/twbs/bootstrap-sass/commit/df7be4f0412497e22f1d271d94d6e9835f61113e
+@function fadein($colour, $pct) {
```
nschonni: This might not be needed anymore since the Bootstap version has been bumped

src/_variables.scss
LaurentGoderre: It does

src/base/_bootstrap.scss
LaurentGoderre: What version is needed for it to work? Nonetheless, this is something that can be addressed by another PR

LaurentGoderre: It seems like it's still needed, `_variables.scss` still uses it.... Which is weird because it shouldn't

LaurentGoderre: Actually no, `fadein` is being converted to `fade_in` so it can go.

LaurentGoderre:
@melodyma you are free to start applying your changes now

PR #6434, opened 2015-01-09
Tabs: Fixed resizing issue that could result in an undefined variable error

pjackson28:


LaurentGoderre:
Is there any difference in the code other than whitespace?

pjackson28:
It is nested one level deeper to ensure that the variable is defined.

LaurentGoderre:
Ah ok!

PR #6444, opened 2015-01-14
Fixed the installation of grunt-contrib-imagemin

LaurentGoderre:
This issue is most common when working with themes


LaurentGoderre:
This should fix all the themes and the generator installs

/cc @shawnthompson

Issue #6436, opened 2015-01-12
Validating whether two <input> field values match?


josephbr:
I was looking through the "Form Validation" working example but I could not find any way to check whether two <input> field values match. For example, say I have an "E-mail" input box and then another for "Please re-type your e-maill address" and I want to ensure both contain the same content. How would I use WET Form Validation to accomplish this or is it not possible? I would rather avoid having to include separate validation functionality.

CalvinRodo:
Wet uses the [Jquery Validation plugin](http://jqueryvalidation.org/documentation/) to handle client side validation so you can read their documentation to see how to do this client side. 



I should add that I highly recommend you also do the validation on the server side as you can never be sure the user didn't just bypass your JavaScript validation and is posting whatever they want to the server. 

For email though I think you can probably get away from having to do the email matching as long as you send them a validation email to properly ensure the email address works.

josephbr:
I already plan to include a server side check. My primary concern is that I cannot achieve the same look/feel as the other validation types (i.e. as seen on http://wet-boew.github.io/wet-boew/demos/formvalid/formvalid-en.html where it highlights the field containing the error, labels it, and also displays the error at the top of the screen).

LaurentGoderre:
@josephbr I believe the plugin allows for custom validation rules that integrate with the other ones.

CalvinRodo:
@josephbr I highly recommend you read the documentation for the jquery validation plugin that i posted
you'd see that one of the standard rules that they provide is http://jqueryvalidation.org/equalTo-method
which according to the documentation is **Requires the element to be the same as another one** so you wouldn't even have to write any custom rules you can just use what is built in.

josephbr:
Validating the data should be no problem but the part I am unsure about is how to display the error so that it uses the same look/feel where the field is outlined and the error appears beside it and at the top of the page. What function call was invoked to display the error?

CalvinRodo:
I personally haven't looked too much into it I figured it would just work but I guess if there isn't a rule defined for that or messages defined for that rule it won't just work. 

I'd then look into [the documentation for the Validate() method](http://jqueryvalidation.org/validate), it seems you can pass back a list of messages that will appear if the error occurs 

I believe the plugin should add the correct markup to the error and the field and the validation summary should pick up and linkify anything that is marked up with the correct syntax. 

[...]

PR #6448, opened 2015-01-14
Form Validation: Add examples for two field needing to have the same value

LaurentGoderre:


LaurentGoderre:
Demos #6436

LaurentGoderre:
@pjackson28 any issue with this?

pjackson28:
Just take the colon off the confirm label so it is like the other labels.

PR #6456, opened 2015-01-19
SCSS Variables: Updated src variables file and added customizable variables for Tabs plugin

melodyma:
- Updated formatting of src variables file for easier navigation and customization
  - Ordering of overridden Bootstrap variables is the same as in Bootstrap's variable file
  - WET plugin variables are ordered alphabetically by plugin name
  - Added instructions for overriding Bootstrap variables not already in src variables
- Added customizable variables for Tabs plugin (i.e. colours, border radiuses)
  - Reduced number of greys for tab border

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -3,22 +3,64 @@
  * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
  */
 
-// Colours
+/*
+ * HOW TO USE THIS FILE
+ * Use this file to override Bootstrap variables and WET custom variables.
+ * If there is a Bootstrap variable not shown here that you want to override, go to "../lib/bootstrap-sass-official/assets/stylesheets/bootstrap/variables" to view the variables that you can override. Simply copy and paste the variable and its applicable section (if applicable) from the Bootstrap file into this override file and override the variables as applicable.
+ */
+
+
+//
+// BOOTSTRAP VARIABLES
+//
+
+//== Colours
+//
+//## Gray and brand colors for use across the application.
+
+// Bootstrap Overrides
+$gray-light: #767676;
+
+$brand-primary: #2572b4 !default;
+$brand-success: #1b6c1c !default;
+$brand-info: #4d4d4d !default;
+$brand-warning: #f2d40d !default;
+$brand-danger: #bc3331 !default;
+
+// WET Custom
 $clrWhite: #fff;
 $clrLight: #eee;
 $clrMedium: #ccc;
 $clrDark: #555;
-$gray-light: #767676;
+$clrDarkBlue: #243850;
 
-// Font scales
+
+//== Scaffolding
+//
+//## Settings for some of the most global styles.
+
+//** Global textual link color.
+// Boostrap Overrides
+$link-color: #295376 !default;
+$link-hover-color: #0535d2 !default;
+
+// WET Custom
+$link-visited-color: #7834bc !default;
+
+
+//== Typography
+//
+//## Font, line-height, and color for body text, headings, and more.
+
+// WET Custom font scales
 $font-scale-xxxl: 2.125 !default;
 $font-scale-xxl: 1.625 !default;
 $font-scale-xl: 1.375 !default;
 $font-scale-lg: 1.125 !default;
 $font-scale-sm: 0.875 !default;
 
-// Font sizes
-$font-size-base: 16px !default;
+// Bootstrap Overrides
+$font-size-base: 16px;
 $font-size-large: ceil($font-size-base * $font-scale-lg) !default;
 $font-size-small: ceil($font-size-base * $font-scale-sm) !default;
 
@@ -28,23 +70,27 @@ $font-size-h3: ceil($font-size-base * $font-scale-xl) !default;
 $font-size-h4: ceil($font-size-base * $font-scale-lg) !default;
 $font-size-h6: ceil($font-size-base * $font-scale-sm) !default;
 
+//** Unit-less `line-height` for use in components like buttons.
 // 20/14 -> 23/16 on the basis of ceil((20/14)*16) = 23
 $line-height-base: 1.4375 !default;
 
-//Bootstrap colors
-$brand-primary: #2572b4 !default;
-$brand-success: #1b6c1c !default;
-$brand-info: #4d4d4d !default;
-$brand-warning: #f2d40d !default;
-$brand-danger: #bc3331 !default;
 
-$link-color: #295376 !default;
-$link-hover-color: #0535d2 !default;
-$link-visited-color: #7834bc !default;
+//== Tables
+//
+//## Customizes the `.table` component with basic values, each used across all table variations.
 
+// Bootstrap Overrides
+//** Background color used for `.table-striped`.
 $table-bg-accent: #f5f5f5 !default;
+//** Background color used for `.table-hover`.
 $table-bg-hover: #f0f0f0 !default;
 
+
+//== Buttons
+//
+//## For each of buttons, define text, background and border color.
+
+// Bootstrap Overrides
 $btn-default-color: #335075 !default;
 $btn-default-bg: #eaebed !default;
 $btn-default-border: darken($btn-default-bg, 5%) !default;
@@ -65,6 +111,25 @@ $btn-warning-border: darken($btn-warning-bg, 20%) !default;
 $btn-danger-bg: $brand-danger !default;
 $btn-danger-border: darken($btn-danger-bg, 20%) !default;
 
+
+//== Media queries breakpoints
+//
+//## These additional are additional breakpoints outside of and in addition to Bootstrap's definitions at which your layout will change, adapting to different screen sizes.
+
+// WET Custom
+// Extra extra small screen
+$screen-xxs-max: 479px !default;
+// Large screen
+$screen-lg-max: 1599px !default;
+// Extra large screen
+$screen-xl-min: 1600px !default;
+
+
+//== Pagination
+//
+//##
+
+// Bootstrap Overrides
 $pagination-color: $btn-default-color !default;
 $pagination-bg: $btn-default-bg !default;
 $pagination-border: $btn-default-border !default;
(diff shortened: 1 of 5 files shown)


src/_variables.scss
```diff
@@ -3,103 +3,222 @@
...
-$mm-progress-bg-color: #fff !default;
-$mm-cc-fg-color: #fff !default;
-$mm-cc-bg-color: #000 !default;
-$mm-ctrl-fg-color: #fff !default;
-$mm-ctrl-bg-color: #3e3e3e !default;
-$mm-ctrl-outline-color: #4aafff !default;
-$mm-ctrl-cc-color: #4aafff !default;
-$mm-volume-bg-color: #aaa !default;
-$mm-volume-thumb-color: #fff !default;
+//== Media queries breakpoints
```
LaurentGoderre: These are custom breakpoints that are not defined in BootStrap , I think the text should reflect that

src/_variables.scss
```diff
@@ -3,103 +3,222 @@
...
+//
+//##
+
+$spacing:                 .375em !default;
+$ddRtnWidth:              3.375em !default;
+
+
+//== Multimedia player
+//
+//##
```
LaurentGoderre: What does this mean?

melodyma: The `//##`? Those are just placeholders for comments for formatting consistency similar to the Bootstrap variables file. We can take them out if you'd like.

src/_variables.scss
melodyma: Yes, I can make this change. Is this the only change outstanding? If so, I can do a fix and commit. Otherwise, I'll wait for your comments to come if there are any more. Thanks.

src/_variables.scss
```diff
@@ -3,103 +3,222 @@
...
+$carousel-tabpanel-figcaption-color:         #fff !default;
+$carousel-tabpanel-figcaption-link-color:    $carousel-tabpanel-figcaption-color !default;
+$carousel-tabpanel-figcaption-bg:            $clrDrabBlue !default;
+
+// Carousel Style 2 with Thumbnails
+$carousel-s2-tablist-controls-bg:            $clrLight !default;
+
+$carousel-s2-tablist-controls-color:         $clrDrabBlue !default;
+$carousel-s2-tablist-controls-button-bg:     #fff !default;
+$carousel-s2-tablist-controls-button-shadow: $clrDrabBlue !default;
```
LaurentGoderre: I am wondering if this should be changed to the whole shadow so that you can turn it off completely if you want

```
$carousel-s2-tablist-controls-button-shadow: 0 0 4px $clrDrabBlue !default;
```

src/_variables.scss
```diff
@@ -3,103 +3,222 @@
...
+$carousel-s2-tablist-controls-button-radius: 999px !default;
+
+$carousel-s2-tablist-thumbnail-active-border: #666 !default;
```
LaurentGoderre: I am wondering if this can also be the full border value to add more control

src/_variables.scss
```diff
@@ -3,103 +3,222 @@
...
+$mm-volume-thumb-color:   #fff !default;
+
+
+//== Tabs
+//
+//##
+
+// Tabbed Interface
+$tablist-color:                  #000 !default;
+$tablist-border:                 $clrMedium !default;
```
LaurentGoderre: I am wondering if this can also be the full border value to add more control

src/_variables.scss
```diff
@@ -3,103 +3,222 @@
...
+$tabpanel-border:                $tablist-border !default;
```
LaurentGoderre: I am wondering if this can also be the full border value to add more control

src/_variables.scss
```diff
@@ -3,103 +3,222 @@
...
+$tablist-bg:                     #ebf2fc !default;
+$tablist-hover-bg:               $clrMedium !default;
+$tablist-active-bg:              #fff !default;
+$tablist-active-link-border-top: #666 !default;
+// Tablist at sm max && xxs max
+$tablist-screen-sm-details-border:           $tablist-border !default;
```
LaurentGoderre: I am wondering if this can also be the full border value to add more control

src/_variables.scss
```diff
@@ -3,103 +3,222 @@
...
+//##
+
+// Tabbed Interface
+$tablist-color:                  #000 !default;
+$tablist-border:                 $clrMedium !default;
+$tabpanel-border:                $tablist-border !default;
+$tablist-bg:                     #ebf2fc !default;
+$tablist-hover-bg:               $clrMedium !default;
+$tablist-active-bg:              #fff !default;
+$tablist-active-link-border-top: #666 !default;
```
LaurentGoderre: This seems very specific....is there a way to make this more generic and controllable?

LaurentGoderre: This article mentions that by splitting it up under 3 different properties you can specify different values for each edge

border {
    width: 4px 0 0 0;
    style: solid;
    color: #666;
}

melodyma:
@LaurentGoderre - The Travis CI build failed because of a bundle install issue for rake. It doesn't seem like it was anything to do with my commit. I did a `grunt lint` on my machine and everything passed. How can I run the Travis build again to make this go through? Thanks.

LaurentGoderre:
I can restart builds when you need it but it seems the build for this PR did pass...

melodyma:
Perfect. I've made my changes based on your comments and it's ready for review. Let me know if there are any other issues.

melodyma:
@LaurentGoderre - When do you think you'd be able to review and merge this? I'd like to start working on doing the same for the other plugins. Thanks!

LaurentGoderre:
Can you squash the last two commits?

LaurentGoderre:
On thing I noticed, and it's not your code, is that there is no consistency in the way the colors are handled. We should pick one and stick with it. 

What would you propose? Here are some possible options

* `$purpose-fg-color, $purpose-bg-color`
* `$purpose-fg, $purpose-bg`
* `$purpose-color, $purpose-background`

src/_variables.scss
```diff
@@ -3,103 +3,220 @@
...
+
+//== Calendar
+//
+//##
+
+$cal-text-head:           #fff !default;
+$cal-text-body:           #000 !default;
+
+$cal-dark:                #333 !default;
+$cal-medium:              #555 !default;
```
LaurentGoderre: I know these are not from your but they should be changed to reflected their purpose, not their styles

nschonni:
Awesome stuff :heart: :heart: :heart: , but I don't think the variable values should be aligned. This will cause noise (and muddy git-blame) every time a longer variable name is added the file.

nschonni:
@LaurentGoderre SCSSLint will have some stuff to help in this exercise in their next release https://github.com/causes/scss-lint/issues/337

LaurentGoderre:
@nschonni do you mean aligned in characters? Btw @melodyma where did you get this convention

nschonni:
@LaurentGoderre yeah, I was talking about the alignment https://github.com/causes/scss-lint/blob/master/lib/scss_lint/linter/README.md#spaceafterpropertycolon
It is a fairly common format (https://github.com/twbs/bootstrap/blob/master/less/variables.less) that some IDEs apply by default

melodyma:
@LaurentGoderre - I put it in because it's a file that's meant for external consumption, so it's easier to read and navigate for other users when they are using the file. Bootstrap also uses this convention for their variables file. However, @nschonni makes a good point. I'm happy to change it since it's just a personal preference.

nschonni:
@melodyma maybe that style could be applied once all the "variablization" of the plug-ins is finished.

melodyma:
@LaurentGoderre - For colours, it would be most consistent if all the variables were `$*-color`. Bootstrap eliminates it for `bg`, but because we have `fg`, `cc` and everything in between and beyond, it may be wise to override the Bootstrap defaults to something that is more semantically maintainable for us. However, the downside to this approach is that we'll have to add more code to override the defaults and it may not be obvious to others what's happening right away. Perhaps, we can also meet somewhere in the middle and apply `$*-color` only if it's not a `$*-bg` or `$*-fg`. Thoughts?

@nschonni - I concur with your approach!

LaurentGoderre:
I want what is the clearest and it should maybe be clear for people that are not familiar to CSS as well so explicitly specifying which type of color might be desirable. What do you think?

LaurentGoderre:
We could also use `text-color` `border-color` and `bg-color` to be more explicit

melodyma:
Should all the Bootstrap overrides also be appended with `-color` then?

LaurentGoderre:
It depends what we decide on...

melodyma:
My take is that adding `-color` will be more usable.

LaurentGoderre:
And how do you difference the different color type?

melodyma:
Sorry, could you give an example of different colour types?

LaurentGoderre:
text, border and background

melodyma:
I think if we're going to add `-color`, we might as well do it for all colour-associated variables. The borders one will have to have the color attribute split out from the rest of the attributes.

LaurentGoderre:
Can you apply this convention to all our plugin variables (not the Bootstrap ones)

melodyma:
@LaurentGoderre - Cleaned up this PR as per your comments. Some notes:

* I couldn't get more specific with the calendar colours based on how the variables were extended in the plugin. May be worthwhile to look at how the calendar plugin variables/mixins/css were organized in a separate dedicated PR.
* The border attributes had to be separated into three separate attributes each (width, style, color). The snippet you provided earlier doesn't compile properly and I had to explicitly spell each attribute out as `border-[attribute]`
* Not sure why the Travis build failed again. Seem like an npm package error. I ran a grunt dist locally that showed that my PR is clean.

That's all!

LaurentGoderre:
The build is failing because of the sass linting (which is not available in grunt/node.js). For our project we use tabs instead of space so you need to configure your editor to use tabs for indenting

src/_variables.scss
```diff
@@ -73,33 +138,100 @@ $pagination-hover-color: $btn-default-color !default;
...
+
+$tablist-active-link-border-color: #666 !default;
+$tablist-active-link-border-style: solid !default;
+$tablist-active-link-border-width: 4px 0 0 0 !default;
+
+// Tablist at sm max && xxs max
+$tablist-screen-sm-details-border-color: $tablist-border-color !default;
+$tablist-screen-sm-details-border-style: solid !default;
+$tablist-screen-sm-details-border-width: 0 1px 1px 1px !default;
+$tablist-screen-sm-details-top-border-width: 1px !default;
```
LaurentGoderre: This probably need to be expanded to allow different radius on all edges

src/plugins/tabs/_screen-sm-max.scss
```diff
@@ -9,18 +9,29 @@
...
+				}
 			}
 
 			&:last-of-type {
-				border-bottom-left-radius: 4px;
-				border-bottom-right-radius: 4px;
+				border: {
+					bottom: {
+						left-radius: $tablist-screen-sm-details-border-radius;
+						right-radius: $tablist-screen-sm-details-border-radius;
```
LaurentGoderre: We should investigate if we can remove the bottom so that this can be applied to all edges

src/_variables.scss
melodyma: @LaurentGoderre - Realistically, the top left and right radii will be different from the bottom left and right radii, but left and right radii would never be different from each other. Does it make sense to split them up to 4 separate radii or two pairs? Trying to balance the amount of customization.

src/plugins/tabs/_screen-sm-max.scss
melodyma: I think this ties into my comment above.

src/_variables.scss
LaurentGoderre: Well in the example above `$tablist-screen-sm-details-top-border-width` is not needed since `$tablist-screen-sm-details-border-width` can specify the width on each edge. Can't the same be done for the radius?

```
$tablist-screen-sm-details-border-radius: 5px 0 5px 0;
```

melodyma:
@LaurentGoderre - OK. I think this is good now. I had to refactor the `tabs_screen-sm-max.scss` file to get the borders to show in the right places so that you can control the borders as variables. Let me know if there's anything else!

LaurentGoderre:
@melodyma sounds good. We should clean it up ASAP before people actually start using those variables ;)

src/plugins/calendar/_base.scss
```diff
@@ -132,7 +132,6 @@
 .cal-days {
 	a {
 		@extend %cal-days-a-div;
-		color: $cal-days-link;
```
LaurentGoderre: Why was this removed?

melodyma: It was declared, but wasn't actually used. The links were overrode by the styles from `.cal-wd`. I think the entire calendar plugin needs to be reviewed independently in another PR like the tabs plugin. For this PR, I was thinking of just keeping the variables that were there already that were actively used by the plugin.

screen shot 2015-02-13 at 5 55 47 am

LaurentGoderre: Ok, thanks for the explanation!

PR #6458, opened 2015-01-19
Carousel: Modify working example to show whole panels as clickable

thomasgohard:
cc/@masterbee

### DIFF
--- a/src/plugins/tabs/tabs-carousel-en.hbs
+++ b/src/plugins/tabs/tabs-carousel-en.hbs
@@ -30,10 +30,7 @@
 				<figure>
 					<img src="img/protect-environment.jpg" alt="Panel 1" />
 					<figcaption>
-						<p>
-							Take Note: <a href="http://www.tc.gc.ca/eng/civilaviation/opssvs/general-personnel-changes-1814.htm">Renewal of the Aviation Document Booklet</a>&#160;<br />
-							Learn more about&#160;<a href="http://www.tc.gc.ca/eng/air-menu.htm">air transportation</a> in Canada.
-						</p>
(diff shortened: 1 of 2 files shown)


LaurentGoderre:
This involves a markup change doesn't it? It might be too late for this version but we can merge this immediately after v4.0.11 goes out

shawnthompson:
@thomasgohard From a usability side, does this mean all carousels will be fully clickable no matter what? Do we have the option of only making the link text clickable?

cc/ @pcwsmith @DK-TBS @lucas-hay

LaurentGoderre:
The markup change may be an issue as well...it should work both ways

LaurentGoderre:
We can't merge as is because it doesn't respect the proper use of variables etc. I will open a seperate PR to address the CSS changes

LaurentGoderre:
@thomasgohard please rebase this PR to exclude the second commit

thomasgohard:
@shawnthompson @pcwsmith @DK-TBS @lucas-hay @LaurentGoderre both options can be used: full panel or caption-only. The only thing I changed is that the default example shows the full panel being clickable at @masterbee's request.

thomasgohard:
@LaurentGoderre removed the CSS commit. This should be good to merge once the build is finished.

shawnthompson:
@thomasgohard from a usability perspective would it be a good idea to use both options? On some carousels you can click the full panel and on others it's just the title caption?

thomasgohard:
@shawnthompson from a usability perspective, you should be consistent. Both options are valid. But you should not mix both options on one Web site. Use one or the other.

shawnthompson:
Perfect thanks

PR #6477, opened 2015-01-28
DataTable: Fixed ambiguous page links

LaurentGoderre:


LaurentGoderre:
@pjackson28 any issue with this one?

pjackson28:
Why aria-label? Why not update the link text and hide the page with wb-inv? It should be better supported than aria-label.

Don't we already have the translation for page already for all 33 languages?

pjackson28:
Looks like we have main-page. Perhaps that can help with the other languages.

LaurentGoderre:
It's not an issue to use wb-inv but aria-label would be easier to submit upstream. I will switch it anyways

LaurentGoderre:
@pjackson28 shouldn't we be adding href="#" to all those links so that we are not solely relying on the ARIA state?

pjackson28:
If we add anything it should be href="javascript;" rather than href="#". Having just the "#" can result in unexpected behaviour in some cases. For instance if the default is not prevented it would apply focus to the body rather than keeping focus where it is currently.

LaurentGoderre:
Oh you're right, but we should add something, otherwise it is not registered as a link

LaurentGoderre:
Any issue remaining or can I merge?

PR #6520, opened 2015-02-13
Geomap: Fixed issue where base map attribution text is too large on mobile.

jvanulde:


LaurentGoderre:
@jvanulde if this fix is only for mobile, then it should be added to a mobile specific stylesheet. You can add a `_screen-sm-max.scss` file, add your style and modify the `src/base/_wet-boew.scss` to load that new file

jvanulde:
It's not mobile specific. It addresses an issue that was also problematic on desktop but less so.

LaurentGoderre:
Ok, sounds good

src/plugins/geomap/_base.scss
```diff
@@ -142,6 +142,7 @@
 
 .olControlAttribution {
 	bottom: 5px;
+	font-size: x-small;
```
LaurentGoderre: I don't think these are used anywhere. Would it be possible to convert it to em to be consistent with the rest of the CSS?

jvanulde: We do use xx-small as well in the Geomap CSS. Not ideal. We can change them all now or in the next release. Your call.

LaurentGoderre: Ok, if it's consistent with the plugin itself then I can merge as it but it should be made consistent with the rest of the plugins for next version. Sounds good?

jvanulde: Sounds good to me.

PR #6530, opened 2015-02-16
Tabbed interface: Added SCSS property to remove an unnecessary left margin in medium view and over.

EricDunsworth:
Prevents a generic details/summary property from interfering with the tabbed interface's "full" layout.

### DIFF
--- a/src/plugins/tabs/_screen-md-min.scss
+++ b/src/plugins/tabs/_screen-md-min.scss
@@ -18,6 +18,10 @@
 					}
 				}
 			}
+			
+			* {
+				margin-left: auto;
+			}
 		}
 	}
 }


EricDunsworth:
Just for reference, here are before and after shots of the tabbed interface's example page...

**Medium view (before):**
tabs-medium-before

**Medium view (after):**
tabs-medium-after

Since in small view the tabbed interface reverts to a "standard" details/summary layout, the left margin will continue to appear in that scenario :).

LaurentGoderre:
I have a better fix for this that addresses the issue in mobile view and also optimizes some existing rules

EricDunsworth:
@LaurentGoderre Shouldn't the left margin appear in mobile view, though? In that view the tabbed interface pretty looks exactly like a grouping of standard ``<details>``/``<summary>`` elements with borders around them. That's why I had specifically targeted medium view and upwards in this PR, sine the left margin seems very out-of-place when the full-blown tab interface appears.

In any case, I'm good with whatever :).

LaurentGoderre:
My PR automatically addresses this. If margins are desirable in mobile then the tabbed interface code could be changed slightly

EricDunsworth:
Ok, in that case maybe I'll send over another PR later on to bring back the left margin in smaller views.

PR #6532, opened 2015-02-17
Tabbed interface: Adjusted "row" SCSS class to prevent a scroll bar from appearing when it's used.

EricDunsworth:
The scroll bar was introduced in PR #6302 to fix an unrelated issue.

**HTML5 code example...**

```
<div class="wb-tabs">
    <div class="tabpanels">
        <details id="details-panel1-1">
            <summary>Tab 1</summary>
            <div class="row">
                <div class="col-md-6">
                    <section class="panel panel-default mrgn-tp-md">
                        <div class="panel-heading">
                            <h2 class="mrgn-tp-0 mrgn-bttm-0">Panel</h2>
                        </div>
                        <div class="panel-body">
                            <p>Content</p>
                        </div>
                    </section>
                </div>
                <div class="col-md-6">
                    <section class="panel panel-default mrgn-tp-md">
                        <div class="panel-heading">
                            <h2 class="mrgn-tp-0 mrgn-bttm-0">Panel</h2>
                        </div>
                        <div class="panel-body">
                            <p>Content</p>
                        </div>
                    </section>
                </div>
            </div>
        </details>
    </div>
</div>
```

**Screenshot (before):**
tabs-row-before

**Screenshot (after):**
tabs-row-after

### DIFF
--- a/src/plugins/tabs/_base.scss
+++ b/src/plugins/tabs/_base.scss
@@ -213,6 +213,15 @@ $medGray: #ccc;
 		z-index: 1;
 	}
 
+	.tgl-panel {
+		> {
+			.row {
+				margin-left: 0;
+				margin-right: 0;
+			}
+		}
+	}
+
 	/**
 	 * Style 1 - Basic carousel style
 	 */


pjackson28:
We shouldn't be changing the row class. The problem is that the row class is being used incorrectly. The row class shouldn't be directly nested with the tab panel as it is only meant to nested within a class="container" element. If we change the row class, then we'll mess up everyone's implementation who is using the row class correctly within a tab panel.

Closing since this would break correctly implemented markup.

pjackson28:
Note also that the scroll bar is just a symptom of the row class being used incorrectly. The row class has negative margins/padding which is supposed to counteract the positive margins/padding of the container. If you nest it in an element that doesn't have margins or padding (such as the tab panel) then you'll get the row being wider than the parent container, resulting in scroll bars. So it isn't a problem with the row class, just with how it is being used.

PR #6547, opened 2015-02-19
Adds guidance for share widget right-side placement. Fixes #85.

sviens:
Updated as per @rubinahaddad (https://github.com/wet-boew/wet-boew-styleguide/issues/85).


LaurentGoderre:
I am not sure this belong here....Other themes may want to use it differently. To me, this is a style guide matter.

sviens:
@LaurentGoderre Sounds good.

PR #6556, opened 2015-02-23
SCSS Text Highlight Plugin: Added customizable variables

melodyma:
Added customizable SCSS variables to text highlight plugin. 

Note that the original _base.scss file contained parameters that were overwritten, because it was not specifying that the parameters were for `mark`. The original `background-color` that was defined and not used was also wrong too.

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -236,3 +236,13 @@ $carousel-s2-tablist-controls-btn-radius: 999px !default;
 $carousel-s2-tablist-thumbnail-active-border-color: #666 !default;
 $carousel-s2-tablist-thumbnail-active-border-style: solid !default;
 $carousel-s2-tablist-thumbnail-active-border-width: 10px !default;
+
+
+//== Text Highlight
+//
+//##
+
+$txthl-bg-color: #ff0 !default;
+$txthl-color: #000 !default;
+$txthl-font-weight: 700 !default;
+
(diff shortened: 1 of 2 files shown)


LaurentGoderre:
Which begs the question, why do we need a span around the mark.....could we not use `<mark class="txthl"></mark>` instead? This would simplify the code.

LaurentGoderre:
@pjackson28 do you know why is was created this way?

pjackson28:
No idea, that was the way it was in v3.x. I just ported it over without changing that part.

Why was the colour changed from ff0 to fcf8e3? The fcf8e3 colour is much harder to see than ff0 (so seems counter-productive as highlighting).

melodyma:
ff0 was actually being overwritten by fcf8e3 on the current production version, as with all the other styles declared in _base.scss for the text highlight plugin. Here's a screenshot:
screen shot 2015-02-23 at 9 39 58 am

LaurentGoderre:
@melodyma let's fix these. I will open a PR to fix the markup first, then we can change this PP accordingly.

pjackson28:
Hadn't noticed it being overridden before. The new colour is too difficult to see (didn't even notice the colour until I really looked closely). I don't know if ff0 is the best colour to use but it is definitely easier to see that fcf8e3.

LaurentGoderre:
Seems like this plugin has been broken for a little while....

LaurentGoderre:
@melodyma, can you rebase on top of master and change your CSS to use `#FF0` and use the `.txthl` selector instead of `mark`

PR #6557, opened 2015-02-23
Text Highlight: Removed the redundant `span` around the `mark` element

LaurentGoderre:
/cc @melodyma


LaurentGoderre:
@pjackson28 any objection to this?

LaurentGoderre:
I confirm that it's working in IE8

PR #6590, opened 2015-03-03
Plugin Footnotes: adds alternative footnote mark-up/styling

sviens:
Public Safety is currently use an ordered list for footnotes. Changing to the definition list (as in the current footnote plugin), would require considerable effort, while not providing much apparent value. I consider an ordered list for footnotes to be just as, possibly more, semantic as the definition list option. I'm not suggesting a replacement, just an alternative so that we can use the WET4 Footnote plugin, just in the context of an ordered list.

Thanks!

### DIFF
--- a/src/plugins/footnotes/_base.scss
+++ b/src/plugins/footnotes/_base.scss
@@ -72,7 +72,7 @@
 			ul,
 			ol {
 				margin: 0 $spacing $spacing (($spacing * 2) + $ddRtnWidth);
-				
+
 				&:first-child {
 					@extend %fnote-margin-top-spacing;
 				}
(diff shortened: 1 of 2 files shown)


LaurentGoderre:
That is a lot of markup. I'm very hesitant to bloat the CSS for little benefit.

sviens:
Perhaps there is a way to keep it to a minimum, I'm no SCSS expert. I just copied the `wb-fnote` styles and tweaked what I needed. I'm up for working with someone to reduce the bloat! 

Plus, I've never been sold on the name/value pairing semantics implicit with definition lists over the simple semantics of using an ordered list to mark-up what is an ordered list of notes at the end of a document. :)

So I'm down with either working with someone to streamline the added CSS, *or* replacing the definition list with (to my understanding) a more semantic option altogether.

Thanks for looking into/helping with this!

PS. I noticed the build failed at scss-lint, something to do with the property sort order. I'm not sure what that means but I'll dig into and fix it up when working on the de-bloating.

LaurentGoderre:
I don't quite believe that an ordered list  is more semantic because the order of the notes does not affect the meaning, it's only the order in which they appear in the content. The definition list on the other end offers an explicit relationship that would only be implied by an ordered list. If i recall correctly, not all footnotes are represented by an integer. I think IN have seen footnote  1a and 1b in the past.

LaurentGoderre:
As per the SCSS error, it's lint error. According to the code style rules, CSS property need to be in alphabetical order

```css
background :#000;
border: 1px solid #000;
color: #fff;
```

sviens:
The order of footnotes is an integral part of footnotes: "*Footnotes/endnotes begin with 1 and are numbered consecutively throughout the entire essay.*" ([Trent University's explanation of the Chicago Style for footnotes/endnotes](http://www.trentu.ca/academicskills/documentation/chicago_bibliography.php)). The use of a definition list ignores that footnotes are 'sequential' by nature; hence the need for them to be better semantically represented through an ordered list. Also, regardless of the symbol representing the footnotes, they are still sequential in nature - a great match for ordered lists! :)

Since definition lists are list types that support name/value pairing - technically "a term and a description" ([W3C 10.3 Definition lists](http://www.w3.org/TR/html4/struct/lists.html#h-10.3)) - the use of a definition list is giving the **terms** *Footnote x* a **definition** of *Footnote information*. Even going a bit generic and saying each name of *Footnote x* has a value of *Footnote information* doesn't provide the same semantic value of an ordered list; as in *here is a sequential list of the footnotes in the preceding document*.

Additionally, the footnotes are already demarcated through both the use of a header element (`h2` of Footnotes) and the reference link text `Footnote x`. Thus, the name/value pairing is also redundant while increasing individual page weight for no added benefit.

Thanks for clarifying the property order error! I'll fix that up before re-submitting.

LaurentGoderre:
None of this means that the order of the footnotes impacts the meaning, quite the contrary. In the example provided, the assumption is that a user reads a document from the first word to the last word but the reality is that users may jump to the 3 section of the document and start reading there and may encounter footnote 15 which is relevant to what he is reading. How many footnotes there was before it is irrelevant. In reality, the sole purpose of the number is to create a relationship between a section of text and a related piece of information

sviens:
I made no assumptions about how a user interacts with a document, their interaction is independent of the order of the footnotes, which are to be "*numbered consecutively throughout*". I'm not arguing about meaning, I'm suggesting that the semantics (the relationship) of the ordered list of footnotes in a document is better represented by using an element designed with ordering items in mind in the footnotes section at the end of a document.

Either way, I'm proposing that an alternative to the bulky, and unnecessary, definition list be provided when marking-up footnotes. [W3C even uses an unobtrusive `<p>` element for their footnote example](http://www.w3.org/TR/html5/common-idioms.html#footnotes), and I'm cool with that option as well. 

Basically anything that moves away from the definition list example, or at least an alternative should be provided. :)

andre-gh:

andre-gh:

PR #6602, opened 2015-03-06
Lightbox: Fixed positioning of close button for AJAX sources

### LINKED ISSUE #6600: AJAX Ligtbox close button Problem
I found a small bug in the modal lightbox.  I could only find this in the one instance of this plugin but the close button ends up moving the right side of the screen.  You can actually see it on the demo page:  http://wet-boew.github.io/v4.0-ci/demos/lightbox/lightbox-en.html

I was looking at the Ajax Example 1.

modallightboxerror

Any thoughts on the fix?

pjackson28:
Fixes #6600

### DIFF
--- a/src/plugins/lightbox/lightbox.js
+++ b/src/plugins/lightbox/lightbox.js
@@ -210,19 +210,22 @@ var componentName = "wb-lbx",
 				},
 				parseAjax: function( mfpResponse ) {
 					var urlHash = this.currItem.src.split( "#" )[ 1 ],
-						$response = $( "<div>" + mfpResponse.data + "</div>" );
+						$response;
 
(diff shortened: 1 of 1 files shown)


src/plugins/lightbox/lightbox.js
```diff
@@ -210,19 +210,22 @@ var componentName = "wb-lbx",
...
+					} else {
+						$response = $( mfpResponse.data );
 					}
 
 					$response
 						.find( ".modal-title, h1" )
-						.first()
-						.attr( "id", "lbx-title" );
+							.first()
+								.attr( "id", "lbx-title" );
```
LaurentGoderre: Not sure this indendation change is necessary in the future.

PR #6605, opened 2015-03-06
Data-AJAX: Added ability to filter by URL hash or selector

### LINKED ISSUE #6563: Data Ajax: pulling in content by id not working
In WET version 4.0.3 and prior, there was a function available for the data ajax script to directly pull content from a page by ID, class or tag.  For example the ajax call would be something like 
="ajax/data-ajax-extra-2-en.html #ajax-content"
that would pull content wrapped with the id of "ajax-content".  You can further customize it by pulling in by class and/or certain tags within that content.
As of version 4.0.4, that function has disappeared due to a change in the ajax-fetch and data-ajax scripts in the wet-boew.js script.   Is it possible to get this function back for a future version of WET4?  Or is there a new way to pull in content based on id/class/tag??

pjackson28:
Fixes #6563

### DIFF
--- a/site/pages/docs/ref/data-ajax/data-ajax-en.hbs
+++ b/site/pages/docs/ref/data-ajax/data-ajax-en.hbs
@@ -6,7 +6,7 @@
 	"categoryfile": "plugins",
 	"description": "A basic AjaxLoader wrapper that inserts AJAXed-in content.",
 	"altLangPrefix": "data-ajax",
-	"dateModified": "2014-08-04"
+	"dateModified": "2015-03-09"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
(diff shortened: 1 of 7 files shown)


LaurentGoderre:
Before merging, we need to add a unit test for this scenario. I can take care of that if you want

pjackson28:
Yes please.

pjackson28:
@LaurentGoderre Do you have an estimate of when the unit testing will be ready? If it won't be ready before v4.0.12 then we should merge this PR in as is and bring in the unit testing later (as soon afterwards as possible) since there is a potential for the related issue to be a blocker for some migration.

LaurentGoderre:
While implementing the test, I found something unclear about this....If no filtered content is found, the plugin should not include all the rest of the content, it probably should ignore the request.

LaurentGoderre:
@pjackson28 I opened a PR on your branch for the test.

pjackson28:
It was intentional to return the content but if it is better to return nothing then you can switch it to that. I was just concerned about the URL hash being used for another purpose and inadvertently filtering out all content.

pjackson28:
Note that the same was done for the Lightbox PR.

LaurentGoderre:
I'm closing this one in favor of #6625

PR #6612, opened 2015-03-09
Tables: Fixed the pagination display where there are more than 6 pages

### LINKED ISSUE #6610: Tables: Issues with Pager when using Server Side Processing
We're trying to use the Tables control (http://wet-boew.github.io/wet-boew/demos/tables/tables-en.html) to display a large dataset...so we needed to enable server side processing.  It works actually quite well, however the pager at the bottom of the table seems incorrect.  If you select the next/previous links, its cycles through 33,000+ records just fine, calling for them a page at a time from the server:
image

You can see however, that the pager shows the final page as 6 in this case, when it should be page 3,334 as shown by the example below which shows the identical dataset/JSON call to a server side application/page but just loads the table data into a table from the DataTables example pages/documentation:
image

I'm not sure if we've done anything incorrectly, but since we've done nothing to the control and have simply set this to use server side processing, is it possible the the WET "adaptation" to this control has somehow caused a bug with the display of pagination links?

pjackson28:
Fixes #6610

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -159,8 +159,8 @@ $document.on( "init.dt draw.dt", selector, function( event, settings ) {
 		} )
 		.not( ".previous, .next" )
 			.attr( "aria-pressed", "false" )
-			.html( function( index ) {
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
HUH! I didn't know you could do that!

PR #6614, opened 2015-03-10
Charts - Thousand and decimal separator defined in i18n and documentation improvement

### LINKED ISSUE #3267: Pie Charts - Decimal Issue
http://wet-boew.github.io/wet-boew/demos/charts/pie-en.html

If you go in the Decimal Precision section, charts will only work properly with a period(.) as the decimal separator, not with a comma(,) (as used in French). 

Eng: 12.3
Fra: 12,3

With comma as decimal separator, the chart will show, but won't consider decimals in calculation, meaning the data in the chart won't reflect exactly those in the table.

### LINKED ISSUE #4433: Chart Graph-Simple and Easy--Bar Charts
For this particular bar chart, there's not enough separation b/w one x-axis category to the other (e.g, "General" bar for corporate is really close to the "Lawyer" bar for "Information").  Can it be personalized to increase the separation for easier viewing?

duboisp:
Resolve #3267
Note: the label rendering on the pie charts depends on the browser local and not on how the number was formatted in the table.

Close #4433
The documentation was improved


LaurentGoderre:
I think the first issue is not the right one.

Can you provide a description of this change because it's massive.

Also, if there is an i18n for separator, shouldn't you use it the demos?

duboisp:
Thanks @LaurentGoderre I fixed the pull request description with the right issue number.

I also modified an existing working (charts customization first example) to show how the support of the i18n thousand separator.

LaurentGoderre:
What about if they use comma as thousand separator instead of decimal seperator.

In essence, all of these should work:

1,234.12
1 234.12
1 234,12

LaurentGoderre:
@duboisp the following function resolves all of the above

```js
var parseFloatI18n = function(number, base) {
  base = base || 10;
  
  return parseFloat(number.replace(/(\d{1,3}(?:(?: |,)\d{3})*)(?:(?:.|,)(\d{1,2}))?$/, function(a,b, c){
    return b.replace(/ |,/g, "") + "." + c || "0"
  }), base);
}
```

duboisp:
Thanks @LaurentGoderre it work well, I will implement it, I will remove the i18n entry I did and then I will update my commit.

LaurentGoderre:
Any news on updating this PR?

duboisp:
@LaurentGoderre  It was already updated and tested with the code you provided in your previous comments.

LaurentGoderre:
So it's ready to merge?

duboisp:
yes

PR #6641, opened 2015-03-25
Form Validation: Added support for a single required checkbox

### LINKED ISSUE #6638: Form validation: required checkbox that is on its own and not in a fieldset asterisk issue
We've come across instances where we require a single checkbox not in a fieldset and it is a required field. It seems that the asterisk bleeds into the checkbox. We've played with the CSS a bit as follows to get the red asterisk to appear before the checkbox but not sure if this can maybe be properly added to WET framework too? Thank you.

'''
div.checkbox.required:before{
        content:"* ";
        color: red;
        margin-left:-.665em
}
'''

LaurentGoderre:
Fixes #6638

### DIFF
--- a/src/base/forms/_base.scss
+++ b/src/base/forms/_base.scss
@@ -11,7 +11,9 @@ $asterisk-width: 0.665em;
 }
 
 label,
-legend {
+legend,
+.checkbox,
+input[type="checkbox"] {
 	&.required {
 		&:before {
 			@extend %forms-color-red-italic;


LaurentGoderre:
@thomasgohard @pjackson28 any objection to this?

thomasgohard:
Should we also include `input[type="checkbox"]` for when the `checkbox` class isn't applied?

LaurentGoderre:
That has the risk of being to intrusive. Isn't the checkbox class from bootstrap?

thomasgohard:
Yes, the `checkbox` class is from Bootstrap. Bootstrap applies most styling to both `input[type="checkbox"]` and the `checkbox` class. So doing the same here would be consistent.

All the `checkbox` class seems to do differently is to remove the bold treatment on the label element.

PR #6650, opened 2015-03-26
Multimedia & Feeds: Replace the loading image with a glyphicon

LaurentGoderre:


LaurentGoderre:
@pjackson any issue with this?

Issue #6570, opened 2015-02-26
Tabbed Interface: Geomap: Don't play nice together


shawnthompson:
I tried to put a geomap inside a tabbed interface and the map would not load.

Here's an [example page](http://servicecanada.github.io/GCWeb/contact/find-en.html).

This page is a draft and might be updated or deleted so if someone goes to check it and it's not there, let me know and I'll put it back up.

This has been tested on the tabs demo page and same thing happens there.

LaurentGoderre:
@jvanulde can you look into this?

pjackson28:
@shawnthompson Could you please describe what the incorrect behaviour is? I'm assuming it has something to do with Geomap injecting some of the content outside of the tabbed interface but just wanted to confirm what is the actual issue (also makes for more helpful issue tracking).

shawnthompson:
The geomap will not show up inside the tabs, I had to move it outside of the tabs to show the client what it would look like if it did show up.

pjackson28:
@shawnthompson Did it show up for you at all in the tabs? When I tried it, some content (not the maps) appeared in the tap panel while the rest (including the maps) appeared below the tab panel. Did you experience that behaviour as well?

shawnthompson:
@pjackson28, It didn't show up at all but I'll try it again and let you know

shawnthompson:
@pjackson28 , I updated my [sample page](http://servicecanada.github.io/GCWeb/contact/find-en.html)

shawnthompson:
same results

pjackson28:
I think I know what the issue is. I think Geomap caches a reference to the map container (and possibly other containers) which gets lost when the Tabbed interface gets enhanced, since the enhancement involves DOM changes. Those DOM changes are invalidating the previous reference to the map container (since it is a descendent of the elements that underwent DOM changes). Since the reference to the map container is lost, Geomap has nowhere to inject the map.

@jvanulde The way we've dealt with this in other plugins is we store the id of the element rather than a reference to the element itself. It is a very inexpensive operation to find an element by id so we don't have to cache an object indefinitely. It is fine to cache the element for the duration of a function or a sequential (uninterrupted) set of steps but any time you are starting the executing of code as a result of an event or outside trigger then you should get the element fresh with the id.

pjackson28:
[...]

PR #6663, opened 2015-04-02
Core: Improved handling of nested plugin/polyfill elements.

pjackson28:
This adds support for different plugins/polyfills nested within one another where previously only plugins/polyfills of the same type could be nested. Also it enables the top-level plugin/polyfill to care care of initializing all its descendents (so you would only need to do a wb-init on the top ancestor).


pjackson28:
Fixes #6570

LaurentGoderre:
Looks good to me!

PR #6728, opened 2015-04-28
Base: change look & feel of  global-loading indicator

ESWAT:
I found the current spoke style loading indicator, used for Multimedia and Twitter plugins right now, to be very jarring and dizzying to look at for videos that take a long time to buffer.

I wanted to create an indicator that shows there’s ongoing activity in a more contemporary fashion. In this case I replaced it with a slowly-rotating, quarter-filled, thick circle (a less fancier-version of the kind of indicators Google has been using). I went with this instead of altering the existing spoke animation because of the complexity that would involve to do properly (wanted to keep the simple CSS rotation implementation because it works).

screen shot 2015-04-28 at 10 46 45 am


LaurentGoderre:
#6650 is actually replacing it with a glyphicon so we don't have to load a seperate image

ESWAT:
Ah, OK. Feel free to close.

Issue #6719, opened 2015-04-23
Tabbed Interface: how to deal with long tab labels? ie. make the text wrap


robwloch:
We have an existing home grown tabbed interface solution in CLF 2.0 which we are converting to WET4 so we're switching to WET4's tabbed interface. The issue we've run into is that our current solution has several tabs with longer sentences and so in CLF 2.0 we made it span two rows on the tab label. We noticed that this works in WET4 too by using `<br>` but it doesn't look the greatest. Just wondering what you recommend or if there could maybe be an option parameter added to the widget to force all tabs to be the same width and thereby automatically wrapping long labels without having to hardcode `<br>`'s? Thank you.

LaurentGoderre:
@robwloch have you tried using Bootstrap grids to enforce a width to the container?

robwloch:
We're just using the full width of the div to hold the tabbed interface. Which container are you referring to?

LaurentGoderre:
If you add a grid to the tabbed interface it should stay the same width for all tabs

pjackson28:
@robwloch How long is the text that you are including in the tabs? Normally the text in tabs is supposed to be kept short as the tabpanel is where all the text is supposed to reside.

@thomasgohard What is the normal convention regarding long text and tabs? Are tabs still used when the text is long or should an alternate solution be used (e.g., accordion)? If long text is still used, what is normally the best practice for dealing with the long text?

thomasgohard:
Agreed, tab labels should be kept short: They're supposed to support scanning. If tab labels start getting long, it's an indication that tabs are probably not the right way to present the content.

The one use case I can think of where longer tab labels and an accordion-style interface would work is a multiple step process within a single page like the one discussed in #6367.

To avoid the situation where tabs start wrapping when their combined labels are too long or there are too many tabs, we could force a `max-width` on all tabs based on the number of tabs (i.e., available width divided by number of tabs). The tab labels should then simply wrap within the tabs.

robwloch:
Hi guys, this is our CLF 2.0 version that we're trying to move to WET4. The tab text is quite short in my opinion but I guess not short enough:
![capture](https://cloud.githubusercontent.com/assets/5721679/7373906/40759930-ed9d-11e4-9af3-de27da562f75.PNG)

thomasgohard:
@robwloch yes, that's not excessively long.

I think I have a solution. I'm testing it out and will update this thread with the results.

thomasgohard:
[...]

PR #6730, opened 2015-04-28
Tabbed interface: Stop the list of tabs from wrapping when there are too many tabs or when their labels are too long.

thomasgohard:
Fix for #6719.

This will ensure that tabs themselves don't wrap and stay in a single row. The labels of the tabs will now wrap instead.

### DIFF
--- a/src/plugins/tabs/_base.scss
+++ b/src/plugins/tabs/_base.scss
@@ -3,6 +3,8 @@
  * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
  */
 
+$tab-margin-width: 10px;
+
 %tabs-position-relative {
 	position: relative;
 }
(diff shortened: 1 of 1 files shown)


src/plugins/tabs/_base.scss
```diff
@@ -135,6 +135,8 @@
 	}
 
 	[role="tablist"] {
+		border-spacing: 10px 0;
```
LaurentGoderre: Can you use a variable for this

src/plugins/tabs/_base.scss
```diff
@@ -150,6 +152,8 @@
 			}
 			color: $carousel-tablist-tabcount-color;
 			cursor: pointer;
+			display: table-cell;
+			left: -10px;
```
LaurentGoderre: ...and reuse the variable here `-$variable`

thomasgohard:
@LaurentGoderre done.

Anything else or can we merge?

PR #6741, opened 2015-05-03
Overlay: Ensure that elements with focus outside an overlay are visible.

thomasgohard:
cc/@masterbee


masterbee:
@thomasgohard - this looks good and well done. It is green on my end. If there are no issues I would like to merge.

thomasgohard:
@masterbee if there's anything else we find, we can fix with another PR. I'm sure it could use some more testing.

I should also document somewhere that it will only work for the top/bottom bars. For the side bars it won't work as we can't scroll the page horizontally.

PR #6745, opened 2015-05-05
Adding browser support page.

ghost:


pjackson28:
A French page is needed as well, otherwise we'll have a 404 when the language link is used.

LaurentGoderre:
Actually, the build would fail because of the broken link

pjackson28:
Not sure why the build isn't broken but we shouldn't be providing an English page without an equivalent French page. It would even be better to have the French page with the English text (using the appropriate lang attribute) rather than no French page at all (although it should still be translated).

site/pages/docs/browser-support-en.hbs
```diff
@@ -0,0 +1,18 @@
+---
+{
+	"title": "Browser Support",
+	"language": "en",
+	"description": "Browser support for the Web Experience Toolkit.",
+	"altLangPrefix": "index",
```
LaurentGoderre: This is the reason why it's not failing. It's pointing to another page which might actually be worse. I would recommend copying the english page, wrapping in a `<div lang="en"></div>` in the meantime and change this value to `browser-support`

pjackson28:
Can the missing language page and incorrect language link be fixed so this can be merged in? Also the message should be changed to "Documentation: Added browser support page"

shawnthompson:
Can we finish this? and would this be the "official" list?

andre-gh:

andre-gh:

PR #6766, opened 2015-05-12
Overlay: Added no JavaScript fallback

### LINKED ISSUE #6763: Overlay: Question before allowing it's use...
Has this been vetted for accessibility and usability? I ask only because in WET3 there was a vaguely similar plugin that was used for a table of contents but it wasn't very well received by the Accessibility group, and I believe there were some usability concerns as well. I would just like to ensure these are an issue with this overlay plugin and it will be safe to use going forward.

pjackson28:
Fixes #6763

### DIFF
--- a/src/base/noscript.scss
+++ b/src/base/noscript.scss
@@ -4,11 +4,17 @@
  */
 @import "../variables";
 
+%noscript-display-block-important {
+	display: block !important;
+}
+
 /*
  Plugins
  */
 @import "../plugins/tabs/noscript";
 @import "../plugins/menu/noscript";
+@import "../plugins/overlay/noscript";
+@import "../plugins/lightbox/noscript";
 
 /*
  Polyfills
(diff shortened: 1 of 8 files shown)


nschonni:
I think this needs to stay opt-in. Otherwise noscript people would suddenly have their pages littered with error message pop-ups, etc...

pjackson28:
That is no different than with tabbed interface, details/summary, etc where all hidden content are made visible by default. This only makes visible content that is already embedded in the page but hidden with CSS. The issue here is that unless the implementer knows to opt-in, they will have dead links when JS is disabled (which is worse than littering the page with previously hidden content). If they don't want this content to be visible when JS is turned off, then they can force it to remain hidden with CSS (although those would likely be edge cases).

LaurentGoderre:
I just talked with @pjackson28 and this would only affect overlay that are embed on the page. Dynamic overlay should really be injected via Javascript.

PR #6775, opened 2015-05-13
[WIP] Data-inview: Modifications to deal with data-inview content that is dismissable

thomasgohard:
**Do not merge: This needs a lot of work.**

cc/@masterbee


LaurentGoderre:
Does this need to be merge before the release or can it wait after?

/cc @lucas-hay

masterbee:
@LaurentGoderre - Yep it needed to make it into the mix. So I have merged, since there is some demand for dismissable content and this bug has some negative impacts on user experience.

Thanks :+1:

PR #6786, opened 2015-05-22
Dismissable content: Changes to ID handling

thomasgohard:
- Eliminate the whitespace before generating an ID.
- When an ID has been set, use that ID.


thomasgohard:
cc/@masterbee.

Issue #6751, opened 2015-05-08
Geomap bugs


bsouster:
1. When using the search box with the keyboard, I cannot move focus down through the options.

2. The map starts without the blue focus border but as soon as I give the map focus it maintains the blue border even when the map isn't in focus or arrowkey-traversable. It does this for me in Chrome 42 and Firefox (dev edition) v39. Sometimes the minus key zoom-out functionality doesn't work while the plus key zoom does.

CC @jvanulde

jvanulde:
@bsouster item 1 in your list of bugs should be fixed now. Please test and close.

PR #6808, opened 2015-05-28
Plugin Geomap: Fix active border toggle on map element.

jvanulde:
Addresses in part issue #6751. Now using native OpenLayers events to catch mouseover and mouseout on the map.


LaurentGoderre:
Looks good. Is it ready to go?

jvanulde:
Good to go.

PR #6820, opened 2015-05-29
jQuery UI - Sortable, draggable and other ui features

pascal-bigras:
I would like to suggest the implementation of jQuery UI to the template. I experimented with and it is very easy to use and very powerful. You can make lists or tables sortable, draggable, droppable, resizable, etc. There are some other nice features like a date picker, dialog, progress bar, slider and you can even make custom animations. I know that must not be fully accessible but the use of jQuery UI is manly to improve the UX of the users. It would be meant for complementary use more the primary. Here is the link to find more about it : https://jqueryui.com/.


LaurentGoderre:
jQuery UI on top of not being accessible, duplicates functionality provided in native HTML and for which we already have polyfills for. It is very unlikely we will include jQuery UI.

LaurentGoderre:
It doesn't appear that there are any new commits here, you should open an issue instead of creating a pull request

pascal-bigras:
Ok thank you. Do you think that we could still use it to enhance UX? I mean as long as the access to the main content is not affected it would not be an accessibility issue.



Pascal Bigras
IT Developer | Développeur en TI
Web Services | Services Web
COTS Integration Solutions & Services (CISS) | Solutions et Services d'intégration de COTS (SSIC)
Management Systems Division (MSD) | Division Systèmes de gestion (DSG)
Corporate and Entreprise Solutions Directorate (CESD) | Direction des Soutiens et Systèmes d’Entreprises (DSSE)
Information Technology Branch (ITB) | Direction générale de l'informatique (DGI)
Canada Revenue Agency (CRA) | Agence du revenu du Canada (ARC)
3-63, 2465 St-Laurent, Ottawa ON, K1G 6G8
Telephone | Téléphone 613-948-5733

Feel free to reply in the official language of you choice.
Sentez-vous à l’aise de répondre dans la langue officiel de votre choix.

pjackson28:
@pascal-bigras Another issue is the weight of jQuery UI. It is a very heavy framework and by no means optimized for mobile. With over 33% of visitors using mobile devices we can't afford such heavy downloads and processing. A long time ago we integrated jQuery UI but had to strip it out because of the major performance problems it was causing. We instead went with native methods and polyfills as we get equivalent functionality with much better performance (and accessibility).

Issue #6809, opened 2015-05-28
Changes to WET Core and Themes to simplify implementation by variants


LaurentGoderre:
To simplify the implementation of WET in variant I am proposing the following changes:

* [x] Merge wet-boew.css into theme.css and deprecate the use of wet-boew.css
* [x] Merge the unmin resources with the minified resources
* [x] Move theme resources into their own subfolder

/cc @pjackson28 @masterbee @sylus @jeffdavidgordon @upsonp

pjackson28:
One risk is that this will result in an across the board markup change (merging two CSS files). If people forget to remove the old theme CSS file then it will override any newer changes in the merged CSS file. Should this be done globally at this point or just as part of the AEM variant? If we are doing it globally, how would it be versioned, communicated and managed?

I am unclear what is meant by merging the unminified and minified resources. Why would they be merged? Wouldn't that cause a lot of overlap and confusion having minified and unified files in the same directory (since that means duplicate JS, CSS, HTML, etc. files). Shouldn't we only be sending production (minified files) to a production server?

Not sure what you mean by moving theme resources into their own sub-folder. What would be the before and after? Are you meaning in the source repository or the dist repository or both?

How would we create streamlined packages? Isn't that the responsibility of variants rather than core and the themes? Or do you mean that variants would be responsible for creating the streamlined packages? Also what files would be included and what would be left out and where would this be done?

pjackson28:
Also this seems to involve some pretty significant changes for a maintenance release (since this is tagged v4.0.15). What would be the impact on each of the repos as well as implementers?

LaurentGoderre:
This should have no effect on the implementers, here is why:

1) The WET styles are integrated inside the theme file. The existing WET CSS files are kept so that it doesn't work even if you don't update your markup.

2) We would be merging the folders, having the minified and unminified source in the same folder

3) This would introduce a small markup change but this is something every variant needs. We may be able to address this without introducing the markup change.

pjackson28:
[...]

PR #6840, opened 2015-06-04
Build: Moved core and themes resource in their own subfolders

LaurentGoderre:
Partially implements #6809


LaurentGoderre:
GTG!

site/includes/footerresources.hbs
```diff
@@ -1,17 +1,17 @@
 <!--[if gte IE 9 | !IE ]><!-->
-<script src="https://ajax.googleapis.com/ajax/libs/jquery/{{environment.jqueryVersion}}/jquery{{environment.suffix}}.js"></script>
-<script src="{{assets}}/js/wet-boew{{environment.suffix}}.js"></script>
+<script src="https://ajax.googleapis.com/ajax/libs/jquery/{{environment.jqueryVersion}}/jquery.js"></script>
```
nschonni: It appears that jQuery is always server unminified after this change

LaurentGoderre: Really? I thought I fixed that

LaurentGoderre: We would need to move the UseMinAssets to a separate task, it gets really annoying to update it on every theme.

Issue #6812, opened 2015-05-28
Horizontal description lists


Bgolden67:
Silly issue really.. but it's caused frustration in attempting to use this.
it seems that no matter what we try to over ride the alignment, the <code>dt</code> item is aligned right. This causes a white space gap on the left side of the lists.
Is there a way I am missing to force this left aligned or can the <code> dl-horizontal</code> class be adjusted to default left instead?

thomasgohard:
As indicated in https://github.com/twbs/bootstrap/issues/14180, we would need to implement an override for `.dl-horizontal dt` as it has greater specificity than `.text-left`.

Right-aligned text is harder to scan so the default should be left-aligned. There are also other problems with the style, including the fixed width for the `dt` element and long `dt` elements being truncated when they don't fit the fixed width. The styling of the horizontal description lists needs to be entirely redone.

hsrudnicki:
@Bgolden67 @thomasgohard : Are you ready to close this issue or otherwise?

Bgolden67:
thank you

PR #6842, opened 2015-06-05
CSS: Improvements to the horizontal description list styling.

thomasgohard:
For #6812.

### DIFF
--- a/src/base/bootstrap-overrides/_base.scss
+++ b/src/base/bootstrap-overrides/_base.scss
@@ -45,8 +45,24 @@ a {
 }
 
 .dl-horizontal {
-	dt {
-		white-space: normal;
+	@media (min-width: $screen-sm-min) {
+		dt {
+			border-top: 1px solid $clrMedium;
+			padding: 10px 10px 10px 0;
+			text-align: left;
+			white-space: normal;
+		}
+
+		dd {
+			border-top: 1px solid $clrMedium;
+			margin-bottom: 3px;
+			margin-left: 160px;
+			padding: 10px 10px 10px 0;
+		}
+
+		dt + dd {
+			padding-bottom: 0;
+		}
 	}
 }
 


hsrudnicki:
Just by occasion, can this issue be closed as well?

LaurentGoderre:
Please don't nest media queries inside selector. If you need media query, create a new file and include it in the `_wet-boew.css`

LaurentGoderre:
Where does the 768px come from?

LaurentGoderre:
Ah nevermind. Bootstrap seems to be doing this a lot.

src/base/bootstrap-overrides/_base.scss
```diff
@@ -45,8 +45,24 @@ a {
 }
 
 .dl-horizontal {
-	dt {
-		white-space: normal;
+	@media (min-width: 768px) {
```
pjackson28: Shouldn't we be using $screen-sm-min instead?

LaurentGoderre: Bootstrap uses this media query a lot in it's code. It seems like he just copied it.

pjackson28: Maybe so, but it is also defined in a variable so shouldn't we stick to the variables? If Bootstrap changes from 768px to 769px then they are responsible for updating all their variables and code but by us using a variable then we wouldn't have to make a change if that scenario were to happen.

LaurentGoderre: Aaaah good call, I only looked at the generated CSS. You are right!!

andre-gh:

andre-gh:
@thomasgohard Since this fixes an issue are you willing to resume this pull request or should I take it over?

thomasgohard:
@andre-gh I've rebased the PR. It's ready for review and merging.

andre-gh:
Thanks, so much much better aligned right.

andre-gh:
Er, aligned correctly

Issue #6791, opened 2015-05-25
Lightbox: Transparency overlay


shawnthompson:
When using the lightbox gallery and adding text below the image it makes it hard to read if the overlay has black text beneath it.

![lightbox-issue](https://cloud.githubusercontent.com/assets/2575266/7800050/61c95478-02dd-11e5-8b54-a89c34f310de.JPG)

cc: @trtustian

thomasgohard:
Agreed. There should be a solid background behind the text.

Can you submit a pull request with the fix?

hsrudnicki:
@shawnthompson @thomasgohard  @LaurentGoderre : Are you ready to close this issue?

pjackson28:
So we now have two options, one changing the bottom text to a grey background with black text and the other increasing the opacity of the background overlay. Which should we pick? Should we offer both options? If so, how do we offer both options?

Also, this is something that should also be considered for addressing upstream.

hsrudnicki:
Just labeled "Going Stale" to close this issue when resolved/fixed completely, and 
please @LaurentGoderre and others help in making decisions/answers to the last comment's questions from @pjackson28, thank you

hsrudnicki:
Can be this issue closed as resolved/fixed or is this still an issue?  Either way, please update this issue's status, whether this issue can be closed or otherwise? Thank you
[Note that the both referenced PRs have been already merged.]

RobJohnston:
I think this can be closed since https://github.com/wet-boew/wet-boew/pull/6856 got merged and the change can be seen on the [Lightbox working example] (http://wet-boew.github.io/wet-boew/demos/lightbox/lightbox-en.html).

shawnthompson:
Thanks @RobJohnston

PR #6843, opened 2015-06-08
Plugin lightbox : add a background color to the description 

stevenbriscoeca:
I added a background color to the lightbox in reference of this issue :
https://github.com/wet-boew/wet-boew/issues/6791

I tested the contrast, you can see the result here :
http://jxnblk.com/colorable/demos/text/?background=%23ccc&foreground=%23243850

### DIFF
--- a/src/plugins/lightbox/_base.scss
+++ b/src/plugins/lightbox/_base.scss
@@ -61,14 +61,26 @@
 
 /* Should fix upstream in Magnific Popup rather than overriding in WET */
 .mfp-bottom-bar {
+	background-color: $clrMedium;
+	margin-bottom: 5px;
+	padding: 10px;
+
+
 	.mfp-title {
+		color: $clrDarkBlue;
+		display: inline-block;
 		padding-right: 5px;
 		width: 75%;
+
 	}
 
 	.mfp-counter {
+		color: $clrDarkBlue;
+		display: inline-block;
 		font-size: 1em;
+		position: static;
 		text-align: right;
+		vertical-align: top;
 		width: 25%;
 	}
 }


LaurentGoderre:
To avoid PR getting like this, we recommend to always start by checking out master, pull the latest ansd then create a new branch.

For now, you can try fixing the PR by doing 
```
git pull upstream
git rebase upstream/master
```

LaurentGoderre:
The build error is because CSS property need to be ordered alphabetically

shawnthompson:
Maybe rebase / squash your commits

stephenbe:
I just figured out my error, I did update my master, but I only pushed my branch instead of pushing my master first...Thank you for your patience.

stephenbe:
@LaurentGoderre Everything is now fixed and ready to go!

@thomasgohard any thoughts on this?

hsrudnicki:
So, can this issue be closed?

shawnthompson:
Not until it's merged. Still waiting on comments.

masterbee:
@stephenbe - is there any intention on trying to push this upstream to Magnific Popup rather than overriding in WET as your comment hints at? Just want to deduce if there is any value in adding it to WET if you are already pushing this upstream.

stephenbe:
@masterbee - I usually don't modify plugin code as much as possible, just in case you would want to update this plugin. You won't have to deal with putting back in the code I modified. In addition, I believe that this modification is for WET needs and has nothing to do with the plugin itself.

Nothing is black and white but in this case I think it's more appropriate to add the css into WET. If there was bugs or issues with the plugin, then I think it would be appropriate to modify the code or make a PR.

Hope that answers your question!

pjackson28:
Which PR should we be going forward with, this one or #6856? They both address the sam3 issue but have a different impact on the resulting appearance. I think the fix in #6856 is less noticeable of a change so should be more easily accepted (even upstream).

andre-gh:

andre-gh:

PR #6859, opened 2015-06-12
Theme: Moved custom SCSS variable declarations before bootstrap's ones.

EricDunsworth:
Prevents bootstrap's !default variables from overriding WET's ones.


EricDunsworth:
Closing since #6855 already achieves the same thing as this.

Issue #5981, opened 2014-08-27
jQuery Timeline script


crochefort:
Good day everyone, I notice that the style-tab-3 from the TAB script has not been port to WET 4.

So I did some research and found a script that might be a good addition.

http://www.csslab.cl/2011/08/18/jquery-timelinr/

I just want your input on how much effort it would take to bring it compliant. 

On my DEV server I start working on it but when i'm resizing the screen to get a mobile view the script dont resize correctly so it need some work on it.

@pjackson28  As you request :) thanks!

hsrudnicki:
@crochefort @thomasgohard @pjackson28 : Can be this issue closed?

thomasgohard:
No. It is presently in discussion.

hsrudnicki:
Based on the above merged PR, and no further discussion/status updates, this issue made "Going Stale" and closed or if otherwise, please update again, thx

PR #6861, opened 2015-06-15
Tabbed interface: Added class to allow tabs to wrap.

thomasgohard:
@crochefort this should do the trick as a quick fix for the problem we discussed the other day. Just add the class `allow-wrap` to your tab list and it'll revert to the previous behaviour before #6730.

### DIFF
--- a/src/plugins/tabs/_base.scss
+++ b/src/plugins/tabs/_base.scss
@@ -220,6 +220,16 @@ $tab-margin-width: 10px;
 				top: 1px;
 			}
 		}
+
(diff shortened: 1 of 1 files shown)


crochefort:
Thanks Thomas! it's working! :+1: 
At least I will be able to go forward with the Pathfinder!

To keep track : https://github.com/wet-boew/wet-boew/issues/5981

thomasgohard:
Thanks @crochefort. Merging in. The fix will be in the next release of WET. In the meantime, you can always take the development version.

Issue #6246, opened 2014-11-06
Splash Page - Math Polyfill Default Values...


lafleurp:
If the JQuery's library is unavailable, one of the polifill default values kicks in and we get weird text.

External CDN was blocked by error. That's how I ended having that issue!

Screenshot:

![image](https://cloud.githubusercontent.com/assets/8491916/4941470/77cd106c-65e3-11e4-92d7-17b96ecaa551.png)

hsrudnicki:
Are you ready to close this issue or otherwise?

RobJohnston:
I kind of liked the StackOverflow idea of a fallback.  Just today I had to revamp an application that is about to be released because it is going onto https://secure.cic.gc.ca/ instead of http://cic.gc.ca/.  The secure server can't see the Google CDN.

lafleurp:
Yep, wish they would implement it!

pjackson28:
Note that PR #6874 hides the MathML check with CSS after the Modernizr check completes. So it should prevent the xx/yy from being visible any more.

As for your issue with not seeing the Google CDN, then you should instead host the jQuery file on a server that can be seen and update the URL accordingly. We should minimize the use of extra loading through JS (particularly jQuery) as that introduces performance issues. You should still use a CDN wherever possible, but sometimes that may not be possible so should be handled on a case-by-case basis rather than globally.

pjackson28:
Another thing to consider is that you should be using https for the jQuery URL rather than http if you application uses https (Google CDN supports both http and https).

PR #6874, opened 2015-06-17
MathML: Hides Modernizr MathML check with CSS after the check has completed

pjackson28:
Related to #6246

### DIFF
--- a/src/base/views/_screen.scss
+++ b/src/base/views/_screen.scss
@@ -6,6 +6,22 @@
 	display: none !important;
 }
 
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
@hsrudnicki can you look for related issues for this. It came up a few times

hsrudnicki:
It's not exactly but into consideration, here is the next issue where is a suggestion for using of MathML:
https://github.com/wet-boew/wet-boew/issues/7013

PR #6909, opened 2015-06-30
Add the WET theme to the cdn repo and added license files to the packages

LaurentGoderre:


LaurentGoderre:
@pjackson28 @masterbee any issue with this?

Gruntfile.coffee
```diff
@@ -1324,7 +1337,27 @@ module.exports = (grunt) ->
 					base: "<%= coreDist %>"
 					message: ((
 						if process.env.TRAVIS_TAG
-							"Variant files for the " + process.env.TRAVIS_TAG + " maintenance release"
```
pjackson28: Should we just use "release" rather than "maintenance release" to cover non-maintenance releases or is there a separate message for those types of releases?

pjackson28:
Otherwise looks fine to me.

Gruntfile.coffee
LaurentGoderre: That is the message that is used for everything (it was the text in the release instructions that inspired most of this). We can change it but we will have to change in multiple places.

LaurentGoderre: @pjackson28 can we merge as is and perhaps revist the message in another issues since it spans multiple repos?

pjackson28: Yes that's fine. It can be dealt with separately.

Issue #6933, opened 2015-07-10
How to close lightbox overlay through javascript


jfbourgon:
I'm able to open modal inline content throught javascript using the following code
```javascript
$document.trigger( "open.wb-lbx", [
			[
				{
					src: "#inline_content_modal",
					type: "inline"
				}
			],
			true
		]);
```
I'm wondering how I could also invoke closing of the opened overlay throught javascript.

pjackson28:
You can use <code>$.magnificPopup.close();</code>. According to the Magnific Popup documentation it will close the currently open Lightbox.

jfbourgon:
I already tried this code in the first place but I was getting a javascript error about $.magnificPopup being undefined.

I then tried to import the library manually with something like:
```html
<script src="./theme-gcwu-fegc/unmin/js/deps/jquery.magnific-popup.js"></script>
```
I got ride of the error but the `$.magnificPopup.close();` call had no effect on the dialog box. 

Looks like I can't get an handle on `$.magnificPopup` from outside the WET plugin.

LaurentGoderre:
You shouldn't include the file manually. Is it possible that you your `$.magnificPopup.close();` was not actually executed when a lightbox was open and before the plugin was initialized?

jfbourgon:
Are you suggesting that the `$.magnificPopup.close();` directive is called too soon after the open call so this is why it is yet undefined.

I can try to but some delay between each call. How can I ensure the lightbox dialog is completed when this one is trigger by javascript?

LaurentGoderre:
So you want to close a lightbox without any user action?

jfbourgon:
Yes. To give a bit of context my web application is doing ajax call that may take few second to execute. I was expecting to open a modal lightbox with some kind of loading animation prior to execute the ajax call and then close it on success response . Maybe there could be other alternatives?

pjackson28:
If you want to close a modal then that is covered in the documentation with <code>popup-modal-dismiss</code>:
* Identifies the button that is to be used to close a lightbox behaving like a modal dialog.
* Add popup-modal-dismiss to the link or button that will be used to close the lightbox behaving like a modal dialog.

That is the recommended way of dealing with modal closure. If that isn't sufficient then you should wait on the <code>wb-ready.wb-lbx</code> event before trying to close anything (to ensure the Magnific Popup object is present). It will be fired after the Lightbox has finished initializing.

Here is some JS that should help:
```js
[...]

PR #6934, opened 2015-07-10
Lightbox: Fixed firing of the wb-ready.wb-lbx event when opening a Lightbox through JavaScript

pjackson28:
Related to #6933


LaurentGoderre:
Nice!

PR #6935, opened 2015-07-13
Session timeout: Add request method, additional data, custom callback

bmadzar:


LaurentGoderre:
Can you squash the commits and make the commit message the title of this PR?

bmadzar:
Done!

LaurentGoderre:
Wouldn't it be simple to keep the same call like this?

```
$.post( settings.refreshCallbackUrl, settings.additionalData, function( response ) {

}, "text" );
```

This should do the same thing while staying shorter

andre-gh:
It might be cleaner to set `context` to `$elm` and use `.done(keepaliveresponse)` and move the callback out of the function scope using `this` to access the `$elm`

bmadzar:
The main motivation for this pull request was to allow the request method used to be set as a configuration option. We currently have to apply a workaround on the server side to be able to handle this request using POST. This is the reason for `$.ajax` instead of `$.post`.

LaurentGoderre:
@bmadzar I understand. WHat I am saying is that you can pass data by simply modifying the existing $.post() call.

What is the use case to return results to the plugin? I doubt you could do anything with it without modifying the plugin.

bmadzar:
@LaurentGoderre What I meant was that it would be nice if the response checking was customizable; i.e. a custom callback function that would accept the response text from the server and return true or false to the plugin. The requirement of the server having to return the literal value "true" is a bit rigid.

We've faced three minor issues at various times while using this plugin:
1. POST is required, whereas for this type of request we would use GET (fixed in this PR)
2. No ability to send additional data to the server (fixed in this PR)
3. No ability to implement custom response checking (not sure how best to do this)

Hope that helps clarify the motivation a bit.

LaurentGoderre:
AAh so you want the ability to sometime use GET? If so that would make more sense.

LaurentGoderre:
You could move

```js
if ( response && response.replace( /\s/g, "" ) === "true" ) {
	$elm.trigger( resetEvent, settings );
}
```
 to a function called isSessionValid like such

```js
settings: {
	...
	isSessionValid: function( response ) {
		return response && response.replace( /\s/g, "" ) === "true"
	}
}
```

This way, you can override it with your own function. You would then change the callback to:

```js
// Session is valid
if ( settings.isSessionValid( response ) ) {
	$elm.trigger( resetEvent, settings );

// Session has timed out - let the user know they need to sign in again
} else {
...
}
```

bmadzar:
@LaurentGoderre I've updated the patch, adding an optional custom callback.

src/plugins/session-timeout/session-timeout.js
```diff
@@ -36,7 +36,13 @@ var $modal, $modalLink, countdownInterval, i18n, i18nText,
 		refreshCallbackUrl: null,	/* refresh callback if using AJAX keepalive (no default) */
 		logouturl: "./",			/* logout URL once the session has expired */
 		refreshOnClick: true,		/* refresh session if user clicks on the page */
-		refreshLimit: 200000		/* default period of 2 minutes (ajax calls happen only once during this period) */
+		refreshLimit: 200000,		/* default period of 2 minutes (ajax calls happen only once during this period) */
+		method: "POST",				/* the request method to use */
+		additionalData: null,		/* additional data to send with the request */
+		sessionaliveCallback:		/* callback function used to check the server response */
+		function( response ) {
```
LaurentGoderre: Can you make this be on the previous line and indent the content of the function

```
sessionaliveCallback: function( response ) {		/* callback function used to check the server response */
	return response.replace( /\s/g, "" ) === "true";
}
+		
```

bmadzar:
Fixed spacing

LaurentGoderre:
Good to go on my end

LaurentGoderre:
Oooh, actually, could you document the callback in your PR?

bmadzar:
Updated documentation, example page

LaurentGoderre:
This is good to go on my end. @masterbee @pjackson28 any issue with this?

pjackson28:
Looks good to me. Have all the normal use cases been tested to ensure they aren't breaking? Just want to make sure that this won't negatively impact existing implementations.

LaurentGoderre:
The existing unit tests all pass. So as far as we know it won't break anything.

Issue #6937, opened 2015-07-14
Scope issue with  'mouseleave' on the menu plugin


ptsimard:
We encountered a weird scope issue where by the time the setTimeout is executed, the event.currentTarget was changed to 'document' instead of the menu.

This caused all .active class on the page to be removed.

@LaurentGoderre proposed the following change which retains  the reference to the correct currentTarget.

From:
```javascript
$document.on( "mouseleave", selector + " .menu", function( event ) {

	// Clear the timeout for open/closing menus
	clearTimeout( globalTimeout );

	globalTimeout = setTimeout( function() {
		menuClose( $( event.currentTarget ).find( ".active" ), true );
	}, hoverDelay );
} );
```
To : 

```javascript
$document.on( "mouseleave", selector + " .menu", function( event ) {
	var $currentTarget = $(event.currentTarget);
	// Clear the timeout for open/closing menus
	clearTimeout( globalTimeout );

	globalTimeout = setTimeout( function() {
		menuClose($currentTarget.find( ".active" ), true );
	}, hoverDelay );
});
```

hsrudnicki:
@ptsimard @LaurentGoderre @pjackson28 : Please update the status of this issue since the done PR is merged and can be closed or otherwise?

LaurentGoderre:
This was fixed with #6938

PR #6938, opened 2015-07-14
Update menu.js to fix a changing currentTarget on mouseleave 

ptsimard:
We encountered a weird scope issue where by the time the setTimeout is executed, the event.currentTarget was changed to 'document' instead of the menu.

This caused all .active class on the page to be removed.

@LaurentGoderre proposed the following change which retains  the reference to the correct currentTarget.


ptsimard:
This should fix issue #6937

LaurentGoderre:
Any issue with merging this @pjackson28?

pjackson28:
Looks good to me.

PR #6942, opened 2015-07-16
Build: Use new Travis-CI Docker infrastrucutre

nschonni:
Moves the build worker to the new container based version http://docs.travis-ci.com/user/migrating-from-legacy/

Notes;
- Not ensure that the Bundler caching is working full, but it did seem to speed up my build by 30% after the first build primed the asset cache.
- Since we run `npm install` rather than `npm update` in the scripts, this will keep the versions at a set install version while the semver is still respected
- The global npm installs (grunt-cli, bower) don't appear to be cached right now, but since the location is dynamic based on the version of Node that nvm loads, I parked trying to address that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,3 +26,4 @@ notifications:
 cache:
   directories:
   - node_modules
+  - lib
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "0.10"
+  - "0.12"
 
 sudo: false
 
@@ -9,7 +9,6 @@ env:
     - TRAVIS_COMMIT_MSG="$(git log --format=%B --no-merges -n 1)"
 
 install:
-  - npm -g install npm@next
   - ./script/setup
   - bundle install
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: node_js
 node_js:
   - "0.10"
 
+sudo: false
+
 env:
   global:
     - TRAVIS_COMMIT_MSG="$(git log --format=%B --no-merges -n 1)"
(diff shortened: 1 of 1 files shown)


nschonni:
First run here was " ran for 14 min 56 sec"
Reset just to sample the difference

.travis.yml
```diff
@@ -20,3 +22,10 @@ after_success:
 notifications:
   irc:
   - "chat.freenode.net#wet-boew"
+
+cache:
+  bundler: true
+  directories:
+  - node_modules
+  - lib
+  - vendor/cache
```
LaurentGoderre: Does this folder exists anywhere? Also how does it know how to invalidate the cache?

nschonni:
After restarting "ran for 9 min 39 sec", so ~30% faster on subsequent builds to a branch. This may not speed up single build PRs though

.travis.yml
nschonni: You can see it here https://travis-ci.org/wet-boew/wet-boew/caches, or by selecting "Caches" under the Settings dropdown

LaurentGoderre:
The benefit of not caching is the build would catch problems introduce by children dependency updated., basically it's acting like a new developer setup would.

nschonni:
> The benefit of not caching is the build would catch problems introduce by children dependency updated

Yeah, I mentioned that in the description.

>> Since we run npm install rather than npm update in the scripts, this will keep the versions at a set install version while the semver is still respected

Changing to `npm update` would bump the out of date installs, but not the dependencies that are already ok.

LaurentGoderre:
But not the nested dependencies, which is usually where problems come from. I guess it is a minor issue because we still wipe our node_modules in the rare cases where this happens

Issue #6770, opened 2015-05-13
Update grunt-sass (maybe)


LaurentGoderre:
Libsass resolved the bug that was preventing us to move to the new version and node-sass 3.1.0 includes that version. Just need tro see if grunt-sass supports it now.

/cc @nschonni

LaurentGoderre:
@EricDunsworth @rlambert27 when this is done, you will be able to upgrade back to Node v0.12

EricDunsworth:
@LaurentGoderre Awesome, thanks :)!

hsrudnicki:
@EricDunsworth @rlambert27 @LaurentGoderre : Can this issue be closed?

LaurentGoderre:
No. It still needs to happen

PR #6944, opened 2015-07-17
Build: Bump grunt-sass to 1.0.0

nschonni:
I used KDiff3 to compare the dist outputs from the grunt/node-sass to the last build from the wet-boew-dist, and here is what I noticed:
- rgba(0, 0, 0, 0) to "transparent"
- single quotes to double in some cases
- selectors moved to new lines
- `@charset "UTF-8"` added automatically if it wasn't added manually
- strip leading zeros from values
- second green digit rounded down in some cases in the `.text-` classes in Bootrap. This may just be a precision difference for the `darken($color, 10%)`, since there is no difference to things like the grids
- removed media query blocks that contain only comments
- removed duplicate selectors when found on a class declaration. Ex: `.foo, bar, .foo` becomes `.foo, bar`

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "grunt-jscs": "^1.5.0",
     "grunt-mocha": "~0.4.12",
     "grunt-modernizr": "~0.5.2",
-    "grunt-sass": "^0.14.1",
+    "grunt-sass": "^1.0.0",
     "grunt-saucelabs": "^8.5.0",
     "grunt-wet-boew-postbuild": "wet-boew/grunt-wet-boew-postbuild#v0.1.2",
     "grunt-wget": "~0.1.0",


nschonni:
This also means people can run the build with Node 0.10 , 0.12, or iojs since the old node-sass dependency only shiped binaries for 0.10.

LaurentGoderre:
:+1: Thanks for taking the time for making the diff. I think we should merge in v4.0.17 since v4.0.16 comes out next Tuesday and if this changes something, it doesn't give us much time to catch it.

LaurentGoderre:
Can you create a gist for the difference?

LaurentGoderre:
Fixes #6770

EricDunsworth:
Btw thanks :)!

LaurentGoderre:
Here is the diff https://gist.github.com/LaurentGoderre/a2521cf8c75ce336eb3e

LaurentGoderre:
Perhaps it's time to revisit those TODOs and try to implement `@extend` where we couldn't before.

PR #6945, opened 2015-07-17
Build: Fix regex when replacing ".min." assets

nschonni:
The regex exluded the jQuery CDN link when it addressed the issue
explained in #5734. When the other method of using `environment.suffix`
was removed, there was no more replacement catching the jQuery links.


LaurentGoderre:
Are you sure this works??? I don't see how adding github int he regexp fixes the google cdn link. Also this fix needs to be applied to all themes so perhaps it's time to move to the repo that was created for this task.

nschonni:
The previous fix to exclude the references to the GitHub references also catches the Google CDN since it excludes at just the protocol level. This just makes it more specific to the case it was trying to avoid.
As mentioned, this wasn't an issue because it was covered by an alternate lookup for the suffix.

PR #6970, opened 2015-07-23
Build: Optimized the uglify task and added sourcemap for the javascript

LaurentGoderre:
/cc @nschonni

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -385,6 +385,7 @@ module.exports = (grunt) ->
 
 			i18n:
 				options:
+					banner: ""
 					process: ( src, filepath ) ->
 						lang = filepath.replace grunt.config( "coreDist" ) + "/js/i18n/", ""
 						# jQuery validation uses an underscore for locals
@@ -711,12 +712,14 @@ module.exports = (grunt) ->
 
 		# Minify
 		uglify:
+			options:
+				preserveComments: (uglify,comment) ->
+					return comment.value.match(/^!/i)
+
 			polyfills:
 				options:
-					report: "min"
 					banner: "<%= banner %>"
-					preserveComments: (uglify,comment) ->
-						return comment.value.match(/^!/i)
+					sourceMap: true
 				expand: true
 				cwd: "<%= coreDist %>/js/polyfills/"
 				src: "*.js"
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
Any issue with this @masterbee @pjackson28 ?

pjackson28:
@LaurentGoderre Are there any differences in the output, particularly with the comments?

LaurentGoderre:
Here is the diff. It  actually removes the double header in the i18n file and it adds a comment for the source map to work at the end of the file.

https://gist.github.com/LaurentGoderre/f4ebd57c7a59e41a0300

LaurentGoderre:
This will help us help other developers because they automatically kick in when using dev tools in all modern browsers, allegedly even IE11. This means we won't necessarily to ask people to switch to unmin, the dev tool will correctly report the right error and line number

Gruntfile.coffee
```diff
@@ -711,12 +712,15 @@ module.exports = (grunt) ->
 
 		# Minify
 		uglify:
+			options:
+				report: "min"
```
nschonni: I think this can be dropped now since it looks looks like it's the default https://github.com/gruntjs/grunt-contrib-uglify#report

LaurentGoderre: I wish there was an actual output but I agree

LaurentGoderre:
So is everyone ok with merging this?

pjackson28:
Could you post the comment blocks for each of the files (rather than just the diff)? Unsure if the comment blocks were right in the first place. For instance it looks like the Bootstrap license block is now missing from theme.min.css. Want to make sure we are complying to all the licenses (also please make sure to fix theme.min.css either here or in another PR).

LaurentGoderre:
Should the javascript use the bootstrap license?

pjackson28:
Not the JavaScript, since we don't use any JS from Bootstrap, but we need to make sure any license info from 3rd party JS is included (which I think is just Modernizr in the JS). The Bootstrap license needs to be in the CSS though, since we are certainly using the Bootstrap CSS.

LaurentGoderre:
None of this affect this PR though. I only touch our own files. Can we merge this and open separate issues for those?

pjackson28:
If this PR only affects the JS, then the CSS can be dealt with separately. It doesn't look like any 3rd party JS licenses are being stripped out by the build, only in the CSS (although not as a result of this PR).

PR #6973, opened 2015-07-24
Bump autoprefixer & fix linear-gradient errors

nschonni:
Bumping the dependency highlighted errors in the polyfills linear-gradient syntax. This was likely caused during my original conversion of the upstream CSS to SCSS.

The proper result can be viewed in IE and iOS Safari, since they are the last browsers to require the polyfill.


LaurentGoderre:
:+1:

Issue #6981, opened 2015-07-27
Major i18n overhaul


LaurentGoderre:
The way we handles translation makes it very difficult for variants to manage those translations. To better empower variant developer we need to revisit the way we manage our translations. For this, we should implement #5894 and #6003.

As part of this, we can look at translation services for FOSS that would allow us to manage our strings more efficiently and would allow variant to reuse the work done instead of having to redo it all.

/cc @masterbee @nschonni @pjackson28

nschonni:
Cross linking #2094, #2986, and #6726

LaurentGoderre:
Why did you add the Chrome tag?

nschonni:
Hit the enter key when using the "Labels" filter so I removed it right after.

LaurentGoderre:
Ah ok

LaurentGoderre:
Transifex has a nice client for synchronizing translations. Only downside is that it's in Python.

hsrudnicki:
Please continue your thought(s) and update this issue status whether this issue with others related can be closed or otherwise?

hsrudnicki:
Further, please update this issue's status whether this issue can be closed or otherwise? Thx

hsrudnicki:
Based on the above, the support PR is already closed but there are other referenced issues, lingering for over 4 months, to be resolved/fixed with this issue or if otherwise?
Either way, please update this issue's status, as well the referred issues, thank you

hsrudnicki:
Is this still an issue? Please update the status of this, whether this can be closed or otherwise? Thx

PR #6982, opened 2015-07-27
[WIP] i18n: Flatten the i18n file

LaurentGoderre:
Supports #6981


LaurentGoderre:
I am wondering if we should merge this..... If we intend to move to JSON, we can still have nesting I guess but I have never seen nesting in i18n anywhere else.

PR #6951, opened 2015-07-17
CSS: Made details/summary look clickable.

thomasgohard:
Fixes https://github.com/wet-boew/wet-boew-styleguide/issues/222.

PR #6989, opened 2015-07-28
CSS: Fix the presentation of details/summary in the menu.

### LINKED ISSUE #6984: Issue with the mobile menu expand/collapse styles
expandcollpasemobilemenu

thomasgohard:
Fixes #6984.

### DIFF
--- a/src/plugins/menu/_base.scss
+++ b/src/plugins/menu/_base.scss
@@ -91,6 +91,20 @@
 			z-index: -1;
 		}
 	}
+
+	details {
+		border: 0;
+
+		summary {
+			border: 0;
+			color: #fff;
+
+			&:focus,
+			&:hover {
+				text-decoration: none;
+			}
+		}
+	}
 }
 
 %menu-mb-pnl-margin-bottom-5px {


LaurentGoderre:
Can't we adopt the Bootstrap way instead?? This looks like the WET3 days of fighting for specificity. We should have a class to add for content expand/collapse (like bootstrap tables) so that we don't have to reset a plugin in another plugin

thomasgohard:
@LaurentGoderre the appearance given by #6951 should be the default appearance for the details/summary, otherwise the details/summary is unusable. It should not be necessary to add a class to get that appearance.

So the menu plugin is in effect overriding the default appearance for the details/summary.

@masterbee and I will be reviewing all the CSS in the next few weeks. If there's a cleaner way to do this, we'll make the change then.

LaurentGoderre:
It should be the default appearance in the content area but if another plugin uses it, they will have to undo everything making it very counter-productive. The table class in Bootstrap should be the default for actual table they are smart in knowing that blindly applying styles to an element can have significant impact.

hsrudnicki:
Please continue and update this issue status whether this issue can be closed or otherwise?

EricDunsworth:
@LaurentGoderre Wouldn't something along the lines of a "details" class work better in the context of WET 4.1-5.0 as opposed to a 4.0 maintenance release?

Suddenly introducing the need for a class to induce desired default styling would make life difficult for pre-existing 4.0 implementers. IMO, updating to the latest maintenance should generally just be a matter of dropping in a new /dist/ folder. Retroactively adding a class into tons of pages (potentially hundreds/thousands) for what should be a simple update isn't realistic.

andre-gh:
I have decided to merge this version of this fix as suddenly adding a class to get released functionality back to a maintenance breaks our versioning policy.

@LaurentGoderre Could you investigate #7137 and add a fix for that in the mean tiem as well?

We should going forward add any changes like this in classes so that this never happens again and should fix existing instances like this in `4.1` 

#7023 needs to happen in 4.1 along with any other problems like this

Merging this is only a temporary fix until v4.1

andre-gh:
Sorry I meant @thomasgohard or @LaurentGoderre on investigating #7137.  If not, I can investigate the issue.

Issue #6232, opened 2014-11-03
WET 4 - Session login/logout Guidance


chyne:
Is there any guidance for WET 4 related to positioning of login/logout links, as there was for WET 3.1?

shawnthompson:
@pcwsmith would adding this to the GCWeb templates be a good thing?

shawnthompson:
@pcwsmith or @DK-TBS is there anything in the works for something like this?

DK-TBS:
Nothing in the works on our end.  We've had a look Thomas' sketches over here, and everyone thinks it's tops.  No complaints from TBS!  Run with it  :)

RobJohnston:
So using existing styles, what's the best markup to accomplish this today (or close to it)?  Some thoughts...

```html
<div class="well bg-info full-width text-right">
        Signed in as User Name
        <button class="btn btn-default" type="button">Account settings</button>
        <button class="btn btn-primary" type="button">Sign out</button>
</div>
```
or

```html
<div class="row bg-info">
    <div class="pull-right mrgn-tp-md mrgn-rght-lg mrgn-lft-md">      
        <ul class="list-inline">
        <li>Signed in as User Name</li>
        <li><button class="btn btn-default" type="button">Account settings</button></li>
        <li><button class="btn btn-primary" type="button">Sign out</button></li>
        </ul>
    </div>
</div>
```

I think the first looks better but the second responds more like the screenshot.

robwloch:
Just wondering if the Logout button should maybe have the colour btn-danger instead of btn-primary. Thoughts?

thomasgohard:
@RobJohnston I'm out of town right now. I have the markup on a separate computer I don't have with me. I'll post it as soon as I'm back.

@robwloch `btn-danger` should be reserved for truly dangerous actions. The sort that you cannot reverse. I think `btn-primary` is appropriate here as it is the primary action for that form.

robwloch:
@thomasgohard To me "Sign out" does seem to fit the criteria for a dangerous action especially inside a web application where you could potentially lose all of your work; another candidate would be a "Delete" button perhaps. In the mock-up screenshot above wouldn't the "Submit" button be the appropriate primary action button for that page? Or is it per form where the page contains multiple forms?

Thank you, Rob.

thomasgohard:
@robwloch applications should warn you if you're going to lose your data from a particular action. Signing out isn't irreversible like delete it. You can sign back in.

[...]

PR #6990, opened 2015-07-28
WIP: Added the sign-in/sign-off example

LaurentGoderre:
TODO:
- [x] Add i18n
- [x] Add the french example


site/includes/sign-on-off.hbs
```diff
@@ -0,0 +1,16 @@
+<div class="container">
+	<div class="row">
+{{#is signed "off"}}
```
thomasgohard: There should be no `off` mode here. The register/login interface is not in the sign-on information bar but directly in the content area.

LaurentGoderre: I think it makes more sense in the header sine it's not directly related to the content.

thomasgohard: The register/login interface is the content as it would be a login page, e.g., https://clegc-gckey.gc.ca/j/eng/l?ReqID=s25ee43853adc59e8b95f29aef5afdf398227277ea.

LaurentGoderre: That is not the interface...That is for application that provides links to the sign on interface. You even did it this way in #6232

P.S. your link doesn't work

LaurentGoderre:
signedoff
signedon

site/includes/sign-on-off.hbs
thomasgohard: Not sure why the link isn't working for you. Here's a screenshot:
screenshot-clegc-gckey gc ca 2015-07-28 15-27-33

And in #6232, I only include a state for signing out. That's because when signing in, the user should be presented with a sign-in page, not have to find the sign-in at the top of the page.

LaurentGoderre: That may not always be desirable or possible so my use case is still valid and implementers have been asking for just this for a long time. Seen the screenshot below.

LaurentGoderre:
@pjackson28 @masterbee any issue with this ?

LaurentGoderre:
Waiting for the translations mentioned in #7022

site/includes/sign-on-off.hbs
chrismajewski: The link changes, start with https://clegc-gckey.gc.ca/ and click forward.

I'm not sure precedent and modern design back the exclusive "in content" argument.  I grant there are some design flows where the site requires credentials to function, where you funnel users to a registration page.  The CKAN registry doesn't work like that, it's more like twitter or google.

CKAN allows for near full functionality as a guest and as such they don't need to be clobbered over the head with login information.  They should be able to seamlessly use the site without realizing you can login till they reach a point where it's required.  In those models it's nearly convention that the optional login information be top right.

Google:
image

Twitter:
image

I'd even prefer it further top then right.  Meaning if you take the canada.ca look I'd hope for top white bar and this intranet look top grey.  To the left side of the page or just left of the language.

image

image

What Laurent proposes may be the best compromise considering mobile, accessibility and other considerations.  His example could be better with smaller buttons but the least disruption to the page is to the left of the language and would be my preference.

One thing that is sure, we need this optional credential functionality as Laurent proposes.  How it's implemented seems to be the only question.

pcwsmith: Twitter and Google are single sign-on scenarios, where the user is globally signed in to the entire system. The GC ecosystem does not have that, you can only sign in to specific apps - and in the vast majority if not all scenarios, it is unfortunately impossible to carry over your signed in state from one app to another. So it's actually misleading to include the login/out controls up in the global header.

LaurentGoderre: That's a really good point!! My solution is probably the best compromise

LaurentGoderre:
Can we merge this?

andre-gh:
Very low risk, sure

PR #6993, opened 2015-07-29
Site: Removed unnecessary role attributes (banner, navigation, main and contentinfo) from HTML5 code.

EricDunsworth:
Prevents warnings in the W3C's NU HTML validator related to those roles serving the same purpose as the elements they've been set on.

Example of the warnings:
https://validator.w3.org/nu/?showsource=yes&doc=http%3A%2F%2Fwet-boew.github.io%2Fv4.0-ci%2Findex-en.html

### DIFF
--- a/site/layouts/default.hbs
+++ b/site/layouts/default.hbs
@@ -18,7 +18,7 @@
 	</li>
 {{/is}}
 </ul>
-<header role="banner">
+<header>
 	<div id="wb-bnr">
 		<div id="wb-bar">
 			<div class="container">
@@ -55,7 +55,7 @@
 		</div>
 	</div>
 {{#isnt sitemenu "false"}}
-	<nav role="navigation" id="wb-sm" data-ajax-replace="{{#if page.altMenu}}{{page.altMenu}}{{else}}{{#if site.menuRoot}}{{site.menuRoot}}{{else}}{{assets}}{{/if}}/ajax/sitemenu{{/if}}-{{language}}.html" data-trgt="mb-pnl" class="wb-menu visible-md visible-lg" typeof="SiteNavigationElement">
+	<nav id="wb-sm" data-ajax-replace="{{#if page.altMenu}}{{page.altMenu}}{{else}}{{#if site.menuRoot}}{{site.menuRoot}}{{else}}{{assets}}{{/if}}/ajax/sitemenu{{/if}}-{{language}}.html" data-trgt="mb-pnl" class="wb-menu visible-md visible-lg" typeof="SiteNavigationElement">
 		{{>sitemenu}}
 	</nav>
 {{else}}
@@ -63,7 +63,7 @@
 {{/isnt}}
 {{#isnt breadcrumb "false"}}
 	{{#isnt assets "."}}
-		<nav role="navigation" id="wb-bc" property="breadcrumb">
+		<nav id="wb-bc" property="breadcrumb">
 			<h2>{{{i18n "you-are-here"}}}</h2>
 			{{>breadcrumbs}}
 		</nav>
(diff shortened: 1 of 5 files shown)


LaurentGoderre:
I'm not sure that version is as up to date as the one at http://validator.nu

It may be a false positive

EricDunsworth:
IMO those warnings make sense. Setting roles on elements that already have the same semantic meanings doesn't achieve anything apart from better support for (really?) old browsers/assistive technologies.

There's a W3C working draft document out there called [Notes on Using ARIA in HTML](http://www.w3.org/TR/aria-in-html/). Section 2.8 specifically recommends using the W3C's version of the NU validator to check ARIA markup. Section 4 recommends against using ARIA roles redundantly.

Older versions of the aforementioned document used to [recommend scenarios in which ARIA roles should be used on elements that have the same semantic meaning](http://www.w3.org/TR/2014/WD-aria-in-html-20140626/#recommendations-table) to account for poor built-in support in browsers/ATs, but that's no longer the case now.

Lastly, I think the W3C's version of the NU validator has some customizations that go beyond the "normal" one. I've seen other W3C NU-specific errors/warnings in the past. I've also read some things in the past that indicated both variants of the NU validator are generally the same, but sometimes one variant implements new features before the other.

pjackson28:
Those roles are likely still needed as some screen readers do not respect the implied roles of the HTML5 elements and need them explicitly stated through WAI-ARIA. So taking those roles off could decrease screen reader support. That is why the best practice (at least for now) is to pair implicit roles with explicit roles.

EricDunsworth:
@pjackson28 Is it only older versions of the screen readers you're referring to that don't respect implied roles? If so, what will be the determining factor in deciding to remove redundant explicit roles down the line?

nschonni:
This was changed in https://github.com/validator/validator/issues/39, but that hasn't been released yet for our checking in `grunt-html` yet.

pjackson28:
@EricDunsworth Yes, it was older screen readers that had issues with just the HTML5 elements alone. We were just following the best practice from a few years ago (e.g., https://dequeuniversity.com/assets/html/jquery-summit/html5/slides/landmarks.html#best_practice_use_both). Although it is interesting that it is Steve Faulkner (The Paciello Group) that recently recommended that the explicit roles be removed (which caused the change in the validator).

@dfarough @mysticjeff @SBleeks What are your thoughts on this? When would it be safe to drop explicit roles where they are explicitly set (e.g., dropping role="navigation" from an HTML5 nav element)? Is there still a need to to keep the explicit roles now that they are being flagged as HTML5 validator warnings?

dfarough:
Although there are more options for free screen readers available these days, it is prudent to remember that many people may be forced to use older versions of this software due to language localization issues.  Therefore  these users may still need the benefits provided by this redundant role specification.

SBleeks:
Agreed.  Although dropping the role specification would satisfy our baseline testing, if it is not impeding accessibility or general user experience the benefit exists for some. In other words,  it is likely beneficial to users of older access technologies.

LaurentGoderre:
Closing since we should hold on to those roles for a little bit longer

sylus:
@LaurentGoderre as it has been a year any chance to take a look at this again? I only ask cause Drupal is now evaluating this.

https://www.drupal.org/node/2655764

EricDunsworth:
@sylus You probably already know this by now, but according to https://www.drupal.org/node/2862904#comment-12161565, all versions of IE are incapable of passing on implicit ARIA semantics for HTML5 elements to Windows - which negatively affects many screen readers. Thankfully, Edge is capable of passing on those semantics.

My guess is that it won't be practical to stop hardcoding redundant ARIA roles in WET's HTML markup until WET officially drops support for IE10-11. Since it might take years for that to happen, maybe there'd be value in implementing some kind of stop-gap solution in WET 4.1+?

I believe early versions of WET 4.0 claimed to only officially support IE8 when JavaScript was enabled, since at the time it was so far behind other browsers that it couldn't properly render 4.0's hardcoded HTML markup. Maybe something similar could be done with IE10-11 in WET 4.1+ with regards to ARIA roles? Like moving away from hardcoding redundant roles into HTML5 elements that have the same implicit ARIA semantics and using a JavaScript polyfill to automatically add in the implicit roles if IE is detected.

That way, WET 4.1+'s page sizes would be reduced and implementations of it would run into way fewer HTML validation warnings. Not to mention that implementers wouldn't need to worry about manually removing hardcoded ARIA role after IE support is eventually dropped.

LaurentGoderre:
WE'd have to test but it seems like a good idea to me!

PR #7009, opened 2015-08-03
Proposal to replace facebook feeds with facebook page plugin

andre-gh:
Facebook has removed the ability to read atom/rss feeds and the real Facebook API requires passing around a "secret" API Key.

This pull request prototypes replacing Facebook feeds with the pages plugin transparently.  This method is really intended for pages which only embed a single Facebook feed inside the feed as Facebook does not directly support multiple pages in a single page embed.

It should be possible to re-implment this such that multiple Facebook pages are automatically placed in tabs or that a small height is used in order to stack them if the demand is there.

It should also be possible to re-implement this as a standalone plugin and instead require folks migrate to it explicitly to avoid any issues like that.

embed

### DIFF
--- a/src/plugins/feeds/_base.scss
+++ b/src/plugins/feeds/_base.scss
@@ -15,6 +15,9 @@
 }
 
 .feeds-cont {
+	.fb-wrapper {
+		text-align: center;
+	}
 	&.waiting {
 		min-height: 100px;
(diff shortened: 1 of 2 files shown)


andre-gh:
Closing, going to open again as an isolated plugin

Issue #6888, opened 2015-06-23
Bug found in WET feed reader - Facebook


ehcee:
Hi,

I noticed a bug in the feed reader for the facebook component:
http://wet-boew.github.io/wet-boew/demos/feeds/feeds-en.html

By default the feed reader works as expected returning results that appear on the facebook wall news feed it's pointing to, however when a facebook page updates it's "Cover Photo", the entry that appears in the facebook wall for "Cover photo update" breaks the facebook feed reader and it returns nothing but the default filler text.

Facebook - Healthy Canadians
https://www.facebook.com/HealthyCdns
https://www.facebook.com/feeds/page.php?id=209857686718&format=atom10

Page with reader:
http://www.healthycanadians.gc.ca/drug-prevention-drogues/index-eng.php

As of writing this post the 2nd entry is a cover photo update, pointing the facebook reader to this page returns nothing but the filler text, if I change the default limit-1 (ie feed to just before the cover photo update) it will display properly.

I've tested this on a few other gov of Canada. social media feeds and have found consistent results, as soon as it see's a cover photo update on the news feed wall it does not work.

2nd entry as of writing this is a cover photo update and won't work
https://www.facebook.com/4CdnDiv4DivCA

No cover photo updates (as of writing this) and works fine:
https://www.facebook.com/AANDCanada

Wondering if this is something that can be fixed in a future update or it's out of WET templates control.

hsrudnicki:
@ehcee @LaurentGoderre : Can this issue be closed or otherwise, pls update this issue status?

LaurentGoderre:
No, this is still an issue

hsrudnicki:
again, please update this issue status and close when fixed, thx

andre-gh:
#7010 WIP

hsrudnicki:
Please update this issue's status again whether this issue, lingering for about 4 months, can be closed as resolved/fixed or otherwise?

hsrudnicki:
Based on the above, can be this issue closed?  Is this still an issue?

kodecount:
@masterbee This can probably be closed. But the Facebook code should be removed from the JavaScript.

PR #7010, opened 2015-08-03
Proposal for Facebook plugin

andre-gh:
As Facebook feeds functionality no longer functions, this plugin is a proposal to replace the functionality.  

This plugin uses Facebook embedded pages similar to the twitter plugin to allow embedding of Facebook pages which were previously used as feeds.

Updating the feed plugin to use the newer API is not suitable as it would require a private API key distributed for client side functionality.

My first revision of this attempted to integrate directly with the feeds plugin transparently but it did not integrate very well and I had concerns about the large visual change which would be placed on clients using the feed plugin for Facebook previously without notice.  Thus, I split it out.

Translation of docs still has to happen, however, the proposed functionality is all there.

Feeds documentation needs to be changed to point to the Facebook plugin.

fb

### DIFF
--- a/site/pages/docs/ref/facebook/facebook-en.hbs
+++ b/site/pages/docs/ref/facebook/facebook-en.hbs
@@ -1,0 +1,79 @@
+---
+{
+	"title": "Facebook embedded pages",
+	"language": "en",
+	"category": "Plugins",
+	"categoryfile": "plugins",
+	"description": "Helps with implementing Facebook embedded pages.",
+	"altLangPrefix": "facebook",
+	"dateModified": "2015-08-03"
+}
+---
+<span class="wb-prettify all-pre hide"></span>
+
+<section>
+	<h2>Purpose</h2>
+	<p>Helps with implementing Facebook embedded pages.</p>
+</section>
+
+<section>
+	<h2>Working example</h2>
+	<ul>
+		<li><a href="../../../demos/facebook/facebook-en.html">English example</a></li>
+		<li><a href="../../../demos/facebook/facebook-fr.html">French example</a></li>
+	</ul>
+</section>
+
+<section>
+	<h2>How to implement</h2>
+	<ol>
+		<li>Create an embedded page on the <a rel="external" href="https://developers.facebook.com/docs/plugins/page-plugin">Facebook Page Plugin</a> documentation.</li>
+		<li>Press the "Get Code" button and copy only the HTML portion in the second section of the popup</li>
+		<li>For each area that will display a Facebook embedded page, create a <code>div</code> element with <code>class="wb-facebook"</code>.
+			<pre><code>&lt;div class="wb-facebook"&gt;&lt;/div&gt;</code></pre>
+		</li>
+		<li>Inside the div paste the HTML code from the "Get Code" button in the earlier step</li></ol>
+</section>
+
+<section>
+	<h2>Events</h2>
+	<table class="table">
+		<thead>
+			<tr>
+				<th>Event</th>
+				<th>Trigger</th>
+				<th>What it does</th>
+			</tr>
+		</thead>
+		<tbody>
+			<tr>
+				<td><code>wb-init.wb-facebook</code></td>
+				<td>Triggered manually (e.g., <code>$( ".wb-facebook" ).trigger( "wb-init.wb-facebook" );</code>).</td>
+				<td>Used to manually initialize the Facebook embedded pages. <strong>Note:</strong> The Facebook embedded page will be initialized automatically unless the required markup is added after the page has already loaded.</td>
+			</tr>
+			<tr>
+				<td><code>wb-ready.wb-facebook</code> (v4.0.5+)</td>
+				<td>Triggered automatically after the Facebook embedded page initializes.</td>
+				<td>Used to identify where the Facebook embedded page was initialized (target of the event).
+					<pre><code>$( document ).on( "wb-ready.wb-facebook", ".wb-facebook", function( event ) {
+});</code></pre>
+					<pre><code>$( ".wb-facebook" ).on( "wb-ready.wb-facebook", function( event ) {
+});</code></pre>
+				</td>
+			</tr>
+			<tr>
+				<td><code>wb-ready.wb</code> (v4.0.5+)</td>
+				<td>Triggered automatically when WET has finished loading and executing.</td>
+				<td>Used to identify when all WET plugins and polyfills have finished loading and executing.
+					<pre><code>$( document ).on( "wb-ready.wb", function( event ) {
+});</code></pre>
+				</td>
+			</tr>
+		</tbody>
+	</table>
+</section>
+
+<section>
+	<h2>Source code</h2>
+	<p><a href="https://github.com/wet-boew/wet-boew/tree/master/src/plugins/facebook">Facebook embedded pages plugin source code on GitHub</a></p>
+</section>
--- a/src/plugins/facebook/facebook.js
+++ b/src/plugins/facebook/facebook.js
@@ -1,0 +1,59 @@
+/**
+* @title WET-BOEW Facebook embedded page
+* @overview Helps with implementing Facebook embedded pages.
+* @license wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
+* @author @pjackson28
+*/
+( function( $, window, wb ) {
+	"use strict";
+
+	/*
+	* Variable and function definitions.
+	* These are global to the plugin - meaning that they will be initialized once per page,
+	* not once per instance of plugin on the page. So, this is a good place to define
+	* variables that are common to all instances of the plugin on a page.
+	*/
+	var componentName = "wb-facebook",
+	selector = "." + componentName,
+	initEvent = "wb-init" + selector,
+	$document = wb.doc,
+	fbinited = false,
+
+	/**
+	* @method init
+	* @param {jQuery Event} event Event that triggered the function call
+	*/
+	init = function( event ) {
+
+		// Start initialization
+		// returns DOM object = proceed with init
+		// returns undefined = do not proceed with init (e.g., already initialized)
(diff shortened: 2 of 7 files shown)


site/pages/docs/ref/facebook/facebook-en.hbs
```diff
@@ -0,0 +1,79 @@
...
+});</code></pre>
+				</td>
+			</tr>
+		</tbody>
+	</table>
+</section>
+
+<section>
+	<h2>Source code</h2>
+	<p><a href="https://github.com/wet-boew/wet-boew/tree/master/src/plugins/facebook">Twitter embedded timeline plugin source code on GitHub</a></p>
```
nschonni: Twitter should be Facebook in the link text

src/plugins/facebook/facebook.js
```diff
@@ -0,0 +1,62 @@
...
+		// returns undefined = do not proceed with init (e.g., already initialized)
+		var ele = wb.init( event, componentName, selector );
+
+		if ( ele ) {
+
+			Modernizr.load(
+				{
+					load: [ "//connect.facebook.net/" + wb.lang + "_CA/sdk.js" ],
+					complete: function() {
+						initfb();
```
nschonni: Function could be inlined here since it isn't used outside this scope

src/plugins/facebook/facebook.js
```diff
@@ -0,0 +1,62 @@
...
+		// Start initialization
+		// returns DOM object = proceed with init
+		// returns undefined = do not proceed with init (e.g., already initialized)
+		var ele = wb.init( event, componentName, selector );
+
+		if ( ele ) {
+
+			Modernizr.load(
+				{
+					load: [ "//connect.facebook.net/" + wb.lang + "_CA/sdk.js" ],
```
nschonni: The `_CA` wouldn't apply for non-English/French Canadian content

src/plugins/facebook/facebook.js
```diff
@@ -0,0 +1,62 @@
...
+	selector = "." + componentName,
+	initEvent = "wb-init" + selector,
+	$document = wb.doc,
+	fbinited = false,
+
+	// Init facebook API if not inited already
+	initfb = function( ) {
+		if ( !fbinited ) {
+			window.FB.init( {
+				version: "v2.4"
```
nschonni: Looks like the `version` needs to be in the query string
```js
<div id="fb-root"></div>
<script>(function(d, s, id) {
  var js, fjs = d.getElementsByTagName(s)[0];
  if (d.getElementById(id)) return;
  js = d.createElement(s); js.id = id;
  js.src = "//connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.4";
  fjs.parentNode.insertBefore(js, fjs);
}(document, 'script', 'facebook-jssdk'));</script>
```
Not sure if this should be put somewhere as an option like `data-wb-facebook-version="v2.4"` or something

src/plugins/facebook/facebook.js
```diff
@@ -0,0 +1,62 @@
...
+
+		if ( ele ) {
+
+			Modernizr.load(
+				{
+					load: [ "//connect.facebook.net/" + wb.lang + "_CA/sdk.js" ],
+					complete: function() {
+						initfb();
+						window.FB.XFBML.parse( ele[ 0 ] );
+					}
```
nschonni: Think the tests are failing because you're missing the `ready` event like in https://github.com/wet-boew/wet-boew/blob/ce779c60525d0fe85ffb87c1d9be4f10e7877774/src/plugins/twitter/twitter.js#L39

src/plugins/facebook/facebook.js
andre-gh: `version` may be in the query string or at init time (depending on if you want to init fbxml on load or not, which we do not as we want to do it dynamically later as well).  The problem is that we can only init a single version per page and a data-web-facebook-version would be ambiguous if you have two on a page....

src/plugins/facebook/facebook.js
andre-gh: _US is just as useless is it not?  Facebook does not accept just "en"

andre-gh: (and WET is not inited with anything except en or fr)

andre-gh:
Updated with fixes

src/plugins/facebook/facebook.js
```diff
@@ -0,0 +1,58 @@
...
+
+		// Start initialization
+		// returns DOM object = proceed with init
+		// returns undefined = do not proceed with init (e.g., already initialized)
+		var ele = wb.init( event, componentName, selector );
+
+		if ( ele ) {
+			Modernizr.load(
+				{
+					load: [ "//connect.facebook.net/" + wb.lang + "_CA/sdk.js" ],
```
nschonni: `en_CA` actually redirects to `en_US` since there is no localization https://connect.facebook.net/en_CA/sdk.js
Might want to grab the protocol sniffing from here https://github.com/wet-boew/wet-boew/blob/master/src/plugins/twitter/twitter.js#L35 mainly applicable for local debugging.

src/plugins/facebook/facebook.js
```diff
@@ -0,0 +1,58 @@
...
+					complete: function() {
+						if ( !fbinited ) {
```
nschonni: This check and the variable shouldn't be needed anymore since the init is only called after the initial script load

andre-gh: What if someone triggers `wb-init.wb-twitter` on content loaded after the initial load?

andre-gh:
Updated and rebased

src/plugins/facebook/facebook.js
pjackson28: The wb.init function checks to see if the element has already been initialized. If it has already been initialized then it returns undef so the load doesn't proceed any further. You can see the code here: https://github.com/wet-boew/wet-boew/blob/master/src/core/wb.js#L165-L191

andre-gh: You might have multiple on a page though

andre-gh: You might have multiple on a page though, the FB init is global per page, the specific element init is not and needs to happen on a per element bases.  I can have two facebook page embeds on one page and they could be dynamically inited.  The FB init itself is an SDK init and only needs to happen once.

andre-gh: In other words, the `window.FB.init` call only needs to happen once and might error if it happens more than once.  Once that call is made then we are free to call `window.FB.XFBML.parse` as many times as we like on multiple facebook page embeds.

pjackson28: Then make sure to check to make sure it is only called once like we do with the i18n. See how we do it in the Overlay plugin: https://github.com/wet-boew/wet-boew/blob/master/src/plugins/overlay/overlay.js#L43-L53

andre-gh: This is exactly what I am already doing, or am I totally misreading my own code?

andre-gh: I have a variable called `fbinited` and I only call the facebook SDK init if that is false then set it to true, and then regardless of that I call the facebook pages DOM parser...or have I gone code blind?

pjackson28: I was just answering your questions regarding <code>wb-init.wb-twitter</code>. The <code>fbinited</code> approach is consistent with the i18n approach that I was referring to.

nschonni:
I took a kick at flattening the logic a little here https://gist.github.com/nschonni/554104336cf679d4f657#file-facebook-js

I'm not sure what to do about the language since the only "API" to get the available languages/format would be to parse https://www.facebook.com/translations/FacebookLocales.xml, but that is a CORS issue that I couldn't find a solution for (when requesting XML).

andre-gh:
Having trouble getting to gist, will look later today...unless it is blocked?  If it is, then it is a recent change.

andre-gh:
The language issue is more of an issue with WET or WET users not being more specific about the language code than a Facebook plugin issue specifically.

hsrudnicki:
Based on the above, can be this issue closed as resolved/fixed or if otherwise?  Please update this issue's status for this current year, thx

EricDunsworth:
Are there any plans to merge in this PR at some point, or enhance it if it still needs more work? If the only thing that's still wrong with it by this point is a lack of French translations, why not just add "needs translation" notices in the same manner as other WET pages that lack translations?

Some of my colleagues recently ran into issues attempting to to implement a campaign page on Canada.ca that was meant to use a Facebook feed. If this 1-2 year old PR's plugin had already been in place, things would've been much more straightforward for us.

The Canada.ca IA spec claims that social media content (with a specific mention of Facebook) can be included in campaign pages. I presume that covers Facebook feeds, especially since [GCWeb's campaign page template](http://wet-boew.github.io/themes-dist/GCWeb/campaign-en.html) already contains one (btw it stopped working months/years ago when Facebook removed its feed support).

WET already has an [embedded timeline plugin for Twitter](http://wet-boew.github.io/v4.0-ci/demos/twitter/twitter-en.html). This PR's plugin is basically a Facebook equivalent.

kodecount:
@EricDunsworth 

Here is some more information from https://github.com/wet-boew/wet-boew/issues/6888 and https://github.com/wet-boew/wet-boew/issues/7679

Facebook doesn't support RSS/XML feeds. So this PR should be used (no sense reinventing the wheel). Someone could cherry pick this if there any updates need.

Also before this is merged someone should strip Facebook from the old feed plugin as it is useless.

shawnthompson:
Working example:
https://shawnthompson.github.io/andre-gh-fbplugin/demos/facebook/facebook-en.html

gh-pages repo:
https://github.com/shawnthompson/andre-gh-fbplugin

shawnthompson:
I've pushed this change to my fork. Should I create a new PR from that branch instead?

https://github.com/shawnthompson/wet-boew/tree/andre-gh-fbplugin

lucas-hay:
Closing as this was sent as another PR (#7992) with update by @shawnthompson

PR #7014, opened 2015-08-04
Feeds: Escape quotes on properties.

### LINKED ISSUE #6997: Flickr API Returning Invalid JSON
The Flickr API is returning invalid JSON, it escapes single quotes. WET-BOEW v4.0.9 seemed to have dealt with this issue properly but not v4.0.15. Is there a work around until this is fixed?

See http://wet-boew.github.io/wet-boew/demos/feeds/feeds-en.html

compare with http://www.international.gc.ca/bci-test/results-resultats.aspx?lang=eng

andre-gh:
Fixes #6997 and a potential issue with youtube videos with `'` in their title.  Ideally these would be generating DOM directly but I am not sure if there are speed concerns there.

### DIFF
--- a/src/core/helpers.js
+++ b/src/core/helpers.js
@@ -1101,6 +1101,10 @@
 		} );
 	};
 
+	wb.propsafe = function( str ) {
+		return str.replace( /'/g, "&#39;" ).replace( /"/g, "&#34;" );
+	};
+
 } )( wb );
 
 ( function( $, undef ) {
--- a/src/core/helpers.js
+++ b/src/core/helpers.js
@@ -1101,7 +1101,7 @@
 		} );
 	};
 
-	wb.propsafe = function( str ) {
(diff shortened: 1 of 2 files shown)


LaurentGoderre:
I feel like the function name should be more descriptive of the use case. The use case is to process a valid JSON object and convert it in a way that is safe to include in the DOM. maybe something like wb.JSONDomStringify. Alternately, we clould extend the prototype of the JSON object to include a `JSON.stringifyDOM()` method.

andre-gh:
The use case is to process any string regardless of what it is and make it safe for use inside a `'` or `"` property..not just JSON.  We just happen to be using it for JSON currently.

LaurentGoderre:
maybe escapeAttribute would be a better name then

andre-gh:
Sounds great to me, renamed

LaurentGoderre:
:+1:

Issue #6991, opened 2015-07-29
Collapsible alert CSS issues


EricDunsworth:
Just wanted to bring up another CSS issue that started with PR #6951.

It doesn't seem to play nicely with collapsible alerts:
http://wet-boew.github.io/v4.0-ci/demos/collapsible-alerts/collapsible-alerts-en.html

@thomasgohard Any thoughts on what the "proper" look for these should be now? Should more overrides just be added to hide the borders?

LaurentGoderre:
One more reason why the details styles should only applied when the details class is added to the element like Bootstrap did.

hsrudnicki:
Please continue your thoughts, and update this issue status whether this issue can be closed or otherwise?

hsrudnicki:
Please update this issue's pull request and status, thx

hsrudnicki:
Based on the above, this issue, labeled "Going Stale" and closed as resolved/fixed, due to referenced PR and the other issue closed, or if otherwise?
If you think otherwise, please update this issue's status again, thank you

PR #7023, opened 2015-08-10
Prevent details/summary polyfill UX enhancements from breaking other plugins using it

### LINKED ISSUE #6984: Issue with the mobile menu expand/collapse styles

LaurentGoderre:
Fixes #6984 and #6991

### DIFF
--- a/src/polyfills/details/_base.scss
+++ b/src/polyfills/details/_base.scss
@@ -18,29 +18,31 @@ summary {
 }
 
 details {
-	border: 1px solid #ddd;
-	border-radius: 4px;
-
-	summary {
+	.details {
 		border: 1px solid #ddd;
 		border-radius: 4px;
-		color: #295376;
-		padding: 5px 15px;
 
-		&:focus,
-		&:hover {
-			background-color: transparent;
-			color: #0535d2;
-			text-decoration: underline;
-		}
-	}
-
-	&[open] {
 		summary {
-			border: 0;
-			border-bottom: 1px solid #ddd;
-			border-bottom-left-radius: 0;
-			border-bottom-right-radius: 0;
+			border: 1px solid #ddd;
+			border-radius: 4px;
+			color: #295376;
+			padding: 5px 15px;
+
+			&:focus,
+			&:hover {
+				background-color: transparent;
+				color: #0535d2;
+				text-decoration: underline;
+			}
+		}
+
+		&[open] {
+			summary {
+				border: 0;
+				border-bottom: 1px solid #ddd;
+				border-bottom-left-radius: 0;
+				border-bottom-right-radius: 0;
+			}
 		}
 	}
 
@@ -69,6 +71,7 @@ details {
 	}
 }
 
+//TODO: Move this to the tabbed interface CSS
 .tabpanels {
 	> {
 		details {


src/polyfills/details/_base.scss
```diff
@@ -18,29 +18,31 @@ summary {
 }
 
 details {
-	border: 1px solid #ddd;
-	border-radius: 4px;
-
-	summary {
+	.details {
```
pjackson28: Wouldn't this need an <code>&.details</code> instead? Otherwise this only applies to elements with the "details" classname inside of a details element.

pjackson28:
This would require all existing implementations to have the "details" class added. If this is the direction that we need to go, then this transition should be a coordinated rollout to avoid styles dropping off where they shouldn't. This should also be tested to make sure any other uses are not dependent on some of the styles covered by this (e.g., accordion, mobile menu). Of course there may be an opportunity to drop some extra CSS that was used to rollback some of these global changes.

@masterbee Suggest you be involved on this as it would mean work for implementers.

pjackson28:
@thomasgohard fyi since you were involved as well.

LaurentGoderre:
This would not drop anything except for the rare case where people are on v4.0.16 already. For everyone else, it would look like the default polyfill (made to look like other browsers).

hsrudnicki:
Please continue and update this issue status whether this issue can be closed or otherwise?

andre-gh:
Does this conflict with #6989?

andre-gh:
(And if so, has a decision been made on which implementation to go with).

LaurentGoderre:
Yes it does and no a decision has not been made but I strongly recommend the Bootstrap, no conflict approach. As plugins start using HTML5, we don't want to break someone else's plugin because of our aggressive styling.

andre-gh:
What I am confused about is: Does this change mean that old content would have to put a class in?

LaurentGoderre:
It means that if you have details/summary in the content, you will need to do the same you would do with a table and do `<details class="details"></details>`

andre-gh:
That is all well and good but that means anyone upgrading WET will unexpectedly have to update (often generated static content) to include that class, which is not acceptable IMO.  I totally understand your reasoning for not changing the default behavior of an HTML element and how it can cause messy problems with other content or plugins or external deps etc using details/summary, however, I personally can not justify merging this and I recommend against merging this unless we want to release a `4.1` release.

My personal recommendation is to wait until a 4.1 release is planned for this change.

See also: http://wet-boew.github.io/wet-boew/docs/versions/version-info-en.html

LaurentGoderre:
Except that it's only the last two maintenance release that have that behavior. The majority of site don't even have that behavior yet. The original commit should not have been merged in the first place.

LaurentGoderre:
Also, they will need to make that change at one point anyways. This doesn't break anything

andre-gh:
After reading through the original issue (wet-boew/wet-boew-styleguide#222) and considering down the road where things will head I believe it to be a bug that it was merged in the first place.

In my personal opinion we should not be making enhancements like that without an extra class, especially if those enhancements are changed again then all resets or devs making web applications and plugins will have to change their resets again and that is not somewhere we want to head.  This is especially true with something dynamic like details is where folks (we even do so for two different plugins) are going to re-purpose it a lot.

However, I have the following concerns still:

1) This breaks our versioning promise (although technically no HTML has to change, it does remove functionality we added and require you to add a class if you want it back)

2) Most content is going to receive no usability enhancement for a long time and this enhancement will be virtually useless.

3) If anyone was testing usability with this enhancement in place, their testing could be undone with a bug fix upgrade which should not happen imo.

I am not the final say on this, there are other folks who could still merge this, but that is where I land.

EricDunsworth:
Wasn't the original commit appropriate for a maintenance release since it only adjusted default CSS (no HTML5 markup changes)?

This PR's change would require HTML5 markup changes across almost every details implementation, which goes into minor/major release territory. When departments upgrade to "real" minor/major releases, they're coming into it with the expectation that HTML5 markup changes will be needed. But that's not the expectation when upgrading to the latest maintenance release. No one's likely to be very keen on upgrading to a maintenance release if it requires changing HTML5 code on upwards of hundreds/thousands of pages.

IMO the idea of a "details" class makes sense, but it's just too major of a change to sneak into a maintenance release.

andre-gh:
@EricDunsworth It should not have been merged because enhancing the style and usability in this particular way is a can of worms. If we want to change that usability enhancement again we need to then change every plugin which is resetting that again AND inform devs of any local usage of details that they might have to change theirs again.  This ideally would not be the case.

andre-gh:
But yes, markup changing changes belong in `4.1`

LaurentGoderre:
I don't think that constitute a markup change because 1) only the ones who are on the last two have such feature and 2) it's not removing any functionality or breaking anything. If they don't update their HTML, everything still works as expected, it just doesn't have a border.

EricDunsworth:
@andre-gh I agree that it's a can of worms, but technically it's appropriate according to the versioning page's maintenance version specs. Addressing conflicts between it and other features is still possible without requiring HTML markup changes (even though it totally bloats the CSS with overrides).

@LaurentGoderre Since the intent behind the original change was to improve the default visual usability of the details element, the only way to retain the improved default styling in your scenario would involve markup changes to hundreds/thousands of pages.

IMO it would be really weird if things got to a point where normal details implementations didn't look consistent across-the-board.

LaurentGoderre:
@EricDunsworth they had to go through pages to add the `table` class for tables to display as table. Imposing defaults on HTML component is asking for trouble because more often than not, these HTML5 components are used in other plugins which will break or are force to create overrides.

Bootstrap understands this and that's why you need to add a class to get the Bootstrap styles to your element, the detail/summary should be no different.

Finally, it's far easier to go in your content and search and replace `<details>` with '<details class="details>` or change your CMS than it is to try to fix all the widget's code that the details summary and override these styles.

andre-gh:
In the case of many sites the "simple search and replace" is through a whole ton of generated pages and old posts etc through various different apps.  I completely I understand your reasoning, however, The next release is not until December, and there have been two releases with the details enhancement so we should not break our release promise just because few people are using the enhancement yet.  For that reason such a change needs to wait until `4.1`

andre-gh:
See my comment on #6989

PR #7028, opened 2015-08-11
Assemble templates improvements

LaurentGoderre:
<del>TODO: Modify the themes assemble to not double include markup</del>

### DIFF
--- a/site/includes/brand.hbs
+++ b/site/includes/brand.hbs
@@ -1,4 +1,6 @@
-<a href="{{assets}}/index-{{#contains language "fr"}}{{language}}{{else}}en{{/contains}}.html">
-	<object type="image/svg+xml" tabindex="-1" data="{{assets}}/../{{site.theme}}/assets/logo.svg"></object>
-	<span>{{{i18n "site-title"}}}<span class="wb-inv">{{{i18n "comma-space"}}}</span><small>{{{i18n "site-tagline"}}}</small></span>
-</a>
+<div id="wb-sttl" class="col-md-8">
+	<a href="{{assets}}/index-{{#contains language "fr"}}{{language}}{{else}}en{{/contains}}.html">
+		<object type="image/svg+xml" tabindex="-1" data="{{assets}}/../{{site.theme}}/assets/logo.svg"></object>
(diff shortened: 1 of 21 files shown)


LaurentGoderre:
Just making a final check but this is ready to merge. Any issuw with it @masterbee @pjackson28 ?

pjackson28:
Could you post a diff of the before and after generated HTML in a gist? These changes should result in no changes to the resulting HTML so it is important that we confirm that.

LaurentGoderre:
Here are the diffs: 

Core: https://gist.github.com/LaurentGoderre/3e566e794971bf25b791?w=0
Base Theme: https://gist.github.com/LaurentGoderre/e59ff438d38d45cc99ba
GC Intranet: https://gist.github.com/LaurentGoderre/edb370f14f537d7f1ffc
GCWU Theme: https://gist.github.com/LaurentGoderre/f8473b6598f4b4263329
OGPL Theme: https://gist.github.com/LaurentGoderre/f5f6725bd6034b3b9134
GCWeb Theme: https://gist.github.com/LaurentGoderre/fdf8c41ce0a4006059e9

There is a markup change in GCWeb but it is non breaking.

LaurentGoderre:
Can we merge?

pjackson28:
The diffs are driving my browser into the ground. What is the markup change you are seeing? Could you please summarize it in this thread?

LaurentGoderre:
For all but the GCWeb theme, the only difference is in whitespace. I could try to generate the diff excluding the unmin folder.

For GCWeb,The h2 is move out of the div and under the section to bring it inline with the other themes. However it existing implementation, it makes 0 difference wheter it is one place or the other.

pjackson28:
Whitespace changes shouldn't be an issue but @masterbee will have to weigh in on the h2 move in GCWeb (his decision to make).

LaurentGoderre:
I can make a version that doesn't have those change in GCWeb and another PR to drop the custom file need to not have the markup change

LaurentGoderre:
@pjackson28 I updated the diff to exclude the unmin folders.

PR #7037, opened 2015-08-18
Build: Include jQuery for offline/internal sites or application

LaurentGoderre:


masterbee:
@LaurentGoderre this is a bit of a reverse direction since we decided in the beginning versions to stay with parameterless. While I am not opposed to it, is there a reason why it should be folded back in to the mix again.

LaurentGoderre:
Parameterless? We have intranet application that do not have intranet connectivity so it would be easier if the WET package included the right versions of jQuery instead of implementors having to figure out the version number and downloading it manually.

LaurentGoderre:
Btw, this does not switch the URL, it just includes the files in case you need them for site or apps that can't access the CDN

masterbee:
@LaurentGoderre - ok I can see the benefit there. The issue was more that adding pre-canned jQuery can sometimes create problems in term of applications and CMS since jQuery and developers would have to do that research regardless.

Issue #7030, opened 2015-08-13
Feeds - Internally hosted RSS/ATOM


ryanhyma:
Is it possible to use the feeds widget to format internally hosted RSS/ATOM feeds?

I think the answer is no, it seems to rely on a Google API that requires an external link that converts the XML into JSONP....

Would it be possible to refactor the code a little to allow for the ajax-data attribute to supply the URL of the internally hosted JSONP? It almost works as it is, but seems to set the feedType incorrectly.

I could probably make a pull request if this is a desirable feature....

Thanks.

LaurentGoderre:
The feeds widget is a mess anyways so even if it's patchwork, I don't think it matters that much. We will need to refactor it to make it more flexible.

ryanhyma:
Okay, the code didn't look that bad IMHO... (maybe it tries to do a little too much)

So if I take the time to at least set the feed type to generic when using the data-ajax parameter, would you accept a pull request?

LaurentGoderre:
We always consider PR. I think there is value in it until we make it flexible.

ryanhyma:
My pull request was accepted, so I guess this can be closed. If there are any other issues I can help with, please ask. I'm fairly competent. I just need some direction...

PR #7042, opened 2015-08-19
Allow generic JSONP call through data-ajax

ryanhyma:
Small change to allow internal JSONP calls using the plugin. 
Basically, it allows an internal JSONP source to be used as the datasource for the feeds. 

For reference: https://github.com/wet-boew/wet-boew/issues/7030


ryanhyma:
The pull request and build process has changed since my last commit, I hope I followed it properly...

PR #7044, opened 2015-08-20
Multimedia: Fixed the play overlay not disappearing when playing

### LINKED ISSUE #6956: Multimedia player: Play icon continues to overlay the video even after the video is playing
Reproduced on the live working examples in multiple browsers. The big play icon overlaying the video in the multimedia player temporarily turns into the spinning circular arrows while the video starts playing but then returns to the big play icon when the video is actually playing. Doesn't make a difference whether I used the play button in the control bar or the play icon overlaying the video to start the video. Regardless of how the video is started, the big play icon should disappear until the video is paused or stops.

@LaurentGoderre @crichard

LaurentGoderre:
Fix #6956

TODO: Add styles for IE8

### DIFF
--- a/src/plugins/multimedia/_base.scss
+++ b/src/plugins/multimedia/_base.scss
@@ -60,15 +60,6 @@ $mm-volume-thumb-width: 10px;
 	}
 }
 
-@mixin multimedia_overlay_hidden {
-	.display {
-		&:before,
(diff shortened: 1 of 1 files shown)


masterbee:
@LaurentGoderre : thanks this :+1:

PR #7051, opened 2015-08-24
Add release notes for v4.0.17+1

andre-gh:


andre-gh:
#7060 is a similar approach to this except it only updates links instead of treating it as a full release, not sure which is more proper for a  +1 release with no actual changes to core.

Issue #7067, opened 2015-08-28
Duplicate validation errors if using "require_from_group"


rachelywang:
I added a customized validation to ensure at least one field (telephone number, email) must be filled, using jQuery "require_from_group". Now the errors are displayed multiple times at the top of the page.

![snap5](https://cloud.githubusercontent.com/assets/12052626/9547845/15c9ed7a-4d6b-11e5-81b0-d17cc3b3ca78.jpg)

My code:

<div class="row">
            <div class="col-md-6 pull-left">
                <div class="form-group"><label for="submitter_telephoneNumber" class="required"><span class="field-name">Telephone (including area code): </span><br> <strong class="required">(Telephone OR Email is required)</strong> , both is preferred</label>
                     <input class="form-control required_group" type="tel" id="submitter_telephoneNumber" name="submitter_telephoneNumber" ng-model="submitter.telephoneNumber" placeholder="Telephone" maxlength="50"/>
                </div>

```
        </div>
        <div class="col-md-6 pull-right">
            <div class="form-group"><label for="submitter_email" class="required"><span class="field-name">Email: </span><br> <strong class="required">(Email OR Telephone is required)</strong> , both is preferred</label>
                 <input class="form-control required_group" type="email" id="submitter_email" name="submitter_email" ng-model="submitter.email" placeholder="email@example.com" maxlength="50"/>
            </div>
        </div>
    </div>
```

...

<script src="./wet-boew/js/wet-boew.min.js"></script>

<script>
wb.doc.on( "wb-ready.wb-frmvld", ".wb-frmvld", function( event ) {
    $(".required_group").each(function () { 
        $(this).rules('add', {
            require_from_group: [1, ".required_group"],
            messages: {
                require_from_group: 'An email or telephone is required.'
            }  
        });
    });
});

I'm using Firefox 40.0.3 and latest WET4 version 17.

I tried previous versions, and this issue appeared since version 12 when the fix of summary links was introduced.

cgroves:
Problem started occurring in version 12

wangyuxiao:
My code:

<div class="row">
            <div class="col-md-6 pull-left">
                <div class="form-group"><label for="submitter_telephoneNumber" class="required"><span class="field-name">Telephone (including area code): </span><br> <strong class="required">(Telephone OR Email is required)</strong> , both is preferred</label>
[...]

PR #7087, opened 2015-09-08
formvalid: Fix the issue with displaying duplicate validation errors …

rachelywang:
…if using require_from_group.

Issue #7067.


LaurentGoderre:
What is the reasoning behind this change?

rachelywang:
If you use jQuery validation "require_from_group", the validation errors will be displayed multiple times.

LaurentGoderre:
I know but how does breaking the caching of element fixes that?

rachelywang:
I don't even know why use the caching of elements initially (from release notes of version 12: to fix the issue with a click on a summary link being ignored when focus is in a form field). I'm thinking it may relate to racing...

pjackson28:
This may no longer be needed with the fix in PR #7161. Should check to see if that PR fixes the duplication issue before considering the merging in of this PR.

masterbee:
closing since this is duplication P.R and has gone stale

PR #7102, opened 2015-09-21
Tabbed interface: Opens panel when linking from a differnt page to an anchor within that panel and corrects the scroll position

### LINKED ISSUE #7084: Anchors not working properly in tabs
We are having issues when trying to link to an anchor in content that uses the wb-tabs functionality.

When clicking on the link, the page focus is in content way above the anchor itself. This is happening in both Firefox and IE 11. Note that when we hit enter on the URL in the address bar, the page focus is correct.

Thanks.

pjackson28:
Fixes #7084

### DIFF
--- a/src/plugins/tabs/tabs.js
+++ b/src/plugins/tabs/tabs.js
@@ -53,12 +53,12 @@ var componentName = "wb-tabs",
 		// returns DOM object = proceed with init
 		// returns undefined = do not proceed with init (e.g., already initialized)
 		var elm = wb.init( event, componentName, selector ),
-			hashFocus = false,
 			isCarousel = true,
 			open = "open",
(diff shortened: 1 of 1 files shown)


pjackson28:
@masterbee This PR passed the build checks. Should it be merged in v4.0.18 or wait until v4.0.19?

andre-gh:
This seems like a new feature rather than a bug fix, does this fix broken functionality or does it add new functionality?

pjackson28:
It addresses an issue when deep linking from a separate page that does not exist when deep linking from the same page, namely ensuring the linked to content is scrolled into view. It is an issue because when linking from another page, the browser resolves the scroll position before the loading has been completed (which is not an issue on the same page as the loading has already completed).

Issue #7104, opened 2015-09-22
Lightbox: Alternative image title (text with a hyperlink)


Christopher-O:
I am testing out a page that has alot of images with descriptions of what actions are going on in each image.  I am using the lightbox with the alternative image title function so that if a user clicks on the thumbnail image, the image is enlarged and the title contains the description of the action.  Therefore, the user can just cycle through the images with the arrows instead of opening/closing/opening/closing...each image to get a clearer picture.  I use the description of the image as the alternative text (adding a unique id to the div). 

The issue is when the description (Alternative image title) contains a URL (a tag), the lightbox thinks it's another image and tries to load it but it fails.  Obviously I do not want the hyperlink to load as an image, so I am wondering if there is way to tell the script to ignore that hyperlink, or an alternative way around this issue.   Thank you.

MaximMartel:
Can you please give us an example of the code?

Christopher-O:
Here is a basic overview of code that will create the issue
`<div class="wb-lbx lbx-gal">`
 `<p><a data-title="alt_img1" href="images/img01.jpg" title="Image title"> <img src="images/img01.jpg" alt="Image title" /></a></p>`
 `<p id="alt_img1">This is alternative text</p>`
 `<p><a data-title="alt_img2" href="images/img02.jpg" title="Image title"> <img src="images/img02.jpg" alt="Image title" /></a></p>`
 `<p id="alt_img2">This is alternative text</p>`
 `<p><a data-title="alt_img3" href="images/img03.jpg" title="Image title"> <img src="images/img03.jpg" alt="Image title" /></a></p>`
 `<p id="alt_img3">This is alternative text with a <a href="#">hyperlink</a></p>`
`</div>`

Please note that I realize that the alternative text is inside the lightbox, and is what is causing the issue, but due to the page structure, it needs to follow the image for reading order purposes.  What I am hoping is some sort of class that could be created so the gallery skips over the hyperlink, or some other way that the gallery ignores non image type hyperlinks.  

If that is not possible, I could place all the images together in one group and add the alternative text afterwards, but I am wondering if data-title is JAWS friendly and when the JAWS reader lands on the hyperlink surrounding the image, if they hear the title text, or the data-title text.

andre-gh:
Right, I understand now, give me bit to investigate how a downstream dependency of the lightbox functions and if there is a way to skip some links within the gallery target element.

hsrudnicki:
[...]

PR #7118, opened 2015-09-28
lightbox: Add a feature to skip gallery links.

andre-gh:
Use case as identified by #7104


andre-gh:
Can someone sanity check this for me?  @LaurentGoderre?

PR #7120, opened 2015-09-29
Tabs: Fix issue with closed tabs having height.

### LINKED ISSUE #7111: Tabbed Interface: Anchors away!
We are using the tabs on the Canada.ca site and we are seeing a bug with using anchors on a second tab.

Here's a sample of the markup we are trying to use:
http://shawnthompson.github.io/wet-boew/unmin/demos/tabs/tabs-en.html

If you switch to the second tab and click on one of the letters anchors (W), the content above the anchor disappears into the tab.

image

The first tab panel works fine.

cc/ @therealolivergross

andre-gh:
Fixes #7111 although I do not really understand why the browser is scrolling the way it does when the other tab has height.  The reverent spec is http://www.w3.org/TR/cssom-view/#scroll-an-element-into-view.

### DIFF
--- a/src/base/transitions/_base.scss
+++ b/src/base/transitions/_base.scss
@@ -104,9 +104,10 @@
 		animation-duration: 125ms;
 		animation-name: fadeout;
 		z-index: -1;
+
 		&.noheight {
-			max-height: 0;
 			animation-name: fadeoutnoheight;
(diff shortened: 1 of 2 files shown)


andre-gh:
Can someone sanity check this for me?  @LaurentGoderre

PR #7129, opened 2015-10-01
Make secnav selector less restrictive (menu.js)

ptsimard:
Some systems (like Drupal) need to have DIVs wrapping the secondary menu UL element. This makes the selector less restrictive. Otherwise the secondary menu will not get populated in the mobile menu.

### DIFF
--- a/src/plugins/menu/menu.js
+++ b/src/plugins/menu/menu.js
@@ -260,7 +260,7 @@ var componentName = "wb-menu",
 			// Add the secondary menu
 			if ( $secnav.length !== 0 ) {
 				allProperties.push( [
-					$secnav.find( "ul > li > *:first-child" ).get(),
+					$secnav.find( "ul" ).filter( ":not(li > ul)" ).find( " > li > *:first-child" ).get(),
 					"sec-pnl",
 					$secnav.find( "h2" ).html()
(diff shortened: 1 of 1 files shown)


ptsimard:
Recommended by @sylus

andre-gh:
Sorry for the delay, there are just a few priorities here which came up.  I am just trying to determine if that selector was restrictive for a reason, I had a thought about nested ULs and need to verify that works as expected.

andre-gh:
@ptsimard Yeah, this selects differently.

With the secnav example:

         $('#wb-sec').find( "ul > li > *:first-child" ).length
         9
         $('#wb-sec').find( "> ul > li > *:first-child" ).length
         1

I think maybe you want (Is there a selector to do this?)

       $('#wb-sec').find( "ul:not(ul ul) > li > *:first-child" ).length
       1

LaurentGoderre:
Actually, the line would be the following, as demonstrated by my pen

```javascript
$('#wb-sec').find("ul").filter(":not(li > ul)").find(" > li > *:first-child" )
```

http://codepen.io/LaurentGoderre/pen/PPjmpx

sylus:
Woot thanks everyone!

PR #7185, opened 2015-11-06
Fixed various JSCS error not catched because of travis cached modules

LaurentGoderre:
/cc @shawnthompson

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -3,7 +3,7 @@
 	// Preset overrides
 	"validateLineBreaks": null,
 	"requireCamelCaseOrUpperCaseIdentifiers": null,
-	"requireCapitalizedComments": null,
+	"requireCapitalizedComments": false,
 	"es3": true,
 	"maximumLineLength": 500,
 	// Additional preset tightening
--- a/src/plugins/charts/charts.js
+++ b/src/plugins/charts/charts.js
@@ -183,7 +183,7 @@
 									parseFloat( cellValue.replace( /(\d{1,3}(?:(?: |,)\d{3})*)(?:(?:.|,)(\d{1,2}))?$/, function( a, b, c ) {
 										return b.replace( / |,/g, "" ) + "." + c || "0";
 									} ), 10 ),
-									cellValue.match ( dataCellUnitRegExp )
+									cellValue.match( dataCellUnitRegExp )
 								];
 							}
 						}
--- a/src/plugins/deps/geomap-lib.js
+++ b/src/plugins/deps/geomap-lib.js
@@ -1347,7 +1347,7 @@ var componentName = "wb-geomap",
 											atts = {};
 											for ( name in layerAttributes ) {
 												if ( layerAttributes.hasOwnProperty( name ) ) {
-													atts[ layerAttributes[ name ] ] = $row.find ( name ).text();
+													atts[ layerAttributes[ name ] ] = $row.find( name ).text();
 												}
 											}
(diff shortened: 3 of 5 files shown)


LaurentGoderre:
The build is failing because of the cached module

masterbee:
@LaurentGoderre, you build has an error in the JSCS file. I agree with the following https://github.com/jscs-dev/node-jscs/issues/338 thread that mentions exactly the issue you have in that setting ```false``` actually trips up the rule set. It has to be set to ```null``` to be false. I did an earlier PR yesterday when I noticed the build was failing from code style issues.

Thank you for this.

LaurentGoderre:
I wish you had read the comment......this is caused because Travis caching the modules, my PR does build when npm install is run instead of using the outdated cached modules

andre-gh:
@LaurentGoderre I am out on the loop on this

> my PR does build when npm install is run instead of using the outdated cached modules

Do we have to wait for something to uncache?  What do you mean?

andre-gh:
@LaurentGoderre Could you update me on what is going on with this?  Last I heard on Thursday the gcweb build was failing for odd reasons.  It looks like @masterbee tried to manually patch the lint failure out by fixing the lint issue (which just hides the problem under the rug to appear a different day).  Does this fix the issue or not?  What is the issue? Why did @masterbee close it?

andre-gh:
@LaurentGoderre Ah, I understand now, @masterbee fixed the issue in his own PR.  What is this about cached modules though, is something else broken?

andre-gh:
@LaurentGoderre Also, do you know why this issue suddenly came up, did jscs update?

LaurentGoderre:
https://github.com/wet-boew/wet-boew/commit/179feecb0603b104c20685309384d766e4a3a9f8

LaurentGoderre:
@andre-d I guess @masterbee did not read the subtext about the real issue and duplicated half of my PR to fix the apparent problem. However, the build will still fail for all developers who clone a fresh copy of the repo and do `npm install` Basically, the symptoms were fixed but not the real problem and it's going to reoccur

andre-gh:
@LaurentGoderre Ok, but why does changing null to false fix the real issue? What is the real issue?  You didn't change anything else different.

LaurentGoderre:
`null` doesn't do anything in grunt-jscs version 2 and high. You are all using an older version of the plugin, but the second you update, it's going to complain about 50 issues. The same problem happens with Travis. It used to do a clean install, now someone set it up to cache the node_modules that's why you have 2 different behavior on WET and on GCWeb

andre-gh:
@LaurentGoderre Sorry for the confusion, I was not around for any of this discussion and @masterbee was misunderstanding.  Could you pull request just this one change isolated now?

andre-gh:
@LaurentGoderre So we have to change a setting in Travis to not cache?

LaurentGoderre:
Yep, although @nschonni might oppose to doing so

andre-gh:
@LaurentGoderre If it is breaking the build twice somehow, there is not much justification for leaving it enabled.

LaurentGoderre:
We could turn it on for only the bower deps since those don't change that often and don't have the same crazy nested dependencies

andre-gh:
@LaurentGoderre We are going to disable it, its breaking too much transparently

LaurentGoderre:
Remove this:

https://github.com/wet-boew/wet-boew/blob/master/.travis.yml#L25-L30

PR #7187, opened 2015-11-09
updated imagemin to new version to fix issue with GCWeb build

shawnthompson:
cc/ @andre-d

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "grunt-contrib-csslint": "~0.4.0",
     "grunt-contrib-cssmin": "^0.9.0",
     "grunt-contrib-htmlmin": "^0.4.0",
-    "grunt-contrib-imagemin": "~0.9.2",
+    "grunt-contrib-imagemin": "~0.9.4",
     "grunt-contrib-jshint": "^0.11.2",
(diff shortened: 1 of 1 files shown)


andre-gh:
Alright, tomorrow I will do a full out dist build and make sure we haven't broken anything with these recent dep updates.

LaurentGoderre:
Can we just get rid of that one.....I added that module way back thinking it would be cool to keep the original in the repo and compress at build time but this module has given us so much problem. We could have a folder for source images when we want to keep non compressed source and compress the other images once and not every build.

Just a thought

shawnthompson:
I'm ok with that

andre-gh:
@LaurentGoderre Updating this did not even fix the issue, I am concerned there is something else wrong...any ideas why when running through grunt init it does not pick up the PNGs suddenly?

LaurentGoderre:
Same thing as ther other issue.....the npm dependencies are being cached

andre-gh:
@LaurentGoderre How do I fix it?  Sorry, you keep saying that but I am not understanding the implications of that.

shawnthompson:
Can we just remove the caching option, @LaurentGoderre you said it wouldn't make that much of a difference anyways.

Issue #7191, opened 2015-11-12
Multi-day calendar events start one day late, duplicate on end date


nedyrb:
As seen on the example page:
https://wet-boew.github.io/v4.0-ci/demos/cal-events/cal-events-en.html

An event that starts on March 22, 2013 and ends April 26, 2013 will begin showing on the calendar on March 23rd, and will have a duplicate entry on April 26. Issue shows in Chrome Version 46 and IE Version 9.

PR #7192, opened 2015-11-12
fixes 7191

nedyrb:
Variable named 'date' was inside a loop and being overwritten in the next iteration of the loop. Fixed by setting date to date = new Date (date.getTime())


andre-gh:
Although this likely does fix #7191, the build is failing due to whitespace style issues, are you available to fix the style issues it is complaining about?

nedyrb:
Ah, excuse me for that. This is my first time using CI like this. It should be fixed now..

andre-gh:
Thanks, TravisCI is being slow today, I will take a look once it passes.

PR #7200, opened 2015-11-18
Fixed the offline jQuery version being switched

### LINKED ISSUE #7199: jQuery 1.11.1 and 2.1.4 in Wrong Directories
I'm in a position where I need to load jQuery locally instead of from google. I was under the impression that the copies of jquery in js/jquery/1.11.1 and js/jquery/2.1.4 could be used for this purpose. Is this correct? 

Looking at the headers of the source files, it looks like they've been reversed accidentally. ie.: 2.1.4 is in the 1.11.1 directory and vice versa. I will download my own copies and use those, but logging this in case others are having similar problems. (Wet 4.0.18)

LaurentGoderre:
Fixes #7199

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -1020,12 +1020,12 @@ module.exports = (grunt) ->
 					expand: true
 					flatten: true
 				,
-					cwd: "lib/jquery-oldIE/dist"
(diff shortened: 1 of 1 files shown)


andre-gh:
Ill merge once travis passes, thanks

PR #7216, opened 2015-11-23
Tabs: Fix tabs equalheight adding both on and off class at the same time

andre-gh:
I'm not even sure if equal heights is even working correctly with this fix, or if it ever worked.  I am looking into this tonight to see if I can make it work in earlier versions in any browser.

### DIFF
--- a/src/plugins/tabs/tabs.js
+++ b/src/plugins/tabs/tabs.js
@@ -676,9 +676,8 @@ var componentName = "wb-tabs",
 						}
 
 						// Enable equal heights for large view or disable for small view
-						if ( isSmallView !== $elm.hasClass( equalHeightOffClass ) ) {
-							$elm.toggleClass( equalHeightClass + " " + equalHeightOffClass );
-						}
+						$elm.toggleClass( equalHeightClass, !isSmallView );
+						$elm.toggleClass( equalHeightOffClass, isSmallView );
 
 						$summary.attr( "aria-hidden", !isSmallView );
 						$tablist.attr( "aria-hidden", isSmallView );


andre-gh:
@pjackson28 What is the intended behavior of the equal height?  I assume it is to make all the tabs have the same height in the large view, but I can not seem to make it actually do that in any of the earlier versions even when equal height inits correctly in the large view.

pjackson28:
@andre-gh The intent of equalheights, when applied to the tabbed interface, was to ensure the tabpanels were all the same height in "desktop" view (medium and up) to avoid "jumping" when going from tabpanel to tabpanel. It had to be disabled in "mobile" view because the tabbed interface turned into an accordion so the equalheights functionality was pointless and did more harm than good.

pjackson28:
@andre-gh Note that equalheights was a plugin built separately from Tabs (was intended for generic containers) so may or may not have been fully compatible with tabs (since they were developed and maintained separately). The code in question was just to deal with the issue that equalheights could be applied to an accordion in "mobile view" so the code just made sure that equalheights was disabled to avoid any negative effects.

andre-gh:
@pjackson28 Thanks, when I try older releases like 4.0.3 and 4.0.0 beta 1 the tabs appear to have the equal height class but do not appear to actually have the same height when switching between tabs.  We can look into fixing the equal height issue itself after the release on Monday, but I am trying to confirm this has not actually worked this long rather than a super recent issue which should be fixed it the next release...and if it has been broken that long.

pjackson28:
@andre-gh I don't recall there being any explicit testing for the equal heights functionality on tabs as the v4 equalheights plugin wasn't really designed for that use case. It may have worked correctly at one point (since we had to introduce the fix to prevent it from messing up the accordion) but don't have any data regarding which versions it worked correctly in (if any).

If I had to guess, I'd assume that the switch to CSS transitions for tabpanel changes (rather than JS animation) may have knocked out the equalheights (if it was working at some point).

andre-gh:
@pjackson28 Thanks, I will investigate this further for the next release since I can not actually make this work correctly (Although it appears to do something as this fix was fixing an issue triggering chrome small view to have a min height)

PR #7222, opened 2015-11-25
wb: Allow for specific directory paths via data attrs.

andre-gh:
This is the start of some work to test things for integration with frameworks like AEM better where they would prefer to store things in their own structure.

Allows for the following (for now undocumented) properties on the body element:

`data-wb-homepath`: Sets the root path to base all other folders off of
`data-wb-jspath`: Sets the path for all js to load from
`data-wb-csspath`: Sets the path for all css to load from

`data-wb-assetpath`: Sets the path to load assets from for plugins, it appears like this works correctly but css based assets will not change
`data-wb-templatepath`: Path for templates, not actually sure what uses this attribute
`data-wb-deppath`: Supposed to be the path deps are loaded from, plugins do not actually use this and must be a subfolder "deps" of the js path

@masterbee Do not merge, for your testing

### DIFF
--- a/src/core/wb.js
+++ b/src/core/wb.js
@@ -57,7 +57,7 @@ var getUrlParts = function( url ) {
 	 * @variable $src
 	 * @return {jQuery Element} of wb script element
 	 */
-	$src = $( "script[src*='wet-boew.js'],script[src*='wet-boew.min.js']" )
+	$src = $( "script[src*='wet-boew.js'],script[src*='wet-boew.min.js'],script[data-wb-core]" )
 		.last(),
 
 	/**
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
This is likely to break every other variant other than AEM on top of making it very difficult to maintain the CDN compatible approach we implemented. For example, this is probably going to create a lot of grief to the Drupal variant.

WET is supposed to be framework agonstic

andre-gh:
@LaurentGoderre What specifically would break?  This does not remove the existing method of reading the src path, this only allows one to define it manually as a override. Am I misunderstanding your concern?

LaurentGoderre:
Ok, as long as it's not the default!

andre-gh:
@LaurentGoderre nope, not the default, this only uses the given path if it exists and otherwise works the same as it already does and allows for pulling some strings for specific paths if you so choose.

LaurentGoderre:
I think the approach that other variants adopted are easier to manage than this however

andre-gh:
@LaurentGoderre Could you give me some details on that approach?  I am willing to change this approach, the only catch is that it can not be a build-time definition, and can not be a script blob defining overrides.

andre-gh:
@masterbee Updated to use the script element rather than the body, if the script element has data-wb-core it will use that (for situations where you do not have a wet-boew named js file).

The properties are now `data-home`, `data-js`, `data-css`, `data-asset`, `data-dep`, etc

If you wish to use the regular logic just from a different location you should use `data-home`, if you wish to point everything individually specifically you may use the specific attributes.

andre-gh:
@masterbee Centralized the path generation as per your request

LaurentGoderre:
The approach that the Drupal variant uses is in my mind much easier to manage by having a package for core and a package for each theme. From what I remember, AEM should be able to do this.

PR #7223, opened 2015-11-25
equalheight: Fix issues with multiple equalheight groups.  By @duprasf

### LINKED ISSUE #7153: Problem with Equal Heights
I wanted to create a equal height "menu list" similar to the one on [canada.ca](http://www.canada.ca/en/index.html), because my text is of different height I get a different visual result but if you inspect the DOM, [canada.ca](http://www.canada.ca/en/index.html) has the same problem I'm having so I'm going to use it in my example.

If you inspect the list of "Most requested services and information" items, you will see that you have (in large view) 3 colones using col-lg-4. The first three elements are ok, but the forth element has a ```max-height: 0px```, the following 3 elements (including element 7 on the next row) have ```max-height: 156px;``` and the last two have ```max-height: 133px;```.

Because the text is in the item with 0px max-height is smaller than the other items on it's row it's not a big problem but if it was longer floating problems would occur.

I had 4.0.16 but upgrated to 4.0.18 before opening this issue.

andre-gh:
Via @duprasf

Fixes #7153

### DIFF
--- a/src/plugins/equalheight/equalheight.js
+++ b/src/plugins/equalheight/equalheight.js
@@ -56,7 +56,7 @@ var componentName = "wb-eqht",
 	 * @method onResize
 	 */
 	onResize = function() {
-		var $elm, $children, $anchor, currentChild, childCSS, minHeight, i, j,
+		var $elm, $children, $anchor, currentChild, childCSS, i, j,
 			$elms = $( selector ),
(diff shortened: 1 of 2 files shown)


andre-gh:
@duprasf your code appears to be failing unit tests

PR #7233, opened 2015-12-02
Fix performance issue with Collapsible alerts

Pacoup:
This fixes a performance issue with Collapsible alerts: https://github.com/wet-boew/wet-boew/issues/7125

The main culprit was https://github.com/wet-boew/wet-boew/blob/master/src/plugins/collapsible-alerts/collapsible-alerts.js#L61

As far as I understand, the way this was written, wb.ready would be called on $document and loop several times through the init unnecessarily, but I'm not sure I understand why. This is basically copy pasted from another plugin, so if someone could explain it to me, that'd be great.

On L69, the code is further optimized to only set native details events on the selector and initEvent.

### DIFF
--- a/src/plugins/collapsible-alerts/collapsible-alerts.js
+++ b/src/plugins/collapsible-alerts/collapsible-alerts.js
@@ -16,7 +16,7 @@ var componentName = "wb-collapsible",
 	selector = "details.alert",
 	initEvent = "wb-init." + componentName,
 	$document = wb.doc,
-	details, key,
+	$elm, key, elm,
 
 	/**
 	 * @method init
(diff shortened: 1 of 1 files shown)


nschonni:
If you leave the variable name as `details` it would be easier to tell the actual change you're proposing.

Pacoup:
Oh, sorry, I figured I'd make it use the same style as the other plugins.

Pacoup:
I still think it should use the new variable name though. It's pretty obvious, and like I said, in line with other plugins.

I'm not gonna change it back.

Pacoup:
Think of it as, the variable name change is also part of the change I'm proposing.

Pacoup:
(I'll gladly split the request if you insist.)

Pacoup:
See: https://github.com/wet-boew/wet-boew/pull/7234

PR #7234, opened 2015-12-02
Fix and optimize Collapsible alert performance

Pacoup:
This fixes a performance issue with Collapsible alerts (https://github.com/wet-boew/wet-boew/issues/7125) and optimizes code on L69 so that it only sets native details events on the selected element and initEvent.

### DIFF
--- a/src/plugins/collapsible-alerts/collapsible-alerts.js
+++ b/src/plugins/collapsible-alerts/collapsible-alerts.js
@@ -16,7 +16,7 @@ var componentName = "wb-collapsible",
 	selector = "details.alert",
 	initEvent = "wb-init." + componentName,
 	$document = wb.doc,
-	details, key,
+	details, key, $details,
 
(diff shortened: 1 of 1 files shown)


andre-gh:
I believe the correct solution is to change 

       $document.on( "timerpoke.wb", function() {

to

     $document.one( "timerpoke.wb", function() {

The rest of the changes are unrelated to the performance issue

hsrudnicki:
Can be this pull closed due to pass results?

Pacoup:
I am no longer able to commit to fixing this issue myself. It also seems other people already know how to fix the issue, so I will let them commit a fix instead. You may do with this PR as you please.

hsrudnicki:
Based on the above, can be this pull closed or otherwise?

masterbee:
closed since P.R abandoned by owner due to time constraints

PR #7237, opened 2015-12-02
Feeds: Use Yahoo YQL to fetch rss.

andre-gh:
@thomasgohard @pjackson28 @zachfalsetto @lucas-hay @masterbee


masterbee:
@andre-gh thanks for this.

PR #7238, opened 2015-12-02
Remove long broken facebook demo from feeds.

andre-gh:


masterbee:
@andre-gh - Thanks for this

PR #7297, opened 2015-12-15
Form validation: Fixed undefined variable

codejunkyard:
$summaryContainer was declared but undefined. The error happened when the $errors.length was equal to 0.

### DIFF
--- a/src/plugins/formvalid/formvalid.js
+++ b/src/plugins/formvalid/formvalid.js
@@ -167,7 +167,8 @@ var componentName = "wb-frmvld",
 								prefixEnd = i18nText.colon + " </span>",
 								separator = i18nText.hyphen,
 								ariaLive = $form.parent().find( ".arialive" )[ 0 ],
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
I am not sure how solid this fixed and the line ends up being called twice (the original line is here) https://github.com/GabrielVautour/wet-boew/blob/master/src/plugins/formvalid/formvalid.js#L243
The reason for the delay is probably to solve some race condition.

I would say that this only addresses the symptoms and that this fix is not mature enough to be merged in.

GabrielVautour:
What I initially noticed was that the summary panel didn't reset properly. The last error reported didn't clear from the summary.

You can see the error in action here: https://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-en.html

GabrielVautour:
I just noticed that @rachelywang has already committed a fix for this on September 8th, but her pull request was never accepted to the main stream. https://github.com/rachelywang/wet-boew/commit/b1e10308e1d03338288fc1a452dc921efa557583

PR #7302, opened 2015-12-16
DataTable: Added functionality to update a datatable or add new data

LaurentGoderre:


src/plugins/tables/tables.js
```diff
@@ -147,6 +153,19 @@ var componentName = "wb-tables",
 // Bind the init event of the plugin
 $document.on( "timerpoke.wb " + initEvent, selector, init );
 
+// Bind the update event of the plugin
+$document.on( updateEvent, selector, function( event ) {
+	var $elm = $( event.currentTarget );
+
+	$elm.DataTable().destroy();
+
+	if ( event.newData ) {
```
andre-gh: newData is a parameter to clear the table? Why?

LaurentGoderre: For web application this is a common use case....

andre-gh: Right, but how do I actually add new data using your update?

LaurentGoderre: If you send data using `data-wb-tables` and then want to modify this data and add new one, you would use this.

shawnthompson:
cc/ @masterbee

Issue #7035, opened 2015-08-18
SASS compile errors in tabs plugin?


rmwbweb:
Was compiling WxT using the SCSS source grabbed from GitHub Aug. 17 (using the default theme.scss) and got the following two errors:

> error src/plugins/tabs/_base.scss 
> Line 430: ".wb-tabs.carousel-s2 figure img" failed to @extend "%tab-width-100-height-auto".
> The selector "%tab-width-100-height-auto" was not found.

and ...

> error src/plugins/tabs/_screen-xxs-max.scss
> Line 13: ".wb-tabs.carousel-s1 li[role="presentation"]" failed to @extend "%accessible-invisible-xxSmallView".
> The selector "%accessible-invisible-xxSmallView" was not found.

If I made those those statements !optional everything else compiled just fine.

Any hints as to the source of those errors?

pjackson28:
It doesn't look like either of those are needed and there seems to be no impact if they are removed (other than stopping the SCSS compile errors). So I suggest that they be removed.

I checked back through the history and looks like it has been an issue for a long time, just never flagged as an error. Perhaps a recent update to the SASS compiler is surfacing these as errors.

masterbee:
@rmwbweb  and @pjackson28 - second that after the review. @rmwbweb would you like to make a PR to remove them?

rmwbweb:
@masterbee :  Nor sure I should risk taking down canada.ca with my first pull request :-) But here's the diff for the two files in question with the complete removal that's working for us.

**./src/plugins/tabs/_base.scss:** (at line 428)
````
 428,431d427
 <
 <                       img {
 <                               @extend %tab-width-100-height-auto;
 <                       }
````

**./src/plugins/tabs/_screen-xxs-max.scss:** (at line 10)
````
 10,17d9
 <       &.carousel-s1 {
 <               li {
 <                       &[role="presentation"] {
 <                               @extend %accessible-invisible-xxSmallView;
 <                       }
 <               }
 <       }
 <
````

LaurentGoderre:
I am not getting any compile errors, which project are you trying to compile?

rmwbweb:
@LaurentGoderre  Initially, the WxT theme right out of the box, but anything that uses that .scss above gives me the error because those extended selectors don't seem to exist anywhere.

This may be because I am compiling with vanilla sass and it's picky in that situation.

LaurentGoderre:
[...]

PR #7308, opened 2015-12-18
Sass Compile Errors

kristremblay:
Applied fixes outlined in https://github.com/wet-boew/wet-boew/issues/7035#issuecomment-165894964.

### DIFF
--- a/src/plugins/tabs/_base.scss
+++ b/src/plugins/tabs/_base.scss
@@ -431,10 +431,6 @@ $tab-margin-width: 10px;
 		figure {
 			@extend %tabs-position-relative;
 			@extend %tab-background;
-
-			img {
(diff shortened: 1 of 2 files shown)


LaurentGoderre:
Before merging this, we should find out why Travis's and almost everyone else's don't fail on this

andre-gh:
Something lies in

> This may be because I am compiling with vanilla sass and it's picky in that situation.

I am not sure what that implies, but that sounds like the answer.

LaurentGoderre:
Ok, I reproduced by installing the ruby version of sass

LaurentGoderre:
Here is the bug report: https://github.com/sass/libsass/issues/871

andre-gh:
@LaurentGoderre Looks like our issue then, good to merge?

PR #7312, opened 2016-01-04
added important to border removals because of specificity restraint

shawnthompson:


LaurentGoderre:
tsk tsk tsk...slippery slope....

shawnthompson:
I was going to put it in the theme but there are a few places where we want to remove the borders of an element, like details / summary and another one I can't remember right now, which I overwrote in the theme CSS and removed the borders. I think this is a good case where `!important` is used properly.

LaurentGoderre:
Make sure it doesn't start proliferating ;)

masterbee:
@LaurentGoderre +1 love the proliferating..

shawnthompson:
Can someone merge this?

shawnthompson:
Thanks! :+1:

PR #7322, opened 2016-01-07
Details/Summary: Refactor to duplicate the spec API for the element

LaurentGoderre:
Added the open attribute and the toggle event

/cc @shawnthompson

### DIFF
--- a/src/polyfills/details/details.js
+++ b/src/polyfills/details/details.js
@@ -30,6 +30,11 @@ var componentName = "wb-details",
 
 		if ( summary ) {
 			details = summary.parentNode;
+			Object.defineProperty( details, "open", {
+				get: isOpen,
+				set: setOpen
+			} );
+			details.summary = summary;
 			summary.setAttribute( "aria-expanded", ( details.getAttribute( "open" ) !== null ) );
 
 			if ( !summary.getAttribute( "role" ) ) {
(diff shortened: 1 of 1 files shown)


shawnthompson:
:+1:

src/polyfills/details/details.js
```diff
@@ -30,6 +30,11 @@ var componentName = "wb-details",
 
 		if ( summary ) {
 			details = summary.parentNode;
+			Object.defineProperty( details, "open", {
```
andre-gh: Are we using this pattern anywhere else in the WET code?  It looks great, but it concerns me with the amount of pattern mixing the WET codebase already has that we would just start using a new pattern again.

andre-gh: Or was the intention of this PR to implement this property?

LaurentGoderre: The purpose of this PR is to make the use of details summary uniform so that an application could use the polyfill and native implementation in the same way and this pattern is the only way to do this. @shawnthompson already needs this

shawnthompson:
So what I was trying to do is change the text of the `summary` once the `details` has expended.

image

And change to this using a `data-summary-replace="Hide extra services and information"` attribute.
image

**This is not an approved change to the topic page, it was an option I was trying out and we decided to go another route but I'm sure we could be using this in other instances of `details/summary`.**

src/polyfills/details/details.js
andre-gh: Ok, sounds good

LaurentGoderre:
Can we get this merged then?

PR #7326, opened 2016-01-11
Details/Summary: Fixed the double description in the french demo that crashes the build

LaurentGoderre:


andre-gh:
Why did this just start crashing now?

LaurentGoderre:
Because a nested dependency of assemble was updated and probably started doing more through front-matter validation

andre-gh:
Right, that happens a lot

LaurentGoderre:
Weclome to dependency management.....the opposite problem is still way worse

PR #7327, opened 2016-01-11
Fix equalheights from data ajax, fix error in eqht with no content

### LINKED ISSUE #7324: Console error when using GoC Activites and Features
Hello,

A developer using the CDN pointed out a console error after upgrading to the latest version of WET (v4.0.20) on the CDN. The error is triggered when the GoC activities and initiatives is being pulled in from cdn.canada.ca.

URL with console error : http://www.esdc.gc.ca/includes/op-so/cdn-demo/content1-en.shtml
URL without console error (no GoC activities and initiatives pull) : http://www.esdc.gc.ca/includes/op-so/cdn-demo/content1-fr.shtml

I used the unminified WET files for easier debugging. Spoke with @EricDunsworth and maybe a "ajax-replace vs wb-eqht incompatibility" in v4.0.20.

Thought I would check on canada.ca (https://www.canada.ca/en.html) but it's not using data-ajax-replace to pull in the GoC activities and initiatives.

I couldn't find another website using v4.0.20 already to replicate the issue. I couldn't use the demo site because the feed is not used.

I think this may need to be resolved earlier than later given the console errors it will produce on websites running the v4.0.20.

Thoughts?

Roch

andre-gh:
Fixes #7324 @LaurentGoderre

### DIFF
--- a/src/plugins/data-ajax/data-ajax.js
+++ b/src/plugins/data-ajax/data-ajax.js
@@ -27,6 +27,7 @@ var componentName = "wb-data-ajax",
 	selector = selectors.join( "," ),
 	initEvent = "wb-init." + componentName,
 	updateEvent = "wb-update." + componentName,
+	contentUpdatedEvent = "wb-contentupdated",
 	$document = wb.doc,
 	s,
 
@@ -130,6 +131,8 @@ $document.on( "timerpoke.wb " + initEvent + " " + updateEvent + " ajax-fetched.w
 
 				//Resets the initial jQuery caching setting
 				jQuery.ajaxSettings.cache = jQueryCaching;
+
+				$elm.trigger( contentUpdatedEvent, { "ajax-type": ajaxType, "content": content } );
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/plugins/data-ajax/data-ajax.js
```diff
@@ -130,6 +131,8 @@ $document.on( "timerpoke.wb " + initEvent + " " + updateEvent + " ajax-fetched.w
 
 				//Resets the initial jQuery caching setting
 				jQuery.ajaxSettings.cache = jQueryCaching;
+
+				$elm.trigger( contentUpdatedEvent, { type: ajaxType, content: content } );
```
LaurentGoderre: I'm wondering if the type overwrites the event name. It probably is safer like: `$elm.trigger( { type: contentUpdatedEvent, content: content } );`

LaurentGoderre: If you want to send the ajax type, the parameter should be `ajax-type` instead of `type`

andre-gh:
@LaurentGoderre thanks, updated

LaurentGoderre:
Looks good to me

PR #7328, opened 2016-01-11
Toggle: Updated to the new details polyfill

LaurentGoderre:
Since the details polyfill now acts like a real details summary, cange the toggle logic and test to accomodate for the change


andre-gh:
Nice and clean, love it

LaurentGoderre:
Yeah, that is exactly the kidn of things I want to get rid of by making the details polyfill to spec

andre-gh:
All for having less special cases throughout all the WET codebase

PR #7333, opened 2016-01-12
menu: Inherit color value for summary.

### LINKED ISSUE #7330: wb-menu Colour
Between v.4.0.18 and v4.0.19, it appears that the colour of the summary item in the mobile menu has been changed to white, which makes it more difficult to read.  Here is the CSS from theme.css:

.wb-menu details summary, .wb-menu details[open] summary {
    border: 0 none;
    color: #fff;
}

And here is how the mobile menu appears:

wbmenu

Is this what the menu is expected to look like?

andre-gh:
@LaurentGoderre  Fixes #7330

### DIFF
--- a/src/plugins/menu/_base.scss
+++ b/src/plugins/menu/_base.scss
@@ -100,7 +100,7 @@
 
 			summary {
 				border: 0;
-				color: #fff;
+				color: inherit;
 
 				&:focus,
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
Interesting. It might not be enough but it's worth the try I think

Issue #7342, opened 2016-01-21
Polyfill: Date picker


E1D1:
Good day,

I have an issue with the Date Picker 

I use 2 date field on the page (start date and end date). When the page loads the first one(start) is showing and I'm hiding the "End Date" until the user select a "start date"

once the "start date" is selected (for example 2016-01-15), I then use JQuery Or javascript  to change the min attribute to min="2016-01-25" of the "end date"  to prevent the user to select an "end date" earlier than the "start date" 

Then I display the second calendar.

When I inspect the code with firebug I can see the attribute was added but when the "end date" calendar opens the days before the set min date are not greyed out (still available to select)

Any suggestions will be appreciated.

Thank You

E1D1:
The process describe above was working with the previous version of the calendar.

Thank You

LaurentGoderre:
For performances reasons, the min and max settings are cached. So changing the values afterhand is definitely and issue.

E1D1:
any other options / work around ?

LaurentGoderre:
This was not a documented use case so I can add it to the unit test and change the approach to support it.

E1D1:
Thank You

neoinsight:
@E1D1 We will be usability testing the revised WET datepicker again soon, and I am interested in having participants try your Start and End date solution. Can you provide a link to the page or is it within a protected application?  Thanks

E1D1:
It's within a protected application

neoinsight:
Ah ok, thanks.

LaurentGoderre:
I just added the test for this issue

hsrudnicki:
Does the test pass? Either way, please update the status of this issue, whether this issue can be closed as resolved/fixed or if otherwise? Thank you

E1D1:
I Will close this Since I approached my issue on a different angle 

Thank you

LaurentGoderre:
I still think this should be fixed

E1D1:
How can I test it

LaurentGoderre:
It has been fixed yet. I created a unit test for it though so we just need to write the code to make the unit test pass

hsrudnicki:
Does this remain still an open issue? Either way, please update the status of this again, whether this can be closed as resolved/fixed or otherwise? Thx

shjryan:
Has this issue been addressed yet?
```javascript
$('{input selector}').get(0).state.minDate = new Date();
```
[...]

PR #7357, opened 2016-01-27
Fixed the date picker polyfill not using updated min and max dates in the DOM

LaurentGoderre:
#7342


LaurentGoderre:
@nschonni can you review?

nschonni:
Looks fine, but looking at the code, i would have thought this was already done around L107 in the `init`

kodecount:
The issue is when some changes the min or max using JavaScript after the polyfill has been initialized.

PR #7416, opened 2016-02-23
Menu: Enhancements to hover behaviour

derpsoul:
Prevents opening another sub menu if mouse re-enters the currently opened sub-menu within the hover delay

### DIFF
--- a/src/plugins/menu/menu.js
+++ b/src/plugins/menu/menu.js
@@ -506,6 +506,13 @@ $document.on( "mouseleave", selector + " .menu", function( event ) {
 	}, hoverDelay );
 } );
 
+// Prevent opening another menu if mouse re-enters already opened menu
+$document.on( "mouseenter", selector + " .sm", function() {
+  if ( $( this ).attr( "aria-expanded" ) === "true" ) {
+    clearTimeout( globalTimeout );
+  }
+} );
+
 // Touchscreen "touches" on menubar items should close the submenu if it is open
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
Did you know that you could update your pr by pushing new commits two it?

Can I also ask you to squash these two commits if the build passes?

derpsoul:
How often are these types of requests merged? There are a number of gov't sites using wetbeow for menus where their menu is wide, causing issues. 
To explain the issue a bit better, trying to follow the red cursor path in this image will result in  the "Business" menu from popping up, when the user intends to select something from the dorpdown. [http://i.imgur.com/jnB5G4s.jpg](url)

here's an example where the dropdown menu takes the entire width:
[http://i.imgur.com/lLVFW6i.jpg](url) . Blue being the typical cursor path, red being the path someone has to take to prevent the 'business & development' menu from popping up...

derpsoul:
How often are thes types of requests merged? There are a number of gov't sites using wetbeow for their menus that have this issue.
To explain the issue a bit further in this image http://i.imgur.com/jnB5G4s.jpg following the red cursor path will result in the _Business_  menu from opening up.

It's worse when the dropdown menu is wider than the button that opened it. For example;
http://i.imgur.com/7rO2bid.jpg The blue cursor path is typical, but following this path will result in the "business & development" menu to drop down. Instead they have to follow the red path, which is not ideal...

PR #7383, opened 2016-02-09
Theme: Fixed various RTL breadcrumb issues.

EricDunsworth:
- Prevents arrows from getting mispositioned between items.
- Prevents text from unexpectedly getting cutoff and replaced with ellipses in WebKit browsers.

**PS:**
If anyone happens to have a working build system, I'd recommend quickly testing out this PR before merging.

I wasn't able to do a test build of these changes, since as usual, I'm unable to setup the build system no matter how much I troubleshoot. Having said that, I did test this PR's changes by manually modifying local copies of unminified theme CSS files. Seems to work fine in Firefox/Opera/IE11/Chrome.

**PS2:**
Once this gets merged in, I'll port the changes to the other theme repos.

PR #7423, opened 2016-02-26
Breadcrumbs: Reworked into conditional base SCSS, added variables and fixed RTL issues.

### LINKED ISSUE #7422: Should generic breadcrumb SCSS be moved into /src/base/?
Recently I noticed some RTL issues in WET's breadcrumbs and came up with a fix... but because breadcrumb SCSS is contained within the individual themes, I had to open several PRs to all of them (#7383, wet-boew/theme-gc-intranet#152, wet-boew/theme-ogpl#119, wet-boew/theme-base#125, wet-boew/theme-gcwu-fegc#234, wet-boew/GCWeb#1066).

From what I saw, every theme's breadcrumb SCSS files are virtually identical. The only differences seem to be related to some themes not keeping up-to-date with wet-boew and a couple of colour overrides.

@pjackson28 @LaurentGoderre @shawnthompson @masterbee Would you be on board if I moved all generic breadcrumb SCSS into WET's base folder and only leave minor overrides in the individual themes?

EricDunsworth:
- Managing generic breadcrumb styles should now be easier since they're no longer repeated across all themes.
- Added enable variable to allow themes to optionally exclude breadcrumb SCSS.
- Added variables for inner/outer background colours and arrow colours.
- RTL fixes:
  - Prevents arrows from getting mispositioned between items.
  - Prevents text from unexpectedly getting cutoff and replaced with ellipses in WebKit browsers.

Fixes #7422 and closes #7383.

@LaurentGoderre @shawnthompson IMO this should've been good to go, except when I try running `grunt css`to build it, I get a warning that falsely claims my `breadcrumb-print` mixin doesn't exist :S. AFAIK I didn't botch that mixin's declaration, include, nor imports to its file. Any ideas?

**Screenshot:**
breadcrumb-print-mixin-warning


LaurentGoderre:
It's likely the print css doesn't have access to the mixins. The print css should be minimalist so it perhaps shouldn't call the mixin

EricDunsworth:
@LaurentGoderre I've come up with a workaround that seems to work :). In src/base/_wet-boew.scss, ``@import "breadcrumb/base";`` is now directly followed by ``@import "breadcrumb/print";``. So the print mixin declaration is no longer stuck within the confines of ``@media print``.

LaurentGoderre:
Please don't do that. It needlessly duplicate the code. The print is only for override of the base. You shouldn't call the mixin form that file.

LaurentGoderre:
Actually, the mixin should all be in the same files, maybe in a `mixins.scss` in this folder 
https://github.com/wet-boew/wet-boew/tree/master/src/base

And then in `wet-boew.scss`, you include that mixin file

EricDunsworth:
@LaurentGoderre I've noticed that https://github.com/wet-boew/wet-boew/tree/master/src/base/_wet-boew.scss already contains global placeholder selector declarations prior to the imports. Do you think maybe it'd be better to just dump these mixins into there? Might be kind of strange to have a separate file for mixins but not for placeholders :S.

LaurentGoderre:
I still think it would be clearer if it was in a separate file.

src/base/breadcrumb/_base.scss
```diff
@@ -0,0 +1,55 @@
+/*
+  WET-BOEW
+  @title: Breadcrumb CSS
+ */
+
+@mixin breadcrumb($wb-bc-background-color: null, $wb-bc-ol-background-color: null, $wb-bc-arrow-color: #333) {
```
nschonni: The parameters should be in the _variables.scss

src/base/_wet-boew.scss
```diff
@@ -101,6 +101,8 @@
 
 /*! WET-BOEW Core and Plugins */
 @import "accessibility/base";
+@import "breadcrumb/base";
+@import "breadcrumb/print";
```
nschonni: Don't import print into the overal SCSS

src/base/breadcrumb/_print.scss
```diff
@@ -0,0 +1,20 @@
+/*
+  WET-BOEW
+  @title: Breadcrumb print CSS overrides
+ */
+
+@mixin breadcrumb-print {
```
nschonni: Why is this a mixin?

nschonni: If you want to conditionally include it, maybe another variable "$wb-bc-print-enable"?

EricDunsworth:
@nschonni I like the idea of a ``$wb-bc-print-enable`` variable, but maybe it'd be worth taking things a step further with an enable variable that impacts both base and print breadcrumbs (like ``$wb-bc-print-enable``)?

Maybe it'd be better if I just abandoned the idea of mixins for this and stick solely with variables and src/base/breadcrumb/* files containing ``@if`` conditions? That way breadcrumb mixins wouldn't need to be referenced by any of the themes. They'd just need to worry about declaring a couple of variables and wet-boew's core SCSS would deal with the rest.

Thoughts?

nschonni:
@EricDunsworth that probably makes more senses since `@mixin`'s are more for generic functions/reusablity

EricDunsworth:
@nschonni Ok, awesome :)! Any thoughts on my latest commit? Once everyone's cool with what I have, I'll squash it all into one commit and maybe revise this PR's title/OP message.

theme/_variables.scss
```diff
@@ -1,5 +1,6 @@
 // Theme Variable overrides
 
-// Add variables here
+//Breadcrumb list background colour (Boostrap override)
+$breadcrumb-bg: #FFF;
```
nschonni: The rest of the variables use lower case for the colour codes

nschonni: It looks like it was `#f9f9f9` in the old base.scss

EricDunsworth: @nschonni 

* I'll change the colour code into lower-case soon.
* ``$wb-bc-background-color`` is set on ``<nav id="wb-bc">``. It controls the breadcrumb's background colour outside the content area. It was originally ``#f9f9f9``, but that's already the default background colour of the WET theme's ``<body>`` element, so IMO there's no value in re-declaring it.
* ``$breadcrumb-bg`` is a Bootstrap variable that's used on ``<ol class="breadcrumb">``. It controls the breadcrumb's background colour inside the content area. Most themes don't override Bootstrap's default colour for this (the WET and base themes are the exception).

EricDunsworth:
I've just squashed the commits into one and revised the OP to reflect what this PR has morphed into. IMO it should finally be good enough to merge now.

src/base/breadcrumb/_base.scss
```diff
@@ -0,0 +1,54 @@
...
+
+		li {
+			max-width: 100%;
+			overflow: hidden;
+			text-overflow: ellipsis;
+			white-space: nowrap;
+
+			&:before {
+				color: $wb-bc-arrow-color;
+				content: "\e092";
```
nschonni: Should the icon be customizable as well?

nschonni:
LGTM, just that one comment about whether it makes sense to expose the breadcrumb icon too (and the RTL version)

EricDunsworth:
@nschonni Good idea :). I'll sneak it in and rename the arrow colour variable to be more generic.

LaurentGoderre:
I still think it should be a mixin that theme can chose to include. If you want to create a them with your own breadcrumb, you would have to turn off the default breadcrumb with a variable and then provide your own breadcrumb which would be weird.

With that being said, I won't oppose to merging this since it's an improvement over what we have.

EricDunsworth:
Updated :). Btw once this gets merged in, I'd highly recommend quickly merging in the PRs I sent to all the themes to avoid conflicts between the old and new ways of dealing with breadcrumbs.

@LaurentGoderre Hopefully the separator variables I've just added will help alleviate that risk to some degree. In addition to shape variables, I've also added font family/size variables.

src/_variables.scss
```diff
@@ -16,6 +16,19 @@ $wb-assets-path: "#{$wb-core-path}/assets" !default;
...
+
+// WET Custom
+$wb-bc-enable: true !default; // Enable or disable WET's breadcrumb SCSS. Themes that don't use breadcrumbs can set this to false.
+$wb-bc-background-color: null !default; // Outer background colour. Override $breadcrumb-bg (Bootstrap variable) to change the inner background colour.
+$wb-bc-separator-color: #333 !default; // Separator colour.
+$wb-bc-separator-font-family: "Glyphicons Halflings" !default; // Separator font family.
+$wb-bc-separator-font-size: 0.7em !default; // Separator font size.
+$wb-bc-separator-content: "\e092" !default; // Separator content.
+$wb-bc-separator-content-rtl: "\e091" !default; // Separator content (right to left).
+
```
LaurentGoderre: I know this is knitpicking but could you move that block to before the tables variables?

EricDunsworth: @LaurentGoderre Done :). Btw just out of curiosity, what was the reasoning behind your suggestion?

When I initially placed the breadcrumb variables before the colour ones, I was thinking of putting things in alphabetical order. Felt kind of strange though, since it came right after the start of the Boostrap variables block. But on the other hand, that block already contained a mix of Bootstrap overrides and WET-specific variables, so it didn't feel any weirder than what was already being done.

LaurentGoderre: That block is for general overrides wjile the section where you moved it is for component specific variables. In the output it makes no difference but I wanted that file to be well organized, even if we don't have any formal specification for it.

LaurentGoderre:
There probably should be a `wb-bc-print-enabled` to specifically control of the breadcrumb but that is not a blocker for this. It can be done in another PR

EricDunsworth:
@LaurentGoderre IMO that might be overkill. If anyone wants to use WET's built-in breadcrumb SCSS, I can't think of a legitimate reason to keep only the base styles and forego the print ones. I'd consider it to be an all-or-nothing deal. The print styles also only override 2 properties.

Btw thanks for merging it all in :)!

LaurentGoderre:
I can see a theme not wanting to have breadcrumbs appear in print view but we can leave it for now

LaurentGoderre:
We'll have to test all the themes once the builds are done

EricDunsworth:
@LaurentGoderre Oh, so what you had in mind was using one of those variables to outright hide the breadcrumbs in print view?

LaurentGoderre:
A new variable yeah. We can wait until someone requests it though since it's central now

EricDunsworth:
@LaurentGoderre Ok :). Otherwise, maybe I could send off a quick PR for that within the next few days. I originally thought you just wanted the ability to exclude the extra print properties lol.

LaurentGoderre:
I can undersand now why you didn't want that! It would be pretty terrible :P

EricDunsworth:
@LaurentGoderre Just checked out the other themes and their breadcrumbs look ok :). Most of them just use Bootstrap's default breadcrumb background colour.

Btw I've just opened PR #7448 for the print variable.

PR #7441, opened 2016-03-08
Tabs Carousel - Stop carousel if only 1 panel fixed#7428

TTVXStarWake:
I did run grunt dist and all check passed.

### DIFF
--- a/src/plugins/tabs/tabs.js
+++ b/src/plugins/tabs/tabs.js
@@ -71,6 +71,16 @@ var componentName = "wb-tabs",
 			$panels = $elm.find( "> .tabpanels > [role=tabpanel], > .tabpanels > details" );
 			$tablist = $elm.children( "[role=tablist]" );
 			isCarousel = $tablist.length !== 0;
+
+			// Stop carousel if only 1 panel
+			if ( isCarousel && $panels.length === 1 ) {
+
+				// Remove tabs roles
+				$tablist.remove();
+				$panels.removeAttr( "role" );
+				return true;
+			}
+
 			activeId = wb.jqEscape( wb.pageUrlParts.hash.substring( 1 ) );
 			hashFocus = activeId.length !== 0;
 			$openPanel = hashFocus ? $panels.filter( "#" + activeId ) : undefined;


shawnthompson:
cc/ @masterbee

MaximMartel:
cc/ @LaurentGoderre

src/plugins/tabs/tabs.js
```diff
@@ -71,6 +71,16 @@ var componentName = "wb-tabs",
 			$panels = $elm.find( "> .tabpanels > [role=tabpanel], > .tabpanels > details" );
 			$tablist = $elm.children( "[role=tablist]" );
 			isCarousel = $tablist.length !== 0;
+
+			// Stop carousel if only 1 panelc
+			if ( isCarousel && $panels.length === 1 ) {
```
shawnthompson: remove the c in the comment
add, amend, force push your commit

LaurentGoderre:
Would it be possible to have that condition before the creation instead of letting do and then undo?

shawnthompson:
What if we merge this fix now and work on a larger change in a new PR. We need this fix now without changing the source code.

duboisp:
@LaurentGoderre "That condition" is located in the `init` after the variables required to operate this conditional are initialized. Furthermore, "that condition" is located before the creation of any tabs control and before any setting got extended.

There is only one line of "creation" code before "that condition". It is about backwards compatibility and it is clearly commented "Should be removed in WET 4.1". So I can presume that line of code will be deprecated in a near future.

@shawnthompson I think this fix is good as is. The "larger change" will be to teach author (web editor) to not use tabs (carousel) widget when they have only one tab.

Cheers

:+1:

masterbee:
@MaximMartel, @duboisp - I like the approach but kind of agree with @LaurentGoderre in terms of seeing if there was a way to prevent the execution of the plugin if the condition is not met.

What is everyone thoughts on tweaking the selector variable in the plugin to ensure only the tabs with 2 or more tabpanels get the treatment. This way we are not mucking with the internals of the plugin.

Here is my suggestion:

currently :
```js
selector = "." + componentName,
```

change this to:
```js
selector = "." + componentName + ":has(.tabpanels:nth-of-type(2),[role=tabpanel]:nth-of-type(2),details:nth-of-type(2))",
```

Thoughts?

MaximMartel:
@masterbee @LaurentGoderre 

Thanks a lot for those comments and I really like the approach that masterbee is taking. Laurent, i totally agree with you as well. No dought.

I will work on this issue this morning and get it done asap. 

If the author remove 2 images in the carousel, but leave the code as a carousel and now has only one item, we need to keep our fix in so that the DOM will be changed in that case? 

@duboisp Thanks.

LaurentGoderre:
Do you have an example that illustrate the problem before the fix?

duboisp:
@masterbee I really like your solution, but it do not remove the `[role=tabpanel]` and the elements set to `[role=tablist]` originally set by the authors. 

The authors intentionally called the tabs carrousel widgets and I think the widget should at least be initialized. 

The ideal solution will be to somehow let the authors know that it should not use the tabs widget when there is only one panel.

Thoughts?

shawnthompson:
We don't want to have to change the source HTML for now. The goal is to change the source HTML in the carousel for WET 4.1 thought.

 @MaximMartel is doing a gh-pages version for you right now.

shawnthompson:
Here's what the carousel looks like with only one in it.

http://shawnthompson.github.io/wet-tabs/demos/tabs/tabs-carousel-en.html

LaurentGoderre:
It's not really a bug though.....it just looks stupid

MaximMartel:
well, the play/pause button is irrelevant now ?

shawnthompson:
Here's what the fix does
http://shawnthompson.github.io/wet-tabs-1/demos/tabs/tabs-carousel-en.html

MaximMartel:
http://www.universallabs.ca/WET-PR-7441/demos/tabs/tabs-carousel-en.html
http://www.universallabs.ca/WET-PR-7441/demos/tabs/tabs-carousel-fr.html

LaurentGoderre:
Can't we create more flexible CSS to be able to create this without using the JS at all?

shawnthompson:
What's the status on this? I'd like to merge this unless someone has a better option?

duboisp:
@LaurentGoderre I think we can not avoid the usage of js because the HTML need to be cleaned. If any js get executed that means the web editor as intentionally enhanced the content with the carousel widget.

Ideally, the carousel widget should not be used when there is only one tab. 

@LaurentGoderre and others what do you think if we add an aside note in the carrousel working example that said the carrousel should not be used when there only one tab. And we can add an example that show how to use only style (css) to produce a similar results.

I propose for the next version of the tabs / carrousel widget that aria role get added by the widget instead of having the web editor to code those aria role.

Thoughts?

shawnthompson:
@duboisp, I agree that the _ARIA_ should be added by the _JS_ instead of the author. I also agree that the carousel shouldn't be used for just one panel but sometimes with systems like Drupal, Teamsite and AEM use components to control the carousels and if the author just wants to add one, our implementation should be flexible enough to allow authors to do that. The only thing semantically marking it up as a carousel / panel is the role attributes which we should be putting in with _JS_ so we can only add them if the panel is larger than 1. The `ul` should / could also be created with _JS_, if there is only one, the list wouldn't be there. Having said all that, I guess the systems could also build the components with conditions to use different markup depending on how many elements are added... ARRRRG "ce n'est pas facile la vie d'un cow-boy"

LaurentGoderre:
To me, that is the editor's responsibility. We should not complicated our codebase for soemthing that is better done in the editor.

masterbee:
@LaurentGoderre , @MaximMartel , @duboisp - Let me chime in since I think there are several discussions on this thread happening.

- Leveraging JS to clean up HTML: I happen to agree with @LaurentGoderre that we do not want to start using JS to handle poorly written HTML. To me that makes things both hard to maintain ( from a framework perspective ) and really does not do any editor any favors since we are basically masking potential accessibility pitfalls. WET, IMHO, has worked to enhance code base with additional items like ARIA but requires good markup and semantics to know where items are. This what makes working with WET so neat, you almost have to learn a few things about accessibility to understand how it is built.
- Carousel should no function trigger with only one item: I agree with this approach simply because overtime a section of webpage that holds a carousel will flux in terms of items it has to display. It should not becomes the burden of the editor and/or application change the core markup each and every time there is only one item. I can understand people may want a different treatment of that carousel space if there is only one item, but that is a separate discussion, IMHO. Most feature rich framework components handle this internally.

@LaurentGoderre - My recommendation was to modify the selector of the plugin to ensure it only fired when there was sufficient items in the carousel markup pattern. This I believe speaks to your concern earlier about not complicating our codebase, but more ensuring that the codebase only runs on the markup that supports it correctly. 

@MaximMartel , @duboisp - This thread has been open for a bit so want to make sure this solution still works since there was a few last minute commits to tweaks from the testing that has been going on for a while. Is that the case?

LaurentGoderre:
The selector approach works for me.

MaximMartel:
@masterbee I have been running test this morning with this proposed fix, and all is fine.

LaurentGoderre:
If you are releasing soon, I would recommend waiting post-release to merge

PR #7449, opened 2016-03-11
Fixed bold in th to be consistent with bold in all browsers

shawnthompson:
Fixes https://github.com/wet-boew/GCWeb/issues/1075

Thoughts @LaurentGoderre @pjackson28?

Should this be in the theme or base?


shawnthompson:
I was wrong... @EricDunsworth found the fix in https://github.com/wet-boew/GCWeb/pull/1070

PR #7453, opened 2016-03-15
Fix character encoding

nschonni:
Playing around with Codacy and saw these get flagged


LaurentGoderre:
Very interesting....what were they? Non breaking space?

nschonni:
Not sure, I had to flip the files over to a hex editor to even tell which ones were broken, but I didn't dig into which type it was. It was http://eslint.org/docs/rules/no-irregular-whitespace that flagged them, so I'll make sure I enable that rule in the PR to switch to ESLint

PR #7474, opened 2016-03-29
Fixed an issue with the date picker returning incorrect dates

### LINKED ISSUE #7470: Calendar issues on daylight saving time days between (midnight - your local timezone offset) to midnight, only in some browsers
The unit tests have revealed a bug with the calendar plugin and dependencies (calendar of event and date picker polyfill) where selecting a date sometimes gives you the wrong date. It also means that build fail between 7PM EDT and 11:59 EDT.

LaurentGoderre:
Fixes #7470


LaurentGoderre:
To truly see if this fixed the problem, the build will need to be restarted after 7pm

LaurentGoderre:
To truly see if this fixed the problem, the build will need to be restarted after 7pm

Issue #7435, opened 2016-03-04
ARCHIVED - Archived information: 


Mass45:
Hi,
Couple of questions for Paul Jackson. I hope you can help us clarify a few points from page http://www.tbs-sct.gc.ca/ws-nw/mo-om/ts-st/p9-eng.asp:
1- Where it says "Where the content of a page/screen is archived in its entirety, adding the word "archived", in the language of the content, to the beginning of the screen title.", does it mean:
a) to insert "Archived - " in the tag only?
b) to insert "Archived - " in the tag only? (e.g. http://www.tbs-sct.gc.ca/reports-rapports/cp-rc/index-eng.asp)
c) both? (e.g. http://www.tbs-sct.gc.ca/report/res_can/rc-eng.asp)
d) none (e.g. http://www.tbs-sct.gc.ca/reports-rapports/tbsol-losct/tbsol-losct-eng.asp)

2- The old way of taking care of "Links to archived content are identified by...", was to add "ARCHIVED - " to the hyperlinked text pointing to the archived document. Now, we don't provide this information before or on the link. Now of course, we follow the directive: Adding the word "archived", in the language of the link, to the end of the link text.
To convey the archived information to people using JAWS or NVDA that might click on the link without reading the archived label, we recommend inserting a <span class="wb-inv">Archived </span> in-front of the text as shown on the test page http://www.statcan.gc.ca/eng/test/archive.

The inconvenience of JAWS mentioning "Archived" twice outweighed the disadvantage of clicking on the link and ending up on an archived page.</p> <p>The second bullet on the page was using the aria-describedby as an option, but the accessibility team here didn't like it as much as the first option.

If you could please provide me with an answer for Point 1 and give me your comments for point 2.
Thanks.

pjackson28:
@Mass45 

1. It should be added to both.

2. The minimum required is adding the word archived at the end of the link, but technically there is nothing preventing it from being added at the beginning as well. The rationale for not having it only at the beginning was that it made the links harder to find through a screen reader link list (since all those links would be grouped with the other links that started with "A"). It also was more difficult for users to scan (since all links started with "Archived"). But as you pointed out, there are minor issues either way.

Mass45:
@pjackson28 Thanks for the very quick response.
1- Thanks, will do.
2- Are you Ok with the way we coded the first bullet (using the invisible Archived code)?
Mass45

MaximMartel:
Hi,

2- For the first bullet, you used :

[...]

PR #7502, opened 2016-04-15
Fixes - #7435 : Updated Archived

TTVXStarWake:


LaurentGoderre:
It seems that perhaps the site was down or at least the api

PR #7506, opened 2016-04-17
[Saucelabs] - disable saucelabs

masterbee:
- Consistant failure of saucelabs tunnel from travis is preventing community contributions
- Working with Saucelabs Open via similar ticket to investigate the issue. [https://support.saucelabs.com/customer/portal/questions/14324286-sauce-connect-seems-to-be-failing-to-make-a-connection]
- Adjust JSON formatting to align with .editorconfig settings.


package.json
```diff
@@ -1,67 +1,67 @@
...
-    "grunt-wet-boew-postbuild": "^0.1.3",
-    "grunt-wget": "~0.1.0",
-    "load-grunt-tasks": "^3.2.0",
-    "mocha": "^1.21.5",
-    "sinon": "~1.12.2",
-    "time-grunt": "~1.0.0"
-  }
-}
\ No newline at end of file
+    "name": "wet-boew",
```
nschonni: This will get formatted back next time, because 2 spaces is the NPM default

nschonni:
Sauce doesn't run on community contributions, only for merges to master by contributors

LaurentGoderre:
There was an api error last week

masterbee:
@nschonni - hmmm then we need to figure out something for the .editorconfig file cause my IDE auto formatted since I have it set to listen to those files.

Additionally I understand he role sauce labs plays and when it fires in contributions but accepting PR's as of late has been littered with tunnel and api issues that was already highlighted. It is not the first time there has been an issue in WET around Saucelabs failing systematically so we are taking the time after a stable release to look into issue with Sauce Open to see if there is anything that can be improved on our side.

nschonni:
@masterbee for your IDE changing the spacing, it sounds like your plugin is broken, because these files are dealt with explicitly in the config https://github.com/wet-boew/wet-boew/blob/master/.editorconfig#L13-L15

Issue #4775, opened 2014-03-10
Multimedia Player - Video: How to distinguish controls for multiple videos on a page


dfarough:
@wet-boew/access-working-group 
How can we Indicate which video is being controlled by a given set of controls?  In the past I have by accident started more than one video and then had to guess at which Play /pause button would stop the video. Would use of a field set for the video be a good  way to do this?
Or
. I am thinking of something like Preceding each button in the set with a number representing the position of the video on the page. Probably separating the number and the button name with a Colin would be a good Idea so that the synthesizer could pause between the number and the name.

hsrudnicki:
Are you ready to close the issue or otherwise ?

dfarough:
@wet-boew/access-working-group 
There is still no way to know which video is being controlled by the player controls when there is more than one multi media player on the page.

When using Jaws and Internet explorer 10 adjusting the volume slider changes the progress of the video not the volume.  This does not occur when using Firefox.

This issue should not be closed in my opinion.

LaurentGoderre:
@dfarough for controlling the volume, you are using the up/down arrows right?

dfarough:
yes I am using up and down arrows.  this does seem to work with Firefox though.

LaurentGoderre:
very strange

hsrudnicki:
@dfarough and others : Please update this issue's status whether this issue had been fixed and ready to be closed or otherwise, still needs be open? Thank you

hsrudnicki:
Close this issue, just labeled "Going Stale" or otherwise? Please update this issue's status, thank you

hsrudnicki:
Due to no further discussion/status updates, this issue is closed.

mysticjeff:
Should it beÉ

-----Original Message-----

hsrudnicki:
@LaurentGoderre : Again, please help in making the decision whether this issue keep closed or re-opened?

SBleeks:
If the ARIA label is associated with the UI controls (play/pause for example), the association isn't being recognized by JAWS or NVDA.  For pages with multiple videos, the could the ARIA describedby or labeledby be used on the UI controls to correspond with the heading of the video.

As far as volume control, I am able to increase and decrease volume with the up and down arrows using IE11 JAWS 16 and NVDA latest build.

dfarough:
I think that using aria-labeled or aria described by would be a good option for these controls.

[...]

PR #7507, opened 2016-04-17
[Multimedia] - ARIA addition

masterbee:
- Adding `aria-labelledby` to contextualize buttons on media player
- proposal for https://github.com/wet-boew/wet-boew/issues/4775#


LaurentGoderre:
Hmmmm not sure this would work. It's not pointing to a textual element but a div no?

hsrudnicki:
Can be this pull closed based on the 'pass' results?

masterbee:
@LaurentGoderre - the aria-labeled by is actually targeting the video element which has the title attribute that the recommendation is asking for. We are in development between builds so I think it would be good explore this so @dfarough can take a look to see if the screen readers behave as we hope they will.

PR #7550, opened 2016-05-13
Tabbed interface: Replaced selector with namespace in timerpoke init event.

EricDunsworth:
Prevents timerpoke from endlessly adding and removing IDs on several elements, due to its use of the :has() selector.

:has() isn't a native CSS selector. It's a jQuery extension. When it's used, jQuery temporarily adds and removes ID attributes to the elements that get checked by that selector. It's not desirable for timerpoke to use it since it runs every 500 milliseconds... which will result in continuous, unnecessary DOM manipulations. It also hurts the performance of browser DOM inspectors.

Thanks @fsnoddy for figuring out what :has() was doing :).

### DIFF
--- a/src/plugins/tabs/tabs.js
+++ b/src/plugins/tabs/tabs.js
@@ -722,7 +722,7 @@ var componentName = "wb-tabs",
 	};
 
  // Bind the init event of the plugin
- $document.on( "timerpoke.wb " + initEvent + " " + shiftEvent + " " + selectEvent, selector, function( event ) {
+ $document.on( "timerpoke.wb " + initEvent + " " + shiftEvent + " " + selectEvent, namespace, function( event ) {
 	var eventTarget = event.target,
 		eventCurrentTarget = event.currentTarget,
 		$elm;


pjackson28:
@EricDunsworth This seemingly simple change looks like it could break a lot use cases (e.g., nested tabbed interface and accordions) so it will be important to ensure any change you implement has the same coverage as the current selector.

nschonni:
@pjackson28 true, but the commit that added the `:has()` https://github.com/EricDunsworth/wet-boew/commit/e3363c348b3a1609c1e8b78b447f4787d836619c actually hasn't been included in a release yet.

pjackson28:
@nschonni Oh okay, so it it effectively reverting it to the way it was before.

masterbee:
@EricDunsworth - thanks for this and I have to say that this is a interesting side effect of ```:has()```. the namespace approach is a clean fix, but wonder if there should not be a more extension to the timerpoke logic since it would not be un common to leverage the ```:has()``` modifier to qualify children elements via selectors. 

Thoughts?

EricDunsworth:
@masterbee AFAIK, no other existing WET plugins/polyfills rely on the state of their children in order to initialize. So I don't think it's a common use case.

Btw Canada.ca is currently using a v4.0.22-development build from April 28 and is in a much worse state than GitHub's demo pages. Literally every single HTML element on all of Canada's pages are currently getting IDs inserted and removed by ``:has()`` every 500 milliseconds. As a result, it's hurting performance and turning browser DOM inspectors into flickering light shows.

IMO there'd be value in merging in this PR and updating Canada.ca's version of v4.0.22-development shortly afterwards to sort this out.

EricDunsworth:
@masterbee @nschonni @pjackson28 Just to better illustrate the issue this PR is fixing, I've taken a few screenshots of what's currently happening when I open up Firefox Developer Edition's DOM inspector on sites/pages that use v4.0.22-development. It affects all pages - not just ones that use the tabbed interface.

Yellow-highlighted lines represent DOM manipulations and are currently flickering like crazy because ``:has()`` is endlessly adding and removing IDs from them every split second.

**[GitHub - WET 4.0 demos page](http://wet-boew.github.io/v4.0-ci/demos/index-en.html) (``html``, ``body`` and ``header`` elements are affected):**
wet4-demos_has_dom-inspector

**[Canada.ca - Home page](https://www.canada.ca/en.html) (virtually every single HTML element on the page are affected):**
canada ca-home_has_dom-inspector

I should also note that the CSS panel in IE11's DOM inspector is completely unusable in v4.0.22-development at the moment. It continuously refreshes every split second because of the DOM manipulations, which makes it impossible to manually-edit any CSS properties within that browser.

Is there any chance this PR could be merged in ASAP to fix it?

PR #7580, opened 2016-06-14
wb-telLink: adds hyperlinks to all phone numbers on mobile devices.

atamsingh:
The plugin locates any 10 digit phone number with and without
hyphenation, brackets etc. and adds an a tel: link as href to the phone
number.


nschonni:
Thanks, you might want to take a look at the https://github.com/wet-boew/wet-boew/blob/master/src/plugins/texthighlight/texthighlight.js for an example of a plug-in structure. 
Except for polyfills, which I don't think this is since native mobile browsers handle this already https://davidwalsh.name/phone-link-protocol, plug-ins usually act on a decorator class like `wb-tel`.

atamsingh:
@nschonni Thanks for your comment! I have been working on making telephone link plugin comparable to other wet plugins and seemed to have luck with it. The only issue I am facing the need for including component name as a class name in the webpage. Is there any mechanism in wet that allows for running plugin regardless of the selector being present? (Maybe this is a very stupid question with an obvious answer, but I couldn't figure it out :S) Thanks!

As to your concern for native mobile browsers already implementing this protocol, it seems to me that with the exception of iOS, tel href links are required for click to call functionality. Also with the multiple third party browsers lacking auto detect functionality, this plugin might be very useful! What do you think?

nschonni:
@atamsingh the only plug-ins that run without a specific selector are those that polyfill native browser features. 
PS: Looks like you've merged some other stuff into this PR's branch, but if you're thinking about using the microdata as a selector that might be considered a polyfill, but that's not my call :wink:

RobJohnston:
@atamsingh Just an FYI, in addition to iOS, BlackBerry also auto detects phone numbers, and there's a way to turn it off.  From [mobilexweb](http://www.mobilexweb.com/blog/click-to-call-links-mobile-browsers):

> The BlackBerry browser and Safari for iOS (iPhone/iPod/iPad) automatically detect phone numbers and email addresses and convert them to links. If you don’t want this feature, you should use the some meta tags.
> ```
> For Safari
> 
> <meta name="format-detection" content="telephone=no">
> 
> For BlackBerry
> 
> <meta http-equiv="x-rim-auto-match" content="none">

chrismajewski:
Lets not forget that the **Standard on Web Interoperability** is still in effect even if under review for web renewal.
http://www.tbs-sct.gc.ca/pol/doc-eng.aspx?id=25875

Modification to metadata must include HTML Data as described in Appendix E in that standard.  That's RDFa Lite 1.1+ and Schema.org.  Class changes that inject semantic understanding to the data is included and can be added as a secondary/alternate form of metadata.

If you need assistance on how to implement feel free to contact me, I'm sure someone here knows how to find me :)

atamsingh:
@nschonni Thanks a lot for that! Not going to lie, I wasn't mush aware of the concept behind polyfills. Now that I know, it seems like the correct way to implement click to call functionality. Also, I added the additional files by mistake. Will correct. Thanks a lot for your help! :smiley: 

@RobJohnston Thanks for bringing that to my attention. Will be helpful in creating the polyfill!

@chrismajewski Thank you for your comment but I'm a little confused about which metadata you are referring to. Were you implying the comment by @RobJohnston about turning off auto detect? (In which case I think I'm not going to turn off the auto detect suing meta tags, will create an exception for iOS and BlackBerry OS instead. ) or were you be referring to the schema plugin I uploaded by mistake? (In which case I am implementing that using dynamically generated RDFa Lite and SchemaLD) Sorry for the confusion! :confused:

chrismajewski:
@atamsingh 

Hyperlinking telephone numbers is applying semantic understanding to data on the page.  Machine description of data, in Government of Canada HTML content, is done by applying "HTML Data" as previously described.

Alternate definitions are acceptable but the primary form must be RDFa lite and Schema.org.

If this isn't being applied to GC content than I'm mistaken, if it is we'd need to apply an schema object such as contact point.

https://schema.org/ContactPoint

hsrudnicki:
Based on the 'pass' results, can be this pull closed or otherwise?

LaurentGoderre:
I would say :-1: on this because WET never tried to guess content, only enhance properly marked up content. 

Potential problems i see are phone numbers being applied to tables with numbers and this conflicting with plugins that something similar like the skype one.

lucas-hay:
Discussed at PR review meeting. We will not merge. The way the plugin is coded now is specific to north american formats. We are also concerned that perhaps site developers may not want this. Phone numbers should coded by coders. This could cause adverse affects.

PR #7587, opened 2016-06-22
Changes to fix issues when serving page as application/xhtml+xml

oldhen:
Changes to fix issues when serving page as application/xhtml+xml

### DIFF
--- a/src/plugins/menu/menu.js
+++ b/src/plugins/menu/menu.js
@@ -204,7 +204,7 @@ var componentName = "wb-menu",
 				sectionHtml + "</ul></nav>";
 		}
 
-		return panel.replace( /['"]?list-group-item['"]?/gi, "\"\"" ) + "</div>";
+		return panel.replace( /['"]?list-group-item['"]?/gi, "\"\"" );
 	},
 
 	/**
--- a/src/plugins/favicon/favicon.js
+++ b/src/plugins/favicon/favicon.js
@@ -6,7 +6,7 @@
  *
  * This plugin provides the ability to add and update the favicon's on a web page. Its default behaviour is to add a mobile favicon to web pages that have a favicon defined by a `<link rel='icon'>` element.
  *
- * The mobile favicon's file name, rel, path and sizes can be set with data attributes on the `<link rel='icon'>`:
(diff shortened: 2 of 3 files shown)


masterbee:
@oldhen - Just chiming in to take a look at these P.R. I am all for the 1st P.R for adding the "/" for xhtml compatibility but not sure about the extra div. Can you confirm the mobile menu is in fact generating a rogue open "<div>" with not close since none of my post JavaScript test on my browser showed mal-formed <div>'s.

oldhen:
@masterbee Your DOM will never have malformed HTML.  This is because inserting malformed HTML either gets cleaned-up on parsing (in the case of a `text/html` document), or it will cause an error (in the case of an `application/xhtml+xml` document).  For example, running this in a text/html document:

`var test = document.createElement("div"); test.innerHTML = "<div></div></span>"; console.log(test.innerHTML);` 

Will output `<div></div>` (not `<div></div></span>` as you might expect) when you run it in a `text/html` document.  Running it in a `application/xhtml+xml` document however will simply cause an error (`The provided markup is invalid XML`).

Compare these pages, whose only difference is the Content-Type:
[text/html](https://oldhen.github.io/MenuTest/index.html)
[application/xhtml+xml](https://oldhen.github.io/MenuTest/index.xht)

If you have debugging on, you'll see the error caused by the stray </div> on line 6782 (invalid XML).  If debugging is off, you'll simply have no mobile menu and the top menu bar will be visible because the script failed to complete.

src/plugins/menu/menu.js
```diff
@@ -204,7 +204,7 @@ var componentName = "wb-menu",
 				sectionHtml + "</ul></nav>";
 		}
 
-		return panel.replace( /['"]?list-group-item['"]?/gi, "\"\"" ) + "</div>";
+		return panel.replace( /['"]?list-group-item['"]?/gi, "\"\"" );
```
LaurentGoderre: The other fixes I understand, but not this one

LaurentGoderre:
Thanks for this @oldhen! IN the future, no need to add WET-BOEW to the start of your commits

PR #7625, opened 2016-07-29
Created a script to clean the wet build files

shawnthompson:
@nschonni what do you think? I'm always doing this on my work computer because bower keeps timing out due to slow connections.

@lucas-hay @MaximMartel @EricDunsworth

### DIFF
--- a/script/wet-clean
+++ b/script/wet-clean
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+# Does a clean install of the wet build
+
+rm -rf lib &&
+rm -rf node_modules &&
+rm -rf dist &&
+bower cache clean &&
+npm cache clean &&
+npm install


nschonni:
Might be easier to add another target to [`clean`](https://github.com/wet-boew/wet-boew/blob/master/Gruntfile.coffee#L1189) since that would be more cross platform

shawnthompson:
Can we clear the npm and bower cache with grunt?

EricDunsworth:
@shawnthompson According to http://stackoverflow.com/questions/10456865/running-a-command-in-a-grunt-task/13186882#13186882, it could be done with either grunt-shell or grunt-exec. But that would mean adding further dependencies (although I bet those are highly unlikely to ever break)...

shawnthompson:
Wouldn't that be an issue with cross platform also?

shawnthompson:
@nschonni What if I put it in the `package.json` as a npm script? Would that be cross platform?

nschonni:
If you want to do it through an NPM script, take a look at [rimraf](https://www.npmjs.com/package/rimraf) since that is what grunt-contrib-clean uses.

LaurentGoderre:
Closing in favor of #7785

PR #7636, opened 2016-08-16
Adds class to change font-weight to normal on elements that are bold

shawnthompson:


shawnthompson:
@masterbee @lucas-hay @MaximMartel

PR #7671, opened 2016-09-22
Change the defaults setting for Calendar -GP.

jilusan:


shawnthompson:
This is not passing the Sauce lab test cases created for the calendar plugin.

You can find the details on Travis. The link is beside the notice below.

jilusan:
@shawnthompson 
Do we have access to the Sauce lab test cases and change it?

shawnthompson:
@jilusan, I think @MaximMartel or @duboisp  might have it

MaximMartel:
I have checked with @duboisp and we don't have access to saucelabs. Looping in @nschonni to see if he could help ? running the saucelab test cases locally would be ideal.

jilusan:
Yes the ideal should be to have the possibility to run a specific test case locally without doing a PR.

duboisp:
@jilusan try this grunt command line: ```grunt test-mocha```

duboisp:
@jilusan the issue that you are facing is may be related of when you have changed the date in the existing example.

jilusan:
@duboisp  Thank you!

jilusan:
@LaurentGoderre 
I made some minors changes in calendar to have the calendar set on the actual date. 
https://jilusan.github.io/dist-calendar/demos/cal-events/cal-events-fr.html
Unfortunately it had an impact on passing the Sauce lab test case.
I already made some changes in test.js but have some difficulties with "advances keyboard navigation" tests. 
I would like to discuss with you about the changes I made and the one I have some difficulties.
Do you have any time? Thanks.

LaurentGoderre:
Hey @jilusan You shouldn't be changing the calendar plugin to modify the calendar of event. The calendar plugin is only a building block for creating a calendar for the calendar of events and the date polyfill. If you need to change the calendar of events, you need to modify the `cal-events` plugin

MaximMartel:
so, I will close this PR, @jilusan will make his changes on cal-events.js and we will PR the changes.  Thank @LaurentGoderre.

shawnthompson:
Why close this issue? You can just modify your branch and force push it back up.

MaximMartel:
ok great thought @shawnthompson, we will do that instead

jilusan:
And thank you @LaurentGoderre !

src/plugins/cal-events/cal-events.js
```diff
@@ -4,415 +4,422 @@
...
-			break;
-	}
-} );
-
-// Add the timer poke to initialize the plugin
-wb.add( selector );
-
-} )( jQuery, window, wb );
+(function($, window, wb) {
+    "use strict";
```
nschonni: The formatting was changed from tabs to spaces here, you need to revert that. Guess there is a problem with the ESLint config because that should have failed the build

nschonni (changes requested):

src/plugins/cal-events/cal-events.js
LaurentGoderre: @nschonni that's why I said that ESLint is being a bit lax, but i"m working on it

jilusan:
@nschonni @LaurentGoderre 
Not sure what I need to do...
I did a HTML prettify for cal-events.js because I had this error when grunt dist:
"error  Mixed spaces and tabs  no-mixed-spaces-and-tabs".
Do I need to get the previous format?
Thanks.

src/plugins/cal-events/cal-events.js
jilusan: @nschonni @LaurentGoderre 
Not sure what I need to do...
I did a HTML prettify for cal-events.js because I had this error when grunt dist:
"error Mixed spaces and tabs no-mixed-spaces-and-tabs".
Do I need to get the previous format?
Thanks.

LaurentGoderre: What editor are you using? I recommend using Atom which has plugins for ensuring all coding conventions are respected.

jilusan: I use Sublime. Recommended by Shawn. Do you suggest I switch to Atom?

LaurentGoderre: I'm thinking to create a config set for WET for Atom

jilusan:
@nschonni 
Do you want me to format the file as it was before, or it is fine?
Thanks.

nschonni:
Yeah, the format used in the project is Tabs. Since you're using Sublime, you should be able to install the EditorConfig plugin for it https://github.com/wet-boew/wet-boew/blob/master/.editorconfig

jilusan:
Ok. Thanks.

jilusan:
@nschonni @shawnthompson 
Do you have any idea on what is the next step to solve the conflict?
Thanks a lot.

LaurentGoderre:
@jilusan if you run you can force the conflict and resolve it

```
git fetch upstream
git rebase upstream master
```

src/plugins/cal-events/cal-events.js
```diff
@@ -80,22 +79,31 @@ var componentName = "wb-calevt",
 	},
 
 	processEvents = function( $elm ) {
-		var settings = $.extend( {}, window[ componentName ], $elm.data( dataAttr ) ),
```
LaurentGoderre: You can't get rid of this line because it would remove the ability to extends the calendar

src/plugins/cal-events/cal-events.js
```diff
@@ -80,22 +79,31 @@ var componentName = "wb-calevt",
...
+			currMonth = currDate.getMonth(),
+			events, minDate, maxDate, containerId, $container;
 
 		events = getEvents( $elm );
 		containerId = $elm.data( "calevtSrc" );
 		$container = $( "#" + containerId ).addClass( componentName + "-cal" );
 
 		minDate = events.minDate;
-		year = settings.year || minDate.getFullYear();
-		month = settings.month || minDate.getMonth();
```
LaurentGoderre: `settings.year` and `settings.month` need to be preserved because they are overrides. This needs to be rewritten in a way that the default settings are only used if they are not present in the settings object

LaurentGoderre (changes requested):

src/plugins/cal-events/cal-events.js
```diff
@@ -18,7 +18,6 @@ var componentName = "wb-calevt",
 	initEvent = "wb-init" + selector,
 	evDetails = "ev-details",
 	setFocusEvent = "focus",
-	dataAttr = componentName,
```
LaurentGoderre: This needs to be preserved as well

LaurentGoderre (changes requested):

jilusan:
I did all the requested changes. What the next?

src/plugins/cal-events/cal-events.js
```diff
@@ -80,22 +80,36 @@ var componentName = "wb-calevt",
 	},
 
 	processEvents = function( $elm ) {
-		var settings = $.extend( {}, window[ componentName ], $elm.data( dataAttr ) ),
-			year, month, events, minDate, containerId, $container;
+		var settings = $.extend( {}, window[ componentName ], $elm.data( dataAttr ) ), // Ability to extend the calendar
```
nschonni: Dropping the comment at the end will minimize the diff

src/plugins/cal-events/cal-events.js
```diff
@@ -80,22 +80,36 @@ var componentName = "wb-calevt",
...
 		events = getEvents( $elm );
 		containerId = $elm.data( "calevtSrc" );
 		$container = $( "#" + containerId ).addClass( componentName + "-cal" );
 
 		minDate = events.minDate;
-		year = settings.year || minDate.getFullYear();
-		month = settings.month || minDate.getMonth();
+		maxDate = events.maxDate;
+
+		if ( minDate.getTime() > currDate.getTime() ) {
```
nschonni: It looks like this already being done in the loop below https://github.com/jilusan/wet-boew/blob/0dd8f5cf4f893a27992ab9138d849bcfa0f1bc8b/src/plugins/cal-events/cal-events.js#L237

src/plugins/cal-events/cal-events.js
jilusan: Yes. No problem.

src/plugins/cal-events/cal-events.js
jilusan: Not sure I understand. Could you explain?

duboisp: What @nschonni said is the value you are retrieving is from "events.minDate". And that value was set by the returned value of the local function "getEvents()" called at [line 89](https://github.com/jilusan/wet-boew/blob/0dd8f5cf4f893a27992ab9138d849bcfa0f1bc8b/src/plugins/cal-events/cal-events.js#L89) in the code. Note that is the only place where the function "getEvents()" is called. So inside the function "getEvents()" starting at [line 150](https://github.com/jilusan/wet-boew/blob/0dd8f5cf4f893a27992ab9138d849bcfa0f1bc8b/src/plugins/cal-events/cal-events.js#L150) there is already some calculation made in order to set the proper minDate and maxDate (between the [line 242](https://github.com/jilusan/wet-boew/blob/0dd8f5cf4f893a27992ab9138d849bcfa0f1bc8b/src/plugins/cal-events/cal-events.js#L242) and [line 248](https://github.com/jilusan/wet-boew/blob/0dd8f5cf4f893a27992ab9138d849bcfa0f1bc8b/src/plugins/cal-events/cal-events.js#L248)). 

So the question is why those calculated minDate and maxDate are not good? Why to modify/re-evaluate them between [line 96](https://github.com/jilusan/wet-boew/blob/0dd8f5cf4f893a27992ab9138d849bcfa0f1bc8b/src/plugins/cal-events/cal-events.js#L96) and [line 102](https://github.com/jilusan/wet-boew/blob/0dd8f5cf4f893a27992ab9138d849bcfa0f1bc8b/src/plugins/cal-events/cal-events.js#L102)?

duboisp (changes requested):

duboisp:
@jilusan what issue this PR are solving?

MaximMartel:
The intention of @jilusan, correct me if I am wrong here, is to have the calendar of event start date to be the current day, not the first event happening in the calendar. In the example page, they all start on March 2013 since its the first event. We want to default the current day. 

Make sense?

duboisp:
@MaximMartel yes it's make sense. 

So I am thinking that could be the default behavior, but there should be way allow that the start date will be the first event happening in the calendar. I can see it useful for archived multi-event/campaign page. What do you think?

LaurentGoderre:
There needs to be a clear decision map for this behavior of what si to happen in all the possible scenarios.

MaximMartel:
@duboisp Yes, make sense that the default upon loading the page would be the current date, but with an option to setup the calendar to the first event. The 2 code example should be well documented in the working example

src/plugins/cal-events/cal-events.js
jilusan: In case the lower event is something like 2013, we want the calendar to be set as the date of today. Presently the fonction getEvents does do this. But I agree, the fonction should return it instead of (re)modifying. What do you think?

jilusan: My mistake. :-) I beleive the fonction getEvents does the job.

jilusan: The fonction getEvent returns the object {minDate: Thu Mar 23 2017 00:00:00 GMT-0400 (EDT), maxDate: Wed Mar 11 2020 00:00:00 GMT-0400 (EDT), ...} when the multi-Day event has been set to March 22nd, 2017 to April 26th, 2017. And we want minDate to be the date of today. This is the problem. Does getEvents work correctly?

LaurentGoderre: You could write a test to see if it does.

jilusan: Merci @duboisp pour la discussion au tel. 
I understand I need to modify the fonction getEvent to produce the new values for minDate et maxDate.

jilusan:
Please reviewers @nschonni @LaurentGoderre @duboisp  tell me if everything is fine.

MaximMartel:
@jilusan Can you make it default to current date, but to have an option to set it to first occurrence of any event ? If anyone would like to contribute in helping here, that would be awesome.

MaximMartel:
@nschonni @LaurentGoderre @duboisp 

I need to have this PR merged in, and would like to know if your requested changes still needs to be addressed ?

The enhancement that @jilusan has made do not break anything, just put the start date to the current date. Later enhancement could be implemented as discussed on this thread to have an option to have the date as the first event date, but looking to get that code in asap. 

Thoughts ?

duboisp:
@MaximMartel as you may know reviewing this PR is not my priority but call me.

LaurentGoderre:
We don't really know that it doesn't break anything because there is no tests

MaximMartel:
@LaurentGoderre So, in order to have this PR merged in, he will need to create a test in Saucelab?

LaurentGoderre:
Ideally but that would be long

MaximMartel:
can we just test ourselves ?

LaurentGoderre:
Yes but you need to take into account multiple scenarios.... a calendar that have events all in the past months, a calendar that has event in the past, present and future and a calendar that have all events in the future months

MaximMartel:
right, Thanks for clarifying Laurent, really appreciated.

src/plugins/cal-events/cal-events.js
```diff
@@ -256,6 +259,15 @@ var componentName = "wb-calevt",
 		// End of loop through objects/events
 		}
 
+		// Set the min and max Date according to the current date.
+		if ( events.minDate.getTime() > currDate.getTime() ) {
+			events.minDate = currDate;
```
duboisp: Same issue as noted noted previously https://github.com/wet-boew/wet-boew/pull/7671#discussion_r86915152

It's still looks like this already being done in the loop before: https://github.com/jilusan/wet-boew/blob/e772fc680be81434b25dade9f19945fbbd31d6a0/src/plugins/cal-events/cal-events.js#L226

duboisp (changes requested):

MaximMartel:
@nschonni @LaurentGoderre  @duboisp  

Are all the requested changes still valid ?

duboisp:
@MaximMartel oui les changements demandé sont encore valide

duboisp:
If I understand correctly

The current plugin behavior is:
* The calendar min date and max date is based on date retrieved from his associated list of events.
* The default view is the min date.
* The displaying default month can be forced by using the configuration ```year``` and ```month```

So, I think the intention of this PR is to change the behavior like:

* If Min-Date < Today AND Max-Date > Today Then 
  * The min date will be the event min date
  * The max date will be the event max date
  * The default view will be Today month

* If Min-Date < Today AND Max-Date < Today Then
  * The min date will be the event min date
  * The max date will be Today
  * The default view will be Today month

* If Min-Date > Today AND Max-Date > Today Then
  * The min date will be Today
  * The max date will be the event max date
  * The default view will be Today month


**Do you want the user to be able to navigate the calendar of events:**
 * **Option 1** - Limited from the date range determined by the list of events. So when the wb-calevt is initialized the default view is Today month (if not forced through the configuration) but only navigable for the events that is in the events date range.
 * **Option 2** - Extended from the date range determined by the list of events where the min date and max date are adjusted depending of Today date. So Today date is always in the navigable range. Then the default view will be Today month (if not forced through the configuration).
* **Something else?**

My preference will be option 1.

Thought?

MaximMartel:
What y'all think ? 

I am voting for option 1 as well, simple yet effective. On init, the calendar is todays date. Thats what the clients wants, he wants the current month to be displayed at first.

Thanks @duboisp

lucas-hay:
I agree also as this is the most common default behaviour of any calendar
of events I have come across.

duboisp:
I have worked on the **Option 1** solution, it still incomplete, but here where I am:

Code: https://github.com/duboisp/wet-boew/tree/calendar-pr7671
Working example: (see additional example at the bottom)
* http://universallabs.org/wet/labs/alt-pr7671/unmin/demos/cal-events/cal-events-en.html
* http://universallabs.org/wet/labs/alt-pr7671/unmin/demos/cal-events/cal-events-en.html

Currently I am at debugging an failing test for the *calendar* test suite. It's failing in *reinitialize a calendar* block for the test *should trigger a wb-navigate event with the new year and month*
see: http://universallabs.org/wet/labs/alt-pr7671/unmin/test/test.html

duboisp:
At the PR meeting today, we have discussed about the calendar of event behaviour.

So we will look to have the following behavior:
If Event < Today > Event Then by default show Today
If Event < Today Then show max date of Event
If Today > Event Then show min date of Event

duboisp:
@lucas-hay, @MaximMartel is that the behaviour that you need?

Check the last working example here
*    http://universallabs.org/wet/labs/alt-pr7671/unmin/demos/cal-events/cal-events-en.html
*    http://universallabs.org/wet/labs/alt-pr7671/unmin/demos/cal-events/cal-events-fr.html

If that is ok with you, I will submit a new PR

PR #7685, opened 2016-10-04
Details: Prevented summary disclosure arrows from disappearing in Firefox if JS is disabled.

EricDunsworth:


cybersubtype:
This is a problem even with no changes to Firefox setting the detail/summary toggle arrow is gone.

PR #7715, opened 2016-10-26
Improve ESLinting

LaurentGoderre:


LaurentGoderre:
@nschonni

nschonni:
Should probably use this https://github.com/jquery/eslint-config-jquery since that was the original idea I think. Although that one looks like it's missing the tab setting too.

LaurentGoderre:
AAAAh the jQuery one works great but now there's more stuff to fix.... :(

nschonni:
Try the `--fix` flag since there are alot of rules ESLint will fix for you

LaurentGoderre:
@nschonni almost there but the last remaining error is scaring me...not even sure what it's supposed to be or fix it

LaurentGoderre:
Here's the offending line https://github.com/wet-boew/wet-boew/blob/master/src/core/wb.js#L109-L110

LaurentGoderre:
Ok, fixed it without having to risk the logic

LaurentGoderre:
@nschonni @shawnthompson ok with merging this?

shawnthompson:
@LaurentGoderre I'm ok with it but @nschonni is an expert.

nschonni (approved):
Might be a good time to rename to .eslintrc.json
Looking at some of the old overrides like `max-len` can be done later

LaurentGoderre:
I just kept the ones that differed with jQuery

Issue #7713, opened 2016-10-25
Possible Secure / Unsecure issue in feed.js


crochefort:
Just doing some test and saw that some call to youtube are http instead of the usual //

Like this line : https://github.com/wet-boew/wet-boew/blob/master/src/plugins/feeds/feeds.js#L87

https://github.com/wet-boew/wet-boew/blob/master/src/plugins/feeds/feeds.js#L151

and this line : https://github.com/wet-boew/wet-boew/blob/master/src/plugins/feeds/feeds.js#L444

Got my finding from this page... 
https://www.whynopadlock.com/

So might be good to take a look on the source code to see if it's possible to remove all http:// and juste put // instead...

Since not good with GitHub ... a PR from me is not an option! 

@masterbee @LaurentGoderre what do you think??

Might cause future secured content to not show the padlock if canada.ca is https

kodecount:
Issue #7583 probably relates. Also see #7679.

masterbee:
@crochefort - thanks for this. I agree it would probably be best that we address this issue the same way we have address the original feed with ```wb.pageUrlParts.protocol ```

kodecount:
@masterbee The problem with `wb.pageUrlParts.protocol` is it is also returning `http` which can break the feed plugin. Just look at  #7583.

kodecount:
@masterbee I did some testing on a Government Network and on my personal network. It seems if `wb.pageUrlParts.protocol` produces `http` and it is being block by SmartFilter application.

Here is an example with Pinterest (generated URL):
`http://query.yahooapis.com/v1/public/yql?q=select%20*%20from%20feed%20where%20url%20%3D%20%27https%3A%2F%2Fwww.pinterest.com%2Fhealthycdns%2Fdrug-prevention.rss%27%20limit%205&format=json`

masterbee:
@kodecount - thanks I was under the impression that yahoo query had an http counterpaart but it looks like probably not. Appreciate the test and will tweak this.

kodecount:
@masterbee It does it just not reliable. It seem some networks accept it and some block it. Not sure why.

masterbee:
@kodecount - ahh ok. I have been meaning to look at a more native solution since CORS is now more main stream. What are your thoughts on the native alternative? 

[...]

PR #7719, opened 2016-10-28
removed hard-coded URL values for page specific URL properties

masterbee:
- addresses https://github.com/wet-boew/wet-boew/issues/7713
- increased readability on larger concated strings

### DIFF
--- a/src/plugins/feeds/feeds.js
+++ b/src/plugins/feeds/feeds.js
@@ -69,7 +69,10 @@ var componentName = "wb-feeds",
 				};
 
 			// due to CORS we cannot default to simple ajax pulls of the image. We have to inline the content box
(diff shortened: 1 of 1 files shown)


src/plugins/feeds/feeds.js
```diff
@@ -93,7 +100,8 @@ var componentName = "wb-feeds",
 		 * @return {string}    HTML string of formatted using a simple list / anchor view
 		 */
 		pinterest: function( data ) {
-			var content = fromCharCode( data.content ).replace( /<a href="\/pin[^"]*"><img ([^>]*)><\/a>([^<]*)(<a .*)?/, "<a href='" + data.link + "'><img alt='' class='center-block' $1><br/>$2</a>$3" );
+			var content = fromCharCode( data.content ).replace( /<a href="\/pin[^"]*"><img ([^>]*)><\/a>([^<]*)(<a .*)?/, "<a href='" +
```
kodecount: Would it be possible to change this line before merge? Instead of `data.content` to `data.description`?

ghost (changes requested):

Issue #7709, opened 2016-10-25
WET 4.0 - Colour contrast failure: "Search and menus"


michael-milette:
Good day,

When viewed on mobile devices, the heading "Search and menus" seen at the top of the search/menu fails WCAG 2.0 Level AA minimum colour contrast ratio requirements as seen below.
## Government of Canada Usability theme

Working Example: http://wet-boew.github.io/themes-dist/theme-gcwu-fegc/docs/ref/theme-gcwu-fegc/theme-gcwu-fegc-en.html
![image](https://cloud.githubusercontent.com/assets/3808579/19673745/aff410a0-9a50-11e6-9a48-8f6ebfdacafe.png)
## Government of Canada Intranet theme

Working Example: http://wet-boew.github.io/themes-dist/theme-gc-intranet/docs/ref/theme-gc-intranet/theme-gc-intranet-en.html
![image](https://cloud.githubusercontent.com/assets/3808579/19673774/e85ea518-9a50-11e6-8c67-74ad6c3c2f83.png)
## Base theme

Working Example: http://wet-boew.github.io/themes-dist/theme-base/docs/ref/theme-base/theme-base-en.html
![image](https://cloud.githubusercontent.com/assets/3808579/19673963/84075572-9a52-11e6-9b1f-537753ec12a4.png)
## OGPL theme

Working Example: http://wet-boew.github.io/themes-dist/theme-ogpl/docs/ref/theme-ogpl/theme-ogpl-en.html
![image](https://cloud.githubusercontent.com/assets/3808579/19673822/63666a84-9a51-11e6-8689-e3aa826f8928.png)

I believe that the font in each case should be white as it currently is with the GCWeb and WET themes which seem to be correct.

Let me know if you have any questions.

Best regards,

Michael Milette

masterbee:
@michael-milette  - good catch and thanks for this. Interested in doing a small P.R to tweak the CSS for each of themes or just raising awareness?

michael-milette:
Hi @masterbee - I am currently implementing the https://github.com/wet-boew/wet-boew-moodle project on a voluntary basis - wish I was being paid :-).  This is an open source Moodle theme featuring the WET-BOEW 4.0 framework. As I was testing the mobile views, I just happen to notice it.

By the way, any CodeSprint days coming up in the near future?

michael-milette:
Hi @masterbee, 

For the fix, see pull request https://github.com/wet-boew/wet-boew/pull/7720

[...]

PR #7721, opened 2016-10-29
Fixes colour contrast failure: "Search and menus"

michael-milette:
Reference issue #7709 - updated.

Ensures that the title text "Search and Menus" which appears at the top of mobile menus is white by default. This is the colour used by all current themes.

Michael Milette

### DIFF
--- a/theme/banner/_base.scss
+++ b/theme/banner/_base.scss
@@ -18,6 +18,10 @@
 	color: #fff;
 	min-height: 2.2em;
 	z-index: 1;
+
+	#mb-pnl header.modal-header div.modal-title {
+		color: #fff;
+	}
 }
 
 #wb-lng ul,


LaurentGoderre:
Still not the right place for this.

michael-milette:
Where would you suggest @LaurentGoderre ? I am open to your recommendations.

duboisp:
@michael-milette I do think [base SASS file for lightbox](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/lightbox/_base.scss) should be updated instead of the base theme sass file.

I am waiting for feedback in #7731 before establishing to use white as the color for the dialog.

duboisp:
Please do not merge this PR, as the issue is mainly caused by a regression introduced from PR #7668

LaurentGoderre:
Closing in favour of #7721

PR #7730, opened 2016-11-01
change `gc-elps` to `wb-elps` to use WET namespace

shawnthompson:


shawnthompson:
@neoinsight @coprisanu this is / was being used for the new search styles.

duboisp:
@shawnthompson what that style do? I am not able to find any reference in our [style guide](http://wet-boew.github.io/wet-boew-styleguide/v4/index-en.html)?

shawnthompson:
@duboisp it's not in the style guide. I asked @cfarquharson to add it. That's when we noticed it was using the wrong namespace. 

It adds ellipses / truncates the text to the width of the containing element.

duboisp:
Thanks  @shawnthompson :-)

Issue #7667, opened 2016-09-19
Overlay: background color of bars


shawnthompson:
Why are the overlay bars background-color set to #000?

https://github.com/wet-boew/wet-boew/blob/478d2c628cf006cd90187bc4737f15456582837c/src/plugins/overlay/_base.scss#L111

I'm trying to make an alert style for the bar but I don't want to add a lot of CSS overrides in my GCWeb theme to make it happen.

I will be putting in a PR to remove them but I'd like to see if anyone knows why they were there in the first place.

CC/ @andrewc778

kpduncan:
You also have to change [Line 98](https://github.com/wet-boew/wet-boew/blob/478d2c628cf006cd90187bc4737f15456582837c/src/plugins/overlay/_base.scss#L98) cause it is also giving the header a background color.

nschonni:
If you look at `blame` it's been around since the start of 4 https://github.com/wet-boew/wet-boew/blame/478d2c628cf006cd90187bc4737f15456582837c/src/plugins/overlay/_base.scss#L111, but even that change looks like it was just a reformat and not the original change

shawnthompson:
yeah I don't think it will brake anything and be an improvement. Thanks!

hsrudnicki:
also see and clarify #7512

duboisp:
@shawnthompson I am trying to find the "alert style for the bar" that you are referring in GC web, do you know if it was implemented?

I just want to do an impact analysis before to undo PR #7668

(related to #7731 and #7725)

shawnthompson:
@duboisp The colour was moved to the theme's css because when it was put it base css. Can we add the colour to the demo css so theme's can add their own colours?

duboisp:
A valid default should be at set in wet. What do you think of setting the color #fff where the background color dark blue is set?

shawnthompson:
But it's hard to override that in a theme file without using _!important_.

shawnthompson:
Can we use `inherit` or something?

If we add this back we would have to reopen all these:
 - #7667
 - #7512
 - #6550

duboisp:
ok, I will investigate to find a solution that doesn't require using '''!important'''

PR #7740, opened 2016-11-07
Fix contrast issue in dialog title

### LINKED ISSUE #7709: WET 4.0 - Colour contrast failure: "Search and menus"

### LINKED ISSUE #7731: Lightbox - Accessibility contrast issue in dialog title
The contrast of the title in the dialog do not meet WCAG.

To reproduce:
Start here: http://wet-boew.github.io/v4.0-ci/demos/lightbox/lightbox-en.html
Click on the second item under "Singles items" heading named "AJAX - Example 1"
The contrast ratio of the dialog title is 1.55 where the background color is #2e5274 and the color is #333.

I propose to set a default color like this:
```
.modal-title {
	color: #fff;
}
```
in https://github.com/wet-boew/wet-boew/blob/master/src/plugins/lightbox/_base.scss
or in https://github.com/wet-boew/wet-boew/blob/master/src/plugins/overlay/_base.scss

@cfarquharson what do you think of using the color white on a background #2e5274?

Cheers

duboisp:
fixes #7731
fixes #7709
close #7721
related to #7668

Note: An issue remain as essential CSS was moved to demo CSS of overlay and data-inview plugin see #7667

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -280,3 +280,27 @@ $carousel-s2-tablist-thumbnail-active-border-width: 10px !default;
 $txthl-bg-color: #ff0 !default;
 $txthl-color: #000 !default;
 $txthl-font-weight: 700 !default;
+
+
+//== Overlay
+//
+//##
+$overlay-background-color: #fff;
+$overlay-header-background-color: #2e5274;
+$overlay-header-color: #fff;
+$overlay-bar-background-color: #000;
+
+
+//
+// WET CUSTOM THEME VARIABLES
+//
+
+//== Mobile menu
+//
+//##
+$mb-pnl-background: #193451;
+$mb-pnl-header-background: $overlay-header-background-color;
+$mb-pnl-header-color: $overlay-header-color;
+$mb-pnl-body-background: #0e4164;
+$mb-pnl-body-color: #fff;
+$mb-pnl-body-link-color: #fff;
--- a/src/plugins/data-inview/demo/data-inview.scss
+++ b/src/plugins/data-inview/demo/data-inview.scss
@@ -11,18 +11,6 @@
 	padding: 15px;
 }
 
-.wb-overlay {
-	background-color: #fff;
-}
-
-.overlay-def {
-
-	header {
-		background-color: #000;
-		color: #fff;
-	}
-}
-
 .bar-demo {
 	border: 1px solid #ddd;
 	border-radius: 4px;
@@ -42,8 +30,6 @@
 
 .wb-bar-t,
 .wb-bar-b {
-	background-color: #000;
-
 	header {
 		@extend %data-inview-bar-demo-inline-block;
 	}
--- a/src/plugins/overlay/_base.scss
+++ b/src/plugins/overlay/_base.scss
@@ -8,6 +8,7 @@
  */
 .wb-overlay {
 	background-clip: border-box;
+	background-color: $overlay-background-color;
 	border: 0;
 	border-radius: 0;
 	display: none;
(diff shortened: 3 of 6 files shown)


LaurentGoderre:
I think this is a quick fix that doesn't properly address the whole issue. The overlay and menu plugin probably need some variables.

duboisp:
Thanks @LaurentGoderre I will do

duboisp:
@shawnthompson as per your comments https://github.com/wet-boew/wet-boew/issues/7667#issuecomment-258534446 is this PR will allow you to "override that in a theme file without using !important"?

LaurentGoderre:
Nice! We'd have to add those variables to the themes after this gets merged. Can you also look at possibly having variables for the menu, even if the menu variable in base point to the overlay variable?

duboisp:
I added some variable for configuring the menu

src/_variables.scss
```diff
@@ -280,3 +280,27 @@ $carousel-s2-tablist-thumbnail-active-border-width: 10px !default;
...
+
+//
+// WET CUSTOM THEME VARIABLES
+//
+
+//== Mobile menu
+//
+//##
+$mb-pnl-background: #193451;
+$mb-pnl-header-background: #2e5274;
```
LaurentGoderre: Since those don't end up in the css, could have please make the variable names more verbose.

Also, this variable should point to matchless overlay variable

LaurentGoderre (dismissed):

src/_variables.scss
duboisp: ```mb-pnl``` is the id used in WET theme for the mobile menu.

What do you propose?

LaurentGoderre: `menu-background`

duboisp: Is that means we would need to change the id used in the template also? Do you know the rational of why it used ```mb-pnl``` ?

duboisp: I already know that, but it doesn't explain the rational. Do you know the rational of why ```mb-pnl``` id is used for identifying the mobile menu in the markup?

LaurentGoderre: I don't quite know why that one in particular was chosen but all id were chosen to be short, my guess is that it stands for mobile panel, which now come to think of it, `mobile-menu-background` is way better and more descriptive.

LaurentGoderre: Well, now that I see it, there doesn't seem to be any convention for variable...which is terrible but I am partial;ly to blame for that as well.

LaurentGoderre:
This is not ideal, but neither are the rest of the variable. We will need to clean that up in the next major version

duboisp:
@LaurentGoderre yep, I will update my PR later tonight

LaurentGoderre:
Awesome!

src/_variables.scss
duboisp: I am in for using convention naming if it is documented.

PR #7748, opened 2016-11-08
JSON fetch and data JSON plugin

duboisp:
Insert content extracted from a JSON file. Data are selected as per JSON Pointer URI Fragment, [RFC6901](https://tools.ietf.org/html/rfc6901)

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -1022,6 +1022,7 @@ module.exports = (grunt) ->
 						"flot/jquery.flot.pie.js"
 						"flot/jquery.flot.canvas.js"
 						"SideBySideImproved/jquery.flot.orderBars.js"
+						"jsonpointer/src/jsonpointer.js"
 						"jquery-validation/dist/jquery.validate.js"
 						"jquery-validation/dist/additional-methods.js"
 						"magnific-popup/dist/jquery.magnific-popup.js"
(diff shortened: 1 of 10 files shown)


LaurentGoderre:
You are using this to display raw JSON on a page?

duboisp:
@LaurentGoderre by "raw JSON" I assume you means a "value in a JSON". If so yes. I have various use case from various department that support this particular feature. For example, take department that need to show the same fees for one particular services on multiple static pages. Right now, in order to avoid legal pursuit of publishing outdated fees, they simply put a link to a page with big list of uncomprehensive schedules of fees. And that is not good for the user experience.

Cheers

:-)

LaurentGoderre:
If that is really the case, why not create a plugin on top of the data-ajax plugin that only deals with the JSON traversal instead of duplicating almost all the code?

duboisp:
That was what I tried at the first instance but ajax-fetch used by data-ajax :
* do not support JSON pointer fragment selection, as per RFC6901
* put the response in a DOM object, as that action it not needed for a JSON file
* do not have a local cache control, like if there is more than one use of the data-json plugin that is about querying the same json file, then it was resulting that ajax-fetch downloaded twice the json file.
* Code become messy and heavier when a plugin needs to do two ajax-fetch, like one to get a template and the second to get the data.

PR #7750, opened 2016-11-09
Fixes print issue with tabs

ghost:
DNM


kodecount:
Don't merge. There seems to also be print issue with details when print tabs. Further investigating.

kodecount:
Don't merge. Current issue: overlapping text.

kodecount:
@nschonni @masterbee Now working. :smile:

PR #7753, opened 2016-11-10
Added autoplay to Multimedia Player - Youtube

TTVXStarWake:


nschonni:
WCAG failure, this is also a good read http://www.punkchip.com/autoplay-is-bad-for-all-users/

duboisp:
@nschonni What WCAG Sucess Criteria is failing?

Is the propose change from @MaximMartel do not provide a way to pause the HTML5 media element? I thought the role of using the WET multimedia player plugin was for that.

Although I can understand auto playing a video or a sound could be very distracting, but that is not about accessibility (WCAG) but more about usability research recommendation. So may be the multimedia player documentation could advertise that auto playing a video is not recommended by usability research recommendation.

Cheers

nschonni:
The controls on the regular player are not a sufficient technique https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-dis-audio.html since it will interfer with screen reader software as soon as the page is loaded. That is besides the usablitiy problems.

duboisp:
@nschonni I understand it is not a best industry practice and it should be avoided as much as possible, but it is not a WCAG failure. May be we can ask @MaximMartel, to include with his PR to update the multimedia player working example and documentation with adding a clear warning about using the autoplay feature which will include all the necessary reference to WCAG and relevant best practice documentation.

As you referenced the [SC 1.4.2](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-dis-audio.html), there is a note about playing audio automatically. 

> _Note:_ Playing audio automatically when landing on a page may affect a screen reader user's ability to find the mechanism to stop it because they navigate by listening and automatically started sounds might interfere with that navigation. Therefore, we discourage the practice of automatically starting sounds (especially if they last more than 3 seconds), and encourage that the sound be _started_ by an action initiated by the user after they reach the page, rather than requiring that the sound be _stopped_ by an action of the user after they land on the page.

From my point of view it is only a discouraged practice not a failure. The failure is when [there is no mechanism to turn of a sound that are playing longer than 3 seconds (F23)](http://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F23) and for the [absence of a way to pause or stop an HTML5 media element that autoplays (F93)](http://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F93)



_FYI - Techniques and failures (W3C Working Group Note) are updated on a 6 months release cycle._

duboisp:
As the [sufficient technique G170](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G170) the multimedia player could be updated to provide a mechanism to the user to stop autoplay near the beginning of the page.

LaurentGoderre:
Autoplay is not appropriate in almost all situations but there may be some cases where it is appropriate. I'm not opposed to getting this in and the onus is on the content author to use it properly.

MaximMartel:
Just a thought here, but if we have a dismiss able content box right underneath the h1, that mention to the user that they can dismiss the autoplay features on this page by clicking a button ?

kodecount:
@MaximMartel Random thought could it be possible to add a disable sound link to the wb-disable list only if a page has a auto-play video? Also could advise user to have a alert with `role="alert"` so the screen readers well automatically read the warning.

MaximMartel:
Great thought @kodecount. Only disabling the sound on a page that contains a autoplay video should do the trick. More work needs to be done prior to re-opening this issue. Thanks to all that commented on this PR so far, this is great.

LaurentGoderre:
This seems overkill and likely to not work the way you expect. I would say that user should only use autoplay if there's only one video per page and only in a video gallery or if it's opened via an overlay where it's clear that a video will play

duboisp:
@MaximMartel May be not a dismiss able content box as we also want to discourage that practice. I do believe the "play/stop" control should be on the page automatically when one of the multimedia player in the page have the property "autoplay" or another widget are auto playing something (like the carousel). 

Under the ```<h1>```  are fine with me, if the controls are immediately next to it. May be the button could located on the right so it will interfere less with the page content. But that presentation would fail if the video is played in a overlay.

@LaurentGoderre indeed, may that could be more instruction in the documentation about the auto play functionality.

kodecount:
@LaurentGoderre It could be yes...

@MaximMartel @duboisp What about having the video player sound automatically mute when auto playing and unmute when the user hovers over the video with mouse. This seems to be a more common practice.

duboisp:
> What about having the video player sound automatically mute when auto playing and unmute when the user hovers over the video with mouse. 

That is not the behavior as described in the HTML5.2 spec - https://www.w3.org/TR/html52/semantics-embedded-content.html#the-video-element
> ```autoplay``` - Hint that the media resource can be started automatically when the page is loaded 

@kodecount The behavior that you are describing is not when the page is loaded. Although the behavior that you described could be added to the multimedia player like by using a dada-* attribute but I am not a fan of adding that kind of functionality when we try to discourage the use of autoplay.

:smiley:

PR #7759, opened 2016-11-15
Fix print issues with tabs

### LINKED ISSUE #7373: Print active tabbed interface not working well in v4.0.19+2
We are having issues with the Tabbed interface we use for the Travel advice and advisories at Travel. It seems to be responding erratically to printing in different browsers. We are currently using v4.0.19+2 and implemented the Print-active class to print only one tab, but it is not indicative to the user what the other tab titles would be or even if there is content there. The Print-active class is supposed to be good for v4.0.15+. but some browsers have printed some content but not all as well. ANy help would be appreciated, thanks.

The default doesn't work well for print.

Here is a link to the page: http://travel.gc.ca/destinations/azores 

See figure:
acores

ghost:
Fixes #7373

### DIFF
--- a/src/plugins/tabs/_print.scss
+++ b/src/plugins/tabs/_print.scss
@@ -19,6 +19,7 @@
 	[role="tabpanel"] {
 		display: block !important;
 		opacity: 1 !important;
+		overflow: visible !important;
 		position: static !important;
 		visibility: visible !important;
 
@@ -29,6 +30,14 @@
 		summary {
 			display: list-item !important;
 		}
+
+		&.noheight {
+			max-height: none;
+		}
+	}
+
+	& > .tabpanels {
+		overflow: visible !important;
 	}
 
 	&.print-active {


src/plugins/tabs/_print.scss
```diff
@@ -18,7 +18,9 @@
 
 	[role="tabpanel"] {
 		display: block !important;
+		max-height: none !important;
```
EricDunsworth: There might be value in doing something like this for ``max-height`` (to avoid needing to resort to ``!important``):

```
.wb-tabs {
	[role="tabpanel"] {
		...

		&.noheight {
			max-height: none;
		}
	}
}
```

EricDunsworth:
Btw this PR benefits carousels in a few ways when printing:
* Prevents vertical scroll bars from appearing to the right of multimedia players in Blink browsers and in all panels in IE11.
* Prevents latter carousel panel examples from getting cut off in Blink browsers. Previously, the ``noheight`` class' ``height: 0;`` property was probably causing Blink to miscalculate the page's total height.

src/plugins/tabs/_print.scss
kodecount: All updated. Thanks.

LaurentGoderre:
Can you please amend your PR to only focus on the printing issue sicne the background color issue has been addressed by #7858

kodecount:
@LaurentGoderre File has been removed.

lucas-hay (approved):
Approved by PR review group.

PR #7770, opened 2016-11-17
Fix datepicker in a overlay

ghost:
DNM

### DIFF
--- a/site/pages/docs/versions/rdmp-en.hbs
+++ b/site/pages/docs/versions/rdmp-en.hbs
@@ -28,7 +28,7 @@
                 <td>v4.0.24 release</td>
             </tr>
 			<tr>
-				<td><time>2016-02-14</time></td>
+				<td><time>2016-11-16</time></td>
 				<td>v4.0.23 release</td>
 			</tr>
 		</tbody>
--- a/site/pages/docs/versions/rdmp-fr.hbs
+++ b/site/pages/docs/versions/rdmp-fr.hbs
@@ -28,7 +28,7 @@
                 <td>Lancement de v4.0.24</td>
             </tr>
 			<tr>
-				<td><time>2016-11-14</time></td>
+				<td><time>2016-11-16</time></td>
 				<td>Lancement de v4.0.23</td>
 			</tr>
 		</tbody>


src/polyfills/datepicker/datepicker.js
```diff
@@ -243,8 +243,21 @@ var componentName = "wb-date",
...
+		if ( isOverlay( $container, "modal-content" ) ) {
+			$container.attr( "style", "overflow: hidden;" );
+		} else {
+			$container.attr( "style", "top:" + ( position.top + field.offsetHeight ) + "px;left:" + position.left + "px" );
+		}
+	},
+
+	isOverlay = function( elm, className ) {
+		while ( ( elm = elm.parent() ) && !elm.hasClass( className ) ) {
+			if( elm.is( "main" ) ) {
```
kodecount: This prevents an infinite loop when the class is not found.

src/polyfills/datepicker/datepicker.js
```diff
@@ -243,8 +243,21 @@ var componentName = "wb-date",
 		var field = calendar.field,
 			position = calendar.$field.offset();
 
-		$container
-			.attr( "style", "top:" + ( position.top + field.offsetHeight ) + "px;left:" + position.left + "px" );
+		if ( isOverlay( $container, "modal-content" ) ) {
+			$container.attr( "style", "overflow: hidden;" );
```
kodecount: This hides the overlay close button which also dismiss the actual overlay it is inside of.

src/polyfills/datepicker/datepicker.js
```diff
@@ -151,7 +151,7 @@ var componentName = "wb-date",
 		// Disable the tabbing of all the links when calendar is hidden
 		$container.find( "a" ).attr( "tabindex", "-1" );
 
-		$( "main" ).after( $container );
+		$( elm ).after( $container );
```
kodecount: Instead of adding the container to the page. Add it below the input area.

LaurentGoderre: This was the way it was before and it was switched to be in the overlay to avoid a bunch of issues, so merging this would bring back these issues. Also, this way, there is only one instance of the datepicker that gets moved to the field once needed,. Applying this would mean that only the last field of the page woudl work and not the others ones

LaurentGoderre (changes requested):

LaurentGoderre:
:-1: we need to deal with multiple overlay interacting with each other instead

Issue #7766, opened 2016-11-16
Legend tag Float:left


StdGit:
@thomasgohard

Is there a special reason why tag Legend is floating left ?

legend { 
   border-bottom: 0; 
  **float: left;**
 } 

If you add a tag like fieldset after it, it will be right aligned

https://github.com/wet-boew/wet-boew/commit/42a76b1bffa30b3103542892b23f4cbf6ca5d14a

https://github.com/wet-boew/wet-boew/blob/a5b906efd6b5cc2d74167d53da9bbe02b4150b11/src/base/bootstrap-overrides/_base.scss#L601

kodecount:
It because [Line 609](https://github.com/wet-boew/wet-boew/blob/a5b906efd6b5cc2d74167d53da9bbe02b4150b11/src/base/bootstrap-overrides/_base.scss#L609) doesn't work other wise.

kodecount:
What about using:

```css
legend {
  margin-bottom: 0;
}

legend:after {
    border-bottom: 1px solid #e5e5e5;
    content: "";
    display: block;
}
```
This well create a border under the legend title if we really need the border?

kodecount:
Here what it would look like with the updated code:

![border_update](https://cloud.githubusercontent.com/assets/23484460/20361878/8edb6d34-ac0f-11e6-90e1-ec6e5dde2fcb.png)

[Line 599-602 and 609-610](https://github.com/wet-boew/wet-boew/blob/a5b906efd6b5cc2d74167d53da9bbe02b4150b11/src/base/bootstrap-overrides/_base.scss#L599) would need to be removed and replaced with above code.

kodecount:
@StdGit What do you think?

StdGit:
@kodecount 
It is ok with me if it doesn't break anything else.
Thx for your help

kodecount:
@StdGit Started the pull request to fix this.

duboisp:
@cfarquharson 
* Do you have any comment about removing ```float:left``` to ```legend``` element?
* What do you think of having the border under the legend instead at the top?

:smiley:

kodecount:
@duboisp To have it on the top if desired just need to change `legend:after` to `legend:before` personally don't think `float:left` was the best way to get a border.

StdGit:
@duboisp @kodecount  
It's ok with me to remove the Float:Left and I prefer the border under the legend

duboisp:
[...]

PR #7771, opened 2016-11-17
Fix issue with legend border

ghost:
Fixes issue #7766

### DIFF
--- a/src/base/bootstrap-overrides/_base.scss
+++ b/src/base/bootstrap-overrides/_base.scss
@@ -620,6 +620,20 @@ fieldset {
 	}
 
 	/*
+	 *  Allows fieldSet/legend to have a bottom border under legend title
+	 *	https://github.com/wet-boew/wet-boew/issues/7766
+	 */
+	&.legend-brdr-bttm {
+		border-top: 0;
+
+		legend {
+			border-bottom: 1px solid $legend-border-color;
+			float: none;
+			margin-bottom: 10px;
+		}
+	}
+
+	/*
 	 *	Fieldset/legend as a label for a group of checkboxes or radio buttons
 	 *	https://github.com/wet-boew/wet-boew/issues/5797
 	 */


kodecount:
@duboisp @LaurentGoderre The form-style guide was updated. Please merge this when able to.

duboisp:
@kodecount I will need to do some testing before to merge your PR. I think your PR remove completely the border-top where it should remain as the default behavior.

kodecount:
DEFAULT VIEW:
default

WITH CLASS VIEW:
class

kodecount:
DEFAULT VIEW:
default

The default view is not completely the same due to the border being present on each field set, but the original way it achieved the effect of only adding a border when there more then one field set was using a CSS hack. @cfarquharson and @LaurentGoderre Thoughts?

CLASS VIEW:
class

kodecount:
@duboisp Fixed it to keep original code

duboisp (changes requested):
Visually it's seem to work fine, but the fieldset keep his border-top where it is hidden because the float of the legend element is changed. I think the ```border-top``` style should be removed as the intention is to not have it.

Can you make adjustment where the class ```legend-brdr-bttm``` would be applied to the ```fieldset``` element instead of the ```legend``` element, then it would be easier to remove the border-top that was set on the fieldset.

kodecount:
@duboisp It has been fixed to use the fieldset and not use the legend to create a border.

kodecount:
DEFAULT VIEW (When their is more then one fieldset group):
default

CLASS VIEW (using the new `.legend-brdr-bttm` on a fieldset):
legend-brdr-bttm

kodecount:
@duboisp Any ideas just did some testing works great in all but Firefox.

kodecount:
@duboisp Firefox isn't working apparent their is a know issue. Have to use:
```
fieldset {
position: relative;
}
legend {
position: absolute;
}
```
or
```
legend {
float: left;
}
```
Their is another fix with `@-moz-document url-prefix()` but that breaks grunt.

duboisp:
@kodecount 
> [...] but Firefox.

:confused:, it do not work as expected in Firefox??

May be it could be because the ```border-bottom``` should be set on the ```legend``` element instead of setting a ```border-top``` on the next elements following the ```legend```. 

FYI - As per the spec, the [```legend```](https://html.spec.whatwg.org/multipage/forms.html#the-legend-element) is the first chlid of a ```fieldset``` element. Also as the the [```fieldset```](https://html.spec.whatwg.org/multipage/forms.html#the-fieldset-element) spec, the ```legend``` is optional.

kodecount:
@duboisp Firefox isn't working apparently their is a know issue. Have to use:
```
fieldset {
position: relative;
}
legend {
position: absolute;
}
```
or
```
legend {
float: left;
}
```
Their is another fix with `@-moz-document url-prefix()` which allows CSS just for Firefox, but that breaks grunt.

kodecount:
@duboisp `ERROR: Unknown @ rule: @-moz-document. This rule looks for recoverable syntax errors. (errors) Browsers: All` this is the error with Firefox rules.

kodecount:
@duboisp The new update should fix Firefox

duboisp:
Thanks @kodecount, your solution seems to work fine, I tested on FF Dev. Chrome, Edge and IE11.

But I wonder the why the following css was used:

Why the top border was removed on the first child element of fieldset and not on the legend? [line 612](https://github.com/wet-boew/wet-boew/blob/master/src/base/bootstrap-overrides/_base.scss#L612)
```
&:first-child {
 		border-top: 0;
 	}
```
Why float was used in the legend? [line 601](https://github.com/wet-boew/wet-boew/blob/master/src/base/bootstrap-overrides/_base.scss#L601)
Why the bottom border was removed to the legend? [line 600](https://github.com/wet-boew/wet-boew/blob/master/src/base/bootstrap-overrides/_base.scss#L600)
```
legend {
	border-bottom: 0;
	float: left;
}
```

Then I just remarked this note:  [line 590](https://github.com/wet-boew/wet-boew/blob/master/src/base/bootstrap-overrides/_base.scss#L590)
```
/*
 *	Fieldset/legend border position
 *
 *	Bootstrap puts the border below the legend which creates a visual
 *	separation between the legend and the fields to which it relates. Putting
 *	the border above eliminates that visual separation. Putting it on the
 *	fieldset element instead ensures that all fieldsets will have a separation
 *	even when no legend is used.
 */
```

So, @LaurentGoderre what do you think of undoing then re-doing practice? Do you think it is ok for this use case?

kodecount:
@duboisp The file you reference is not the code change.

kodecount:
@duboisp The file you reference is not the code change. That is the original code.

What original code does:

The original code goes against bootstrap of having the border under the legend instead applies the border to above each `fieldset` if their is more then one `fieldset` the issue with this is Firefox behaves differently then most browsers and in order for the border to work the float is required. 

What my current code does:

What the latest version of the code change I provided does is apply the border to the `legend` and put the border above the `legend` or below the `legend` dependent on if a class is added to the `fieldset`. The downside to this method is their is no border when a legend is not added. I'm personally okay with this cause if your to separate fields their should be context to why.

duboisp:
@kodecount Personally I am ok with your solution and as I said it works fine during my testing. 

But I just need a second opinion about the practice of undoing style then redoing style for this particular use case before to merge it.

src/base/bootstrap-overrides/_base.scss
```diff
@@ -598,18 +598,19 @@ blockquote {
...
 %border-style-solid {
 	border-style: solid;
 }
 
 fieldset {
-	border-top: 1px solid $legend-border-color;
 	padding-top: 10px;
 
-	&:first-child {
+	&:first-child legend {
```
LaurentGoderre: Please revrwite this in propper SCCSS

```scss
&:first-child {
    legend {
        [...]
    }
}
```

src/base/bootstrap-overrides/_base.scss
```diff
@@ -620,6 +621,15 @@ fieldset {
 	}
 
 	/*
+	 *  Allows fieldSet/legend to have a bottom border under legend title
+	 *	https://github.com/wet-boew/wet-boew/issues/7766
+	 */
+	&.legend-brdr-bttm legend {
```
LaurentGoderre: Please rewrite this in proper SCSS

```scss
&.legend-brdr-bttm {
    legend {
        [...]
    }
}
```

LaurentGoderre (dismissed):

LaurentGoderre:
@duboisp I feel that it's risky and can create some conflicts in the future.

kodecount:
@LaurentGoderre Newer bootstraps are actually removing borders all together.

kodecount:
@LaurentGoderre Reverted code to use original method for top border.

LaurentGoderre:
I like that it's class specific now so less risky

duboisp (approved):

duboisp:
I tested, it work fine and it is [already documented](http://wet-boew.github.io/wet-boew-styleguide/v4/design/forms-en.html#fieldset). I am ok to merge this PR.
 
@LaurentGoderre do you have any other concern about this PR?

LaurentGoderre:
No issue since it's isolated

kodecount:
Don't forget to close #7766

PR #7305, opened 2015-12-18
Tabbed Interface: Documentation Typo

codejunkyard:

PR #7773, opened 2016-11-18
Tabbed Interface: Documentation Typo

duboisp:
close #7305

Same PR but without the unrelated commit.

/cc @gvautour @LaurentGoderre


LaurentGoderre:
Have you tested it ?

duboisp:
no, but it should work as described. 
I look at the tabs.js code and the event exists and require "id" parameter as described in the doc. 

this pr only fix a copy and paste error introduced when the documentation was written. 

I can understand that you might want to ensure that all existing documentation is accurate, I am for it too, but that is not the goal here and verifying the documentation accuracy is a totally different task. 

Cheer

LaurentGoderre:
If you checked the code, then that's fine. I just wanted to make sure it was the right event

Issue #7688, opened 2016-10-11
wb-popup-mid class not horizontal align centre in IE11


Christopher-O:
The newly implemented wb-popup-mid class does not properly align in Internet Explorer 11.  It works fine in Firefox but in IE11, it aligns to the top and the left of the browser.    I have created a IE11 only hack for our custom CSS, but it might be something to add to a future update of WET 4.   Thanks.

@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {  
.wb-popup-mid {
    border-radius: 6px;
    bottom: 0px;
   left: 0px;
    margin: auto;
    right: 0px;
    top: 0px;
   width:90%;
   max-height:90%;
  max-width:90%;

}
}

masterbee:
@Christopher-O - thanks for this. I have tested your issue out on my machine but was not able to replicate. I have a few custom themes and tweaked my windows native theme a bit so that could be the culprit. 

Do you mind putting a screenshot in this thread or the before and after? Also, since you did all the work for the fix, any interest in P.R'ing that is as well? No sense some else get the credit for your hard work 🐹

Christopher-O:
![modalalign](https://cloud.githubusercontent.com/assets/9367162/19892152/1d602b02-a009-11e6-80a7-cafe82d2563d.png)
Here is a screenshot of an example. My apologies, its more just left aligning rather than at the top.  
And also I have never done a Pull Request before so I have no problem someone taking the lead on that.  I would just be happy to contribute.

Christopher-O:
![modalaligncntr](https://cloud.githubusercontent.com/assets/9367162/19892410/5e292b1a-a00a-11e6-97c2-dd5975a81618.png)
And for reference, here is the overlay when the CSS is applied.

cfarquharson:
@masterbee can this be merged into V4.0.23 instead?

kodecount:
@cfarquharson Posted a PR. New to SASS so hope it good enough.

cfarquharson:
@kodecount Did it get merged into WET4.0.23?

kodecount:
@cfarquharson No. PR #7446.

kodecount:
@cfarquharson No. It won't be added till v4.0.24

duboisp:
As discussed in #7775, the solution proposed by @Christopher-O require further investigation to know if an alternative solution, without any CSS hack, can be used. 

[...]

PR #7775, opened 2016-11-22
Overlay CSS Fix (IE11)

ghost:
Fix is related to issue #7688 and based off @Christopher-O suggestion

### DIFF
--- a/src/plugins/overlay/_base.scss
+++ b/src/plugins/overlay/_base.scss
@@ -75,6 +75,7 @@
 	margin: auto;
 	right: 0;
 	top: 0;
+	width: 90%;
 }
 
 .wb-popup-full {


LaurentGoderre:
I don't like the idea of putting in CSS hacks as a permanent solution

duboisp:
@LaurentGoderre I do not like CSS hacks either, but what kind of solution do yo suggest instead? Do you think, the use case (#7688) that support that fix require more research and documentation?

LaurentGoderre:
I think we should investigate further

duboisp (dismissed):
The use case #7688 require further investigation. The goal is to avoid to use any CSS hack

cfarquharson:
@LaurentGoderre who is going to investigate this?  I fear that by delaying the addition of the CSS and saying it will be 'investigated', that no one will do it and then there will be no solution and IE will continue to display the overlay incorrectly.

Christopher-O:
As stated in #7688 the injected code was due to an inability to adjust the current CSS to fix the problem in IE.  Any fix that was done would then "break" the CSS in Firefox.

LaurentGoderre:
The opposite can be said as well, if dirty code gets in, it may stay there for a long time.

duboisp:
Test OK. I was able to reproduce the issue and this PR fix it.

close #7688

LaurentGoderre:
Much better !

PR #7781, opened 2016-11-25
Build: Append the commit ID to the Travis commit messages

LaurentGoderre:


LaurentGoderre:
@nschonni can you please review this?

nschonni (approved):
LGTM, but i think the "maintenance" word can be removed, although I know it was there before

LaurentGoderre:
Interesting idea! If we are to do this, might as well be specific and say which release it is.

LaurentGoderre:
oh, we're already doing this

LaurentGoderre:
OK, done!

PR #7785, opened 2016-11-28
Build: Added targets to remove npm and bower dependencies

LaurentGoderre:
Replaces #7625

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -1189,6 +1189,8 @@ module.exports = (grunt) ->
 
 		clean:
 			dist: ["dist", "src/base/partials/*sprites*"]
+			npm: "node_modules"
+			bower: "lib"
 
 		watch:
 			options:


shawnthompson:
Should document this task?

LaurentGoderre:
Wiping the npm and bower cache goes beyond the scope of what grunt should do in a project. In docker, this becomes a moot point since you don't preserve state between runs.

shawnthompson:
True but not everyone uses docker!

LaurentGoderre:
In any case wiping cache is propably the expected behavior of these tasks. In the rare case where you need to do it, I don't see why you couldn't run the commands manually

Gruntfile.coffee
```diff
@@ -1189,6 +1189,8 @@ module.exports = (grunt) ->
 
 		clean:
 			dist: ["dist", "src/base/partials/*sprites*"]
+			npm: "node_modules"
```
LaurentGoderre: It works because it loads it all in memory before running it

LaurentGoderre:
@nschonni any issue in merging this?

duboisp:
@LaurentGoderre here my journey when I run ```grunt clean``` 

1. I got an error saying:
> Warning: Unable to delete "node_modules" file (EPERM: operation not permitted, unlink 'wet-boew\node_modules\node-sass\vendor\win32-x64-4S\binding.node).

2. Then when I tried a second time, I can not find ```grunt```, which make sense because I just deleted with running the clean task.

3. Then I ran again the script/setup.

4. Then I build wet by running ```grunt```

5. Then I ran again ```grunt clean```

6. I got the same warning error.

----

Is those "special" grunt task are intended to be documented somewhere instead of looking at the actual Gruntfile.coffee code and figure out what it does?

Because I am thinking that instead of automating it, simple instruction could do a better job. It could be short as: 

**Reset your building environment:** Manually delete the "node_modules" and the "lib" folder, then run "script/setup". Doing that operation could fix issue like when .....

LaurentGoderre:
@duboisp that might not be as easy as you think. That task works perfectly with Linux and Mac OS. The issue you have are Windows specific, and you might get issues too if you do it manually

duboisp:
Thanks for the explication @LaurentGoderre Yes I only test-it on windows.

Who and in what circumstance users will use that cleaning task, as it only work for mac and linux?

LaurentGoderre:
There's no point of it doesn't work on Windows, unfortunately

shawnthompson:
Worked fine with my Bash script on windows using git bash. 

Shawn Thompson 
Plansmash@me.com

duboisp:
I used the bash script on Windows 10 Pro and it didn't worked.

shawnthompson:
My original PR

Shawn Thompson 
Plansmash@me.com

Issue #7615, opened 2016-07-19
Overlay print issue


biljanap:
The content of overlay prints on top of the page content rendering both unreadable. This is the case with all types of overlays except Centred popup lightbox that is placed at the start of the page when printed.
It seams that WET4 removes all backgrounds for the print. Is it possible to keep background at least for full screen overlays? And maybe not even print the rest of the page?
Maybe exclude Left-panel, Right-panel, Top bar and Bottom bar from print all together like it's done for main nav, footer and search? Just a suggestion. 
It would be ideal for my current project if we would be able to print just the content of the full screen overlay and ignore the rest of the page. Any suggestions?

kpduncan:
Suggestion would be to have a web page which is just a modal and change the buttons of the modal to redirect to the desired page.

kodecount:
@biljanap Do you mean when print the page? If so I don't believe their is any print support with overlays?

@masterbee @duboisp Do you know if overlays have print support? I tired printing a page while overlay active and it did as described here. The overlay text was over top the of the content under the overlay.

biljanap:
@kodecount yes, it's when printing the page.

kodecount:
@duboisp @LaurentGoderre What about adding css like:

```css
@media print {
.modal-content {
display: block !important;
}
.modal-content.open {
display: none !important;
}
.modal-content.no-print {
display: none !important;
}
}
```

LaurentGoderre:
We are trying to not use `!important` as much as possible to avoid inheritance hell so perhaps it would need to be tweaked a bit

kodecount:
@LaurentGoderre I understand that, but won't work otherwise. This CSS is only called on a print event so shouldn't have any major effects.

kodecount:
@LaurentGoderre Never mind modified original comment.

kodecount:
@LaurentGoderre Updated to reflex your suggestion. Well require at least one `!important`.

LaurentGoderre:
@kodecount could you rewrite in SCSS?

kodecount:
@LaurentGoderre Updated. This would have to be added to a new file called `_print.scss` and added after [Line 250 on _wet-boew.scss](https://github.com/wet-boew/wet-boew/blob/master/src/base/_wet-boew.scss#L250)

LaurentGoderre:
@kodecount close but not quite. You should read about the SCSS syntax.

kodecount:
[...]

PR #7791, opened 2016-11-30
Add print to overlays (Needs Testing)

ghost:
Fix issue #7615 

This setups overlays to print by default and to disable print must add `no-print` class (this could be changed to the opposite effect if required).

### DIFF
--- a/site/pages/docs/ref/helloworld/helloworld-en.hbs
+++ b/site/pages/docs/ref/helloworld/helloworld-en.hbs
@@ -1,0 +1,102 @@
+---
+{
+	"title": "Hello world",
+	"language": "en",
+	"category": "Other",
+	"categoryfile": "other",
+	"description": "A quick tutorial on how you can create your own WET plugin.",
+	"altLangPrefix": "helloworld",
+	"dateModified": "2016-12-06"
+}
+---
+<div class="wb-prettify all-pre hide"></div>
+
+<section>
+	<h2>Purpose</h2>
+	<p>Learn on how to create a simple plugin for WET</p>
+</section>
+
+<section>
+	<h2>Working example</h2>
+	<ul>
+		<li><a href="../../../demos/helloworld/helloworld-en.html">English example</a></li>
+		<li><a href="../../../demos/helloworld/helloworld-fr.html" hreflang="fr">French example</a></li>
+	</ul>
+</section>
+
+<section>
+	<h2>Source code of <code>helloworld.js</code></h2>
+<pre><code>/**
+ * @title WET-BOEW Hello world plugin
+ * @overview Plugin contained to show an example of how to create your custom WET plugin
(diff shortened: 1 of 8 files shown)


LaurentGoderre:
Can you replace all the spaces with tabs? You should use the editorconfig plugin for it so you don't need to worry about it

kodecount:
Sorry the power went out. Get to it later.

kodecount:
@LaurentGoderre Can't seem to get editorconfig working.

LaurentGoderre:
What is your issue?

kodecount:
bug

kodecount:
bug

LaurentGoderre:
I disabled my whitespace plugin I think

kodecount:
@LaurentGoderre Thanks for help I had too enable a indent/white-space indicator and remove white-space manually.

duboisp:
I test it and all the overlay fragment coded in the page are showing where it was coded, that in the print view. 

I think it is expected to show only the active open overlay not all of them isn't it?

Here what it look like in PDF:
* [Currently in WET, when a middle screen overlay is open](http://universallabs.org/wet/labs/temporary/current-middlesceen-overlay.pdf)
* [This fix, when a middle screen overlay is open](http://universallabs.org/wet/labs/temporary/PR7791-middlescreen-overlay.pdf)
* [Currently in WET, when a full screen overlay is open](http://universallabs.org/wet/labs/temporary/current-fullscreen-overlay.pdf)
* [This fix, when a full screen overlay is open](http://universallabs.org/wet/labs/temporary/PR7791-fullscreen-overlay.pdf)
* [This fix, when a center popup is open](http://universallabs.org/wet/labs/temporary/PR7791-centerpopup-overlay.pdf)

@kodecount do you think it is feasible to only show the active overlay when printing?
I think that is behavior the user will expect to get when printing an active overlay. 

Just an idea regarding the location. I do not know if it will worth to update the overlay plugin to toggle a class in the ```html``` element when an overlay is open. May be that could help to better manipulate the print CSS when printing an open overlay and to hide the rest of the page content.

kodecount:
@duboisp Yes and no. The problem that occurs when trying to hide the page contents is if the overlay is located in the `<main>` (that and it is also dependent on if the user who codes the page put all the content in the `<main>`) cause when you try to hide the page contents you hide the overlay in some cases.

As for only printing the active overlay I say that is possible. We probably use the `.open` class to determine if the modal is open and as before if the user doesn't want that modal to print they just add the `.no-print` class.

kodecount:
@duboisp Code was updated to print only the active overlay with the page content.

duboisp:
@kodecount sorry I forget to put the PDF file on my server.

By modifying slightly the wb-lbx and wb-overlay plugin, I found a way to only print the active overlay/popup  preceded by the page header and the page title (h1). Also I moved the popup opened with wb-lbx after the h1 where it's now provide more context when printed.

PR #7798, opened 2016-12-05
Added Sub-resources basic integrity generation

LaurentGoderre:
Sub-resource integrity is now a W3C recommendation and supported by many major browsers. https://www.w3.org/TR/SRI/

I also created all the SRIs for all v4 releases including themes here:

https://github.com/LaurentGoderre/wet-boew-sri

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "grunt-sass": "^1.0.0",
     "grunt-sass-lint": "^0.2.0",
     "grunt-saucelabs": "^9.0.0",
+    "grunt-sri": "^0.2.0",
     "grunt-wet-boew-postbuild": "^0.1.3",
     "grunt-wget": "~0.1.0",
     "load-grunt-tasks": "^3.2.0",
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -56,6 +56,7 @@ module.exports = (grunt) ->
 		->
 			if process.env.TRAVIS_PULL_REQUEST is "false" and process.env.DIST_REPO isnt `undefined` and ( process.env.TRAVIS_TAG isnt "" or process.env.TRAVIS_BRANCH is "master" )
 				grunt.task.run [
+                    "sri"
 					"copy:deploy"
 					"gh-pages:travis"
 					"gh-pages:travis_cdn"
@@ -1382,6 +1383,26 @@ module.exports = (grunt) ->
 					branch: process.env.DEMOS_BRANCH
 					message: "<%= grunt.config('gh-pages.travis.options.message') %>"
 					silent: true
+
+		sri:
+            options:
+                pretty: true
+            wet_boew:
+                options:
+                    dest: "<%= coreDist %>/payload.json"
+                cwd: "<%= coreDist %>"
+                src: [
+                    "{js,css}/**/*.{js,css}"
+                ]
+                expand: true
+            theme_wet_boew:
+                options:
+                    dest: "<%= themeDist %>/payload.json"
+                cwd: "<%= themeDist %>"
+                src: [
+                    "{js,css}/*.{js,css}"
+                ]
+                expand: true
 
 		checkDependencies:
 			all:


package.json
```diff
@@ -58,6 +58,7 @@
     "grunt-sass": "^1.0.0",
     "grunt-sass-lint": "^0.2.0",
     "grunt-saucelabs": "^9.0.0",
+    "grunt-sri": "LaurentGoderre/grunt-sri#add-cwd",
```
LaurentGoderre: This is until https://github.com/neftaly/grunt-sri/pull/12 lands

duboisp:
@LaurentGoderre it seems to be a very nice feature, I like it but I am questioning his usefulness. It seems to be very specific to a WET variant implementation in a secure context. Do we really help them by generating the [integrity metadata](https://www.w3.org/TR/SRI/#integrity-metadata)? It seems easy to generate and what about it they have a custom build of WET?

How the series of "payload.json" file that you have created can/will be used?

I do not saw in your PR any modification of the wet template that implement the subresource integrity, like:
```html
<script src="https://example.com/example-framework.js"
        integrity="sha384-Li9vy3DqF8tnTXuiaAJuML3ky+er10rcgNR/VqsVpcw+ThHmYcwiB1pbOxEbzJr7"
        crossorigin="anonymous"></script>
```

As per the spec, it only recommend to use "intergrity" check for pages in a secure context - https://www.w3.org/TR/SRI/#h-note5

For now, only ```<link>``` and ```<script>``` elements support that new attribute - https://www.w3.org/TR/SRI/#elements. 

They recommend to setup the server in a way that they will send the HTTP header ```Cache-Control: no-transform``` for file that their integrity need to be checked - https://www.w3.org/TR/SRI/#proxies 

That feature is not supported yet by IE and Edge as per http://caniuse.com/#feat=subresource-integrity

It is nice to see that other elements, like anchor, may are supported in a future revision of the spec - https://www.w3.org/TR/SRI/#h-note9.

LaurentGoderre:
This addresses one of the biggest concerns to linking to a CDN for WET resources: what if the server is compromised and malicious code is inserted. If you add integrity checks to your markup and the file is compromised, the script won't load.

I don't want to force this on any implementation but rather make it available for those who chose to use it. I am not concerned about custom builds because those who choose to do that have to maintain security on their own. I still think the best way to extend WET is to create a separate script than modifying WET directly. 

Having this only on script and link is all we need for the CDN use case and it's more than likely that CDN servers will be configured correctly with these headers.

The fact that IE and Edge don't support those is beside the point since it just means these browsers are less secure until they implement it.

duboisp:
Thanks @LaurentGoderre and I understand.

But should the integrity metadata be added directly to the script element and the link element on every pages?

rlambert27:
We added SRi for all files that were non-WET that use the CDTS, so basically the jQuery files.

Something that prevented us from applying to all WET files, and this is just a thought for you guys, is the fact that when you perform an upgrade you'll now also need to update the hash.

So now instead of being able to upgrade your WET by simply updating your files this will no longer be sufficient, you'll need to update the hash for each file when you perform an upgrade. Depending on website setup this could be really quick or.... really painful!

duboisp:
So adding those hashes will be up to the implementer and only recommended for secured context implementation. 

@MaximMartel considering the above, I think it will be a need that each future release of WET will require migration instruction page/section with at least the new generated hashes.

rlambert27:
FYI the ones for jQuery used by WET are here:

http://www.gcpedia.gc.ca/wiki/Centrally_Deployed_Templates_Solution_(CDTS)/SRI

LaurentGoderre:
@duboisp it would be up to the implementer and at their discretion. I don't want to recommend it only for secure app because there would be a point for using this in a static site as well.

duboisp:
@LaurentGoderre where this can be documented?

What do you think if we create a new page called "Building WET and grunt task" under the section "reference" in our [main documentation page](http://wet-boew.github.io/wet-boew/docs/index-en.html)?

LaurentGoderre:
Why in build? This is an implementation issue more than a build one since it's automatically generated via the build. I don't think there is a place in the site where the documentation for this would fit. I guess we could create a Security & WET page....

duboisp:
I am ok with a new page named "Security & WET" or if it is not a lot of content, that could be simply a sub-section the main documentation page under a new header "CDN for WET resources".

LaurentGoderre:
Sure! Can we merge and document later though?

duboisp:
I will let @MaximMartel to decide on that. 

Me, I am ok if you can provide a time estimate of when it will be documented in WET.

I know that WET is lacking a lot regarding of his documentations. I was told by others that was a contribution barrier and it was hard to find knowledge from searching in GitHub issue comments. I am trying to at least documents anything new to WET (at least explain what it is, what it does and how to use it) and keep the rest in a backlog for when resources will be available.

LaurentGoderre:
This isn't a barrier to entry or development though. since nothing depends on it.

duboisp:
@MaximMartel I am ok with merging this PR, as it's now include English and French documentation.

LaurentGoderre:
@MaximMartel can we please merge?

shawnthompson:
We would like this option as well for the CDTS.
cc: @rlambert27 @nicolasraymond @andrewnordlund

shawnthompson:
Once this merges, we will need to do it in all the theme repos, right?

duboisp:
@lucas-hay a quick reminder to not forget to include the SRI payload in today release

LaurentGoderre:
@duboisp what do you mean? They should be included in the package automatically

duboisp:
@LaurentGoderre considering the lack in documentating that feature (https://github.com/wet-boew/wet-boew/pull/7798#issuecomment-266066855) I do beleive the SRI key should be at least included in the release note.

LaurentGoderre:
There are more than one key and the key is only generated on release (after the release notes are merged). Since the keys are already in the package I'm not sure what steps are needed beyond ensuring that the commit is in the release notes.

duboisp:
@lucas-hay just a note to at least ensure the generated keys will be easily findable by implementer when they would look at the release note. thanks

LaurentGoderre:
@duboisp you are welcome to create a documentation page for it.

PR #7790, opened 2016-11-30
Fieldflow plugin - Initial commit, transform a basic list into a selectable list.

duboisp:
Note, the following working example link are based on build 4154900

## English working example:
* [Field flow](http://universallabs.org/wet/labs/wb-fieldflow/dist-0.3/demos/fieldflow/fieldflow-en.html)
* [Field flow basic configuration](http://universallabs.org/wet/labs/wb-fieldflow/dist-0.3/demos/fieldflow/basic-en.html)
* [Field flow advanced](http://universallabs.org/wet/labs/wb-fieldflow/dist-0.3/demos/fieldflow/advanced-en.html)

## French working example:
* [Déroulement de champs](http://universallabs.org/wet/labs/wb-fieldflow/dist-0.3/demos/fieldflow/fieldflow-fr.html)
* [Déroulement de champs, configuration](http://universallabs.org/wet/labs/wb-fieldflow/dist-0.3/demos/fieldflow/basic-fr.html)
* [Déroulement de champs, avancé](http://universallabs.org/wet/labs/wb-fieldflow/dist-0.3/demos/fieldflow/advanced-fr.html)

## Documentation:
* [Field flow Documentation](http://universallabs.org/wet/labs/wb-fieldflow/dist-0.3/docs/ref/fieldflow/fieldflow-en.html)

## i18n temporary workaround
Currently the plugin only support English and French and use a work-around for i18n.

The following text will need to be translated in all language supported by WET, then added to the cvs, then I will be able to update the plugin to fully support the i18n of WET.

* **Button**
  * Continue
  * Allez
* **Default selected item**
  * Make your selection...
  * Sélectionnez dans la liste...
* **Required**
  * required
  * obligatoire

----
cc/ @TouficS

PR #7816, opened 2016-12-20
WIP: Added a plugin to filter a list of elements

LaurentGoderre:
TODO:
* [ ] Add i18n
* [x] Add String.prototype.normalize polyfill
* [x] Add demo
* [ ] Add documentation

### DIFF
--- a/src/plugins/filter/demo/post-filter.js
+++ b/src/plugins/filter/demo/post-filter.js
@@ -1,0 +1,17 @@
+( function() {
+window[ "wb-filter" ] = {
+	filterCallback: function( $field, $elm, settings ) {
+		var $sections =	$elm.find( "section" ),
+			sectionsLength = $sections.length,
+			s, $section;
+
+		for ( s = 0; s < sectionsLength; s += 1 ) {
+			$section = $sections.eq( s );
+
+			if ( $section.find( settings.selector + ":visible" ).length === 0 ) {
+				$section.addClass( "wb-fltr-out" );
+			}
+		}
+	}
+};
+} )();
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -1055,6 +1055,7 @@ module.exports = (grunt) ->
 						"DataTables/media/js/jquery.dataTables.js"
 						"proj4/dist/proj4.js"
 						"openlayers/OpenLayers.debug.js"
+						"unorm/lib/unorm.js"
 					]
 					dest: "<%= coreDist %>/js/deps"
 					rename: (dest, src) ->
--- a/bower.json
+++ b/bower.json
@@ -39,7 +39,8 @@
     "proj4": "2.3.3",
     "respond": "1.4.2",
     "selectivizr": "1.0.2",
-    "SideBySideImproved": "https://github.com/pkoltermann/SideBySideImproved.git"
+    "SideBySideImproved": "https://github.com/pkoltermann/SideBySideImproved.git",
+    "unorm": "^1.4.1"
   },
   "moduleType": [
     "globals"
--- a/src/base/_wet-boew.scss
+++ b/src/base/_wet-boew.scss
@@ -120,6 +120,7 @@
 @import "../plugins/collapsible-alerts/base";
 @import "../plugins/dismissable-content/base";
 @import "../plugins/equalheight/base";
+@import "../plugins/filter/base";
 @import "../plugins/footnotes/base";
 @import "../plugins/formvalid/base";
 @import "../plugins/lightbox/base";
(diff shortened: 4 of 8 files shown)


nschonni:
I'm wondering if this needs to be in an aria-live region

LaurentGoderre:
@nschonni I don't think so since no new content is added. Only hidden. The typical flow for a user would be to type in a filter than navigate below in what is left. Of course I plan to get this tested though

duboisp:
@LaurentGoderre what do you think if I add that feature to my PR #7790. 

Base on your (this) PR, I could add a list filtering feature something similar of what I did for [table filtering](http://universallabs.org/wet/labs/wb-fieldflow/v0.2/wb-fieldflow-advanced.php#tblfilter).

What do you think of filtering by using a ```datalist```? I can add that rendering option in Field flow.

aside: what is the signification of "WIP"?

LaurentGoderre:
@duboisp let's ensure that they are compatible but I don't think they should be the same, because this one allows to filter anyone as long as it can be expressed by CSS selectors.

WIP is work in progress

duboisp:
I agree they should not be the same. I was thinking to add a new "action" called "lstfilter" in wb-fieldflow where it will specifically deal with list filtering. I only wanted to highlight the feature similarity.

Field flow already deal with the UI.

I will work on a demo but I may only able to only deliver in about 3-4 weeks.

LaurentGoderre:
I don't think I'm in too much rush but I think I'll have to implement this locally until this lands. I still want ours to be a reflection of what will eventually land.

duboisp:
@LaurentGoderre what kind of user interaction do you need
1. The user type in the data list then click a button "filter" to filter the list. (Here field flow could do a great job, it could allow to filter with multiple criteria and filter with per-determined criteria)
or
2. The user just type directly in a textbox then the list are filtered live. (Here a new plugin could better fit the need)

LaurentGoderre:
THis allows to filter simply by typing and allows to filter any type of content, not just lists.

LaurentGoderre:
@duboisp I don't think our plugins aim at doing the same thing.

duboisp:
@LaurentGoderre ok

So you are looking to do something similar like : https://github.com/javve/list.js - http://listjs.com/

LaurentGoderre:
Yep.

LaurentGoderre:
@pcwsmith if I recall correctly, you needed something like this for Cannada.ca as well didn't you?

src/plugins/filter/demo/post-filter.js
```diff
@@ -0,0 +1,17 @@
+( function() {
+window[ "wb-filter" ] = {
+	filterCallback: function( $field, $elm, settings ) {
+		var $sections =	$elm.find( "section" ),
```
duboisp: Suggestion:
Include that call back directly in the plugin instead. I think it is common to deal with grouping.

So, instead of targeting directly the ```section``` that could be set through config like:
```html
<div class="wb-filter" data-wb-filter='{"group": "section"}'>
   [...]
</div>
```

LaurentGoderre: I suppose we can have one or more default ones but I'm not sure it will be that simple in many cases. Because this can target any content using selectors. I'm trying to be as least opinionated as possible.

duboisp: You can still allow to use a custom callback for those complex case. 

Is triggering a event could be better than using a call back?

pcwsmith:
indeed we have included filter tools for lists in our design patterns for Canada.ca, been hoping that it could get made.

src/plugins/filter/demo/post-filter.js
LaurentGoderre: Better than a callback no. More WET4....maybe. However, the next version of WET needs to shift to using more callback as it's a way more common pattern in the JS world.

eswright:
I see this as a very important development to deter pages of lists being written as tables to meet usability requirements.

DK-TBS:
Hi gang!  How close are we to having a first version of this plugin up and running?  Any chance I can see a demo of this thing in action?  

The C&IA team here is very keen on this one, we need it for a couple of different page templates.  If the main problem is a lack of manpower to push it across the finish line, perhaps we could help by tracking down a volunteer?  We don't have the skils on our team to actively contribute, but if there's any other way we can help, we would jump at the chance  :)

LaurentGoderre:
@SBleeks can you add your comments of what needs to be fixed accessibility wise for this component?

SBleeks:
I can't find a link to the current design here? Can you send me the link?

LaurentGoderre:
@SBleeks http://laurentgoderre.github.io/wet-boew/demos/filter/filter-en.html

PR #7815, opened 2016-12-19
Carousel: Fixed various noscript, wb-disable and print CSS issues.

EricDunsworth:
* Now displaying all carousel panels in noscript, wb-disable and print scenarios.
* Disabled CSS transitions in noscript, wb-disable and print scenarios to prevent rendering issues.
* Prevented space for carousel style 2's controls from appearing when printing (since the controls aren't used in that scenario).
* Added borders around carousel panel link text when printing.
* Prevented carousel panel images and link text from getting split across separate pages when printing.
* Fixes #7445 and resolves wet-boew/GCWeb#1081.

PR #7817, opened 2016-12-20
Multimedia player: Fixed video cut off/overflow issues.

EricDunsworth:
In Firefox/Blink browsers, the multimedia player's CSS was previously causing the bottom of every video to be cut off by 1px. It also caused video placeholder images to overflow on top of figcaptions when printing (which made the top borders of some of #7815's figcaptions disappear).

**Comparison screenshots of the [multimedia player demo page](http://wet-boew.github.io/v4.0-ci/demos/multimedia/multimedia-en.html)'s placeholder image (note: the grey lines are part of the image itself)...**

**Before (bottom line is missing):**
multimedia_player_before

**After (bottom line is present):**
multimedia_player_after

### DIFF
--- a/src/plugins/multimedia/_base.scss
+++ b/src/plugins/multimedia/_base.scss
@@ -61,7 +61,7 @@ $mm-volume-thumb-width: 10px;
 }
 
 %controls-offset {
-	margin-bottom: -7px;
+	display: block;
 	width: 100%;
 }
 


duboisp:
@EricDunsworth wow, nice catch.

I can barely see the difference (~1px) in FF and Chrome but I saw it. IE seems to be ok with and without the fix.

Lucky you or lucky me, when I started to reproduce the issue, the placeholder image first returned me a 404 error in FF and Chrome but not in Edge, so I was able to save a cache version to perform the test.

I confirm this PR fix the ~1px offset issue for FF and Chrome. 

I don't know if is because of server maintenance on the evening, but the following placeholder image  currently used in the multimedia working example return an 404 error code.
* http://www.servicecanada.gc.ca/eng/video/images/te-lj-eng.jpg 
* http://www.servicecanada.gc.ca/eng/video/images/dc-td-eng.jpg (which I had a cached copy)
I will double check tomorrow and if it still broken I will open a new issue.

EricDunsworth:
@duboisp Thanks :)!

I believe the Service Canada video pages the multimedia player's demos are based on were recently decommissioned as a part of that site's migration into Canada.ca. As a result, as you mentioned, the demo pages' placeholder image references are now broken. The demo pages' transcript links are also broken. The videos themselves still work, since they seem to be hosted on a separate subdomain. For all I know, they might eventually get taken down.

I still have access to preview versions of the aforementioned Service Canada pages and all of their supporting files. If needed, I could send over a PR to bring all of the the placeholder images and transcript pages into the wet-boew repo. But I'd want to keep the videos alone since it probably wouldn't be wise to bring such large files into this repo.

LaurentGoderre:
That is an issue with not hosting the video in GitHub, which we control

duboisp:
Thanks for the clarification, let have the discussion on a new issue.

PR #7822, opened 2016-12-21
Details: Fixed remaining summary arrow issues.

EricDunsworth:
Details: Fixed remaining summary arrow issues.

* Firefox:
  * Prevents summary arrows from disappearing in standard details/summary implementations in wbdisable mode.
  * Prevents summary arrows from disappearing in opened tabbed interface tabs in noscript scenarios.
  * Prevents summary arrows from disappearing in all tabbed interface tabs in wbdisable/print scenarios.
* IE11/Edge:
  * Prevents disc bullets from appearing beside summaries in noscript scenarios.
  * Prevents disc bullets from "flashing" beside summaries when loading pages in JavaScript/wbdisable scenarios.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -23,6 +23,7 @@ rules:
         - border-top
         - border-bottom
         - cursor
+        - list-style-type
   no-empty-rulesets: 2
   no-ids: 0
   no-important: 0
--- a/src/polyfills/details/_base.scss
+++ b/src/polyfills/details/_base.scss
@@ -6,6 +6,8 @@
 summary {
 	// Make sure summary remains visible
 	display: list-item !important;
+	list-style-type: none; // Prevent list bullets from appearing in browsers that lack native details/summary support (e.g. IE/Edge)
+	list-style-type: disclosure-closed; // Show summary arrows in Firefox 49.0+ and future versions of Blink/Webkit
 	visibility: visible !important;
 }
 
@@ -44,6 +46,7 @@ details {
 				border-bottom: 1px solid #ddd;
 				border-bottom-left-radius: 0;
 				border-bottom-right-radius: 0;
+				list-style-type: disclosure-open;
 				margin-bottom: .25em;
 			}
 		}
@@ -90,7 +93,7 @@ details {
 		visibility: visible !important;
 
 		> {
-			* {
+			*:not(summary) {
 				display: block !important;
 			}
 		}
--- a/src/plugins/tabs/_base.scss
+++ b/src/plugins/tabs/_base.scss
@@ -514,7 +514,7 @@ $tab-margin-width: 10px;
 				&[open] {
 					> {
 						summary {
-							@extend %global-display-block-important;
+							display: list-item !important;
 						}
 					}
 				}
@@ -528,7 +528,7 @@ $tab-margin-width: 10px;
 						&[open] {
 							> {
 								summary {
-									@extend %global-display-block-important;
+									display: list-item !important;
 								}
 							}
 						}
--- a/src/plugins/tabs/_noscript.scss
+++ b/src/plugins/tabs/_noscript.scss
@@ -15,7 +15,7 @@
 			&[open] {
 				> {
 					summary {
-						display: block !important;
+						display: list-item !important;
 					}
 				}
 			}
--- a/src/plugins/tabs/_print.scss
+++ b/src/plugins/tabs/_print.scss
@@ -27,7 +27,7 @@
 		}
 
 		summary {
-			display: block !important;
+			display: list-item !important;
 		}
 	}
 
--- a/src/polyfills/details/details.scss
+++ b/src/polyfills/details/details.scss
@@ -36,7 +36,7 @@
 		}
 
 		summary {
-			display: block !important;
+			display: list-item !important;
 		}
 	}
 


.sass-lint.yml
```diff
@@ -23,6 +23,7 @@ rules:
         - border-top
         - border-bottom
         - cursor
+        - style
```
EricDunsworth: @nschonni @LaurentGoderre Are you guys ok with what I did here? Does ``- style`` seem too generic?

In ``src/polyfills/details/_base.scss``, I'm currently using nested style properties within the ``list`` namespace. Originally I tried setting this line to ``- list-style``, but got this error: ``Duplicate properties are not allowed within a block  no-duplicate-properties``.

So I think I only have 2 options for what I'm trying to achieve:
1. What I'm already doing (``- style`` on this line and nested properties in the ``list`` name space)
2. ``list-style`` on this line and two consecutive ``list-style`` properties.

Thoughts?

LaurentGoderre: I'm fine with that as long as it outputs fine

EricDunsworth: @LaurentGoderre Ok, awesome :)! Its generated CSS code looks fine from what I saw.

nschonni: You could also just do a line comment opt-out if it's really a one off

LaurentGoderre: It might not be though...

src/polyfills/details/_base.scss
```diff
@@ -6,6 +6,10 @@
 summary {
 	// Make sure summary remains visible
 	display: list-item !important;
+	list: {
```
duboisp: Did you mean ```list-style: disclosure-closed inside``` ???

src/polyfills/details/_base.scss
```diff
@@ -6,6 +6,10 @@
...
+		style: none;
```
duboisp: That seems not conforming according to SASS.

> ```style:``` Sets the style of the CSS output. See [Output Style](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#output_style).

http://sass-lang.com/documentation/file.SASS_REFERENCE.html#style-option

duboisp (changes requested):
You can found the default CSS for details and style here: http://sass-lang.com/documentation/file.SASS_REFERENCE.html#output_style

where:

```css
@namespace url(http://www.w3.org/1999/xhtml);

summary {
  display: list-item;
  counter-increment: list-item 0;
  list-style: disclosure-closed inside;
}
details[open] > summary {
  list-style-type: disclosure-open;
}
```

src/polyfills/details/_base.scss
EricDunsworth: @duboisp That's one of the properties this code block will generate.

To prevent list bullets from appearing in browsers that don't recognize ``disclosure-closed``, I had to set two ``list-style`` properties. The first is set to ``none`` to prevent the aforementioned browsers from showing list bullets (since they'll fall back on the first property they understood). The second is set to ``disclosure-closed inside`` to support versions of Firefox that have native details/summary support (and [future versions of Blink](https://bugs.chromium.org/p/chromium/issues/detail?id=590014)).

I felt like writing ``list-style`` twice in a row wouldn't look very clean. So I used [SASS' nested properties extension](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#nested_properties) for this code block.

src/polyfills/details/_base.scss
EricDunsworth: @duboisp What you linked is meant for use in Ruby/Rails configuration options that are set outside of SCSS files. This code block is situated in an SCSS file and thus is part of a totally different kind of context that has nothing to do with those options.

On this specific line, I'm using a nested ``style`` property within the previous line's the ``list`` namespace, which will generate a ``list-style`` property. That's a proper use of [SASS' nested properties extension](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#nested_properties).

src/polyfills/details/_base.scss
duboisp: ok now I understand, but I were confuse when I quickly looked for "style:" instruction in SASS documentation. That landed me to look deeper to see what the community did compared to your propose solution.

May be for that one particular case, because it use a "generic" keyword, I think it will add more clarity if you code twice ```list-style"``` instead of using the SASS nested property extension.

Also, using twice the same CSS property seems to be buggy or to hide somewhat a bug. I know this is not the case here, but to avoid that confusion, what do you think of adding a comment prior those two line of  CSS that explain it?

src/polyfills/details/_base.scss
duboisp: Got it, but I was confuse seeing two consecutive "style" and seems to me generic at my first look.

As I suggested in my https://github.com/wet-boew/wet-boew/pull/7822#discussion_r97809071 may be to add a comments before explaining that we need to declare in CSS two  list-style for browser that do not recognize ```disclosure-closed``` will add more clarity

EricDunsworth:
@duboisp Regarding your general review comment, I presume you got that list of properties from [WHATWG's rendering specs for the details/summary elements](https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements)?

On a side note, I didn't include ``counter-increment: list-item 0;`` in this PR. I wasn't aware of its existence at the time. But in any case, that property is built-into all versions of Firefox that have native details/summary support. The issue that property fixes doesn't affect browsers that rely on the polyfill (like IE11) nor Blink/Webkit. So there's no value in re-declaring it in WET's CSS.

duboisp:
Yes I extracted that CSS code it from whatwg, sorry I forgot to include a reference to it. Please see the CSS code sample that I provided as a suggestion not as the solution on how it should be implemented. Sorry for the confusion.

My first concern was when I saw two consecutive instruction ```style:``` without additional information. It was confusing to me. Then I went deeper to understand the propose solution.

I agree that SASS' nested property extension could help to add code clarity like in the use case described in the SASS documentation for CSS property starting with ```font-``` as an example but using it like in the propose solution it do not add code clarity but create confusion.

I think a comment followed by two consecutive (in long) ```list-style``` instruction would be easier to read and to maintain in the long term.

EricDunsworth:
> I agree that SASS' nested property extension could help to add code clarity like in the use case described in the SASS documentation for CSS property starting with ``font-`` as an example but using it like in the propose solution it do not add code clarity but create confusion.

IMO it's not any more likely to create confusion than the SASS documentation's example or any other situations where the same property is declared twice in a row. To anyone that's comfortable with SASS, ``list: {style: ...; style: ....;}`` should be cleaner and just as easy to understand as ``list-style: ...; list-style: ...``.

I don't agree with the idea of not using fundamental SASS features in situations where it makes sense to use them, just to be more understandable to CSS-minded contributors that aren't familiar with some of SASS' features. IMO it should be up to them to review SASS' documentation if they don't feel comfortable with it.

Having said that, despite what I just said, I'll update this PR to use ``list-style: ...; list-style: ...``. Not to help CSS-minded contributors, but to be more consistent with what's already being done in .sass-lint.yml. None of ``no-duplicate-properties``' pre-existing exemptions apply to situations similar to ``list: {style: ...; style: ....;}``. Plus for anyone that browses through the linting exemptions, ``list-style`` will give slightly more context than ``style`` would on its own.

> I think a comment followed by two consecutive (in long) ``list-style`` instruction would be easier to read and to maintain in the long term.

Regarding adding a comment for situations where the same property is repeated twice, no precedent seems to exist for that practice in WET.

I tried a quick test where I removed ``no-duplicate-properties``' from .sass-lint.yml and got this:
```
grunt sasslint --force
Running "sasslint:all" (sasslint) task

src/plugins/tabs/_base.scss
   22:2  error  Duplicate properties are not allowed within a block  no-duplicat
e-properties
  232:6  error  Duplicate properties are not allowed within a block  no-duplicat
e-properties

src/polyfills/slider/slider.scss
  147:2  error  Duplicate properties are not allowed within a block  no-duplicat
e-properties
  149:2  error  Duplicate properties are not allowed within a block  no-duplicat
e-properties
  151:2  error  Duplicate properties are not allowed within a block  no-duplicat
e-properties
  246:3  error  Duplicate properties are not allowed within a block  no-duplicat
e-properties
  265:3  error  Duplicate properties are not allowed within a block  no-duplicat
e-properties
  267:3  error  Duplicate properties are not allowed within a block  no-duplicat
e-properties
  269:3  error  Duplicate properties are not allowed within a block  no-duplicat
e-properties

theme/banner/_base.scss
  12:2  error  Duplicate properties are not allowed within a block  no-duplicate
-properties

? 10 problems (10 errors, 0 warnings)

Warning:  Used --force, continuing.

Done, but with warnings.
```

None of those spots had any comments. Maybe because it was assumed that since it's standard behaviour for browsers to revert to the last CSS property value they understood when encountering unfamiliar values, there wasn't any need to add comments.

EricDunsworth:
Just updated this PR to use ``list-style: ...; list-style: ...``. instead of ``list: {style: ...; style: ....;}``.

Didn't add any extra comments since there's no precedent for it in WET for that kind of situation.

duboisp:
May be those duplicate CSS references are actually missing comment that describe why. My concern is about how a future code reviewer will considerate those duplication? Do will need to spent days or weeks to understands why and if it still required or not considering the list of supported software/technology? A simple comment solve that issue so quickly, and they are not added when the code is compiled.

Can you modify the comment line at line 7 of src/polyfills/details/_base.scss for something like:
```html
// Make sure summary remains visible with his marker
```

EricDunsworth:
@duboisp I've added line comments in lines 9-10.

duboisp (approved):

duboisp:
@EricDunsworth I wasn't able to test all the fixes that was listed in the configuration.

I was only able confirm this PR fix the following issue:
> Prevents summary arrows from disappearing in standard details/summary implementations in wbdisable mode.

I remark that in Edge in disabled mode, now I can see a bullet beside the ```summary```, without this fix there is nothing show beside.

I have some diffilculties to reproduce the enumurated issue and what they should fixes.

--- 

So, for anyone that have time to test this PR, I temporary setup a test environment which is a compiled version of this fix rebased with wet-boew/master (as Jan 30).

http://universallabs.org/wet/labs/PR7822/index.html

---

When the testing is completed and be successful in a way that it fixes all the enumerated issues in the description, I will be ok to have this PR merged.

EricDunsworth:
FWIW, I tested this PR in Firefox, IE11 and Opera before submitting it.

> I remark that in Edge in disabled mode, now I can see a bullet beside the ``summary``, without this fix there is nothing show beside.

So you're seeing a disc bullet (black circle) in Edge in wbdisable mode? Did you refresh the page after activating wbdisable?

According to [Edge's platform status page](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/detailssummary/), it doesn't currently natively support ``<details>``/``<summary>``. [MDN's browser compatibility chart for the ``list-style-type`` property](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type#Browser_compatibility) also claims that Edge doesn't currently support the ``disclosure-closed``/``disclosure-open`` values. As a result, with the way this PR is setup, when encountering summaries, Edge should be reverting to the last ``list-style`` property it understood (``none``), which should prevent any list bullets from appearing.

> I have some diffilculties to reproduce the enumurated issue and what they should fixes.

Which specific issues are you still having trouble reproducing?

duboisp:
> FWIW, I tested this PR in Firefox, IE11 and Opera before submitting it.

Before to merge a fix that describe specific issue, I try to reproduce them with the current build then do the same test with the fix applied and compare both result to see if it really do what it should do. In a negative result, either the PR description may required to be edited in order to be more accurate or maybe code of the PR need a few adjustment. The goal here is to have an independent second opinion about functional aspect of integrating new changes.

> Did you refresh the page after activating wbdisable?

I think yes I did, but I can't confirm at 100%. I didn't thought that a manual "refresh" was required to activate wbdisable and if so that could be a major bug for that feature.

> As a result, with the way this PR is setup, 

My last comment, even this one it is not related to the technique you have used to fixed the issue. It is about trying to answer: Is it functional and do what it should do?

> Which specific issues are you still having trouble reproducing?

I wasn't able to reproduce the issue you described here:

> Firefox:
> * Prevents summary arrows from disappearing in opened tabbed interface tabs in noscript/wbdisable/print scenarios.
>
> IE11/Edge:
> * Prevents disc bullets from appearing beside summaries in noscript scenarios.
> * Prevents disc bullets from "flashing" beside summaries when loading pages in JavaScript/wbdisable scenarios.

EricDunsworth:
@duboisp 
> I wasn't able to reproduce the issue you described here:

Here's a more-detailed overview of how to reproduce all 3 of those issues.

If you're able to reproduce them, I'd recommend holding off merging this PR until sometime tomorrow. I'll make a quick update to my commit message tomorrow morning.

> Firefox
> * Prevents summary arrows from disappearing in opened tabbed interface tabs in noscript/wbdisable/print scenarios.

To see that issue in action, open Firefox, disable JavaScript and go to http://wet-boew.github.io/v4.0-ci/demos/tabs/tabs-en.html. No arrow will appear in Example 1's second tab (the pre-expanded one). If you try expanding any other tabs, their arrows will disappear too.

In http://universallabs.org/wet/labs/PR7822/demos/tabs/tabs-en.html, all arrows will properly appear.

In my tests just now, I noticed that this issue affected wbdisable mode in a slightly different way than noscript. If you repeat the same steps on http://wet-boew.github.io/v4.0-ci/demos/tabs/tabs-en.html while in wbdisable mode, no arrows appear at all. I'll update this PR's commit message to clarify that.

> IE11/Edge:
> * Prevents disc bullets from appearing beside summaries in noscript scenarios.

To see that issue in action, open IE11, [disable JavaScript](http://stackoverflow.com/questions/18089298/disable-javascript-in-ie-11-preview) and go to http://wet-boew.github.io/v4.0-ci/demos/tabs/tabs-en.html. Disc list bullets should appear in place of arrows beside the "View code" summaries.

Btw I wasn't able to properly disable JavaScript on my workstation's IE11 install because its security settings are locked. So I made a local copy of that page, stripped its references to JS files and removed ``<noscript>`` elements to simulate what would happen.

In http://universallabs.org/wet/labs/PR7822/demos/tabs/tabs-en.html, no disc bullets nor arrows will appear. That's correct behaviour (since IE11 lacks native support and since the polyfill isn't available in that scenario).

> IE11/Edge:
> * Prevents disc bullets from "flashing" beside summaries when loading pages in JavaScript/wbdisable scenarios.

To see that issue in action, open IE11 and go to http://wet-boew.github.io/v4.0-ci/demos/tabs/tabs-en.html. Press F5 to refresh the page (feel free to do it multiple times). Each time the page is refreshed, disc bullets will momentarily appear beside the "View code" summaries. Once the page finishes loading, the bullets will disappear and be replaced with arrows.

In http://universallabs.org/wet/labs/PR7822/demos/tabs/tabs-en.html, no disc bullets will ever appear at any point.

duboisp:
Thanks  @EricDunsworth for the step by step on how to reproduce and the detailed explanation, it was useful.

For the testing with Edge, I was unable to deactivate javascript. It seems that the javascript engine in Edge are not intended to be deactivated at all. BTW I did my all test on Windows 10 pro where I am fully admin.

> Firefox
> * Prevents summary arrows from disappearing in opened tabbed interface tabs in noscript/wbdisable/print scenarios.

I was able to reproduce in noscript/wbdisable/print scenarios (FF dev) and my test was successful

> IE11/Edge:
> * Prevents disc bullets from appearing beside summaries in noscript scenarios.

I was only able to reproduce in IE11 and my test was successful

> IE11/Edge:
> * Prevents disc bullets from "flashing" beside summaries when loading pages in JavaScript/wbdisable scenarios.

Test was successful in IE 11 but in Edge, instead of seeing from "flashing" now I always see both the arrow and the bullet beside every details/summary
pr7822

> So you're seeing a disc bullet (black circle) in Edge in wbdisable mode? Did you refresh the page after activating wbdisable?

I tested again, with refreshing the page and in Edge when I am in wbdisabled mode, 
* Latest version of WET: There is no bullet 
pr7822-1
* With this fix: There is a bullet beside each summaries 
pr7822-2

---

@EricDunsworth Do you think you are able to fix the issue in Edge where I see both the bullet and the arrow?

src/polyfills/details/_base.scss
```diff
@@ -6,6 +6,8 @@
 summary {
 	// Make sure summary remains visible
 	display: list-item !important;
+	list-style-type: none; // Prevent list bullets from appearing in browsers that lack native details/summary support (e.g. IE/Edge)
+	list-style-type: disclosure-closed; // Show summary arrows in Firefox 49.0+ and future versions of Blink/Webkit
```
duboisp: I remark that in this version you removed the ```inside``` instruction that was after the ```disclosure-closed```. Should that fix the issue I saw in Edge where the bullet was show beside the arrow?

EricDunsworth:
@duboisp Updated this PR earlier :)! It fixes those Edge issues. I've also updated my commit message and the OP's comment to clarify the different kinds of arrow issues that were affecting Firefox's noscript/wbdisable modes.

Just to go into more detail regarding Edge...

Back when I was doing the following, Edge seemed to be ignoring ``list-style: none;``, even though its predecessor (IE11) didn't.
```
summary {
	...
	list-style: none; // Prevent list bullets from appearing in browsers that lack native details/summary support (e.g. IE/Edge)
	list-style: disclosure-closed inside; // Show summary arrows in Firefox 49.0+ and future versions of Blink/Webkit
	...
}
```

When I tried using the ``list-style-type`` property instead, Edge stopped showing list bullets and nothing worsened in any of the other browsers I checked (Firefox, Opera and IE11). So now I'm using the following without any issues:
```
summary {
	...
	list-style-type: none; // Prevent list bullets from appearing in browsers that lack native details/summary support (e.g. IE/Edge)
	list-style-type: disclosure-closed; // Show summary arrows in Firefox 49.0+ and future versions of Blink/Webkit
	...
}
```

src/polyfills/details/_base.scss
EricDunsworth: @duboisp Since I'm no longer using ``list-style`` (shorthand for ``list-style-type``, ``list-style-position`` and ``list-style-image``) and am now solely using ``list-style-type``, the ``inside`` value isn't needed anymore.

I guess this works out for the best since things are now more-targeted than before and are no longer re-declaring as many of Firefox's default ``summary`` properties.

duboisp:
I re-tested and

> Firefox
> * Prevents summary arrows from disappearing in standard details/summary implementations in wbdisable mode.
> * Prevents summary arrows from disappearing in opened tabbed interface tabs in noscript scenarios.
> * Prevents summary arrows from disappearing in all tabbed interface tabs in wbdisable/print scenarios.

Test was successful in Firefox

> IE11/Edge:
> * Prevents disc bullets from appearing beside summaries in noscript scenarios.
> * Prevents disc bullets from "flashing" beside summaries when loading pages in JavaScript/wbdisable scenarios.

Test was successful in IE11 and Edge

Tested in Chrome and work fine.

---

FYI - Now in F11 and Edge in wbdisable mode there is no marker beside summary.

I am fine to address that in a new issue, but I remarked something only in Edge, that in [collapsible alert](http://wet-boew.github.io/wet-boew/demos/collapsible-alerts/collapsible-alerts-en.html) working example I can still see the triangle marker beside summary in wbdisable mode. And I remarked an arrow resize effect when refreshing in standard mode. I think it is because they are adding the character ```\25BA\a0``` ([line 101](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/collapsible-alerts/base.scss#L101)) and ```\25BC\a0``` ([line 110](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/collapsible-alerts/base.scss#L110)) in css.

@EricDunsworth What do you think? Should we change the collapsible alert to match the default behavior or/and should the arrow character be inserted in css to enhance the default behavior?

Personally, I will go with what we have now and merge this PR. Then open a new issue to fix the collapsible alert by a new PR. I think the collapsible alert should match the default behavior. My opinion is neutral in regard of adding the arrow in css for the wbdisabled mode as being the default behavior.

EricDunsworth:
@duboisp Regarding no marker appearing beside ``summary`` elements in IE11/Edge within wbdisable mode, that's expected behaviour. Since those browsers lack native support for ``details``/``summary``, they rely on WET's polyfill. But wbdisable mode disables polyfills... so IE11/Edge will revert to plain old content.

As for the collapsible alerts issue, I don't mind either way. If this PR is still open tomorrow I may attempt to fix it as a part of this one. Otherwise, I don't mind sending out a follow-up PR to try addressing it.

duboisp:
Ok, we will do in a new PR then.

PR #7823, opened 2016-12-21
move sri task into dist to be able to use the payload file in the theme repos

shawnthompson:
CC @LaurentGoderre

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -25,6 +25,7 @@ module.exports = (grunt) ->
 			"demos-min"
 			"htmllint"
 			"bootlint"
+			"sri"
 		]
 	)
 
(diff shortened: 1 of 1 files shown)


Gruntfile.coffee
```diff
@@ -56,7 +57,7 @@ module.exports = (grunt) ->
 		->
 			if process.env.TRAVIS_PULL_REQUEST is "false" and process.env.DIST_REPO isnt `undefined` and ( process.env.TRAVIS_TAG isnt "" or process.env.TRAVIS_BRANCH is "master" )
 				grunt.task.run [
-                    "sri"
+          "sri"
```
LaurentGoderre: You should remove it from here so we don't run it twice

PR #7803, opened 2016-12-08
Overlay Print - Only print the active dialog. Alternate solution of #7791

duboisp:
This should fix #7615

See the conversation on #7791 for more info.

Here PDF showing what it would look like when printed.
* [Centered popup](http://universallabs.org/wet/labs/temporary/center-popup.pdf)
* [Middle screen overlay](http://universallabs.org/wet/labs/temporary/middle-screen.pdf)
* [Full screen overvay](http://universallabs.org/wet/labs/temporary/full-screen.pdf)

It was tested on FF dev, Chrome and Edge

PR #7839, opened 2017-01-16
Bootstrap overrides: Removed abbreviation titles from print SCSS.

EricDunsworth:
Same idea as #7476, but for abbreviation titles.

**Reasons:**
* Printing all abbreviation titles risks bloating printouts of pages that use titles extensively (such as on every single abbr element).
* Showing abbreviation titles in printouts risks presenting meaningful information in that context that isn't available in other contexts. For example, mobile users and keyboard-only desktop users have no means of accessing abbreviation titles.
* Most content isn't written in a manner that lends itself to printing abbreviation titles.

### DIFF
--- a/src/base/_bootstrap.scss
+++ b/src/base/_bootstrap.scss
@@ -14,3 +14,8 @@ $icon-font-path: "#{$wb-core-path}/fonts/";
  */
 @import "../../lib/bootstrap-sass-official/assets/stylesheets/bootstrap";
 @import "bootstrap-overrides/base";
+
+/* Print view */
+@media print {
+	@import "bootstrap-overrides/print";
+}
--- a/src/base/bootstrap-overrides/_print.scss
+++ b/src/base/bootstrap-overrides/_print.scss
@@ -1,0 +1,12 @@
+/*
+  WET-BOEW
+  @title: Bootstrap print overrides for WET-BOEW
+ */
+
+abbr {
+	&[title] {
+		&:after {
+			content: none;
+		}
+	}
+}


duboisp:
@EricDunsworth after reading the referenced issue #4229 in your referenced PR #7476 the argument was that in the next version of bootstrap (version 4) it's removed. 

Is it the same case here for abbreviation titles?

EricDunsworth:
@duboisp Nah. Abbreviation titles are still present in Bootstrap 4's print CSS (at least in the latest alpha version). It doesn't look like any issues have ever been posted in its repo about the possibility of removing abbreviation titles when printing.

Printing link ``href`` attributes had much more significant issues than abbreviation titles, so I can see how it was decided to remove that (although it doesn't seem to be entirely gone - [it's currently commented-out](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_print.scss#L37)).

It looks like both of those link/abbreviation printing styles originated from the [HTML5 boilerplate](https://github.com/h5bp/html5-boilerplate).

duboisp:
Ok thanks for the clarification on bootstrap, 

* Is your PR based on usability research? If so can you include a reference to the report.
* Do you have a sample of published pages that printing the abbreviation titles are causing an issue?

---

For now I am opposed to merge this PR because there currently insufficient documentation that explain printing the abbreviation title are causing a reading issue for the user.

EricDunsworth:
@duboisp Here are my responses...

> Is your PR based on usability research? If so can you include a reference to the report.

My PR isn't based on usability research.

It's based on my personal opinion, understanding of how abbreviation titles aren't accessible to all users, things I've seen in WCAG 2.0 and ways in which I've seen them implemented overtime.

In WCAG 2.0, [success criterion 3.1.4's techniques](https://www.w3.org/TR/UNDERSTANDING-WCAG20/meaning-located.html#meaning-located-techniques-head) cover several valid ways of dealing with abbreviations. Even though it's an AAA criterion (which WET and the GC aren't required to comply with), I believe the vast majority of the content the GC produces already generally complies with situation A's first item via [technique G97](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G97).

However, if a page is coded based on situation A's second item via [technique H28](http://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/H28), they could potentially end up with the following code...

```
<ul>
  <li>I want to use <abbr title="Web Experience Toolkit">WET</abbr> on my website</li>
  <li>I want to customize <abbr title="Web Experience Toolkit">WET</abbr> and design a new <abbr title="Web Experience Toolkit">WET</abbr> theme</li>
  <li>I want to report a bug or ask a question about <abbr title="Web Experience Toolkit">WET</abbr></li>
  <li>I want to fix or improve <abbr title="Web Experience Toolkit">WET</abbr> as a developer</li>
</ul>
```

It would look like the following on user agents with screen viewports (along with dotted underlines below all instances of "WET" and tooltips containing "Web Experience Toolkit"):
> * I want to use WET on my website
> * I want to customize WET and design a new WET theme
> * I want to report a bug or ask a question about WET
> * I want to fix or improve WET as a developer

The only kinds of users that'd be able to access the titles of the "WET" abbreviations would be desktop users that use mice and certain screen reader users. Neither desktop users that only use keyboards nor mobile users would have any means of unveiling the titles (since mice are needed to hover over the abbreviations to make their title tooltips appear).

With the print CSS this PR is trying to remove, my example would look like the following in page printouts:
> * I want to use WET (Web Experience Toolkit) on my website
> * I want to customize WET (Web Experience Toolkit) and design a new WET (Web Experience Toolkit) theme
> * I want to report a bug or ask a question about WET (Web Experience Toolkit)
> * I want to fix or improve WET (Web Experience Toolkit) as a developer

IMO that really bloats the printouts and is unfair to the kinds of the aforementioned users.

If I was using a mobile browser, read through that content without knowing what WET stands for, then tried printing it, I wouldn't expect the printout to contain brand new information that wasn't available to me when I initially read through it in my mobile browser. I probably also wouldn't appreciate all the unnecessary repetition and wasted paper caused by it.

Btw I wasn't able to find any indication that the HTML5 Boilerplate initially adopted the CSS I'm trying to undo as a result of usability research, either. From what I can tell, h5bp/html5-boilerplate@ecd87f7cf58ebeae8bb7590bda5585215ef6fa78 is the very first commit that introduced both link href and abbreviation title printing into the boilerplate. It seems to have been directly committed into the repo (not via a PR - so no further context is available).

As for samples of published pages that are affected, I don't have any off the top of my head, but I'm confident that they're out there.

duboisp:
Thanks for the explanation @EricDunsworth 

I understand the theoretical use case of when the ```abbr``` is used massively that printing the title of abbreviation became to be kind of noise for the content. But I still think that is an isolated case where the probability of having an user that print a such page is very very low.

But theoretical is not a factual evidence, so I am still opposed to have this PR merged. More evidence should be provided.

EricDunsworth:
@duboisp What about the risk of abbreviation titles introducing brand new content in printouts that wasn't available to mobile or keyboard-only desktop users?

What's currently being done isn't default browser behaviour, doesn't appear to have been based on usability research and isn't in line with how other aspects of WET print.

As this time, printouts of WET pages seem to be based on the "live" state of a page in a browser before the user attempted to print it.

For instance:
* If I open a page containing a collapsed ``<details>`` element, then print it, the ``<details>`` element will still be collapsed in the printout. The only way to make it expanded in my printout is to expand it in my browser beforehand.
* In #7803, if I open a page that uses the overlay plugin in standard mode, then print it, no overlays will appear in the printout. But if I open a fullscreen overlay in my browser prior to printing, only the overlay's content will appear in the printout. None of the content "behind" the overlay will appear, which makes sense if the printout is meant to be contextual.

Given all of that, I don't see why abbreviation titles are being treated differently? If they're not available to all users, IMO it's unwise to assume they should always be included in all printouts.

duboisp:
> What about the risk of abbreviation titles introducing brand new content in printouts that wasn't available to mobile or keyboard-only desktop users?

That means that webpage is failing WCAG SC 4.1.1 because the page would not be fully conforming to specification ([G192: Fully conforming to specifications](http://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G192))

The [HTML5.2 specification](https://www.w3.org/TR/html52/textlevel-semantics.html#the-abbr-element) specify that the title attribute set on an ```abbr``` element:
> The attribute, if specified, must contain an expansion of the abbreviation, and nothing else.

I know in general writing it recommend to describe at least the first occurrence of each abbreviation. Do you think that it is possible to achieve that in CSS?

EricDunsworth:
@duboisp The brand new content in the scenarios I described would be the expansion of the abbreviation. For example, for "WET", the brand new content would be "Web Experience Toolkit" (aka its expansion - not something unrelated).

What I mentioned at the start of my previous comment is currently being done on [Canada.ca's home page](https://www.canada.ca/en.html). In the "Focus on" section, check out the "Veterans" subsection. Its content contains "RCMP" (coded as ``<abbr title="Royal Canadian Mounted Police">RCMP</abbr>``). If I read through that page in a mobile browser or on a desktop browser without a mouse, all I'll see is "RCMP". Touching or trying to focus on it won't give me its expansion. If I print the page out, I'll get "RCMP (Royal Canadian Mounted Police)". If I had no way of knowing what RCMP stood for when I read through the page in my browser, why should my printout contain that information? I don't think it's fair that my printout would contain brand new information (i.e. "Royal Canadian Mounted Police") that I had no means of seeing in my browser.

I agree that it's a great writing practice to introduce the first instance of an abbreviation in its fully expanded form, followed by parentheses containing its shortened abbreviation, then using the shortened abbreviation on its own throughout the rest of the content. Unfortunately that isn't always done. Content isn't always written perfectly and content authors aren't always in control of how a developer decides to code their pages' abbreviations. Some developers may not use ``<abbr>`` elements at all, some might only use ``<abbr>`` elements without any attributes, whereas others might use ``<abbr title="...">`` everywhere.

By not attempting to print out the ``<abbr>`` element's ``title`` attribute, WET would stay in line with default browser printing behaviour, encourage better content writing practices, not have to worry about how developers coded their abbreviations and avoid unnecessary bloat. Most importantly, it would prevent confusion/surprises for users that had no means of viewing the expanded abbreviations in their browsers.

> I know in general writing it recommend to describe at least the first occurrence of each abbreviation. Do you think that it is possible to achieve that in CSS?

Given all the variables involved with how content is written and coded, I don't think there would be a good way of achieving what you suggested with CSS only.

duboisp:
> Touching or trying to focus on it won't give me its expansion.

If the title can not be seen on a mobile device or when the page is operated via keyboard that only seems to be only a poor implementation of ```abbr``` in those device/software and that is simply not fair for all users.

> Content isn't always written perfectly and content authors aren't always in control of how a developer decides to code their pages' abbreviations.

We should not fix bad design or bad codding from web editor, instead we should improve on from good design and best practice.

----

If the issue is how web editor code and use ```<abbr>``` then may be the web content style guide should provide more instruction on how to use it, when to use it and what alternative design are available for abbreviation.

lucas-hay:
As discussed at PR review meeting, we will close this PR without merging.

PR #7840, opened 2017-01-16
Base: Hid skip links when printing in wbdisable mode.

EricDunsworth:


duboisp:
@bsouster Is this will interfere with your work in progress for displaying a message when the user are in wbdisable mode.

I do think that the warning message. in wbdisable mode, should be printed as the page are not render the same.

But it look good to me to hide the skip links as it do not add any value of the printed version of the page.

PR #7844, opened 2017-01-17
Menu: In noscript, only show menus/footer in screen viewports.

EricDunsworth:
Prevents those elements from unexpectedly appearing when printing pages in noscript mode.

### DIFF
--- a/src/plugins/menu/_noscript.scss
+++ b/src/plugins/menu/_noscript.scss
@@ -10,22 +10,28 @@
 	display: none !important;
 }
 
-#wb-srch {
-	@extend %noscript-display-block-important;
-}
-
-#wb-sm {
-	@extend %noscript-display-block-important;
-
-	.nvbar {
-		@extend %noscript-display-block-important;
+@media screen {
+	%menu-noscript-screen-display-block-important {
+		display: block !important;
 	}
-}
 
-#wb-sec {
-	@extend %noscript-display-block-important;
-}
+	#wb-srch {
+		@extend %menu-noscript-screen-display-block-important;
+	}
 
-#wb-info {
-	@extend %noscript-display-block-important;
+	#wb-sm {
+		@extend %menu-noscript-screen-display-block-important;
+
+		.nvbar {
+			@extend %menu-noscript-screen-display-block-important;
+		}
+	}
+
+	#wb-sec {
+		@extend %menu-noscript-screen-display-block-important;
+	}
+
+	#wb-info {
+		@extend %menu-noscript-screen-display-block-important;
+	}
 }


src/base/noscript.scss
```diff
@@ -8,6 +8,12 @@
 	display: block !important;
 }
 
+@media screen {
+	%noscript-screen-display-block-important {
```
duboisp: Why to not modify ```%noscript-display-block-important;``` instead of creating a new extend.

duboisp (changes requested):
When I search for the extend that you replaced, I found the 5 reference that you have modified in _noscript and there 1 reference in the lightbox/_noscript and 1 reference in overlay/_noscript. 

So are your propose solution should be also applied to the ligthbox and overlay noscript sass file? In affirmative you only need to do your update for [line 7](https://github.com/EricDunsworth/wet-boew/blob/4aa52fa0f35a67240de167890cc49da1cc6c01d6/src/base/noscript.scss#L7)

EricDunsworth:
> So are your propose solution should be also applied to the ligthbox and overlay noscript sass file? In affirmative you only need to do your update for line 7

@duboisp The reason I didn't modify them was because as things stand (with the Lightbox and Overlay plugins using ``%noscript-display-block-important`` in their _noscript.scss files), it forces their overlays to appear when printing pages in noscript mode. If I had switched them over to use ``%noscript-screen-display-block-important``, that would only impact noscript screen viewports - which would make them stop appearing when printing. I'd consider that to be a regression.

The Lightbox and Overlay plugins' overlays currently only appear when printing in noscript and wbdisable modes. They're not currently appearing when printing in standard mode (i.e. JS enabled and wbdisable=false).

I'll revise this PR to scrap ``%noscript-display-block-important`` and make the Lightbox/Overlay plugins use ``%noscript-screen-display-block-important``. Then I'll send over another PR to force those plugins' overlays to always appear when printing in all modes. That should address both of our concerns :).

duboisp:
@EricDunsworth Thanks for the explanation.

May be instead of adding a new extend, what about to wrap the content in src/plugins/menu/_noscript.scss with a ```media screen```? as your fix is only applied to the noscript menu.

FYI - https://github.com/wet-boew/wet-boew/pull/7803 contain a fix to print overlay/lightbox in standard mode.

src/base/noscript.scss
EricDunsworth: When I first sent this PR, my goal was to change ``%noscript-display-block-important`` to only apply to screen viewports in a handful of scenarios (not all). That's why I created a second variant of it instead of completely replacing it.

But given your feedback and my other comment, I'll go through with your suggestion and stick with the new placeholder selector's name (``noscript-screen-display-block-important``), since it clearly denotes that it only applies to screen viewports.

LaurentGoderre: This is very confusing.....why do you have a `@media` with only a placeholder? I'm not sure this does what you think it does

duboisp: @EricDunsworth the extend ```%noscript-display-block-important``` is only used by the menu, lightbox and overlay. Considering that your fix is limited to the menu, then I adding the ```media screen``` instruction directly in the menu sass is the best solution. 

Sorry but am not able to see how creating a new extend in the global noscript that only apply to menu is usefull. Are you planning to modify that new extend or re-use it in an another theme, like GCweb?

EricDunsworth: @LaurentGoderre Whoops - looks like I totally botched my commit's message (any by extension - this PR's title). It was stating the opposite of what I intended. That's probably what caused (or at least contributed to) all this confusion. Sorry :(.

This is what I should've called it (will change it in a sec):
> Menu: In noscript, only **~~hide~~ show** menus/footer in screen viewports.

My commit's SCSS changes are what I intended. My goal was to use a noscript, screen-only placeholder selector. I researched/tested it beforehand. It also looks ok in compiled noscript.css files and sasslint is happy with it.

@duboisp I wanted to make ``%noscript-display-block-important`` available in the global noscript file since it's generic-enough to potentially be used by more than just the menu plugin. Even if it starts out with only one plugin using it, there's nothing stopping other plugins from potentially adopting it in the future.

I might rework this PR soon to make the Lightbox and Overlay plugins use ``%noscript-display-block-important``.

LaurentGoderre: Can you create a simplified version of this on Codepen or similar. I don't think it's right having a media query with no actual CSS, plus sharing placeholder accros media queries shouldn't work

duboisp: > there's nothing stopping other plugins from potentially adopting it in the future.

I agree, but there currently, **as today**, no use case to do it now. Instead, when we will have other plugin that will repeat the same piece of code, only then centralizing into an extend will be the best solution to take.

EricDunsworth: @LaurentGoderre Here's a few Codepen examples. Both generate the same CSS output.
* [Placeholder selector inside a media query (i.e. what this PR is currently doing)](https://codepen.io/anon/pen/gggmJa)
* [Placeholder selector containing a media query](https://codepen.io/anon/pen/GrrmJo)

I wouldn't mind moving the media query inside ``%noscript-display-block-important`` if you'd prefer.

EricDunsworth: @duboisp @LaurentGoderre I'll revise this PR in the following ways to address your concerns.
* Move my placeholder selector into the menu plugin's noscript file
* Prefix my placeholder selector's name with ``menu-``
* Nest the screen media query inside my placeholder selector

nschonni: You can't extend a placeholder outside a media query http://sass-lang.com/documentation/file.SASS_REFERENCE.html#_in_directives. If you depend on it now, it is something that will likely break when updating node-sass.

EricDunsworth: @nschonni I think what you mentioned is the opposite of what I was doing (and both CodePen examples).

SASS' documentation recommends against declaring a placeholder outside a media query, then extending it from inside media queries (since that technically wouldn't make any sense).

In the original version of this PR, my placeholder selector was declared inside a media query (like my first CodePen example), but the extend directives were situated outside of media queries. Since all of the extend directives were generated into the placeholder selector that was already inside a media query, it worked fine.

My second CodePen example (which I updated this PR to resemble) is still essentially doing the same thing. But it might arguably be better since in my first CodePen example, if the placeholder selector was never extended, the generated CSS would contain an empty media query. With the second example, there's no chance of an empty media query getting generated in the CSS if nothing attempts to extend it.

duboisp:
@EricDunsworth what about something like this?

```sass
/*
 * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)
 * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
 */

/*
 * Menu Sass (noscript)
 */

#wb-glb-mn {
	display: none !important;
}

@media screen {

	%menu-noscript-screen-display-block-important {
		display: block !important;
	}

	#wb-srch {
		@extend %menu-noscript-screen-display-block-important;
	}

	#wb-sm {
		@extend %menu-noscript-screen-display-block-important;

		.nvbar {
			@extend %menu-noscript-screen-display-block-important;
		}
	}

	#wb-sec {
		@extend %menu-noscript-screen-display-block-important;
	}

	#wb-info {
		@extend %menu-noscript-screen-display-block-important;
	}

}
```

LaurentGoderre:
I much prefer what @duboisp suggested. With WET we are explicit with our media queries to not have a thousand of them, potentially competing against each other. The proposed replacement is clearer and more explicit, not to mention more predictable.

EricDunsworth:
@LaurentGoderre Well in any case, as prior revisions of this PR and my CodePen examples demonstrated, in practice all 3 methods we've discussed will generate the exact same output CSS (i.e. a **single** screen media query containing all comma-separated selectors) and not trigger any linting warnings/errors.

When this PR's placeholder selector started out as a variant of ``%noscript-display-block-important`` and was potentially going to be used across multiple plugins, placing it within a screen media query (or a screen media query within it) made perfect sense to me.

Now that everything's within the scope of just the menu plugin's noscript file and literally all of its contents are meant to be screen-only, I don't feel very strongly about which method ultimately gets used. I'll update this PR once again to what @duboisp suggested :).

duboisp:
In my previous comments, I forget to replace this

````sass
#wb-glb-mn {
	display: none !important;
}
```

by this:

````sass
#wb-glb-mn {
	@extend %noscript-display-block-important
}
```

EricDunsworth:
@duboisp I didn't copy/paste your code block in what I updated just now, but made the same mistake lol. Onto another update heh.

src/plugins/menu/_noscript.scss
```diff
@@ -6,26 +6,33 @@
...
-	@extend %noscript-display-block-important;
-}
+@media screen {
+	%menu-noscript-screen-display-block-important {
+			display: block !important;
+	}
 
-#wb-sm {
-	@extend %noscript-display-block-important;
+	#wb-glb-mn {
```
duboisp: Is this selector should be inside or outside the ```@media screen``` query? Because in your previous code revision it was outside I think.

src/plugins/menu/_noscript.scss
```diff
@@ -6,26 +6,33 @@
 /*
  * Menu Sass (noscript)
  */
+
 #wb-glb-mn {
 	display: none !important;
```
duboisp: To be replace by ```@extend %noscript-display-block-important```

duboisp (changes requested):

src/plugins/menu/_noscript.scss
EricDunsworth: @duboisp It was meant to be outside. I've brought it out in one of my latest force pushes.

EricDunsworth:
I believe this PR should now be good to go (just finished my force pushing spree lol).

@duboisp I left ``#wb-glb-mn`` outside the scope of the screen media selector and kept it to ``display: none !important;``. That ID is for mobile mode's hamburger button (which is hardcoded in WET's HTML files).

src/plugins/menu/_noscript.scss
```diff
@@ -10,22 +10,28 @@
...
```
duboisp: To be replace by @extend %noscript-display-block-important

The %noscript-display-block-important is defined in the global [noscript at line 7](https://github.com/wet-boew/wet-boew/blob/master/src/base/noscript.scss#L7) and also used by other plugin.

duboisp (changes requested):

src/plugins/menu/_noscript.scss
EricDunsworth: @duboisp But that placeholder selector has the opposite effect.

I want to keep it hidden (as it's always been). Using ``%noscript-display-block-important`` would forcefully display it in all noscript screen and print views, which would make a random non-functioning hamburger button appear.

duboisp:
@EricDunsworth thanks for clarify that instruction should be declared outside the media query.

Regarding my request of change, it is not about what that line code do, but just to use the ```@extend``` instruction that contain the exact same CSS code declared in the global [noscript at line 7](https://github.com/wet-boew/wet-boew/blob/master/src/base/noscript.scss#L7).

src/plugins/menu/_noscript.scss
duboisp: Sorry I just remark the difference, "block" and "none"

duboisp:
Sorry I just remark the difference, "block" and "none". I will approve by review.

duboisp (approved):

EricDunsworth:
@duboisp Thanks :)!

PR #7850, opened 2017-01-20
Fix lightbox keyboard accessibility issues

LaurentGoderre:


duboisp:
@LaurentGoderre It look that the commit 60b1efe in your PR is the same as 45f93ea052394770f515ab10a4a7a3a1b13c29b3 that I merged yesterday

LaurentGoderre:
Yeah, my rebases are acting weird

src/plugins/lightbox/_base.scss
```diff
@@ -11,6 +11,10 @@
 	outline-offset: -2px;
 }
 
+%lightbox-modal-hide {
+	visibility: hidden!important;
```
nschonni: Space before `!important`

src/plugins/lightbox/_base.scss
```diff
@@ -22,6 +26,30 @@
...
+			}
+		}
+
+		summary {
+			@extend %lightbox-modal-hide;
+		}
+
+		.modal-dialog {
+			summary {
+				visibility: visible!important;
```
nschonni: Space before `!important`

src/plugins/lightbox/_base.scss
```diff
@@ -22,6 +26,30 @@
 	}
 }
 
+body {
+	&.wb-modal {
```
nschonni: How can a body be a modal?

src/plugins/lightbox/lightbox.js
```diff
@@ -18,7 +18,7 @@ var componentName = "wb-lbx",
 	initEvent = "wb-init" + selector,
 	setFocusEvent = "setfocus.wb",
 	dependenciesLoadedEvent = "deps-loaded" + selector,
-	extendedGlobal = false,
+	modalHideSelector = "#tphp, body > header, body > main, body > footer",
```
nschonni: `#wb-tphp`, but I'm not sure I understand how relates to hiding

nschonni (dismissed):
Could you give some more explaination for the changes? I don't really understand what keyboard issues this is addressing

src/plugins/lightbox/_base.scss
LaurentGoderre: The body isn't modal, the body is modal mode...

src/plugins/lightbox/_base.scss
LaurentGoderre: According to which SCSS Lint rule? if the build isn't complaining about it, should we add something to the rule for it?

nschonni: If there is a rule that applies, it's just how it is formatted in the rest of the repo https://github.com/wet-boew/wet-boew/search?utf8=%E2%9C%93&q=%21important

src/plugins/lightbox/lightbox.js
LaurentGoderre: Updated, though for some reason it's not showing up here

LaurentGoderre:
@nschonni when lightbox opens, the plugin disables all tab based navigation and tries to create it's own but it fails. It doesn't recognize many items that can receive focus so these objects don't work with the keyboard (like the details summary element). The other issue is that it takes the tabbing away but a screen reader user can still use list of links or heading and navigate away from the lightbox and the keyboard traps.

LaurentGoderre:
@dfarough @RubyDo can you please test these URL since this fixes the issues you reported?

http://laurentgoderre.github.io/wet-boew/demos/lightbox/lightbox-en.html

/cc @SBleeks

RubyDo:
@LaurentGoderre  Thank you for looking into the issue!
Here are my test results:
**1)** Accessibility of background issue fixed:  
When dialog box is open, the background page is not accessible e.g. user able to access headings available on the dialog box only. Thank you!
**2)** Announcing "entering dialog box" + reading a title for the dialog box: 
When lightbox opens, IE and Firefox reads the H tag (e.g. Title) vs. Chrome read the background page title (Lightbox example...) as the label/title of the dialog box. All three navigators do not announce presence of dialog box "entering the dialog box " 
Here is a demo https://youtu.be/K5k_kk7YEoU (IE vs Chrome)
Here is suggested solution from the Google Accessibility course on https://youtu.be/nLjAqr6YDZE?t=28s

CC @kevinLamb

LaurentGoderre:
@RubyDo would you agree that the point number two would not be a blocker to merging this and we can address that point in another PR?

RubyDo:
@LaurentGoderre Point 2 is not a blocker for merging this commit. Thank you so much for the fix! We have been looking to resolve this issue since long time.

CC @kevinLamb

SBleeks:
this recent fix addresses the issue of screen readers being trapped in the lightbox when using the keyboard TAB key using Firefox browser. this fix also allows keyboard users in firefox to access the "close the overlay" button and the ESC key now also works 100% of the time for closing the overlay.
@dfarough What do you think?

LaurentGoderre:
@masterbee @lucas-hay any objections to merging this?

PR #7855, opened 2017-01-23
Multimedia player: Vertically-centered loading animation in display area.

EricDunsworth:
**Screenshots...**

**Before:**
wet-v4 0-multimedia-loading-anim-position-before

**After:**
wet-v4 0-multimedia-loading-anim-position-after


LaurentGoderre:
Neat that it's so small!

PR #7857, opened 2017-01-23
Base/Theme: Added body/main background colour variables.

EricDunsworth:

PR #7858, opened 2017-01-23
Tabbed interface: Made active tab summary inherit content area's background colour by default.

EricDunsworth:
Depends on #7857.

In standard tabbed interface implementations, the active tab's summary used to be set to white. This looked fine in themes that use a white background colour in their content areas (like WET). But in themes that use other background colours, it caused a mismatch between the summary's white background and the transparent background of its associated details element.

This commit resolves the mismatch by:
* Changing the $tablist-active-bg-color variable's default value to null.
* Adding if/else logic to ensure the active tab's summary uses the content area's background colour (unless a theme has specified a value for $tablist-active-bg-color).

**GCWeb screenshots...**

**Before** (active tab summary's background colour is white whereas details' background is light grey):
v4 0-tabs-active-summary-bg-before

**After** (active tab summary and details' backgrounds are both light grey):
Assumes #7857, wet-boew/GCWeb#1220 and this PR are in place.
v4 0-tabs-active-summary-bg-after

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -238,7 +238,7 @@ $tabpanel-border-width: 1px !default;
 
 $tablist-bg-color: #ebf2fc !default;
 $tablist-hover-bg-color: $clrMedium !default;
-$tablist-active-bg-color: #fff !default;
+$tablist-active-bg-color: null !default;
 
 $tablist-active-link-border-color: #666 !default;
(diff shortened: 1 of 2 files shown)


kodecount:
Well this affect https://github.com/wet-boew/wet-boew/pull/7759 which sets tabs to always have a white background (as well as fix print issues).

Issue #7643, opened 2016-08-23
Problem with the left menu. Version:  GoC.WebTemplate 4.0.21.2


ghost:
The second level of the left menu is showing "Selected" for more than one file.

It should be associated only with the file "Documents.aspx" but it is also "selected" when displaying the file "DocumentsUpload.aspx".

Here is the code that I am using to generate the left menu in both pages:

``` vb
Dim oMenuSection As MenuSection = New MenuSection()

' Level 1
oMenuSection.Name = "Financial Transaction"
oMenuSection.Link = "FinancialTransaction.aspx"

' Level 2
oMenuSection.Items.Add(New MenuItem("ReviewersSignatories.aspx", "Reviewers and Signatories"))
oMenuSection.Items.Add(New MenuItem("Documents.aspx", "Supporting Documents"))
oMenuSection.Items.Add(New MenuItem("HistoricalVersions.aspx", "Historical Versions"))

' Add Menu Section
Page.WebTemplateMaster.WebTemplateCore.LeftMenuItems.Add(oMenuSection)
```

FYI, When removing the breadcrumbs in both pages, this problem disappear.

Thank you

![file_1](https://cloud.githubusercontent.com/assets/21176705/17891187/43e99192-6908-11e6-98dd-40cde8306962.png)
![file_2](https://cloud.githubusercontent.com/assets/21176705/17891188/43f46612-6908-11e6-87a2-39f32e77aac5.png)

rlambert27:
Just to mention that this also fails on GCWeb.

shawnthompson:
Looks like it might be a problem with the [navcurrent plugin](https://github.com/wet-boew/wet-boew/tree/master/src/plugins/navcurrent).

Fu22yWu22y:
We did a Temp  work around and added this to a custom css file 

#wb-sec .list-group .list-group .list-group .wb-navcurr.list-group-item {
    background-color: #335075;
}

hsrudnicki:
Can be this closed as resolved or otherwise? Please update the status of this again, whether this remains still an issue or otherwise, thx

shawnthompson:
We can't close this one, I'd like to solve the problem in the JS and not have to override it with CSS just to mask the issue. If anyone can solve the problem, I can commit it to WET if they want.

BTW @Fu22yWu22y do you have a bear?

kpduncan:
[...]

PR #7859, opened 2017-01-24
Navcurrent - fixing #7643

shawnthompson:
@kodecount @duboisp 

Let me know if this is all that needs to be done.

Thanks

### DIFF
--- a/src/plugins/navcurrent/navcurrent.js
+++ b/src/plugins/navcurrent/navcurrent.js
@@ -39,7 +39,7 @@ var componentName = "wb-navcurr",
 				pageUrlQuery = windowLocation.search,
 				match = false,
 				className = classNameOverride ? classNameOverride : componentName,
-				len, i, j, link, linkHref, linkUrl, linkQuery, linkQueryLen,
+				child, len, i, j, link, linkHref, linkUrl, linkQuery, linkQueryLen,
 				localBreadcrumbLinks, localBreadcrumbLinksArray, localBreadcrumbLinksUrlArray,
 				localBreadcrumbQuery, localBreadcrumbLinkUrl;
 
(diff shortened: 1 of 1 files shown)


duboisp:
@shawnthompson have you try to git cherry pick @kodecount [commit](https://github.com/kodecount/wet-boew/commit/72df780dcf7426257751698c2b1dc4665bb1dbae) instead of copying it?
Prior to cherry pick you will need to add his repo into your local remote (```git add remote RemoteName RemoteURL```) then fetch it and then you should be able to cherry pick

Let me know if you have question

:-)

shawnthompson:
@duboisp Like this?

duboisp:
@shawnthompson  Yes

duboisp:
I successfully reproduced the bug, and I confirm this PR fix it.

LaurentGoderre:
It would be nice to do unit tests instead of manual tests for stuff like this

duboisp:
For this particular issue, it could be difficult because I was able to reproduce the issue by looking at
http://wet-boew.github.io/v4.0-ci/demos/index-en.html compare to http://wet-boew.github.io/v4.0-ci/demos/tables/tables-en.html

but I was unable to reproduce by using this domain
http://wet-boew.github.io/wet-boew/demos/index-en.html compared to http://wet-boew.github.io/wet-boew/demos/tables/tables-en.html

so I manually modified the ajaxed menu in order to reproduce it locally.

But @LaurentGoderre I agree with you that simple thing should have a unit test for.

kodecount:
We actually got a complaint about this during testing of new application multiple users were confused on how to get back to previous page (menu has one link, but it was dark blue as previous page (even though it a sub page of previous page)). We won't see this fix until the next stable version of WET but it goes to show that such small issue could easily confuse people.

PR #7867, opened 2017-01-26
Multimedia player: Removed orphaned WET 3.1 demo files.

EricDunsworth:
@duboisp FYI.


LaurentGoderre:
The player still support those files even if it's not demoed. It's not a WET 3 orphan

EricDunsworth:
@LaurentGoderre I know the WET 4.0 media player still supports XML captions. But I don't see any point in including demo XML files and transcript pages from WET 3.x as hidden "bonuses" that aren't shown or referenced anywhere in the repo.

IMO the demos should be updated to show off support for XML captions. Since the video demo page's first two examples are currently basically the exact same thing for 2 different videos, maybe the second video's example could be reworked into an XML captions example in another PR. I could prep it up if you're cool with the idea.

duboisp:
If it still supported, what about creating a demo that use those files?

EricDunsworth:
@duboisp The [WET 3.1 multimedia player](http://wet-boew.github.io/wet-boew-legacy/v3.1/demos/multimedia/multimedia-en.html)'s demo videos are no longer available. So we wouldn't be able to tie the XML/transcript files to an actual video.

Since the second video example is currently redundant, it should be super straightforward to convert its HTML captions into XML along with a few corresponding adjustments. I can do it in a separate PR if no one's against it.

duboisp:
If those file are still supported, so we should keep them and re-think how we can link to it.

So we can: 
* Create a new demo that show how to use those files. But creating a demo may not best fit the purpose of keeping those file and may require to host/link to more videos.
* Move those file in the [multimedia player documentation](http://wet-boew.github.io/wet-boew/docs/ref/multimedia/multimedia-en.html) and add link to it, as an example, from the docs. The link for those files could make senses in the step 4 of the video heading under "how to implement" where it describe what kind of source caption it is supported by the player. 

I do think improving the docs will be the best option.

duboisp (changes requested):
Move those file in the documentations https://github.com/wet-boew/wet-boew/tree/master/site/pages/docs/ref/multimedia and include link to it from the [multimedia player documentation](http://wet-boew.github.io/wet-boew/docs/ref/multimedia/multimedia-en.html)

EricDunsworth:
@duboisp The goal of this PR was just to remove orphaned files. I don't think it would be wise to rework it to achieve a different kind of purpose (i.e. content rewrites).

I've opened #7901 to revise the second video demo to use TTML captions. That should address some of your concerns. It doesn't add any documentation updates, but FWIW the [multimedia player](http://wet-boew.github.io/v4.0-ci/docs/ref/multimedia/multimedia-en.html)'s documentation at least contains a couple of mentions of TTML support.

duboisp (approved):

duboisp:
New Demo introduced in #7901

Issue #7618, opened 2016-07-19
Feeds: Show description for RSS feeds?


aaornrylow:
Hi,

Is it possible to include the RSS <description> field in the WET4 feeds widget? It was displayed by default in the WET3 version.

Thanks,
-Aaron

crochefort:
are you talking about this http://wet-boew.github.io/v4.0-ci/demos/feeds/feeds-en.html ??

aaornrylow:
Yes. That's the one.

aaornrylow:
My first comment filtered it out because I used `<description>` without "code" markup, but just to be clear I was talking about the RSS description field.

aaornrylow:
Any update on this issue?

hsrudnicki:
Can be this closed or otherwise? Please update the status of this again, whether this remains still an issue? Thx

aaornrylow:
This is still an issue.

aaornrylow:
Any word on this? The WET3 version is currently broken ([even on your site](http://wet-boew.github.io/wet-boew-legacy/v3.1/demos/webfeeds/webfeeds-en.html)). It just shows the spinning "loading" symbol.
We'd like to move to the WET4 version but it's not very practical in our context without also being able to show the description field.

kodecount:
[WET3 support was ended](http://wet-boew.github.io/wet-boew/docs/versions/rdmp-en.html):

> 2014-09-23	
> v4.0.6 released
> v3.1.15 released
> End of support for v3.1.x

Also WET 3 used ajax.googleapis.com which is now deprecated.

kodecount:
On another note what is it your trying to get a feed for? Lot of companies slowly ending support for these type of feeds (like Facebook).

aaornrylow:
We're not asking for further support on the WET3 version. The issue is with the current [WET**4** feeds plugin](http://wet-boew.github.io/v4.0-ci/demos/feeds/feeds-en.html) when using it to pull from a standard RSS (XML) feed ([here's the current feed on our website, for example](http://www.crtc.gc.ca/eng/rss/2017releases-jan.xml)). Unlike the WET3 version it does not have the option to display the **description** field from the RSS file. This field provides key info for us so we need it to display.

LaurentGoderre:
I'm afraid that's not easily done in WET4

duboisp:
@aaornrylow I understand your feature request but it might be only done for ATOM feed not RSS feed.

[...]

PR #7875, opened 2017-02-01
Data-json template and alternative data-ajax syntax

### LINKED ISSUE #7782: Change URL of data-ajax-replace before it gets loaded?
Hello,

Wondering if someone has an idea on how to change the URL of the "Government of Canada activities and initiatives" BEFORE the ajax call is done. We want to add cache busting to the URL so that for example:

```
data-ajax-replace="https://cdn.canada.ca/gcweb-cdn-live/features/features-en.html"
```

Would give this:

```
data-ajax-replace="https://cdn.canada.ca/gcweb-cdn-live/features/features-en.html?cacheBust=(add date.hour here so that it refreshes every hour)"
```

Unfortunatly in our case we can't change it directly in the source code because we're using compiled code that doesn't allow Javascript code to be compiled (it gets escaped) but we can add it in an external JS file.

Ideas?

duboisp:
This PR update the data-json plugin to add new functionality like supporting ```<template>``` element, providing more options on how to insert json data and referring to a dataset in url. The dataset feature will allow to slightly manipulate the data, by a middleware plugin during the fetch operation. The dataset feature is required and will be demonstrated by the upcoming JSON manager that I am currently finalizing it.

Data JSON
http://universallabs.org/wet/labs/data-json/dist-1.1-b1/demos/data-json/data-json-en.html
http://universallabs.org/wet/labs/data-json/dist-1.1-b1/demos/data-json/data-json-fr.html

Template HTML 5
http://universallabs.org/wet/labs/data-json/dist-1.1-b1/demos/data-json/template-en.html
http://universallabs.org/wet/labs/data-json/dist-1.1-b1/demos/data-json/template-fr.html

Documentation
http://universallabs.org/wet/labs/data-json/dist-1.1-b1/docs/ref/data-json/data-json-en.html

---

@LaurentGoderre said in #7618
> That data-json plugin sounds redundant to me. If you want binding like this, shouldn't you use something like angular instead of creating a pseudo angular?

@LaurentGoderre I agree that both may have similar functionality, but angular aren't integrated to wet-boew yet.

I remember that, two-three years ago, there have been informal discussion about interacting with an API in wet-boew but nothing concrete was delivered. The data-json plugin is concrete, tested, functional and ready to be use on production.

If the functionality supported by data-json plugin could be replaced by a better plugin, I am proposing to re-visit it when it will be the time to migrating existing plugins for the next major release which would require major features re-write.

---

This PR also include updates that allow to change the url of data-ajax before it gets loaded. fix #7782

@rlambert27 As discussed in #7782, it's now ready for testing.

I didn't included intentionally an explicit working example for the cache busting feature because a proper server configuration should be use and are very preferable. Instead I added a demo of an alternative syntax of calling the data-ajax plugin. With that alternative syntax you can add the configuration option ```"nocache": true``` for session caching busting  or ```"nocache": "nocache"``` to always reset the cache. That feature is also implemented in data-json.

Data Ajax - (added an alternative syntax)
http://universallabs.org/wet/labs/data-json/dist-1.1-b1/demos/data-ajax/data-ajax-en.html
http://universallabs.org/wet/labs/data-json/dist-1.1-b1/demos/data-ajax/data-ajax-fr.html

Documentation:
http://universallabs.org/wet/labs/data-json/dist-1.1-b1/docs/ref/data-ajax/data-ajax-en.html

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -908,6 +908,18 @@ module.exports = (grunt) ->
 					"dist/unmin/demos/**/ajax/**/*.html"
 					"dist/unmin/assets/*.html"
 				]
+			templates:
+				options:
+					ignore: [
+						"The “details” element is not supported properly by browsers yet. It would probably be better to wait for implementations."
+						"Element “dl” is missing a required instance of child element “dd”."
+						"Element “dl” is missing a required instance of child element “dt”."
+						"Empty heading."
+					]
+				src: [
+					"dist/unmin/demos/data-json/template-en.html"
+					"dist/unmin/demos/data-json/template-fr.html"
+				]
 			all:
 				options:
 					ignore: [
@@ -922,6 +934,8 @@ module.exports = (grunt) ->
 					"!dist/unmin/assets/**/*.html"
 					"!dist/unmin/demos/menu/demo/*.html"
 					"!dist/unmin/test/**/*.html"
+					"!dist/unmin/demos/data-json/template-en.html"
+					"!dist/unmin/demos/data-json/template-fr.html"
 				]
 
 		bootlint:
--- a/Rakefile
+++ b/Rakefile
@@ -6,7 +6,7 @@ task :default => [:test]
 
 task :test do
 	HTML::Proofer.new("dist/", {
-		:href_ignore => [
+		:url_ignore => [
 			"#",
 			# The additional anchor link is picked up from the Geomap JSON, but shouldn't be flagged
 			"\\\"#\\\"",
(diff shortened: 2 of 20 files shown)


duboisp:
Travis Build are only failing on an anchor for having an empty ```href``` attribute that is inside a ```<template>``` element which should be ignored as per the specification the inner content of template element are not intend to be displayed as is.

So, do someone know how I can add an exception for the HTML-Proofer and how I can run those test locally?

LaurentGoderre:
@duboisp we would not integrate angular as part of WET, we would ensure that the two are compatible. I have created angular applications that use WET as the front end which gives you the same functionality as this does with a lot more.

My personal opinion is that there should be a clear divide between WET as a front end layer and MVC functionality like data binding.

duboisp:
Fixed Travis build error

duboisp:
To be moved into the gc web theme, as this plugin execute business logic.

duboisp:
This plugin is merged in GC Web and data alternate syntax is merged
See working example and docs here: http://wet-boew.github.io/themes-dist/GCWeb/index-en.html

PR #7882, opened 2017-02-02
Json manager

duboisp:
New middleware plugin that manage dataset and apply JSON Patch.

* [JSON Manager](http://universallabs.org/wet/labs/json-manager/v1.0-b1/demos/jsonmanager/jsonmanager-en.html)
* [Gestionnaire JSON](http://universallabs.org/wet/labs/json-manager/v1.0-b1/demos/jsonmanager/jsonmanager-fr.html)
* [Documentation](http://universallabs.org/wet/labs/json-manager/v1.0-b1/docs/ref/jsonmanager/jsonmanager-en.html)

----
Depend on PR #7875

PR #7883, opened 2017-02-03
Url mapping

duboisp:
New plugin that execute pre-configured action based on url query string.

English
* [Table filtering](http://universallabs.org/wet/labs/urlmapping/v1.0-b1/demos/urlmapping/tblfilter-en.html)
* [Ajax](http://universallabs.org/wet/labs/urlmapping/v1.0-b1/demos/urlmapping/ajax-en.html)
* [Patching JSON](http://universallabs.org/wet/labs/urlmapping/v1.0-b1/demos/urlmapping/patches-en.html)
* [Documentation](http://universallabs.org/wet/labs/urlmapping/v1.0-b1/docs/ref/urlmapping/urlmapping-en.html)

French
* [Filtrage de tableau](http://universallabs.org/wet/labs/urlmapping/v1.0-b1/demos/urlmapping/tblfilter-fr.html)
* [Ajax](http://universallabs.org/wet/labs/urlmapping/v1.0-b1/demos/urlmapping/ajax-fr.html)
* [Appliquer des correctifs JSON](http://universallabs.org/wet/labs/urlmapping/v1.0-b1/demos/urlmapping/patches-fr.html)

----

Depend on #7882


LaurentGoderre:
This looks like a huge security risk for cross site scripting and I would recommend not merging this unless extensive security testing has been done.

duboisp:
I consider the security risk as the same as delivering a page from a non secure context, like HTTP. There is a few standards out there that can be used/implemented in order to re-enforce the security, for example the [Content Security Policy Level 2](https://www.w3.org/TR/CSP2/),

---
FYI
The following are required to allow the the URL mapping plugin to execute pre-configured action based on url query string:
  * Any action must be explicitly configured.
  * Actions are only executed if they exactly match the mapped URL parameter.
  * By default, only the first action will be executed if a match is found.

If the page need to interact based on a value specified by an URL parameter, the following are required:
  * Any action must be explicitly configured.
  * Actions are only executed if they exactly match the mapped URL parameter.
  * By default, only the first action will be executed if a match is found.
  * The value must match a configured regular expression 
  * The match return value will be used as the value to be replaced by.
  * The replacement location of the replacement value must be explicitly specified.
  * A default value must be set to allow a fall back if the regular expression return no match or an empty string.

LaurentGoderre:
I want to avoid the swiss army/macro plugin that those multiple things. The table filtering is best done in a datatable plugin that can be loaded only when need be and contributed back to the datatable (if such plugin doesn't already exists)

As for the JSON Patch, I feel the same was as #7882. Business logic shouldnt be included in WET.

I feel like you should create an extention repo for your extensions that you can offer to those who want it without forcing those who use angular or other methods of doing the same thing to hav duplicate code. I have tested this with my filter plugin and works really good. We could add the template polyfill to make it easier for you to package your extension.

duboisp:
I am for it to have an extension repo as long I have an official confirmation (from the PP) saying those extension will be available and supported for the content published on Canada.ca website.

duboisp:
To be moved into the gc web theme, as this plugin execute business logic.

duboisp:
This plugin is merged in GC Web
See working example and docs here: http://wet-boew.github.io/themes-dist/GCWeb/index-en.html

PR #7894, opened 2017-02-15
Collapsible alerts: Fixed summary arrow issues in IE11/Edge.

### LINKED ISSUE #7886: Collapsible alert - IE11/Edge - align with default dehavior
The collapsible alert do not exactly behave as the default details/summary in IE11/Edge

As discussed in https://github.com/wet-boew/wet-boew/pull/7822#issuecomment-278213861

In Standard mode: There is an arrow resize effect when refreshing the page.
In wbdisable mode: The arrow is still visible.

Some suggestion:
* Align collapsible alert with default behavior of details/summary,
  * Remove [Line 101](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/collapsible-alerts/base.scss#L101) and [Line 110](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/collapsible-alerts/base.scss#L110)
* Change the default behavior of details/summary for inserting the arrow in wbdisable  mode
  * Add css line like ```.wb-disable.no-details>summary:before{ content: "\25BC\a0"; } ```

EricDunsworth:
* Prevents arrows from resizing/flickering on page load.
* Disables arrows in wbdisable/noscript modes.
* Prevents a CSS selector containing two consecutive open attributes ([open][open]) from getting generated.
* Fixes #7886.

@duboisp FYI I've only tested this in IE11. Feel free to double-check it in Edge if you want. Since this PR basically just added parent classes and didn't modify any CSS properties, it should work identically in both browsers.

### DIFF
--- a/src/plugins/collapsible-alerts/base.scss
+++ b/src/plugins/collapsible-alerts/base.scss
@@ -89,38 +89,39 @@ details {
 	}
 }
 
-.no-details {
-	details {
-		&.alert,
-		&.alert[open] {
-			> {
-				summary {
-					margin-left: 1.4em;
-
-					&:before {
-						content: "\25BA\a0";
-					}
-				}
-			}
-
-			&[open] {
+.wb-enable {
+	&.no-details {
+		details {
+			&.alert {
 				> {
 					summary {
+						margin-left: 1.4em;
+
 						&:before {
-							content: "\25BC\a0";
+							content: "\25BA\a0";
+						}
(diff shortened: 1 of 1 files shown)


duboisp:
Thanks  @EricDunsworth I tested and it work as fine,

That fix make the Collapsible alerts to behave like any other details/summary.

@lucas-hay From my point of view, this PR are safe to be merge in.

PR #6907, opened 2015-06-30
Removal of WET v3.1 notice - close #6802

duboisp:
The notice was move into the documentation main page and a link to it was added in the home page under key resources..

The same text was intentionally kept as is, but the title was renamed to have a parallel structure for the key resources list.

Cheers

PR #7899, opened 2017-02-17
wb-disable - Add a notice to disabled mode

### LINKED ISSUE #7686: Basic mode - wbdisable link
Hi,

We recently had a Webex session with a visually disabled user using our Job Bank website. The user was really confused because he was having trouble creating a new job posting. What we quickly learned when he shared his screen was that he was in "basic" mode. 

We asked him if he knew that he was in basic mode and what the difference was and he had no idea.

I'm wondering if this option is really necessary? Do we have stats on how many people actually use this option?

My suggestion is that this should be optional in WET, especially for web applications that don't want to display that option because they rely heavily on JavaScript for their web application to work. 

I can see this maybe being useful, maybe....., for static websites with simple static content. But for web applications this option is causing confusion!

@masterbee @shawnthompson @EricDunsworth @LaurentGoderre @pjackson28 

Thoughts?

bsouster:
fix #7686 
   
 Updated wbdisable plugin with notice for users who accidentally open
    the page in disabled mode

    Fixed the issue where wbdisable and wbenable buttons would generate
an
    undefined parameter in the URL

    Added fallback for non-English and French pages until translation is
    added to i18n. Wbdisable buttons currently have no translation —
when
    this is added, the temporary fix can be removed

    Here the i18n variable that need to be translated:
    * disable-notice-h
    * disable-notice
    * wb-enable
    * wb-disable


src/plugins/wb-disable/disable.js
```diff
@@ -30,15 +30,29 @@ var componentName = "wb-disable",
 			$html = wb.html,
 			i18n = wb.i18n,
 			pageUrl = wb.pageUrlParts,
-			li, param;
+			li, param,
+			i18nTemporary = {
```
LaurentGoderre: Remove this and move to the i18n file when this ready to go

src/plugins/wb-disable/disable.js
```diff
@@ -30,15 +30,29 @@ var componentName = "wb-disable",
...
+				en: {
+					"disable-notice-h": "Notice: Basic HTML",
+					"disable-notice": "You are viewing Basic HTML view. Some features may be disabled."
+				},
+				fr: {
+					"disable-notice-h": "Avis: Version HTML simplifiée",
+					"disable-notice": "Vous naviguez présentement sur la version HTML simplifiée de cette page. Certaine fonctionalités peuvent être déactivées."
+				}
+			},
+			noticeHeader = i18n( "disable-notice-h" ),
```
LaurentGoderre:
```js
noticeHeader = i18n( "disable-notice-h" ) ||  i18nTemporary[wb.lang]["disable-notice-h"],
```

src/plugins/wb-disable/disable.js
```diff
@@ -30,15 +30,29 @@ var componentName = "wb-disable",
...
+			noticeBody = i18n( "disable-notice" ),
```
LaurentGoderre:
```js
noticeBody = i18n( "disable-notice" ) ||  i18nTemporary[wb.lang]["disable-notice"],
```

src/plugins/wb-disable/disable.js
```diff
@@ -53,11 +67,19 @@ var componentName = "wb-disable",
...
-					li.innerHTML = "<a class='wb-sl' href='" + nQuery + "wbdisable=false'>" + i18n( "wb-enable" ) + "</a>";
+					// Temporary fix for i18n until text are translated
+					if ( !noticeHeader && wb.lang === "fr" ) {
+						noticeHeader = i18nTemporary.fr[ "disable-notice-h" ];
+						noticeBody = i18nTemporary.fr[ "disable-notice" ];
+					} else if ( !noticeHeader ) {
+						noticeHeader = i18nTemporary.en[ "disable-notice-h" ];
+						noticeBody = i18nTemporary.en[ "disable-notice" ];
+						noticehtml = noticehtml + " lang='en'";
+					}
```
LaurentGoderre: Remove this whole block in favor of the changes above

LaurentGoderre (dismissed):
Can you make the following changes to make the changes to the PR smaller when the i18n strings are added to the csv?

src/plugins/wb-disable/disable.js
duboisp: What file? I though you said that the i18n file was an spreadsheet on google drive with edit restricted to a few members.

Is that documented somewhere?

kodecount: Yeah their a google spreadsheet. I don't think it documented with the public.

https://docs.google.com/spreadsheets/d/1BmMrKN6Rtx-dwgPNEZD6AIAQdI4nNlyVVVCml0U594o/edit#gid=1

duboisp: Thanks @kodecount, that's what I thought, but that doesn't help as I can only view it.

kodecount: True. LOL

src/plugins/wb-disable/disable.js
duboisp: Your suggestion would prevent to show the notification and the possibility for the user to return back in standard mode by clicking a link for languages that are not defined in the i18nTemporary. For example if  ```wb.lang === "it"```

src/plugins/wb-disable/disable.js
duboisp: This propose logic will at least nicely fall back on English text if the ```wb.lang``` is not defined in the i18n or in the i18nTemporary. And when the i18n text are falling back to English. the language of the notice section is set to English. That prevent to fail WCAG SC3.1.2 Language of Parts level AA when the current page language is not English or French.

src/plugins/wb-disable/disable.js
duboisp: Your suggestion would prevent to show the notification and the possibility for the user to return back in standard mode by clicking a link for languages that are not defined in the i18nTemporary. For example if  ```wb.lang === "it"```

src/plugins/wb-disable/disable.js
LaurentGoderre: The current i18n framework defaults to the english when no translation is provided and doesn't add the `lang=en` so that is a wider issue.

src/plugins/wb-disable/disable.js
LaurentGoderre: In that case the code would be:

```js
noticeHeader = i18n( "disable-notice-h" ) ||  i18nTemporary[wb.lang !== 'en' ? 'en' : 'fr']["disable-notice-h"],
```

src/plugins/wb-disable/disable.js
LaurentGoderre:
```js
noticeBody = i18n( "disable-notice" ) ||  i18nTemporary[wb.lang === 'fr' ? 'fr' : 'en']["disable-notice"],
```

src/plugins/wb-disable/disable.js
lucas-hay: Did we come to a conclusion on this one guys?

src/plugins/wb-disable/disable.js
duboisp: @LaurentGoderre then what about the WCAG language of part issue? Should be ignored for this PR when we have a way to avoid it?

src/plugins/wb-disable/disable.js
duboisp: @lucas-hay considering the wb-disable is a real core feature of WET, I do think we should not introduce an aware WCAG issue especially when we have a work around for it. 
As it was mentioned in this PR description, that block of code is just a temporary fix until the text get inserted in the i18n file. 
By doing it, we would make fail WCAG all page that are not in English or French.

May be later, we can discuss about a better fallback for text that has been not translated yet.

lucas-hay: @duboisp I agree, a discussion will follow

@LaurentGoderre Do we currently have items that do not have text translated and fall back to English?

nschonni:
You can mock what the Google spreadsheet changes should be by modifying the CSV after the build has pulled it down

src/plugins/wb-disable/disable.js
LaurentGoderre: In that case, remove the `i18nTemporary` altogheter and put it inline here so that only one block of code needs to be removed.

LaurentGoderre: @lucas-hay yeah, quite a lot for languages other than french and english

duboisp:
@nschonni I thought that CSV file was overwritten during the build by whatever the google spreadsheet is.

nschonni:
@duboisp it's static in the repo https://github.com/wet-boew/wet-boew/blob/master/src/i18n/i18n.csv unless someone calls `update-i18n` to sync the latest changes from the Google Docs https://github.com/wet-boew/wet-boew/blob/master/Gruntfile.coffee#L108

lucas-hay:
@duboisp 
Given that we have other issues with this, do you want to remove your code, and we also add this as a discussion item on how to get all of the existing translations that are missing?
If we add this temporary fix, it should really be added for all of the missing translations.

duboisp:
@lucas-hay this fix work for all language, as when it fallback, the '''lang=en''' is added to the notice box. Using the @LaurentGoderre proposal will make it fail WCAG as we are unaware that the language has been fallback.

@nschonni I thought Travis do but I am not sure at 100%

nschonni:
If you look at the `.travis.yml` it just calls https://github.com/wet-boew/wet-boew/blob/master/script/setup then the `travis` prefixed NPM script targets https://github.com/wet-boew/wet-boew/blob/master/package.json#L6

duboisp:
@nschonni see @LaurentGoderre comment about it: https://github.com/wet-boew/wet-boew/pull/6907#issuecomment-117200194

My understanding is by doing update directly to the csv it will make the translation/text vulnerable to be removed by mistake. So I think it is safer to include i18n content directly in the plugin as a temporary measure until it get included in the official spreadsheet.

LaurentGoderre:
@duboisp my proposal doesn't make it fail WCAG since it's already failing on that criteria in many languages

lucas-hay:
At this point guys unless we can agree in the next 15 mins, this will have to be pushed to the next release.

duboisp:
@lucas-hay I think that we should not introduce any WCAG issue that we are aware of and knowing we have a work around for it. Especially for a core plugin. So let's wait for the next release then.

kodecount:
What about my original commit? I updated the i18n in the commit https://github.com/kodecount/wet-boew/commit/c5ed2fa80c148f89033aa5575f606cf2305833a4

src/plugins/wb-disable/disable.js
```diff
@@ -67,12 +89,19 @@ var componentName = "wb-disable",
 						// Store preference for WET plugins and polyfills to be enabled in localStorage
 						localStorage.setItem( "wbdisable", "false" );
 					}
+
+					// Remove variable from URL
+					var lc = window.location.href.replace( "wbdisable=false", "" ).replace( "?#", "#" );
+					if ( lc.indexOf( "?" ) === ( lc.length - 1 ) ) {
+						lc = lc.replace( "?", "" );
+					}
+					window.history.replaceState( "", "", lc );
```
kodecount: Is their no concerns over manipulation of the URL?

kodecount: Is their no concerns over manipulation of the URL?

ghost (changes requested):

src/plugins/wb-disable/disable.js
kodecount: Maybe this should be optional? Like a `data` setting?

kodecount:
https://github.com/kodecount/wet-boew/commit/527df614519467df7fd333910e9b4cef1f75fd98

src/plugins/wb-disable/disable.js
duboisp: I do not see any concern here as it the history information of the current page that is updated for an url that do not include "wbdisable=false".

duboisp:
> What about my original commit? I updated the i18n in the commit kodecount@527df61

@kodecount in the past I made some change to the i18n csv file but [I was asked to removed it](https://github.com/wet-boew/wet-boew/pull/6907#issuecomment-117200194). I do think that still apply today as the i18n methodology has not been change since. 

May be in the future we can leverage two i18n files, one that contain all the language and a second one to temporary support translation until it get officially merge and translated into multiple languages in the master i18n spreadsheet. But before to work on developing a such feature, I will need a little bit more support where I can know it will make it to WET eventually.

LaurentGoderre:
Is everyone ok with this? If so, I will add the i18n.

kodecount:
If you mean adding the i18n to the pull request and not the google document @duboisp express concern at https://github.com/wet-boew/wet-boew/pull/7899#issuecomment-282332936

LaurentGoderre:
To the Google document I mean

colleenhickey:
Is there a timeline for when this will be included in a release?

duboisp:
As discussed at the PR review meeting, I will update the i18n spreadsheet and then @bsouster would be able to update this PR.

LaurentGoderre:
@duboisp you can update the PR too. There's a a doc on github to do this

LaurentGoderre:
@duboisp any update?

bsouster:
The i18n spreadsheet is updated and the PR is ready for review

LaurentGoderre:
The code looks good to me. @shawnthompson can you create an example in gh-pages?

lucas-hay (approved):
Requested changes have been made.

PR #7901, opened 2017-02-20
Multimedia player: Revised second video demos to use TTML captions.

EricDunsworth:
* Added TTML captions files for the second video, based on its HTML captions and the orphaned WET 3.x TTML files that #7867 removed.
  * French TTML file is now using an xml:lang="fr" attribute. The former French TTML file that #7867 removed was using xml:lang="en".
* Renamed and unveiled the demo pages' first two video headings. They were previously visually-hidden.
* Revised the second video's example and code samples in the demo pages.
* Removed HTML caption classes/data attributes from the second video's transcript pages.


LaurentGoderre:
I looked at it quickly, it looks good to me

PR #7908, opened 2017-02-22
Auto add close button to modal dialogs

ghost:
Related:
- https://github.com/wet-boew/GCWeb/issues/1156
- https://github.com/wet-boew/wet-boew/pull/7909

Also AJAX created Modals don't get button created.

### DIFF
--- a/src/plugins/lightbox/lightbox.js
+++ b/src/plugins/lightbox/lightbox.js
@@ -120,6 +120,7 @@ var componentName = "wb-lbx",
 		if ( !i18nText ) {
 			i18n = wb.i18n;
 			i18nText = {
+				fClose: i18n( "close" ),
 				tClose: i18n( "overlay-close" ) + i18n( "space" ) + i18n( "esc-key" ),
 				tLoading: i18n( "load" ),
 				gallery: {
@@ -142,9 +143,12 @@ var componentName = "wb-lbx",
 					var $item = this.currItem,
 						$content = this.contentContainer,
 						$wrap = this.wrap,
+						$modal = $wrap.find( ".modal-dialog" ),
 						$buttons = $wrap.find( ".mfp-close, .mfp-arrow" ),
 						len = $buttons.length,
 						i, button;
+
+					createCloseButton( $modal );
 
 					$document.find( "body" ).addClass( "wb-modal" );
 					$document.find( modalHideSelector ).attr( "aria-hidden", "true" );
@@ -232,6 +236,7 @@ var componentName = "wb-lbx",
 					} else {
 						$response = $( mfpResponse.data );
 					}
+					createCloseButton( $response );
 
 					$response
 						.find( ".modal-title, h1" )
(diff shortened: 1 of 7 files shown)


shawnthompson:
Travis build is fail, needs at PhantomJS test. Need to figure out why.

cc/ @duboisp @LaurentGoderre @kodecount

kodecount:
See a demo: https://kodecount.ca/demos/overlays/

kodecount:
As for the AJAX this maybe out of our control as it being created from another page (DOM). So in this situation the most we could do I think is recommend them to add one unless their another area we call the code.

LaurentGoderre:
It's possible that the build is failing because this PR introduces a JS error.

kodecount:
Could have something to do with mixing pure JavaScript and jQuery if the build is fussy about that.

src/plugins/lightbox/lightbox.js
```diff
@@ -142,10 +143,42 @@ var componentName = "wb-lbx",
...
 						$content = this.contentContainer,
 						$wrap = this.wrap,
+						$modal = $wrap.find( ".modal-dialog" ),
 						$buttons = $wrap.find( ".mfp-close, .mfp-arrow" ),
 						len = $buttons.length,
 						i, button;
 
+					if ( $modal && $modal.length !== 0 ) {
+						var footer = $modal.find( ".modal-footer" )[ 0 ],
+							hasFooter = ( footer && footer.length !== 0 ) ? true : false,
```
kodecount: With `( footer && footer.length !== 0 )` maybe it should be `( footer !== null )`.

LaurentGoderre: Yeah and that line is over-complicated, you can do just:

```js
var $footer = $modal.find( ".modal-footer" )[ 0 ],
hasFooter = footer !== null
```

kodecount: Okay. When I get a chance I'll fix that line.

kodecount:
Figured out how to add it to AJAX content.

LaurentGoderre:
The build is still failing

kodecount:
I'm gonna be offline for a little while (could be a week or more), but when I get back I can work on this.

kodecount:
If this fails I'm at a lost. The code was broken, so should work now.

shawnthompson:
https://shawnthompson.github.io/kodecount-modal-close/unmin/demos/overlay/overlay-en.html

lucas-hay:
@kodecount is there a way you could code this so the close button could be configured by the user. (allow user to add class name, customized text, etc)

kodecount:
If we use `data-wb-lbx` we have to tell it to ignore the settings for button. Easier solution is to use `data-wb-lbx-btn` and keep the current settings as the default if nothing is set.

src/plugins/lightbox/lightbox.js
```diff
@@ -253,6 +258,38 @@ var componentName = "wb-lbx",
...
+					spanTextFtr = footer.innerHTML + i18nText.tClose;
+				} else {
+					footer = document.createElement( "div" );
+					footer.setAttribute( "class", "modal-footer" );
+					footer.style.background = "#fff";
+					spanTextFtr = i18nText.tClose;
+				}
+				spanTextFtr = spanTextFtr.replace( "'", "&#39;" );
+
+				overlayCloseFtr = "<button type='button' id='ftrClose' class='btn btn-sm btn-default pull-left " + closeClassFtr +
```
duboisp: Can you change ```btn-default``` for ```btn-primary``` and remove the ```id='ftrClose'```

src/plugins/overlay/overlay-en.hbs
```diff
@@ -356,7 +355,6 @@ wb.doc.on( "click vclick", "#overlay-open-btn", function( event ) {
 				</ol>
 			</li>
 		</ol>
-		<button class="btn btn-primary popup-modal-dismiss" type="button">Close modal popup</button>
```
duboisp: Can you add a blurb in the overlay/lighbox documentation on how a author can add his own custom button. May be just reuse this working example and specify that it's need to be in a modal-footer.

duboisp (changes requested):

kodecount:
Do we require the ability for the user to customize the button or can we simply explain that by adding their own button to the modal footer well replace the default button? Also should we have the ability to turn this off? In the event we turned this off it would re-enable accessibility issues again.

kodecount:
Also on another note if I'm to change the default visual appeal from `btn-default` to `btn-primary` should this also take affect on the PR that was merged for just overlays?

duboisp:
The following is my opinion, may others have different thought.

I would say to add a new section in the documentation page like "about the close button" and "simply explain that by adding their own button to the modal footer well replace the default button" followed by a code same on how to do it.

I do not think we should allow the ability to turn this off as it is to solve an accessibility issue. If you want you can quickly mention why in the docs under "about the close button".

> Also on another note if I'm to change the default visual appeal from btn-default to btn-primary should this also take affect on the PR that was merged for just overlays?

I would said yes, and that could be a separate PR

kodecount:
Thanks for the input I agree. Not sure if @lucas-hay has any arguments?

kodecount:
I'll produce with add an example if he has none.

lucas-hay:
@kodecount I agree. Thanks for the idea @duboisp

kodecount:
All updated.

duboisp (approved):

PR #7913, opened 2017-02-23
Plugin stepsform (wb-steps): Provides the ability to split a form into steps

ghost:
Currently the issue is long forms on the business side. This well make forms feel shorter for the customer/user. So far the plugin only converts form fields into steps.

Please see Issue #7898 and demo at kodz.ca/projects/wet-boew/demos/forms/steps

Issue #6834, opened 2015-06-03
Validation of input type "file" behaves differently than others


RobJohnston:
This is minor, but it was brought to my attention that the validation of input type="file" behaves differently than others.  For example, on the [Form validation page](http://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-en.html), the required field error message is removed on most controls as soon as the user begins typing.  For the file input it is not removed when the user selects a file to upload but only disappears when the user moves control off the field.

I did find one issue on the jQuery forum about this:  https://forum.jquery.com/topic/jquery-validation-on-input-type-file

Can anything be done?

hsrudnicki:
Can be this issue closed ?

RobJohnston:
@hsrudnicki, You're caught-up to this week's issues.  At this rate, you'll be out of a job by the end of this afternoon.  Congrats!  

Let's give it a bit of time to see if anybody comes up with something clever.

hsrudnicki:
@RobJohnston : fine :+1:

hsrudnicki:
Made "Going Stale" to close this issue as resolved or otherwise?  
Either way, please update this issue's status, thank you

hsrudnicki:
Does this remain an issue? Please update again the status of this issue, in regards to the referenced issue, for the current year, thx

hsrudnicki:
Additionally, see the issue with number: 6245 if applicable

kpduncan:
@RobJohnston  @hsrudnicki  What I think we need for file type is a method like `data-rule-types="[.pdf , .doc]"` and maybe `data-rule-maxsize="3000"`. I could work on this.

RobJohnston:
@kpduncan For the types, one could write `data-rule-extension="pdf|doc"`.  That works now.

For the size, I don't think there is something like `data-rule-maxsize`.  I like it.  In the past, I've used the class `fileMaxSize` with a validator (example for ASP.Net):

```js
//Limit an individual file to 2MB.
jQuery.validator.addMethod("fileMaxSize", function (value, element) {
  var numFiles = $("#<%=FileUpload1.ClientID%>")[0].files.length;
  var iSize = 0;
  for (var i = 0; i < numFiles; i++) {
    iSize = ($("#<%=FileUpload1.ClientID%>")[0].files[0].size / 1024 / 1024); 
    if (Math.round(iSize * 100) / 100 > 2.0) {
      return false;    
    }
  }
return true;
[...]

PR #7921, opened 2017-02-28
Form Validation Update

ghost:
Allows on tab validation (currently need to hit submit before tab validation works).

Also fixes issue discussed in https://github.com/wet-boew/wet-boew/issues/6834 and https://github.com/wet-boew/wet-boew/pull/7913

### DIFF
--- a/src/plugins/formvalid/formvalid.js
+++ b/src/plugins/formvalid/formvalid.js
@@ -267,10 +267,22 @@ var componentName = "wb-frmvld",
 							}
 
 						}, /* End of showErrors() */
+
 						invalidHandler: function() {
 							submitted = true;
+						},
+
+						/* adds on tab validation */
+						onfocusout: function( element ) {
+							this.element( element );
 						}
+
 					} ); /* end of validate() */
+
+					/* fixes validation issue (see PR #7913) */
+					$form.on( "change", "input[type=date], input[type=file], select", function() {
+						$form.validate().element( this );
+					} );
 
 					// Clear the form and remove error messages on reset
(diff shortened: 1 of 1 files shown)


shawnthompson:
https://shawnthompson.github.io/kodecount-frmvld-update/unmin/demos/formvalid/formvalid-en.html

shawnthompson:
@neoinsight What do you think about this functionality?

https://shawnthompson.github.io/kodecount-frmvld-update/unmin/demos/formvalid/formvalid-en.html

neoinsight:
@shawnthompson This looks good - nice to have validation before submit rather than all at once. Although am not sure I'm looking at the right thing because if I tab on the current WET demos page, I get validation right away...

shawnthompson:
The current one does format validation only right away (using letter in a phone number) but not required fields

neoinsight:
Ah - that would be great for screen-reader users! They have been way more likely to miss required fields in our testing, and of course have a harder time handling the error. Nice and good fit with steps.

lucas-hay (approved):
Approveed at PR review meeting.

shawnthompson:
@kodecount @lucas-hay 

I'm using NVDA to test the new form validation required and notice the focus is off once the user tabs out of the required field, the user is notified that it's required but the focus is now on the next focusable element.

Not sure if this is an issue and I'm not sure what to do about it because if we force the focus, we would be holding the cursor hostage in the required field. 

Maybe adding further instructions for non visual users asking them to return to the previous field and fill in the required information???

Just a thought... Can turn this into an issue if we really think it's a problem.

Issue #7936, opened 2017-03-08
fieldset with textbox elements: reopen issue #7283


biljanap:
I would like to reopen #7283 because I have the exact same situation again.

kodecount:
The default method in jQuery Validate is:
```javascript
<div class="wb-frmvld">
<form action="#" method="get" id="validation-example">
<fieldset class="phone-group">
<legend class="required"><span class="field-name">Phone Book</span> <strong class="required">(required)</strong></legend>
<div class="form-group">
<label for="phone1">Cell Phone</label>
<input class="form-control" id="phone" name="phone1" type="text" data-rule-require_from_group='[1, ".phone-group"]' />
</div>
<div class="form-group">
<label for="phone2">Home Phone</label>
<input class="form-control" id="phone" name="phone2" type="text" data-rule-require_from_group='[1, ".phone-group"]' />
</div>
<div class="form-group">
<label for="phone3">Work Phone</label>
<input class="form-control" id="phone3" name="phone3" type="text" data-rule-require_from_group='[1, ".phone-group"]' />
</div>
</fieldset>
<input type="submit" value="Submit" class="btn btn-primary" /> <input type="reset" value="Reset" class="btn btn-default" />
</form>
</div>
```
The only issue is the error message doesn't display properly (displays on the last input).

RobJohnston:
The require_from_group method has had spotty success in WET.  In some versions it works, and it others, it doesn't.  See https://github.com/wet-boew/wet-boew/issues/7067.

kodecount:
Yes and even custom validation for this is tricky since you can't latch onto the error placement call (the error placement call is how WET places the errors on the page).

kodecount:
Did more testing and I can get it to work with WET but before I put a fix in WET the method needs to be fixed in jQuery Validation: https://github.com/jquery-validation/jquery-validation/issues/1972

kodecount:
Here is the fix for WET:
```
var group = $element.hasClass( "require_from_group" ) || $element.attr( "data-rule-require_from_group" );
if ( group ) {
$fieldset = $element.closest( "fieldset" );
if ( $fieldset.length !== 0 ) {
$legend = $fieldset.find( "legend" ).first();
if ( $legend.length !== 0 && $fieldset.find( "input[name='" + $element.attr( "name" ) + "']" ) !== 1 ) {
var strong = $legend.find( "strong.error" ),
id = $legend.attr( "id" );

if ( strong.length === 0 ) {
[...]

PR #7940, opened 2017-03-09
jQuery Validation: require_from_group (error handling fix)

ghost:
Fixes how WET handles the errors from the method `require_from_group`

Well display only 1 error (first element of the group).

Related: https://github.com/wet-boew/wet-boew/issues/7067 and https://github.com/wet-boew/wet-boew/issues/7936

### DIFF
--- a/src/plugins/formvalid/formvalid.js
+++ b/src/plugins/formvalid/formvalid.js
@@ -138,6 +138,7 @@ var componentName = "wb-frmvld",
 						// In this case we will place them in the associated label element
 						errorPlacement: function( $error, $element ) {
 							var type = $element.attr( "type" ),
+								group = $element.attr( "data-rule-require_from_group" ),
 								$fieldset, $legend;
 
 							$error.data( "element-id", $element.attr( "id" ) );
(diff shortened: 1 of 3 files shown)


kodecount:
Should we add an example to the demo page?

duboisp:
A working example could be always helpful or/and you can describe it in the plugin documentation.

kodecount:
I've added it to the docs since the demo seems long as it is.

src/plugins/formvalid/formvalid.js
```diff
@@ -138,6 +138,7 @@ var componentName = "wb-frmvld",
 						// In this case we will place them in the associated label element
 						errorPlacement: function( $error, $element ) {
 							var type = $element.attr( "type" ),
+								group = $element.hasClass( "require_from_group" ) || $element.attr( "data-rule-require_from_group" ),
```
nschonni: The class method of rules in JV was removed awhile ago

kodecount: Well remove it then.

kodecount:
Here is a demo: https://kodecount.ca/demos/formvalid/required_group

lucas-hay (approved):

biljanap:
@kodecount 
Hi there, I wanted to see your demo for group validation, but the page does not exist. Do you still have it?

biljanap:
Would somebody be able to give me an example for this? We don't want to build anything custom if solution is available in WET.
If an example exists, it should be added to Forma validation example page.

kodecount:
Sorry I use it as a testing server and I think I broke it. http://kodecount.ca/demos/formvalid/required_group.php.

kodecount:
Basically right now this exist in WET but it broken I think this was a way to patch this type of validation.

kodecount:
Oh didn’t notice it is merged so should be useable in WET

biljanap:
@kodecount 
Thanks :o)

PR #7948, opened 2017-03-16
[DNM] removed errors about Expected linebreaks. To allow different linebreaks, Windows vs. Unix.

shawnthompson:
https://github.com/eslint/eslint/blob/master/docs/rules/linebreak-style.md

cc/ @LaurentGoderre @nschonni you folks cool with this?


LaurentGoderre:
Ummm, why do you have different line breaks?

shawnthompson:
On the Windows system, might be a setting in Git... when you pull I think it changes them

LaurentGoderre:
Then change your git settings!

shawnthompson:
Thank I get linebreak warnings when I pull!

LaurentGoderre:
We should probably change this file instead https://github.com/wet-boew/wet-boew/blob/master/.gitattributes with settings like mentioned here: https://help.github.com/articles/dealing-with-line-endings/

EricDunsworth:
Wouldn't it be more ideal to add ``end_of_line = lf`` to [.editorconfig](https://github.com/wet-boew/wet-boew/blob/master/.editorconfig)? That way, editors will directly save files with LF in the first place, prior to committing. So if you're liniting locally prior to making a commit, there shouldn't be any issues.

shawnthompson:
I was probably seeing this because I have this set in my `.gitconfig` file

```
[core]
	autocrlf = true
	safecrlf = false
```

LaurentGoderre:
@EricDunsworth both would be better. The gitattribute would enforce it at commit and the editorconfig would enforce it when in an editor that supports it.

EricDunsworth:
@LaurentGoderre Yeah :). I meant to suggest it as a supplement to .gitconfig (as opposed to an alternative).

shawnthompson:
Can we not leave it in `.eslintrc` also. This way is people (like me) don't know about `.editorconfig` and who make changes and run `eslint` locally, won't get the linebreak errors. Any CRLF linebreaks will be converted by git at commit anyways.

shawnthompson:
Can someone explain to me why we can't just use default linebreaks on Windows (CRLF) and default linebreaks on Unix/Linux (LF) and let git convert then on commit / checkout or whatever? I just cloned wet-boew on my Windows machine at home and tried to build (`grunt dist`) and all the js files through an `eslint` error for the linebreaks.

nschonni:
#3019 appears to be relevant

Pacoup:
@shawnthompson Because every ASCII-based system in 2017 uses Unix conventions except for Windows. Web development is heavily dominated by Unix and Unix-like systems, so unless your project is Windows-specific, you're gonna run into issues whenever a web tool makes too many assumptions (which is often).

My general advice is to try to make Windows as Unix-y as possible, instead of conceding for the sake of compatibility. It's just simpler this way.

For instance, in my team, we configure Git this way:

vms_ism

We also configure all of our editors on Windows to default to LF.

The idea of using the first option and having a same project work simultaneously with two different encoding formats is pretty crazy imo.

Ideally, you would checkout as-is and commit as-is, because sometimes you may work on Windows-native projects or have files that specifically use CRLF like iCalendar, but everything defaults to CRLF on Windows, so it's often easy to accidentally introduce CRLF into projects this way.

shawnthompson:
@Pacoup, I agree with you on all of that but we can't be sure all developers in the community are using Unix linebreaks. If we want more people to contribute to this project we need to make sure it works on as many systems as possible with as little configuration needed. Not being OS or IDE dependent. If someone wants to use notepad on Windows with nodeJS and git installed we should they should be able too. Just as long as they respect the spacing rule, that's another debate.

Pacoup:
@shawnthompson You can't be sure they all use LF line endings, but you can specify this as a contribution requirement in the documentation. I don't think it's unreasonable to expect some configurations not to be the default.

A [per-repository setting (.gitattributes)](https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings) can alleviate configuration concerns too.

Windows and all IDEs on Windows can handle LF just fine, as well. We're only talking about restricting use of Notepad.

Notepad has significant problems. Besides not being able to handle LF line endings, Notepad also encodes UTF-8 with byte order marks. This can be properly damaging in some cases. For example, any PHP file with a BOM in a request chain of a PHP script generating an Excel spreadsheet will cause a BOM to be added to the Excel archive, causing it to appear corrupted. This is an isolated case, granted, but it makes no sense to force accept BOMs just for Notepad.

EricDunsworth:
Upon further research, I'd support setting ``"linebreak-style": 0`` to .eslintrc.json and keeping .gitattributes/.editorconfig unchanged.

WET's [.gitattributes](https://github.com/wet-boew/wet-boew/blob/master/.gitattributes) file already contains ``* text=auto``, which is the same as setting ``core.autocrlf`` to ``true``. That overrides the current Git installation's ``core.autocrlf`` setting.

Based on my understanding, that means Git internally uses LF in all of this repo's text-based files, but exposes them as CRLF to Windows. So for example, if I were to literally clone this repo from GitHub without making any changes, then tried opening a random .js file in a stone age editor like Notepad, I'd see line breaks correctly. That indicates the file it read in was already using CRLF prior to being opened. If it has been using LF, Notepad would've shown everything on one line.

If I were to edit and commit the aforementioned .js file, Git would internally store my changed file's line endings as LF, but continue exposing them as CRLF to Windows post-commit. If I proceeded to push the commit to GitHub, send a PR and get it merged into WET, the repo's new version of that file would still exclusively use LF characters. It wouldn't inadvertently introduce CRLF characters into the repo.

Currently, if any Windows user attempts to run ``grunt eslint`` locally in their clone of the WET repo, it attempts to lint all source JS files (/src/* and /theme/*). Since eslint's default settings seem to expect LF, it consequently produces thousands of errors. That goes against the OS-agnostic approach WET's .gitattributes file is already setup to use.

IMO, WET's linters should be configured to not care about what kinds of line endings are used in its source files.

The only scenario I can think of where it might be appropriate to force LF would be when building the project, if there's an expectation that /dist/'s compiled files should use identical line endings regardless of which OS is used when compiling. But I don't think LF should be forced onto source files in Windows without a very good reason.

lucas-hay:
Discussed in PR Review meeting, @shawnthompson to make a few changes.

nschonni:
Looks like the is some more advanced whitespace stuff in gitattributes https://git-scm.com/docs/git-config#git-config-corewhitespace, noticed it while looking at https://github.com/composer/composer/blob/master/.gitattributes#L5

shawnthompson:
cc/ @LaurentGoderre @nschonni, as discussed with @LaurentGoderre I created the new config file for travis and the grunt tasks but I couldn't figure out where to call it so it only builds in the CI environment. Have a look and let me know what you think.

Thanks! 👍

Gruntfile.coffee
```diff
@@ -1240,6 +1240,16 @@ module.exports = (grunt) ->
...
+			travis:
+				options:
+					configFile: '.travis.eslintrc.json'
+					quiet: true
+				all:
+					src: [
+						"src/**/*.js"
+						"theme/**/*.js"
+						"tasks/*.js"
+					]
```
LaurentGoderre: Rewrite this whole block as

```Coffee
travis:
	options:
		configFile: '.travis.eslintrc.json'
		quiet: true
	src: "<%= eslint.all.src %>"
```

LaurentGoderre (changes requested):

Gruntfile.coffee
LaurentGoderre: Actually the block above is the first step. The second step is removing the .travis.eslintrc.json and changing the block with:

```coffee
travis:
	options:
		rules: do ->
  			rules = grunt.file.readJSON('.eslintrc.json')
  			rules['linebreak-style'] = 2
  			rules
		quiet: true
	src: "<%= eslint.all.src %>"
```

LaurentGoderre: One final step would be to remove that block altogether and replace the whole task with the following code:

```coffee
eslint:
	options:
		rules: do ->
  			rules = grunt.file.readJSON ".eslintrc.json"
			if process.env.CI
				rules["linebreak-style"] = 2
			rules
		quiet: true
	all:
		src: [
			"src/**/*.js"
			"theme/**/*.js"
			"tasks/*.js"
]
```

LaurentGoderre: @shawnthompson http://eslint.org/docs/user-guide/configuring#configuring-rules

LaurentGoderre: The issue is that the eslint has more than rules so my approach doesn't work. I wonder if you use both an eslintrc and a rules directive if it merges the two

LaurentGoderre: Turns out the rules property doesn't extend what is in the eslintrc.json

LaurentGoderre: @shawnthompson create a new file called `.eslintrc.ci.json` and use this:

```json
{
{
  "extends": ".eslintrc.json",
  "rules": {
    linebreak-style: 2
  }
}
```

LaurentGoderre: And the grunt task:

```js
eslint:
	options:
		configFile: if process.env.CI == "true" then ".eslintrc.ci.json" else ".eslintrc.json"
	all:
		src: [
			"src/**/*.js"
			"theme/**/*.js"
			"tasks/*.js"
		]
```

Gruntfile.coffee
```diff
@@ -1230,10 +1230,10 @@ module.exports = (grunt) ->
 				files: "site/pages/docs/**/*.hbs"
 				tasks: "pages:docs"
 
+
 		eslint:
 			options:
-				configFile: '.eslintrc.json'
-				quiet: true
```
LaurentGoderre: The new warnings you get are because of this that I accidentally knocked out in my example

shawnthompson: Fixed and pushed

LaurentGoderre (approved):

Gruntfile.coffee
```diff
@@ -1230,9 +1230,10 @@ module.exports = (grunt) ->
 				files: "site/pages/docs/**/*.hbs"
 				tasks: "pages:docs"
 
+
```
LaurentGoderre: Can you remove this blank line?

lucas-hay:
@LaurentGoderre are we good to merge now?

PR #7973, opened 2017-04-18
Update HTML Proofer and use without bundler

LaurentGoderre:


LaurentGoderre:
@nschonni this makes it easier to integrate with Docker

.travis.yml
```diff
@@ -8,9 +8,12 @@ env:
   global:
     - TRAVIS_COMMIT_MSG="$(git log --format=%B --no-merges -n 1)"
 
+before_install:
+  - rvm install 2.2.2
```
nschonni: Not sure what this is for, I believe the machines already have RVM installed

script/docker2/update.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+cp Dockerfile.template Dockerfile
+
+echo "\n#\n# Java\n#\n" >> Dockerfile
+curl -sSL --compressed  https://raw.githubusercontent.com/docker-library/openjdk/9a0822673dffd3e5ba66f18a8547aa60faed6d08/8-jre/alpine/Dockerfile | tail -n +8  >> Dockerfile
+
+echo "\n#\n# Ruby\n#\n" >> Dockerfile
+#curl -sSL --compressed  https://raw.githubusercontent.com/docker-library/ruby/bfc7a48724ceb1917ddbcb713b24c835eca584c8/2.4/alpine/Dockerfile | tail -n +3 | head -n -1  >> Dockerfile
```
nschonni: Not sure why this is included and commented out

script/docker2/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM node:6
```
nschonni: These should be separate from this change

nschonni (dismissed):
I think you can do your Bundler-less version for Docker if you want, but I don't think the Travis and core stuff should.
It was setup this way because this was one of the libraries that would regularly start breaking the builds if we left it as a wildcard dependency

script/docker2/update.sh
LaurentGoderre: These files should not have made it in...

.travis.yml
LaurentGoderre: The latest vesrion of HTML Proofer requires 2.2 but 1.9 is installed by default

LaurentGoderre:
The goal is to transition to a Docker build for CI

LaurentGoderre:
@nschonni can we try it for a while?

lucas-hay (approved):

PR #7982, opened 2017-04-25
Remove Data-JSON and generating an unique session id

duboisp:
As discussed during PR review meeting, it was decided to move the Data-JSON into the GCweb theme.

The unique session id is to support a feature used in data-json and data-ajax

Related to PR #7875


src/core/wb.js
```diff
@@ -255,6 +255,16 @@ var getUrlParts = function( url ) {
 			"#999999"
 		],
 
+		// Get and generate a unique session id
```
LaurentGoderre: Why is this being added?

duboisp: For now it support cache busting in data-ajax and in data-json

nschonni: That's something for a separate PR. Cache busting is also better by just using the date in milliseconds

duboisp: There is different cache busting strategy that will be available in data-ajax and data-json. The author is free to use the one that best fit his scenario. 

The cache busting for data-ajax plugin is already in a different PR, see #7983

nschonni: Ok, the session stuff can be discussed in the other PR, but it doesn't belong with the Data-JSON removal

LaurentGoderre: I agree with @nschonni it should be a separate pr

duboisp: That function is required by data-json which it will be added in GCWeb by PR https://github.com/wet-boew/GCWeb/pull/1237

LaurentGoderre: It can still be dealt with separate PRs.

duboisp: @LaurentGoderre no. The [GCweb PR](https://github.com/wet-boew/GCWeb/pull/1237) that re-add the data-json plugin need that function. Also, that function is needed by PR #7983 which add a cache buster for data-ajax.

Let discuss at the next PR review meeting.

/cc @lucas-hay

LaurentGoderre: If it's needed by #7983 needs this why not move it to that PR?

duboisp: Because it's needed by https://github.com/wet-boew/GCWeb/pull/1237 which are independent from #7983

LaurentGoderre: Then create a new PR on which both wet-boew/GCWeb#1237 and #7983 would be dependent

duboisp: That was already done by the unapproved PR #7875

LaurentGoderre: A PR with just this code.

duboisp: What is the issue, exception of how it was packaged? That insertion is already in a separate commit https://github.com/wet-boew/wet-boew/pull/7982/commits/a42b4cd3ef39f345bbeb2138ddf3ca0beddbed42

LaurentGoderre: The change is unrelated to the title of the PR

lucas-hay (approved):

PR #7990, opened 2017-05-03
Theme index: Revised link text to end with page languages.

EricDunsworth:
* Added hidden link text containing " - [Language]" to the end of all page template links. This makes each link's text unique, which can benefit screen reader users that may encounter them out of context.
* Revised the bilingual 404 error and server message page's links to end with " - [Language]/[Language]" (instead of " ([Language]/[Language])") for consistency.

### DIFF
--- a/theme/site/pages/index-en.hbs
+++ b/theme/site/pages/index-en.hbs
@@ -5,7 +5,7 @@
 	"parentdir": "theme-wet-boew",
 	"language": "en",
 	"altLangPrefix": "index",
-	"dateModified": "2014-08-15"
+	"dateModified": "2017-05-03",
 }
 ---
 <section>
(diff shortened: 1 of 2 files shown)


EricDunsworth:
FYI [WET 3.1](http://wet-boew.github.io/wet-boew-legacy/v3.1/demos/theme-wet-boew/index-en.html#en) and [3.0](http://wet-boew.github.io/wet-boew-legacy/v3.0/demos/theme-gcwu-fegc/index-en.html#en) used to do something similar to this, but in a less elegant way. In those, hidden language text was situated at the start of all links (which could've potentially been kind of annoying since the repetitive part was situated at the start of every link). In this PR, the hidden language text is situated at the end of every link.

LaurentGoderre (approved):

PR #7991, opened 2017-05-04
Data Ajax - Conditional loading based on HTTP referrer

bsouster:
Working example

* http://universallabs.org/wet/labs/dt-ajx-httpref/demos/data-ajax/data-ajax-en.html
* http://universallabs.org/wet/labs/dt-ajx-httpref/demos/data-ajax/data-ajax-fr.html


lucas-hay (approved):
Approved at PR review meeting

PR #7995, opened 2017-05-08
add ability to remove borders in dl

shawnthompson:


duboisp:
@shawnthompson can you send also a PR in https://github.com/wet-boew/wet-boew-styleguide repo that  documents that new style.
May be on this page: http://wet-boew.github.io/wet-boew-styleguide/v4/design/lists-en.html#horizontal

shawnthompson:
I was thinking maybe we should add it to the borders details in the style guide.

http://wet-boew.github.io/wet-boew-styleguide/v4/design/borders-en.html

duboisp:
The following is my opinion. we can discuss further at our next PR meeting

As that feature is only specific to a definition list, I recommend to have it primary documented where the style for definition list is explained. So in the [lists-en.html#horizontal](http://wet-boew.github.io/wet-boew-styleguide/v4/design/lists-en.html#horizontal) page which will also contain a working example.

For clarity you can add link in "[border-en.html](http://wet-boew.github.io/wet-boew-styleguide/v4/design/borders-en.html)" to [lists-en.html#horizontal](http://wet-boew.github.io/wet-boew-styleguide/v4/design/lists-en.html#horizontal)

duboisp (changes requested):
Can this "no border" style be defined with the ```dl-horizontal``` here https://github.com/wet-boew/wet-boew/blob/master/src/base/bootstrap-overrides/_base.scss#L64 instead?

src/base/proximity/_base.scss
```diff
@@ -251,12 +251,20 @@
...
+			border: 0 !important;
+		}
+	}
 }
 
 .brdr-rds-0 {
 	border-radius: 0 !important;
 }
 
+
```
duboisp: Remove the addition of an empty line

duboisp (changes requested):

shawnthompson:
@duboisp Tested and it works. 

Thanks!

duboisp (approved):

duboisp:
I tested too, and it works as expected. 

Next thing is to have that improved feature documented in the style guide.

PR #7998, opened 2017-05-11
Footnotes: Fully disabled plugin in wbdisable mode.

EricDunsworth:
Previously, when basic HTML (wbdisable) mode was engaged, the footnotes plugin's init event wouldn't run, but its click events would remain in effect, resulting in the plugin being in a partially-functional state.

This commit fully disables the plugin in that scenario by moving its click events into init.


lucas-hay (approved):

PR #8006, opened 2017-05-16
Calendar - Show today date or the closest on initialization

duboisp:
Alternate solution of PR #7671

So this PR will make the following behavior:
If Event < Today > Event Then by default show Today
If Event < Today Then show max date of Event
If Today > Event Then show min date of Event

That new behavior is documented in the calendar working example, it is the last working example.

close #7671

### DIFF
--- a/src/plugins/cal-events/cal-events-en.hbs
+++ b/src/plugins/cal-events/cal-events-en.hbs
@@ -297,7 +297,7 @@
 
 		<div id="calendar4" class="cal-cnt-fluid"></div>
 
-		<div class="wb-calevt evt-anchor" data-calevt-src="calendar4">
+		<div class="wb-calevt evt-anchor" data-wb-calevt='{"year": 2013, "month": 2}' data-calevt-src="calendar4">
 			<section>
 				<h4>Events List 1</h4>
 				<ol>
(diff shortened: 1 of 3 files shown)


TTVXStarWake (approved):
Thanks Pierre

LaurentGoderre:
@shawnthompson can you create a gh-pages for this?

The code looks clean! Too bad we don't have unit test

duboisp:
@LaurentGoderre 

See here

http://universallabs.org/wet/labs/alt-pr7671/unmin/demos/cal-events/cal-events-en.html
http://universallabs.org/wet/labs/alt-pr7671/unmin/demos/cal-events/cal-events-fr.html

And unit testing: http://universallabs.org/wet/labs/alt-pr7671/unmin/test/test.html

duboisp:
@LaurentGoderre I would submit a new PR with those unit testing may be this week or next week.

LaurentGoderre:
@duboisp these unit test don't apply to the calendar of events though

duboisp:
I know, sorry for the confusion. I will create new test for it

LaurentGoderre:
What I am concerned are the cases not covered here, what happens if there is only event in the current month? or if there is devent in the current month and future or current month and past

duboisp:
I see, but it should be ok. as if event are in the current month only, then minDate and maxDate should be set to be the current month too. So the current month will be show.

I would write specific unit test for those. Although I just made some test locally and it work as expected for those scenario.

LaurentGoderre:
I think this is good to go

MaximMartel:
Thanks Laurent and Pierre !

PR #8016, opened 2017-05-19
Calendar of events - Add unit testing

duboisp:
As discussed in PR#8006

http://universallabs.org/wet/labs/calevt-unit-test/unmin/test/test.html?grep=event%20calendar%20test%20suite


LaurentGoderre:
Can you please align the tests to other plugins? For example, we only test for initalization once, then tests are specific to the case

lucas-hay (approved):

PR #8055, opened 2017-06-12
Menu: Replaced "data-ajax-fetch" with "data-ajax-replace".

EricDunsworth:


duboisp (dismissed):
What is the rational? What is the issue?

By looking at the menu plugin, it is more efficient to use ```data-ajax-fetch``` than going through the data-ajax plugin

See: https://github.com/wet-boew/wet-boew/blob/master/src/plugins/menu/menu.js#L64

EricDunsworth:
@duboisp The issue is that the example in that documentation is outdated.

The comments before the if/else block you linked to indicate that the whole block only exists for backwards compatibility:
https://github.com/wet-boew/wet-boew/blob/master/src/plugins/menu/menu.js#L52

#6768 changed WET's menu implementations (including the menu plugin's demo page) to use ``data-ajax-replace`` instead of ``data-ajax-fetch``. That change has been in place since 4.0.14. ``data-ajax-fetch`` is still supported for backwards compatibility, but it's been over 2 years since the last time it was officially used in WET and its variants.

duboisp:
@EricDunsworth thanks for the clarification, I will dismiss my review.

duboisp (approved):

PR #8076, opened 2017-07-04
[CSS] Fixes to col-auto functionality based on early feedback

thomasgohard:
@masterbee @lucas-hay @pjackson28 has discussed by email.

### DIFF
--- a/src/base/bootstrap-overrides/_base.scss
+++ b/src/base/bootstrap-overrides/_base.scss
@@ -541,40 +541,6 @@ input {
 	}
 }
 
-/**
- *	Disable pull-right/left on grid columns that do not match the current
- *	breakpoint
- */
-@media (max-width: $screen-xs-max) {
-	.pull-left,
-	.pull-right {
-		&[class*=col-sm],
-		&[class*=col-md],
-		&[class*=col-lg] {
-			float: none !important;
-		}
-	}
-}
-
-@media (min-width: $screen-sm-min) and (max-width: $screen-sm-max) {
-	.pull-left,
-	.pull-right {
-		&[class*=col-md],
-		&[class*=col-lg] {
-			float: none !important;
-		}
-	}
-}
-
-@media (min-width: $screen-md-min) and (max-width: $screen-md-max) {
-	.pull-left,
-	.pull-right {
-		&[class*=col-lg] {
-			float: none !important;
-		}
-	}
-}
-
 /*
  *	Blockquote font size
  */
--- a/src/base/views/_screen-lg-min.scss
+++ b/src/base/views/_screen-lg-min.scss
@@ -10,12 +10,3 @@
 	clear: right;
 }
 
-// col-auto extension to Bootstrap grid
-.col-lg-auto {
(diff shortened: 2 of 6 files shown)


lucas-hay (approved):

duboisp:
@thomasgohard what kind of issue were solve by the commit: "CSS: Remove bootstrap override for pull-right/-left on column classes."?

shawnthompson:
This broke the [institution pages](http://wet-boew.github.io/themes-dist/GCWeb/institution-en.html) in GCWeb.

Thanks to @EricDunsworth for figuring it out.

@masterbee @lucas-hay this your fix.

duboisp:
The col-[xs|sm|md|lg|xl]-auto is not the issue.

@shawnthompson the solution will be to fix the template with the correct code. And then, if necessary, an interim CSS hack could be implemented in GC web specific to that template.

EricDunsworth:
I already alluded to this in wet-boew/GCWeb#1268, but think it merits being repeated here.

Very early on in WET 4.0's life, issue #4191 requested changing the behaviour of the ``pull-left``/``pull-right`` classes under certain circumstances. It requested disabling floating on grid containers that are paired alongside ``pull-left``/``pull-right`` when in smaller views than what's specified in the grid classes.

**Example:**
```
<div class="col-md-6 pull-right>
    ...
</div>
```

* Bootstrap 3 would normally float that grid to the right in all views.
* Issue #4191's suggested behaviour would prevent that grid from floating in view breakpoints that are lower than what's specified in ``col-md-6``. In other words, ``pull-right`` would become disabled in small, extra-small and extra-extra-small views.

PR #4206 subsequently implemented #4191's proposed behaviour into WET 4.0. Since those changes were implemented between WET 4.0.0's first and second beta releases, it established a layout behaviour that's been depended upon ever since. By now, there might be hundreds or thousands of WET 4.0 pages out there that were developed and tested with that behaviour in mind. Especially since several GCWeb page templates relied on it.

This PR didn't just improve the ``.col-*-auto`` classes. It undid #4206's changes and consequently introduced problems in all pages that relied on the previous layout behaviour. Since this PR was almost immediately merged in, it didn't get a chance to be publicly reviewed on GitHub or in a WET PR review meeting. So its side effects weren't discovered until after the fact.

In the future, I think a best practice would be to ensure these kinds of PRs don't get immediately merged in if their changes look like they might have large-scale impacts. Particularly when they involve major SCSS changes like selector removals or big revamps of pre-existing selectors. Backwards-compatibility should be taken into consideration if the changes are intended for a maintenance release (i.e. 4.0.x).

Btw in the ``pull-left``/``pull-right`` classes no longer exist in Bootstrap 4. They've been replaced with [float classes](https://getbootstrap.com/docs/4.0/utilities/float/) (which now have view-specific support). So #4206's behaviour can be achieved in Bootstrap 4 without any custom CSS classes. Although that still wouldn't be enough for GCWeb pages with complex layouts (since they use custom direction/view-specific clear classes that have no equivalents in Bootstrap 4).

PR #8090, opened 2017-07-17
Fix dependency loading with readiness issue

duboisp:
For testing this solution:

**wet-boew theme:**
* Working example index: http://universallabs.org/wet/labs/fixdeps-wet-boew/demos/index-en.html 
* Carousel: http://universallabs.org/wet/labs/fixdeps-wet-boew/demos/tabs/tabs-carousel-en.html
* Data Table: http://universallabs.org/wet/labs/fixdeps-wet-boew/demos/tables/tables-en.html

**GCWeb theme:**
* Working example index: http://universallabs.org/wet/labs/fixdeps-gcweb/unmin/demos/index-en.html 
* Carousel: http://universallabs.org/wet/labs/fixdeps-gcweb/unmin/demos/tabs/tabs-carousel-en.html
* Data Table: http://universallabs.org/wet/labs/fixdeps-gcweb/unmin/demos/tables/tables-en.html
* URL Mapping: http://universallabs.org/wet/labs/fixdeps-gcweb/unmin/demos/urlmapping/tblfilter-en.html

//cc @masterbee @MaximMartel @lucas-hay @thomasgohard

### DIFF
--- a/src/core/wb.js
+++ b/src/core/wb.js
@@ -403,6 +403,18 @@ var getUrlParts = function( url ) {
 
 		stripWhitespace: function( str ) {
 			return str.replace( /\s+/g, "" );
+		},
+
+		// Core function to deal with the dependency racing issue
+		whenLibReady: function( testCallback, readyCallback ) {
+			if ( testCallback() ) {
+				readyCallback();
+			} else {
+				setTimeout( function() {
+					wb.whenLibReady( testCallback, readyCallback );
+				}, 50 );
+			}
+
 		}
 	};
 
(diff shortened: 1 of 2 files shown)


lucas-hay (approved):

LaurentGoderre:
:-1: This is not the best way to do it because many plugins have events we can leverage that would require a lot less DOM operation and loop checking. This approach should be the last resort.

lucas-hay:
@LaurentGoderre Do you have a quick fix?
We need this ASAP, so we are trying Pierre's solution, but can revert if/when another solution is developed.

LaurentGoderre:
It can be a temporary fix but it's definitely not optimal

duboisp:
@LaurentGoderre I agree with you this should be a temporary fix. As it was suggested by Mario before, a longer term solution could be to use requirejs instead of yepnope to load our library.

PR #8094, opened 2017-07-20
Update index-eng.html

ajay004:


index-eng.html
```diff
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
+    <meta charset = "UTF-8">
```
duboisp: Can you remove the space beside the "=" character in the attribute
Can you use tabs for indentation.

duboisp (changes requested):
Can you do the French also?

Thanks

shawnthompson:
I created a working example of the PR.

https://servicecanada.github.io/wet-boew-jvanulde-geomap-improvements/demos/geomap/geomap-en.html

Can we test that the change works.

LaurentGoderre:
This change it not really needed since those are just redirects. I wonder if we should remove them altogheter

lucas-hay:
@LaurentGoderre Should we remove altogether or?

LaurentGoderre:
I would remove them personally. We haven't used this file in over 2 years I think

Issue #8063, opened 2017-06-21
Build - Bower is now deprecated


duboisp:
[Bower](https://github.com/bower/bower) is now deprecated but still maintained

> ..psst! While Bower is maintained, we recommend [yarn](https://yarnpkg.com/) and [webpack](https://webpack.js.org/) for new front-end projects!

LaurentGoderre:
That sucks.....I like having a separate dependency manager for the build and for the libraries

nschonni:
I think NPM deps make sense anyway, I'm just not sure some of our older stuff is available as NPM packages or not

LaurentGoderre:
This is a fairly significant change to the build though

sbrioux:
Just to check... currently I cannot add wet-boew-cdn files to my project via npm/yarn?  Or is Webpack an option?

Excerpt from package.json:
```json
    "dependencies": {
        "wet-boew": "wet-boew/wet-boew-cdn#4.0.25"
    }
```

LaurentGoderre:
We do not publish to npm yet. We still rely on Bower.

PR #8101, opened 2017-07-28
mathjax - fix use of CDN and allow for fail-over if there is no connectivity

### LINKED ISSUE #8078: mathjax CDN has been retired; need to migrate to another CDN and must use https to prevent mixed-mode blocking errors
Currently, when rendering a site with MathML on https, the current use of the mathjax CDN has a number of problems.
1. Does not load because of [mixed-mode blocking](https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content/How_to_fix_website_with_mixed_content) as seen here: https://wet-boew.github.io/wet-boew/demos/mathml/mathml-en.html . The offending line is here: https://github.com/wet-boew/wet-boew/blob/master/src/core/wb.js#L531 :
`

					Modernizr.load( {
						load: "http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=Accessible",
						complete: function() {

							// Identify that initialization has completed
							wb.ready( $document, componentName );
						}
					} );`
2. points to an http version
Further, console says: "WARNING: cdn.mathjax.org has been retired. Check https://www.mathjax.org/cdn-shutting-down/ for migration tips."
3. The "accessible" configuration is deprecated as seen here: http://docs.mathjax.org/en/latest/config-files.html#the-accessible-configuration-file-deprecated
4. Latest.js is used instead of a specific version (which is not recommended anymore) as per notes here: https://www.mathjax.org/cdn-shutting-down/

Suggested resolution: switch to a different CDN such as https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js or something else agreed upon by wet-boew maintainers. In principle, I'd prefer http://www.jsdelivr.com/ but I have an [outstanding](https://jsdelivr.uservoice.com/forums/164147-plugins-submission/suggestions/19875577-mathjax-had-have-the-versions-load-properly-throug) [support requests that needs to be resolved](https://jsdelivr.uservoice.com/forums/164147-plugins-submission/suggestions/2967289-mathjax-http-www-mathjax-org) before we can go with it.

nathanv:
This would close #8078 if accepted.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -39,6 +39,7 @@ module.exports = (grunt) ->
 			"sprites"
 			"css"
 			"js"
+			"string-replace"
 		]
 	)
 
@@ -309,6 +310,7 @@ module.exports = (grunt) ->
 		themeDist: "dist/theme-wet-boew"
 		jqueryVersion: @file.readJSON "lib/jquery/bower.json"
 		jqueryOldIEVersion: @file.readJSON "lib/jquery-oldIE/bower.json"
+		MathJaxVersion: @file.readJSON "lib/MathJax/.bower.json"
 		banner: "/*!\n * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)\n * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html\n" +
 				" * v<%= pkg.version %> - " + "<%= grunt.template.today('yyyy-mm-dd') %>\n *\n */"
 		modernizrBanner: "/*! Modernizr (Custom Build) | MIT & BSD */\n"
@@ -1062,6 +1064,20 @@ module.exports = (grunt) ->
 					expand: true
 					flatten: true
 				,
+					cwd: "lib/MathJax"
+					src: [
+						"MathJax.js"
+						"config/**"
+						"docs/**"
+						"extensions/**"
+						"jax/**"
+						"localization/**"
+						"fonts/HTML-CSS/TeX/woff/**"
+						"fonts/HTML-CSS/TeX/otf/**"
+					]
+					dest: "<%= coreDist %>/js/MathJax/"
+					expand: true
+				,
 					cwd: "lib/jquery/dist"
 					src: "*.*"
 					dest: "<%= coreDist %>/js/jquery/<%= jqueryVersion.version %>"
@@ -1219,7 +1235,10 @@ module.exports = (grunt) ->
 				livereload: true
 			js:
 				files: "<%= eslint.all.src %>"
-				tasks: "js"
+				tasks: [
+					  "js"
+					  "string-replace"
+				]
 			css:
 				files: [
 					"src/**/*.scss"
@@ -1329,6 +1348,15 @@ module.exports = (grunt) ->
 				options:
 					testname: "Local Test - <%= grunt.template.today('yyyy-mm-dd hh:MM') %>"
 
+		"string-replace":
+			inline:
+				files:
+					'dist/wet-boew/js/': 'dist/wet-boew/js/*.js'
+				options:
+					replacements: [
+						pattern: 'BOWER_VERSION_MATHJAX'
+						replacement: '<%= MathJaxVersion.version %>'
+	        ]
 
 		"gh-pages":
 			options:
--- a/src/core/wb.js
+++ b/src/core/wb.js
@@ -473,6 +473,14 @@ yepnope.addPrefix( "i18n", function( resourceObj ) {
 	return resourceObj;
 } );
 
+/**
+ * @prefix: mthjx! - adds the root directory of MathJax resources
+ */
+yepnope.addPrefix( "mthjx", function( resourceObj ) {
+	resourceObj.url = paths.js + "/MathJax/" + resourceObj.url;
+	return resourceObj;
+} );
+
 /*-----------------------------
  * Deps loading, call "complete" callback when the deps is ready if a testReady is defined
  *-----------------------------*/
@@ -561,17 +569,18 @@ Modernizr.load( [
 
 					// Load the MathML dependency. Since the polyfill is only loaded
 					// when !Modernizr.mathml, we can skip the test here.
-					Modernizr.load( {
-						load: "http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=Accessible",
-						testReady: function() {
-							return ( window.MathJax && window.MathJax.isReady );
-						},
+					Modernizr.load( [ {
+						load: "timeout=500!https://cdn.jsdelivr.net/npm/mathjax@BOWER_VERSION_MATHJAX/MathJax.js?config=Accessible",
 						complete: function() {
+							Modernizr.load( [ {
+								test: window.MathJax === undefined,
+								yep: "mthjx!MathJax.js?config=Accessible"
+							} ] );
 
 							// Identify that initialization has completed
 							wb.ready( $document, componentName );
 						}
-					} );
+					} ] );
 				} );
 
 				wb.add( selector );
--- a/bower.json
+++ b/bower.json
@@ -34,6 +34,7 @@
     "jquery-pjax": "1.9.6",
     "jquery-validation": "1.15.1",
     "magnific-popup": "https://github.com/wet-boew/Magnific-Popup.git#1.0.0+keyboard_trap_fix",
+    "components/MathJax": "2.7.1",
     "openlayers": "https://github.com/wet-boew/openlayers-dist.git#2.13.1",
     "proj4": "2.3.3",
     "respond": "1.4.2",
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "grunt-sass-lint": "^0.2.0",
     "grunt-saucelabs": "^9.0.0",
     "grunt-sri": "^0.2.0",
+    "grunt-string-replace": "^1.3.1",
     "grunt-wet-boew-postbuild": "^0.1.3",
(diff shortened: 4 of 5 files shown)


Gruntfile.coffee
```diff
@@ -309,6 +310,7 @@ module.exports = (grunt) ->
 		themeDist: "dist/theme-wet-boew"
 		jqueryVersion: @file.readJSON "lib/jquery/bower.json"
 		jqueryOldIEVersion: @file.readJSON "lib/jquery-oldIE/bower.json"
+		MathJaxVersion: @file.readJSON "lib/MathJax/.bower.json"
```
nschonni: Not sure there is a leading `.` on the `bower.json`

Gruntfile.coffee
```diff
@@ -39,6 +39,7 @@ module.exports = (grunt) ->
 			"sprites"
 			"css"
 			"js"
+			"string-replace"
```
nschonni: maybe add a particular target so it's clearer what is being replaced. EX: `string-replace:MathJaxVersion`

Gruntfile.coffee
```diff
@@ -1329,6 +1345,15 @@ module.exports = (grunt) ->
 				options:
 					testname: "Local Test - <%= grunt.template.today('yyyy-mm-dd hh:MM') %>"
 
+		"string-replace":
+			inline:
+				files:
+					'dist/wet-boew/js/': 'dist/wet-boew/js/*.js'
+				options:
+					replacements: [
```
nschonni: I noticed there is a few query string params on the CDN, should those get exposed here too?

nschonni (commented):
Think this makes sense, but it might be better to use NPM because of #8063

Gruntfile.coffee
nathanv: Great idea. I'll fix that.

Gruntfile.coffee
nathanv: Amazingly there is...

Gruntfile.coffee
nathanv: Hmm good question. That would be important if the default parameters are changed over time as they have been. Good idea. I'll make that fix.

Gruntfile.coffee
nathanv: @nschonni so turns out that string-replace doesn't have multi-task support so I'm going to try to use https://github.com/yoniholmes/grunt-text-replace instead. I've been juggling a few things but I'll try to finish it up asap.

nschonni: Maybe better to stick with `string-replace` even if we can't add the target. Looking at the project it looks like it was more recently maintained and updated for Grunt 1.0 where the other wasn't

Gruntfile.coffee
lucas-hay: @nathanv we are scheduled to release WET next Tuesday (August 15th), any chance this PR will be updated/ready by then?

lucas-hay:
@nathanv any news?

Gruntfile.coffee
nathanv: Hello @lucas-hay . We found a quick workaround internally to get the issue sorted and I had to move onto other stuff. I am working on it right now to get it to a better point hopefully good enough to ship.

Gruntfile.coffee
nathanv: @nschonni okay let's stick with it then

Gruntfile.coffee
lucas-hay: Thanks @nathanv 
Reply here if you get this ready before tomorrow and we can look at merging it.

site/includes/footerresources.hbs
```diff
@@ -1,5 +1,6 @@
 <!--[if gte IE 9 | !IE ]><!-->
 <script src="{{#if offline}}{{assets}}/../wet-boew/js{{else}}https://ajax.googleapis.com/ajax/libs{{/if}}/jquery/{{environment.jqueryVersion}}/jquery.js"></script>
+<script>window.jQuery || document.write('<script src="{{assets}}/../wet-boew/js/jquery/{{environment.jqueryVersion}}/jquery.js"><\/script>');</script>
```
nschonni: Think the previous line covers this off

Gruntfile.coffee
```diff
@@ -1329,6 +1348,15 @@ module.exports = (grunt) ->
 				options:
 					testname: "Local Test - <%= grunt.template.today('yyyy-mm-dd hh:MM') %>"
 
+		"string-replace":
+			inline:
+				files:
+					'dist/wet-boew/js/': 'dist/wet-boew/js/*.js'
```
nschonni: Could target the `concat.core.dest` and `concat.coreIE8.dest` values to glob through less files.

nathanv (commented):
Ready for final review.

Gruntfile.coffee
nathanv: Ready for final review.

site/includes/footerresources.hbs
lucas-hay: @nschonni are you ok with this now?
I will be starting the release at 9:00 and wondering if this should be merged?
@LaurentGoderre @duboisp any comments?

LaurentGoderre: Yeah this line is not very usefull

nathanv:
Without that line, I wasn't able to get Jquery to load in an intranet
environment.

site/includes/footerresources.hbs
nathanv: Without the line I wasn't able to get Jquery to load when blocking Google's Jquery url using Chrome.

LaurentGoderre:
@nathanv these files are autogenerated from a site builder. There is an option for local build to not use the CDN and use a local copy. You may use your line in your own setup or change the paths to local versions but it shouldn't be included in WET.

site/includes/footerresources.hbs
nathanv: Had a quick offline discussion with @laurentgoderre. I will remove the line when I have access to my laptop. Currently out of the office unexpectedly but I'll be able to get something sorted before end of business day.

nathanv:
I removed the one line as was requested by @LaurentGoderre

Gruntfile.coffee
nathanv: @lucas-hay I removed the line as requested by @LaurentGoderre :) so ready for final-final review.

nathanv:
@lucas-hay anything I can do to get this in the next release?

lucas-hay:
@nathanv as long as @LaurentGoderre and @nschonni approve the updates you made as they requested, I will merge it as soon as they approve.

Gruntfile.coffee
nathanv: @nschonni is this the last thing I need to resolve?

site/includes/footerresources.hbs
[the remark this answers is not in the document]
nathanv: @LaurentGoderre I removed the offending line here... is there anything else needed?

lucas-hay:
@LaurentGoderre @nschonni 
Do you guys approved now that the changes you had requested are implemented?

src/core/wb.js
```diff
@@ -561,17 +569,18 @@ Modernizr.load( [
...
+					Modernizr.load( [ {
+						load: "timeout=500!https://cdn.jsdelivr.net/npm/mathjax@BOWER_VERSION_MATHJAX/MathJax.js?config=Accessible",
 						complete: function() {
+							Modernizr.load( [ {
+								test: window.MathJax === undefined,
+								yep: "mthjx!MathJax.js?config=Accessible"
+							} ] );
 
 							// Identify that initialization has completed
 							wb.ready( $document, componentName );
```
LaurentGoderre: This is still a potential race condition since you are not calling this line from the backup

LaurentGoderre (dismissed):

src/core/wb.js
lucas-hay: @nathanv Any news on Laurent's review?

duboisp: @nathanv To avoid the race condition you can run a test with ```testReady``` like:
```js
Modernizr.load( {
	load: "http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=Accessible",
	testReady: function() {
		return ( window.MathJax && window.MathJax.isReady );
	},
	complete: function() {

		// Identify that initialization has completed
		wb.ready( $document, componentName );
	}
} );
```
 https://github.com/wet-boew/wet-boew/blob/master/src/core/wb.js#L566

nathanv:
Thanks for the feedback. I will fix it this weekend. :).

lucas-hay:
Please let us know when fixed so we can merge ASAP so this can be included
in the 4.0.27 release on November 30th.

Gruntfile.coffee
```diff
@@ -1061,6 +1063,20 @@ module.exports = (grunt) ->
 						return dest + "/" + src.replace ".debug", ""
 					expand: true
 					flatten: true
+				,
+					cwd: "lib/MathJax"
+					src: [
+						"MathJax.js"
+						"config/**"
+						"docs/**"
```
Empereol: Should the docs be included? When `bootlint` gets run from the `Gruntfile.coffee`, it fails on the HTML files copied over into this directory. I suspect these files aren't even required and could be removed to easily resolve this issue.

duboisp: @Empereol Please submit a [new issue](https://github.com/wet-boew/wet-boew/issues/new) with your concern. That way it is easier to track and ensure that you will get a follow up.

Empereol: Can do. Issue #8250

PR #8107, opened 2017-08-04
Circular meter - Initial commit

duboisp:
Plugin to render circular meter

Working example:
* English - http://universallabs.org/wet/labs/circular-meter/unmin/demos/meter-circular/circular-en.html
* French - http://universallabs.org/wet/labs/circular-meter/unmin/demos/meter-circular/circular-fr.html
* Docs - http://universallabs.org/wet/labs/circular-meter/unmin/docs/ref/meter-circular/meter-circular-en.html

/cc @lucas-hay  @MaximMartel @masterbee

### DIFF
--- a/.csslintrc
+++ b/.csslintrc
@@ -20,6 +20,7 @@
     "qualified-headings": false,
     "regex-selectors": false,
     "selector-max-approaching": false,
+	"selector-max": false,
     // Some Bootstrap mixins end up listing all the longhand properties
     "shorthand": false,
     "text-indent": false,
--- a/site/pages/docs/ref/meter-circular/meter-circular-en.hbs
+++ b/site/pages/docs/ref/meter-circular/meter-circular-en.hbs
@@ -1,0 +1,124 @@
+---
+{
+	"title": "Meter circular",
+	"language": "en",
+	"category": "Plugins",
+	"categoryfile": "plugins",
+	"description": "Render a meter into a circular meter styled in CSS.",
+	"altLangPrefix": "meter-circular",
+	"dateModified": "2017-08-03"
+}
+---
+<div class="wb-prettify all-pre hide"></div>
+
+<p>Render a meter into a circular meter styled in CSS.</p>
+
+<h2>Purpose</h2>
+<p>Circular meter are not supported yet by browser and seems to not be explicitly defined in the <a href="https://www.w3.org/TR/html52/rendering.html#the-meter-element-rendering">HTML5.2 spec</a> yet. However, the HTML5 spec provide some indication on how to <a href="https://www.w3.org/TR/html5/rendering.html#the-progress-element-0">render a circular progress ring</a> when the progress element is square.</p>
+
+<h2>Working example</h2>
+<ul>
+	<li><a href="../../../demos/meter-circular/circular-en.html">English examples</a></li>
+	<li><a href="../../../demos/meter-circular/circular-fr.html">French examples</a></li>
+</ul>
+
+
+<h2>How to implement</h2>
+<p>Add the class <code>wb-meter-circular</code> to a meter element</p>
+
+<pre><code>&lt;meter class=&quot;wb-meter-circular&quot; value=&quot;10&quot; max=&quot;100&quot;&gt;10 percent&lt;/meter&gt;</code></pre>
+
+<p>If the meter get updated later on through the jQuery function <code>.val()</code>, this circular meter would be immediately refreshed to represent the latest value. Changing the attribute <code>min, max, low, high, optimum</code> would not initiate a refresh, however it will be taken in consideration when a new value is set to the meter element through jQuery.</p>
+
+
+<h3>Showing the value</h3>
+<p>By default the meter value will be displayed in the center of the circular meter.<p>
+<p>Potential issue: A value number with 4 digit or more may not fully pass the <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html">WCAG success criteria 1.4.4</a> (Level AA).</p>
+
+<h3>Add a short descriptive label</h3>
+<p>A short label can be added underneath the actual meter value. What you need to do is to add an <code>aria-label</code> to the meter element.</p>
+<p>Potential issue: A long label may not fully pass the <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html">WCAG success criteria 1.4.4</a> (Level AA).</p>
+
+
+<h3>Modify the appearance</h3>
+
+<p>Built-in personalization for the circular meter is done by applying CSS class at the parent container of the meter element.</p>
+
+<dl>
+	<dt><code>mtr-grayscale</code></dt>
+	<dd>Will render the meter in gray scale</dd>
+
+	<dt><code>mtr-larger</code></dt>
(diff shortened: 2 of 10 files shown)


src/plugins/meter-circular/_base.scss
```diff
@@ -0,0 +1,463 @@
...
+		position: absolute;
+		top: 0;
+		width: 100%;
+	}
+
+	.mtr-l {
+		left: 0;
+
+		.mtr-b {
+			-webkit-transform-origin: center left;
```
nschonni: Autoprefixer should be doing this already

src/plugins/meter-circular/_base.scss
```diff
@@ -0,0 +1,463 @@
...
+		}
+	}
+
+	/* Color */
+	&.mtr-optimal {
+		.mtr-b {
+			border-color: #236423;
+		}
+	}
+	&.mtr-suboptimal {
```
nschonni: These are some odd names for out of the box values

src/plugins/meter-circular/_base.scss
```diff
@@ -0,0 +1,463 @@
...
+	}
+	&.mtr-lessgood {
+		.mtr-b {
+			border-color: #d3080c;
+		}
+	}
+
+
+	/* Value */
+	&.r3m {
```
nschonni: Are these production values or demos?

src/plugins/meter-circular/_base.scss
```diff
@@ -0,0 +1,463 @@
...
+			}
+		}
+	}
+}
+
+/* Animation based on clock minutes */
+
+@keyframes ldg-3m{
+	0%{
+		-webkit-transform: rotate(0deg);
```
nschonni: More autoprefixer stuff

src/plugins/meter-circular/_base.scss
```diff
@@ -0,0 +1,463 @@
...
+		transform: rotate(0deg);
+	}
+	100%{
+		-webkit-transform: rotate(180deg);
+		transform: rotate(180deg);
+	}
+}
+
+@media only screen and (max-width: 990px){
```
nschonni: Use the variables for the screen sizes

src/plugins/meter-circular/circular.js
```diff
@@ -0,0 +1,157 @@
...
+"use strict";
+
+var componentName = "wb-meter-circular",
+	selector = "meter." + componentName,
+	initEvent = "wb-init" + "." + componentName,
+	$document = wb.doc,
+	meterPrefix = "mtr-",
+	selectorValue = "." + meterPrefix + "val",
+	selectorLabel = "." + meterPrefix + "lbl",
+	clsOptimal = meterPrefix + "optimal",
```
nschonni: I'm guessing `cls` is a short form for `class`, but it doesn't make sense to short stuff that the minifier will correct

.csslintrc
```diff
@@ -20,6 +20,7 @@
     "qualified-headings": false,
     "regex-selectors": false,
     "selector-max-approaching": false,
+	"selector-max": false,
```
nschonni: Spacing, but this also means that the compiled CSS has exceeded 32K selectors I believe

site/pages/docs/ref/meter-circular/meter-circular-en.hbs
```diff
@@ -0,0 +1,124 @@
...
+<p>Add the class <code>wb-meter-circular</code> to a meter element</p>
+
+<pre><code>&lt;meter class=&quot;wb-meter-circular&quot; value=&quot;10&quot; max=&quot;100&quot;&gt;10 percent&lt;/meter&gt;</code></pre>
+
+<p>If the meter get updated later on through the jQuery function <code>.val()</code>, this circular meter would be immediately refreshed to represent the latest value. Changing the attribute <code>min, max, low, high, optimum</code> would not initiate a refresh, however it will be taken in consideration when a new value is set to the meter element through jQuery.</p>
+
+
+<h3>Showing the value</h3>
+<p>By default the meter value will be displayed in the center of the circular meter.<p>
+<p>Potential issue: A value number with 4 digit or more may not fully pass the <a href="https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html">WCAG success criteria 1.4.4</a> (Level AA).</p>
```
nschonni: If it fails AA, then the plug-in should reject it

nschonni (changes requested):
Should probably honor the `prefers-reduced-motion` unless they're actually important https://css-tricks.com/introduction-reduced-motion-media-query/
https://www.w3.org/WAI/GL/wiki/Animation_caused_by_user_interaction

LaurentGoderre (changes requested):
I'm strongly against using display none and trying to recreate the element and relationships with ARIA. A plugin like this should rely on the native element.

src/plugins/meter-circular/_base.scss
duboisp: ok, I will fix it

src/plugins/meter-circular/_base.scss
duboisp: Regarding optimal, sub-optimal and less good are the keyword used to define the meter region in the HTML5 spec. Or do you talk about "mtr" prefix ?

src/plugins/meter-circular/_base.scss
duboisp: ok, I will fix

src/plugins/meter-circular/circular.js
duboisp: Yes, I will fix

.csslintrc
duboisp: Ok will fix
Yes the compiled CSS has exceeded the 4095 selector. After reading, that should only causing an issue for IE9 and older version of IE.

Also, that means that some clean up in the CSS for all features would need to completed in a near future.

site/pages/docs/ref/meter-circular/meter-circular-en.hbs
duboisp: We will do some testing in order to determine what are the limits and look at enforcing it in the plugin itself. Just an idea: If the value can not be properly displayed on the circular meter, then it could fallback on the default meter layout.

duboisp:
The circular meter is fully controlled by the native meter element.
I would like to overwrite the Shadow DOM on the meter element but I can't as the spec is still in Draft and mostly because the Shadow DOM spec is not widely implemented.

Regarding the use of ARIA
What do you think if instead I add the ```wb-inv``` at the meter element and then add the role presentation to the circular meter?

Regarding the animation
That will be something I will need to discuss with my client. But I do understand and agree that we should honor ```prefers-reduced-motion```.

An issue that I found will be to replace ```div``` for ```span``` as the meter elements is expected in the phrasing context and to use the title attribute for labeling instead of an aria-label.

In regards of future improvement, may be the circular meter could work better in SVG or in canvas compared to this CSS solution. I did my best with the time constraint I have.

LaurentGoderre:
@duboisp you should disable the default rendering of the meter using the techniques here (https://stackoverflow.com/questions/8094835/how-to-style-html5-meter-tag) and add the circular markup as nested content of the meter.

Garneauma:
@duboisp This has not been touched in 9 years. Is it still relevant? Should we close it?

Garneauma:
@duboisp As discussed, closing as there is no longer a need for it.

PR #8110, opened 2017-08-07
Core: Ensure getUrlParts() always returns slash-prefixed pathnames.

### LINKED ISSUE #8109: Multimedia player: IE 10-11 and older versions of Edge can't playback youtu.be videos.
The extra logic #7835 added to support short YouTube URLs assumes that the pathname property will always consist of a string that begins with a forward slash (/). If a long YouTube URL isn't used, it attempts to treat the URL as a youtu.be one. A valid youtu.be URL's pathname is presumed to always be formatted as "/ID". The extra logic was designed to always strip out the first character, then treat the result as a potential video ID.

It turns out that the extra logic doesn't work properly in any versions of Internet Explorer, nor early versions of Edge. Due to a [browser bug](https://connect.microsoft.com/IE/feedbackdetail/view/1002846/pathname-incorrect-for-out-of-document-elements), pathname's opening slash character is always omitted on generated ``a`` elements. This is currently resulting in the first character of all youtu.be video IDs getting stripped out prior to being retrieved from YouTube.

The end result is that the multimedia player plugin is currently incapable of playing back any youtu.be videos in the aforementioned browsers, since it's requesting invalid video IDs from YouTube.

I'll send a PR very soon to fix it. This issue only exists to give extra context to what my forthcoming PR is trying to resolve.

EricDunsworth:
According to the W3C's HTML specification (https://w3c.github.io/html/browsers.html#dom-location-pathname), the pathname attribute must always start with a forward slash character. Virtually every browser does this. But due to a browser bug (https://connect.microsoft.com/IE/feedbackdetail/view/1002846/pathname-incorrect-for-out-of-document-elements), all versions of Internet Explorer and early versions of Edge omit the forward slash from pathnames retrieved from generated a elements that weren't added to the body.

Since getUrlParts() works by generating an a element that never gets added to the body, any pathname attributes returned by it are affected.

This commit fixes it by inserting a forward slash at the beginning of any pathnames that lack one. It borrows a regex from the NavCurrent plugin that's already in use for the same purpose.

Fixes #8109 and doesn't interfere with the tabbed interface plugin (i.e. the only other WET/GCWeb plugin that uses pathnames extracted from getUrlParts()).

### DIFF
--- a/src/core/wb.js
+++ b/src/core/wb.js
@@ -24,7 +24,7 @@ var getUrlParts = function( url ) {
 			host: a.host,
 			hostname: a.hostname,
 			port: a.port,
-			pathname: a.pathname,
+			pathname: a.pathname.replace( /^([^\/])/, "/$1" ), // Prefix pathname with a slash in browsers that don't natively do it (i.e. all versions of IE and possibly early versions of Edge). See pull request #8110.
 			protocol: a.protocol,
 			hash: a.hash,
 			search: a.search,


src/core/wb.js
```diff
@@ -24,7 +24,7 @@ var getUrlParts = function( url ) {
 			host: a.host,
 			hostname: a.hostname,
 			port: a.port,
-			pathname: a.pathname,
+			pathname: a.pathname.replace( /^([^\/])/, "/$1" ),
```
duboisp: Just add a comment that specify this fix is for IE and Edge specific

duboisp (changes requested):

EricDunsworth:
@duboisp I've added a (sorta long) line comment with a basic overview of what replace() is doing and a reference to this PR.

duboisp (approved):

PR #8133, opened 2017-09-08
EditorConfig: Fixed typos in package.json/travis.yml's section name.

EricDunsworth:
The original section name for those files was written as [{...},{...}], which was invalid and prevented its coding styles from taking effect.

According to EditorConfig's documentation, the correct writing scheme for that kind of section name is [{...,...}]. This commit changes it to that.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,6 +10,6 @@ insert_final_newline = true
 indent_style = tabs
 indent_size = 4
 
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
Interesting!!

nschonni (approved):
Not sure if it changed or was always broken, but I see this style on their home page now

Issue #7752, opened 2016-11-10
Plugin: Equalized (Proposal to remove...)


kodecount:
I wish to propose to remove this plugin from the tab interface. It has caused many issues: #7171, #7385, #7522, #7591.

Also found a new issue when the screen is resized sometimes the height is not adjusted right:
![untitled](https://cloud.githubusercontent.com/assets/22446865/20179279/b341ef56-a72b-11e6-8979-b82014b946c0.png)

duboisp:
Can the equalize plugin be updated to fix all those issue? 

May be equalize plugin simply needs to be improved about how it could better interact with all those plugin that seems conflicting with. A solution, that worth exploring, would be to listening and triggering new events that would fix the issue in a timely matter.

Cheers

kodecount:
@duboisp It is possible, but well require a huge overhaul of the plugin.

kodecount:
Some overhauls too look at:

-How it determines screen size and in certain screen size remove and prevent changes like `min-height`.
-How to determine the current the content inside of the plugin and is the equalizer required with that content?
-What in the equalizer is causing script delays which causes overlaping or flickering.

kodecount:
Has their been any push on fixing the Equalize plugin?

juliemercier:
Any idea when this may be implemented?

kodecount:
@juliemercier I'm not working with the WET group, but as far as I'm aware their looking into, but no work has been started.

PR #8134, opened 2017-09-08
Tabbed Interface: Removal of equal height js on tabs stops flickering during redraw in browser.

### LINKED ISSUE #7950: Tabbed Interface showing/flashing contents from all the tabs on load
Hello,

**Scenario:** Tabbed Interface with a lot of content such as Tables
**WET:** 4.0.23

When you have a Tabbed Interface with tabs with a lot of contents, it will show all contents inside the tabs as it loads. It cycles to all tabs showing all contents in a "flashing" manner.

It should only display the contents when its done loading.

Screenshot during the flashing of all the contents as it loads.
image

Thanks,
John

mercury64:
Specifically, removed 2 variables: equalHeightClass and equalHeightOffClass, as well as removed the .toggleClass( equalHeight... code.

Fixes #7950

### DIFF
--- a/src/plugins/tabs/tabs.js
+++ b/src/plugins/tabs/tabs.js
@@ -22,8 +22,6 @@ var componentName = "wb-tabs",
 	setFocusEvent = "setfocus.wb",
 	controls = selector + " ul[role=tablist] a, " + selector + " ul[role=tablist] .tab-count",
 	initialized = false,
-	equalHeightClass = "wb-eqht",
-	equalHeightOffClass = equalHeightClass + "-off",
 	tabsAccordionClass = "tabs-acc",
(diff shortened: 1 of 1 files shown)


kodecount:
There was an argument to not remove this by @duboisp on https://github.com/wet-boew/wet-boew/issues/7752#issuecomment-259828855.

duboisp:
May be we can add a new configruation option that will prevent to run the "wb-eqht"

So I would assume a tabs that do not want to have the equalize plugin initiatlized will be configured like:

<div class="wb-tabs" data-wb-tabs='{"noeqht": true}'>

Also, this will need to be documented in the tabs documentation page.

Cheers

nschonni (approved):

lucas-hay (approved):

PR #8141, opened 2017-09-12
Base: add color variable in top instead of using color code

nongpp:
add color variable in top instead of using color code

### DIFF
--- a/theme/banner/_base.scss
+++ b/theme/banner/_base.scss
@@ -1,3 +1,14 @@
+/*colour 
+(the lowest number is very dark in the same colour's name 
+Ex darkBlue-1:#114f7a and darkBlue-2:#176ca7)*/
+
+$darkBlue-1:     #114f7a;
+$darkBlue-2:     #176ca7;
+$veryDarkGray-1:   #333;
+$veryDarkGray-2:   #474747;
+$white:          #fff;
+$darkGray:       #777;
+
 /*
  Banner
  */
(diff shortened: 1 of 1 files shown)


lucas-hay:
@nongpp We will be happy to merge this, but could you please update your commit message. (The description you have in the commit description would suffice). see below:
"Base: add color variable in top instead of using color code"

LaurentGoderre:
I think variable should be descriptive of their purpose and not the color themselves

EricDunsworth:
In addition to what @LaurentGoderre said, I'd recommend making the following changes:

1. Move the SCSS variable declarations into [wet-boew/theme/_variables.scss](https://github.com/wet-boew/wet-boew/blob/master/theme/_variables.scss).
2. Refine the comments before the variable declarations to be more in line with what's already done in [wet-boew/theme/_variables.scss](https://github.com/wet-boew/wet-boew/blob/master/theme/_variables.scss).
3. Revise the commit's description to be shorter/more meaningful and prefix it with the commit's topic. I'd recommend going from this:

> Update _base.scss
>
> add color variable in top instead of using color code

To something along these lines:

> Theme: add color variable in top instead of using color code

nongpp:
@lucas-hay Thank you for your respond. And is it ok if I correct all css like this or I set every variable into  wet-boew/theme/_variables.scss.?

lucas-hay:
@nongpp If you wish to add more varaibles, that would be great. Just follow what @EricDunsworth and @LaurentGoderre have mentioned. (use _variables.scss, more descriptive name, comments above, etc.)

nongpp:
@lucas-hay @LaurentGoderre @EricDunsworth thank you for your advice. I will follow your advice.

lucas-hay:
@nongpp should we close this PR? 
It appears you moved this into #8199

PR #8146, opened 2017-09-15
build: Bump grunt-sass@2

nschonni:
Allows prebuilt binaries on Node 8.
Comparing the diff of the built diffs there are a handful of minor precision differences in the built files.

Left in the package-lock.json to help with repeatable builds, but it could be excluded

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,9851 @@
+{
+  "name": "wet-boew",
+  "version": "4.0.27-development",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abbrev": {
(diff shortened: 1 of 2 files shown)


LaurentGoderre:
We're not building on Node 8 yet but :+1:

LaurentGoderre:
@nschonni did you don a diff of the built results to  see what difference there is between the two?

nschonni:
@LaurentGoderre yes, from the overall comment, there was only a few precision differences with a few colour values in the final built CSS.

LaurentGoderre:
Excellent!

LaurentGoderre (approved):

PR #8177, opened 2017-10-19
Core/Theme: Added nav elements to ToCs.

EricDunsworth:


site/pages/docs/versions/dwnld-en.hbs
```diff
@@ -9,7 +9,7 @@
 ---
 <p>See <a href="version-info-en.html">Versioning</a> for the meaning of each version number.</p>
 
-<section>
+<nav role="navigation">
```
duboisp: Is it required to duplicate the role? By definition the [```<nav>``` element](https://www.w3.org/TR/html52/sections.html#ref-for-represent-13) imply the aria role navigation.

Idem for all other instance.

duboisp (dismissed):

site/pages/docs/versions/dwnld-en.hbs
EricDunsworth: @duboisp Since the ``nav`` element's implicit ARIA semantics resolve to ``role="navigation"``, it isn't required. In fact, the [W3C's ARIA in HTML specification recommends against it](https://www.w3.org/TR/html-aria/#nav). It even triggers validation warnings in the W3C's NU HTML validator.

I personally don't agree with placing redundant role attributes on elements that already have the same implicit ARIA semantics. The only reasons I did it in this and its related PRs was because it's already being done throughout WET's page templates and because of the outcome of #6993's discussion. Since 2 years have now passed since then, maybe it's time to revisit whether WET should finally stop using redundant ``role`` attributes. But I didn't feel that it'd be appropriate to have that discussion as a part of this or its related PRs.

lucas-hay (approved):

PR #8178, opened 2017-10-20
Fix date picker closing prematurely

cmeka:
In some cases the focus is lost when disabling the previous and next buttons, this causes the [date picker polyfill](https://github.com/wet-boew/wet-boew/blob/69e2931ff562d253b30a8f84caf4a9b8a5c7061a/src/polyfills/datepicker/datepicker.js#L261) to close prematurely.

I was able to reproduce this bug using the [date picker example page](http://wet-boew.github.io/v4.0-ci/demos/datepicker/datepicker-en.html) on several GoC work computers running IE 11.0.9600.18816 and Windows 7 SP1, however I was unable to reproduce it on my home computer.


cmeka:
Leave a comment

PR #8188, opened 2017-10-31
Fix date picker closing prematurely in some cases

cmeka:
On some systems the focus is lost when disabling the previous and next buttons, this causes the [date picker polyfill](https://github.com/wet-boew/wet-boew/blob/69e2931ff562d253b30a8f84caf4a9b8a5c7061a/src/polyfills/datepicker/datepicker.js#L261) to close prematurely.

I was able to reproduce this bug using the [date picker example page](http://wet-boew.github.io/v4.0-ci/demos/datepicker/datepicker-en.html) on several GoC systems running IE 11.0.9600.18816 and Windows 7 SP1, however I was unable to reproduce it at my home.

### DIFF
--- a/src/core/wb.js
+++ b/src/core/wb.js
@@ -233,6 +233,7 @@ var getUrlParts = function( url ) {
 		ielt8: ( oldie < 8 ),
 		ielt9: ( oldie < 9 ),
 		ielt10: ( oldie < 10 ),
+		ie11: ( !!navigator.userAgent.match( /Trident\/7\./ ) ),
 
 		selectors: [],
 
--- a/src/plugins/calendar/calendar.js
+++ b/src/plugins/calendar/calendar.js
@@ -321,7 +321,10 @@ $document.on( "click vclick touchstart", ".cal-month-prev, .cal-month-next", fun
 		type: navigateEvent,
 		year: date.getFullYear(),
 		month: date.getMonth()
-	} ).trigger( "focusin" );
+	} );
+	if ( wb.ie11 ) {
+		$calendar.trigger( "focusin" );
+	}
 } );
 
 $document.on( "keydown", selector, function( event ) {
--- a/src/plugins/calendar/calendar.js
+++ b/src/plugins/calendar/calendar.js
@@ -321,7 +321,7 @@ $document.on( "click vclick touchstart", ".cal-month-prev, .cal-month-next", fun
 		type: navigateEvent,
 		year: date.getFullYear(),
 		month: date.getMonth()
-	} );
+	} ).trigger( "focusin" );
 } );
 
 $document.on( "keydown", selector, function( event ) {


duboisp:
Can you update you PR and ensure that your fix is only applied for IE 11? I worry that your current request of change created an unwanted effect on the other browser.

lucas-hay:
@cmeka Can you make the change Pierre Requested before end of week?
We can include this in the 4.0.27 next week if so.

EricDunsworth:
FWIW, here's how to reproduce the issue this PR resolves:

**Steps:**
1. Open the [date picker polyfill's demo page](http://wet-boew.github.io/v4.0-ci/demos/datepicker/datepicker-en.html) in IE11.
2. Click the calendar button beside the "Start date (YYYY-MM-DD)" field.
3. Press your keyboard's Tab key. The left (previous) arrow button will gain focus.
4. Press Enter. The date picker widget will switch to April 2013.
5. Press Tab. The right (next) arrow button will gain focus.
6. Press Enter.

**Result without this PR's fix:**
The date picker widget will close.

**Result with this PR's fix:**
The date picker widget will switch to May 2013.

@duboisp Any chance you could try out those steps in Edge? I'm curious if it has the same problem as IE11.

Bte it's likely that IE10 and older might be affected by the same issue. So there might be value in modifying the condition to target any version of IE. Might need to modify it further or possibly even remove it if it turns out Edge is also affected.

shawnthompson:
It doesn't do it in my version of IE which is 11.0.9600.18738 but I can't seem to find anyone else at work using the same version.

EricDunsworth:
FYI I'm currently on IE11 11.0.960.18837 (different from @shawnthompson's version). So maybe the bug this PR fixes only affects one specific version of IE11?

Btw I didn't realize this at the time, but without this PR's fix, my reproduction steps will also fail if I just click the left (previous) and right (next) arrow buttons with a mouse in steps 3-6.

Given all that and since WET 4.0.27 is supposedly slated to launch today, maybe there'd be value in merging in this PR as-is prior to launch?

@lucas-hay @cmeka @duboisp Thoughts?

StdGit:
I guess @EricDunsworth is right. The bug is on every machine with IE11 11.0.960.18837 that I tested. The only machine that doesn't have the bug is @shawnthompson  with 11.0.9600.18738

lucas-hay:
@EricDunsworth  The release is being pushed until tomorrow, I can merge if someone can confirm they have tested and this fixes the issue.

EricDunsworth:
@lucas-hay I've tested this PR's fix in my IE11 install and it works for both of the input types I brought up (keyboard and clicking navigation). Btw my IE11 version is actually 11.0.9600.18837 (9600 as opposed to 960).  So it's newer than @shawnthompson's version.

Btw since I don't currently have IE10, I don't know if it's also affected by the bug this PR fixes. If it's determined later on that it's also affected, the IE11 check this PR added would need to be amended to also target IE10 (or any version of IE since WET technically only supports IE10+ nowadays).

Either way, I don't think the lack of IE10 testing should hold back this PR from getting merged in prior to 4.0.27. IMO the bug is bad enough that it practically makes the datepicker poylfill's widget useless in IE11. It'd be unfortunate to let it linger on for another 3+ months.

**PS:** [According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#Browser_compatibility), Edge 12+ natively supports ``input type="date"``. So all versions of Edge use a native date picker widget instead of WET's polyfill and are therefore unaffected by the bug this PR fixes.

duboisp:
I tested in Edge (Lastest one) and the date polyfill don't load because Edge support ```input[type=date]```

I tested my version of IE11 (11.0.9600.18838) and I was able to reproduce the bug and I confirm it fix it too.

I tested in IE10 mode from my IE11 version and I wasn't able to reproduce the bug.

For testing here a working example with this PR been applied: 
http://universallabs.org/wet/labs/pr-8188/unmin/demos/datepicker/datepicker-en.html

@shawnthompson can you test is if the propose solution is conflicting with your version of IE 11?

lucas-hay:
@duboisp thanks for the link. I confirm it fixes the issue on my browser.

I think we are safe to merge.

Last call for objections...

shawnthompson:
It still works like it should with my version of IE... merge away mighty one!

StdGit:
It works for me too with 11.0.9600.18837  and
http://universallabs.org/wet/labs/pr-8188/unmin/demos/datepicker/datepicker-en.html

 Good job guys.

PR #8197, opened 2017-11-07
Theme: Added footer/cite elements into placeholder content's blockquote.

EricDunsworth:
This addition makes the theme pages' placeholder content demonstrate how Bootstrap 3.3.1's CSS adds a dash in front of blockquote elements that contain footer elements for citations. It may prove useful to anyone that refers to the placeholder content when coding blockquote citations in real content.

Bootstrap 3.3.1's blockquote guidance can be found at https://bootstrapdocs.com/v3.3.1/docs/css/#type-blockquotes.

**Screenshot:**
v4 0-blockquote-citation-v2

### DIFF
--- a/site/includes/placeholdercontent.hbs
+++ b/site/includes/placeholdercontent.hbs
@@ -103,4 +103,5 @@
 </form>
 <blockquote>
 	<p>"<code>blockquote</code>"</p>
+	<footer><code>footer</code> <cite><code>cite</code></cite></footer>
 </blockquote>
--- a/theme/site/pages/content-nosearchlang.hbs
+++ b/theme/site/pages/content-nosearchlang.hbs
@@ -8,7 +8,7 @@
 	"languagetoggle": "false",
 	"breadcrumb3": "true",
 	"breadcrumb3-title": "WET theme",
-	"dateModified": "2014-02-19"
+	"dateModified": "2017-11-22"
 }
 ---
 {{>placeholdercontent}}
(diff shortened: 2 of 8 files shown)


site/includes/placeholdercontent.hbs
```diff
@@ -103,4 +103,5 @@
 </form>
 <blockquote>
 	<p>"<code>blockquote</code>"</p>
+	<footer><cite><code>cite</code></cite></footer>
```
duboisp: What about changing the text "cite" by "footer>cite"? I think that change will be more obvious that the "cite" need to be in a "footer" element

duboisp (changes requested):

site/includes/placeholdercontent.hbs
EricDunsworth: IMO ``footer>cite`` would look kind of strange. I'd prefer not to place anything resembling a CSS selector into the placeholder content.

I could compromise by placing ``footer`` ``cite`` side-by-side. That would at least make it more obvious at first glance that a ``footer`` element is involved in creating that layout. Plus it's technically ok for the ``footer`` element to have content in it separately from ``cite`` in that context.

EricDunsworth:
@duboisp FYI I've updated the following aspects of this PR:

* Used ``footer`` ``cite`` side-by-side (instead of just ``cite``) to make it more obvious that the ``footer`` element is needed to establish that layout.
* Updated the OP's screenshot.
* Updated modified dates in the theme pages that use the placeholder content include.

duboisp (approved):

PR #8198, opened 2017-11-08
Core: Declared all SCSS variables as defaults.

EricDunsworth:
Certain core SCSS variables previously weren't declared as defaults. As a result, themes weren't able to override them.

Since that was most likely unintentional, this commit sets them all as defaults.

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -21,7 +21,7 @@ $wb-assets-path: "#{$wb-core-path}/assets" !default;
 //## Gray and brand colors for use across the application.
 
 // Bootstrap Overrides
-$gray-light: #767676;
+$gray-light: #767676 !default;
 
 $brand-primary: #2572b4 !default;
 $brand-success: #1b6c1c !default;
@@ -30,11 +30,11 @@ $brand-warning: #f2d40d !default;
 $brand-danger: #bc3331 !default;
 
 // WET Custom
-$clrWhite: #fff;
-$clrLight: #eee;
-$clrMedium: #ccc;
-$clrDark: #555;
-$clrDarkBlue: #243850;
+$clrWhite: #fff !default;
+$clrLight: #eee !default;
+$clrMedium: #ccc !default;
+$clrDark: #555 !default;
+$clrDarkBlue: #243850 !default;
 
 
 //== Scaffolding
@@ -64,7 +64,7 @@ $font-scale-lg: 1.125 !default;
 $font-scale-sm: .875 !default;
 
 // Bootstrap Overrides
-$font-size-base: 16px;
+$font-size-base: 16px !default;
 $font-size-large: ceil($font-size-base * $font-scale-lg) !default;
 $font-size-small: ceil($font-size-base * $font-scale-sm) !default;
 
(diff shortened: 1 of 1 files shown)


src/_variables.scss
```diff
@@ -30,11 +30,11 @@ $brand-warning: #f2d40d !default;
...
-$clrWhite: #fff;
-$clrLight: #eee;
-$clrMedium: #ccc;
-$clrDark: #555;
-$clrDarkBlue: #243850;
+$clrWhite: #fff !default;
+$clrLight: #eee !default;
+$clrMedium: #ccc !default;
+$clrDark: #555 !default;
+$clrDarkBlue: #243850 !default;
```
LaurentGoderre: These ones are kind of bad variables because if you change the dark blue to dark green the whole thing doesn't make sense. These ones shouldn't be defaults

LaurentGoderre (dismissed):

src/_variables.scss
EricDunsworth: @LaurentGoderre I see where you're getting at, but there may still be situations where themes might want to intentionally override those kinds of colour variables.

For example, imagine this scenario... Just for argument's sake, let's say ``$clrDark`` was used heavily throughout WET's core SCSS for font colours. Now let's say someone wanted to create a dark WET theme. They'd most likely want their theme to use bright font colours. The simplest, cleanest way of doing it would be to override the ``$clrDark`` variable's colour in their theme's _variables.scss file. However, if core's declaration for that variable prevented it from being overridden, the theme would be forced to use much messier/harder-to-manage alternatives (such as manually overriding all SCSS selectors that rely on the ``$clrDark`` variable).

On a side note, Bootstrap 3.3.1 already does what I'm trying to achieve with this PR. Literally all of [its SCSS variable declarations](https://github.com/twbs/bootstrap-sass/blob/93f42bc913bdd4c9df3d4652bd5168b4ede870c6/assets/stylesheets/bootstrap/_variables.scss) set them as defaults to allow them to be overridden. Even Bootstrap variables that are named after specific colours (like ``$gray``, ``$gray-dark`` and ``$gray-light``) are set as defaults and can be overridden.

In fact, WET core is currently overriding Bootstrap's ``$gray-light`` variable to darken it (via #5600). My goal is to darken it slightly more in GCWeb, but can't currently do so because WET core's override for it isn't setup as a default.

LaurentGoderre: We should have purpose variable and perhaps those variable could override those ones.

EricDunsworth: @LaurentGoderre I agree, although even then there may be situations where it may be desirable for multiple purpose-driven variables to inherit the value of a common colour variable. FWIW, Bootstrap already does that.

Btw I believe the ``$clr`` variables are carryovers from WET 3.0. I think the first ones originated in [WET 3.0's footnotes SCSS file](https://github.com/wet-boew/wet-boew-legacy/blob/v3.0/src/js/sass/includes/_footnotes.scss#L8). Then the variables eventually made their way into WET 4,0's centralized variables file and their naming scheme seems to have been used as a basis to create further variables over time. Maybe it'd be worth revisiting in WET 4.1/5.0.

In any case, I don't think WET's lack of purpose-driven variables should be a show-stopper for this PR. This PR makes WET core more extensible and more in line with how Bootstrap handles the same thing. Plus it'll make it possible for wet-boew/GCWeb#1309's change to take effect.

LaurentGoderre (approved):

PR #8200, opened 2017-11-10
Build: Modified sass-lint settings to require EOF newlines.

EricDunsworth:
Sass-lint doesn't require SCSS files to end with a newline by default. However, WET's .editorconfig file and ESLint's default settings do.

This commit makes WET's sass-lint settings consistent with those in that respect.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -4,6 +4,7 @@ options:
 rules:
   class-name-format: 0
   function-name-format: 0
+  final-newline: 2
   id-name-format: 0
   indentation:
     - 2


nschonni (approved):

EricDunsworth:
@nschonni @LaurentGoderre Once both of you have approved this, I'll port it to the other theme repos.

nschonni:
I'm also looking at https://github.com/editorconfig-checker/editorconfig-checker.javascript for another project. Not sure how well it works yet

EricDunsworth:
@nschonni Wow, that's awesome. I was thinking about trying to implement something along the lines of https://github.com/schorfES/grunt-lintspaces and trying to replicate some of WET's .editorconfig settings in it in order to start validating those kinds of things in HBS/etc files. But that would only be limited to a subset of what can be configured within a .editorconfig file. It wouldn't affect things like encoding.

At first glance, it looks like https://github.com/editorconfig-checker/editorconfig-checker.javascript would be able to enforce all .editorconfig settings in any kind of file. So maybe that'd be a better bet. Btw many HBS files throughout WET's repos currently don't fully respect its .editorconfig settings.

LaurentGoderre:
I think I would prefer an editorconfig linter than individual settings in all other linters.

nschonni:
I'm not sure how great the full EditorConfig-Checker is right now, so I think this can't hurt anyway

EricDunsworth:
Ugh, I just noticed that I didn't insert ``final-newline: 2`` in alphabetical order. I placed it just after ``function-name-format: 0``, but should've placed it before.

Gonna send another PR to address that. Btw my ports to WET's other themes will be in the correct order. I know it's super minor, but all the pre-existing rules were in alphabetical order. Plus I'd prefer to keep WET and its themes' sass-lint settings files as identical as possible.

PR #8202, opened 2017-11-15
Build: Made sasslint task's src globbing patterns more specific.

EricDunsworth:
The "sasslint" task was previously setup to include any SCSS files as sources on a global scale. Exclusion patterns were used to omit anything within the lib, node_modules, dist and wet-boew-dist folders. That resulted in really slow linting performance in Windows environments. That setup was also inconsistent with most of the other tasks. Apart from the "gh-pages" task, all other tasks' source inclusion patterns are more specific and don't rely so heavily on exclusion patterns.

This commit improves the "sasslint" task's performance (especially on Windows) by making its source inclusion patterns more specific and by reducing its reliance on exclusion patterns.

On my Windows 7 notebook (which uses a really slow HDD), the original "sasslint" task would take nearly 9 minutes to run after a cold boot. Subsequent runs would take between 75-90 seconds. Subsequent runs with the updated "sasslint" task only take 2 seconds. That's even faster than the 3.5-5 seconds it usually took for the original version of that task to run in Travis-CI.

@nschonni @LaurentGoderre FYI.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -580,11 +580,9 @@ module.exports = (grunt) ->
 			all:
 				expand: true
 				src: [
-						"**/*.scss"
-						"!lib/**"
-						"!node_modules/**"
-						"!dist/**"
-						"!wet-boew-dist/**"
+						"site/**/*.scss"
+						"src/**/*.scss"
+						"theme/**/*.scss"
 						"!src/**/sprites/**"
 					]
 


LaurentGoderre (approved):

EricDunsworth:
Btw the new commit I just force pushed fixed a small typo in my commit message. I didn't make any further changes to Gruntfile.coffee.

nschonni:
Thanks! Think it's time to give that laptop a magnet massage so they replace it :stuck_out_tongue_winking_eye:

EricDunsworth:
@nschonni Heh :), Well thankfully I think I'm slated to get a new laptop in the not-too-distant future...

PR #8212, opened 2017-11-23
Feature calc

Paul-Doug:
Plugin calc: A calculator helper plugin. 
This plugin was created by the Codesprint group. The requirement for calculations that could be implemented via WET-assisted HTML came about during the migration of web pages where inline Javascript was not supported.


duboisp (changes requested):
* [ ] This PR will need some commits cleanup. 
* [ ] Need to create a French working example

duboisp:
FYI
A independent WET plugin to perform calculation- [https://github.com/canada-ca/digital-playbook-guide-numerique/blob/de904068fc98152977f6315e115bdbbebf63f6cd/assets-atouts/js/calcul.js](https://github.com/canada-ca/digital-playbook-guide-numerique/blob/master/assets-atouts/js/calcul.js)

pjackson28:
FYI
The calcul.js is in use on the following pages:
- [Algorithmic Impact Assessment (draft)](https://canada-ca.github.io/digital-playbook-guide-numerique/views-vues/automated-decision-automatise/en/algorithmic-impact-assessment.html) (to tally the points, calculate the percentage and then trigger results on the page based upon certain conditions related to the calculations)
- [Right Cloud Decision Tool (draft)](https://canada-ca.github.io/digital-playbook-guide-numerique/views-vues/cloud-nuage/en/cloud-decision-tool.html) (similar use case as the other link)

delisma:
> There is several data-* attribute. Can we use only one but to have a JSON object as a value?
> 
> ```
> <output for=”ProvOrTerr priceinput BefOrAft” data-wb-calc=’{
>                 “rounded”: 2,
>                 “eqnum”: “BefOrAft”,
>                 “equation”: [
>                                 “#priceinput * 0.05”,
>                                 “(#priceinput / (1 + 0.05 + #ProvOrTerr)) * 0.05”
>                 ]
> }’></output>
> ```

Is there a polyfill for the `output` tag in WET for IE11?

duboisp:
@delisma 
 
> Is there a polyfill for the `output` tag in WET for IE11?

I don't know, but reading what the element do, it should be quite simple to do for this particular use case. 
Here some references:
* https://html.spec.whatwg.org/multipage/form-elements.html#the-output-element
* https://w3c.github.io/html-aria/#el-output
* https://w3c.github.io/html-aam/#el-output
* https://www.w3.org/TR/wai-aria-1.1/#status

Issue #8088, opened 2017-07-14
JAWS does not work properly


ognil:
Hi guys,

New version of the script wet-boew.min does not include 
aria-posinset  and aria-setsize as follow (it is from old version  v4.0.22 - 2016-08-11) :

`  <li><a role="menuitem " tabindex="-1"  aria-posinset="2" aria-setsize="15"
 href="/24-hour-emergency-numbers">24 hour Emergency `Numbers</a></li>`

As a result, JAWS repeats at least two times the name of the menu item and does not says “two of fifteen”. 

George

nschonni:
Duplicate of #8081

Issue #8081, opened 2017-07-06
Menu item size e.g. X of X no longer reported by screen readers


pbossley:
Verified in working example site: https://wet-boew.github.io/v4.0-ci/index-en.html
Firefox 54.0.1 and NVDA 2017.2, JAWS 18.
When navigating menus with screen readers the set size and position in the menu is no longer reported by screen readers and now just announce 1 of 1 instead of properly reporting position in the set. 
This may be caused by ARIA set size not being defined.

LaurentGoderre:
@shawnthompson this is caused by b9c37a90ad4a4a6c85381638011ec0d92e1cd236

pbossley:
Interestingly, the ARIA Authoring Practices example does expose the size to screen readers. https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-1/menubar-1.html
I'm not sure why that one is working properly as it doesn't appear all that different from what is being done with WET and it also doesn't seem to use set size or position in set... In any case, I would argue that this issue should in fact be solved as users should be given the correct information to orient them to the position in the overall set.

pbossley:
If role=none is applied to the <li> items throughout, screen readers will go back to announcing position properly. Even without the use of ARIA set size. Is there any way that can be done?

shawnthompson:
@pbossley I can work on this with you.

Thanks for your help in this.

pbossley:
Great. Please let me know if you need any further information and I'll try to help.
Pete Bossley
Director, Digital Accessibility Center
The Ohio State University | Office of University Compliance and Integrity

shawnthompson:
@pbossley Here is the code, which `<li>` should the `role="none"` go on?

```html
<nav role="navigation" typeof="SiteNavigationElement" id="sm-pnl" class="sm-pnl wb-menu wb-menu-inited wb-init wb-navcurr-inited">
	<h3>Topics menu</h3>
	<ul class="list-unstyled mb-menu" role="menu">
		<li>
			<details>
			<summary class="mb-item" aria-haspopup="true" tabindex="0"><span role="menuitem">WET project</span></summary>
			<ul class="list-unstyled mb-sm" role="menu" aria-expanded="false" aria-hidden="true">
				<li><a role="menuitem" tabindex="-1" href="http://wet-boew.github.io/v4.0-ci/index-en.html#about">About the Web Experience Toolkit</a></li>
[...]

Issue #7939, opened 2017-03-09
Menu: removing aria-posinset and aria-setsize


shawnthompson:
If you validate a page in WET after JS execution, you will get a bunch of errors with the `aria-posinset` and `aria-setsize` being set to the wrong element. After talking to some accessibility experts at an A11Ymtl meetup, as well as a speaker from the Confoo conference in MTL and looking into the use of these attributes I've come to the conclusion that these attributes are not really needed in our menu and it might be a little overkill to have them there.

I vote for removing them.

Stay tuned for a PR.

cc: @andrewnordlund @masterbee @TouficS @rlambert27

kodecount:
I've had issue on this ages ago because the accessibility software I use for testing flagged them as incorrect.  See issue https://github.com/wet-boew/wet-boew/issues/7795.

shawnthompson:
I'm bad at searching for per-existing issues, my bad. We are going to remove them. Thanks for linking the issues together.

kodecount:
That fine. I closed the issue as their was no interest at the time. Currently WET breaks WCAG by having these on the `<a>` element versus `<li>` anyways.

shawnthompson:
That's why I'm removing them and they aren't really needed because the menu is pretty basic with a standard order and nothing moving inside the drop downs.

duboisp:
I am fine with this as if the menu will remain navigable with common assistive technology.

Issue #7795, opened 2016-12-02
Menu: (Accessibility)


kodecount:
What is the purpose of `aria-posinset="1" aria-setsize="1"` and are they breaking any rule by being attached to a link element.

According to AXE they could break this rule: https://dequeuniversity.com/rules/axe/2.1/aria-allowed-attr?application=axeAPI

duboisp:
> Some states are managed by the user agent, such as ```aria-posinset``` and ```aria-setsize```, but the author can override them if the DOM is incomplete and would cause the user agent calculation to be incorrect. - [2.2 WAI-ARIA States and Properties](https://www.w3.org/TR/wai-aria-1.1/#introstates)

FYI - here the spec of :
* [aria-posinset](https://www.w3.org/TR/wai-aria-1.1/#aria-posinset)
* [aria-setsize](https://www.w3.org/TR/wai-aria-1.1/#aria-setsize)

@kodecount it is about a specific use case? if so can you provide more details.

kodecount:
@duboisp Thanks for the information. I'm using JAWs, WAVE Evaluation Tool, and aXe in Chrome for accessibility testing one of the applications had a issue with the aria controls. I believe it due to the fact that their intended for list elements and not link elements except provide the same purpose in this instance.

kodecount:
@duboisp Thanks for the information. I'm using JAWs, WAVE Evaluation Tool, and aXe in Chrome for accessibility testing one of the applications had a issue with the aria controls.

I think I see why it has a issue with the aria controls. Maybe these should be applied to the `<li>` tag and not the `<a>`? Is their a reason you choice to add them to the link versus the list item?

kodecount:
@duboisp Thanks for the information. I'm using JAWs, WAVE Evaluation Tool, and aXe in Chrome for accessibility testing one of the applications had a issue with the aria controls.

I think I see why it has a issue with the aria controls. Maybe these should be applied to the `<li>` tag and not the `<a>`? Is their a reason WET choose to add them to the link versus the list item?

PR #8230, opened 2017-12-07
Revert "Menu aria validation" #7942

duboisp:
Revert the changes applied by PR #7942 as it introduced a major keyboard navigation issue.

**Current issue** 
On small device, when the user open the site menu, the user can not navigate through the menu items with a keyboard. Because of that, we can highly presume, it cause major issues to assistive technology that are trying to access it from a small device (see #8081 and #8088).

**Immediate solution proposal** 
Revert the changes for now and then take the time to investigate what issue #7942 is trying to solve and then propose a long term solution which are not conflicting with or degrading the existing implemented solution.

**Revert rational**
The information available from PR #7942 , from issue #7939 and #7795 are insufficient. They are not expressing enough research about finding why it was originally implemented like that and they are not showing that an extensive testing was completed with major browser and with assistive technology. 
Please note that at the time the original solution was designed, it was mentioned to me that it was producing a valid HTML markup. We have found that a similar design approach was used and still used by the tabbed interface when entering in mobile mode.

**Future solution**
The proposed solution will need extensive testing with all supported browser and assistive technology.
Prior to merge the new solution, the following task would need to be completed
* Validate or invalidate the need of using ```aria-setsize``` and ```aria-posinset```. Those attribute was added based on bad user experience for navigating the menu (including mobile) with an assistive technology like NVDA and JAW.
* Research and cite the actual reference in the WAI-ARIA specification that explicitly said it's not conforming to spec. Even if the Nu validator are reporting an error, we need to ensure that is not a bug in the validator or a miss-interpretation of the spec by the validator. 
* Documents the markup structure of current menu template and the enhanced menu template.
* Design and propose a few enhanced menu markup structure showing alternate solution along with an explanation of the navigation model.
* Investigate the tabbed interface mobile mode which use a similar mobile design approach and make proposal/recommendation.
* For the implementation of the solution, the developer would need to closely review the code in the menu.js, to look at how the "keydown" event is managed, ensure the new approach is consistent for the megamenu vs the mobile menu and the most important to do a lot of testing. 

----
ps: If this PR is merged, I would open an new official issue to track the progression of the long term solution.

/cc @lucas-hay @shawnthompson @EricDunsworth @masterbee @rlambert27

### DIFF
--- a/src/plugins/menu/menu.js
+++ b/src/plugins/menu/menu.js
@@ -85,6 +85,8 @@ var componentName = "wb-menu",
 			$subMenu = $elm.siblings( "ul" );
 
 			$elm.attr( {
+				"aria-posinset": ( i + 1 ),
+				"aria-setsize": length,
 				role: "menuitem"
 			} );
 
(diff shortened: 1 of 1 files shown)


lucas-hay:
@duboisp I have been discussing this with @masterbee for the last couple days trying to figure out the proper approach.
@masterbee would like @pjackson28's input on this from a policy standpoint. 
We are not comfortable releasing with an accessibility issue, and are open to delaying the release until we have a proper resolution. Although releasing with a validation error (if it is even an error as you mentioned above), is not ideal, we believe it is better than releasing with an accessibility error. 
@pjackson28 can you provide your take on the issues outlined above?

pjackson28:
@lucas-hay From my standpoint, the accessibility issues should be resolved now and a plan should be created for addressing the validation issues in the near future (including proper testing for the planned changes). The summary element-related validation errors came as a result of a recent tightening of requirements in WAI-ARIA which affects any WAI-ARIA overrides applied to summary elements (for WET that is the mobile menu and the mobile version of the tabbed interface). That tightening of WAI-ARIA requirements doesn't change the level of accessibility of the existing solutions but does require additional work to maintain technical compliance.

lucas-hay (approved):

PR #8252, opened 2018-01-12
build: Update Selenium browsers for vendor support

nschonni:
- Used http://gs.statcounter.com/ for some basics on the OS market.
- Dropped IE 9 and 10 as they are no longer supported by MS
- Added Microsoft Edge to the matrix

Note: it looks like the Sauce integration has been off since May https://saucelabs.com/u/wet-boew-bot

### DIFF
--- a/browsers.json
+++ b/browsers.json
@@ -3,7 +3,7 @@
     "platform": "Linux"
 },{
     "browserName": "firefox",
-    "platform": "OS X 10.9"
+    "platform": "OS X 10.13"
 },{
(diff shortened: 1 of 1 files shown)


LaurentGoderre (approved):

nschonni:
@LaurentGoderre do you know what happened to the Sauce bot setup, was it revoked tockens or just part of the docker changes that it got disabled?

LaurentGoderre:
@nschonni I think it was breaking the build too much so it was taken out of the build

PR #8266, opened 2018-01-24
Do not block collapsible component init on whole document

Pacoup:
This uses a technique similar to other plugins (https://github.com/wet-boew/wet-boew/blob/master/src/plugins/filter/filter.js#L94) to, if I understand how this works correctly, prevent the initialization of the component from blocking on the whole document.

In small documents, the difference is negligible, but in large documents with many collapsible alerts, this is much better, which addresses the performance issue here: https://github.com/wet-boew/wet-boew/issues/7125#issuecomment-144518981

### DIFF
--- a/src/plugins/collapsible-alerts/collapsible-alerts.js
+++ b/src/plugins/collapsible-alerts/collapsible-alerts.js
@@ -16,7 +16,7 @@ var componentName = "wb-collapsible",
 	selector = "details.alert",
 	initEvent = "wb-init." + componentName,
 	$document = wb.doc,
-	details, key,
+	key,
 
 	/**
 	 * @method init
@@ -27,9 +27,11 @@ var componentName = "wb-collapsible",
 		// Start initialization
 		// returns DOM object = proceed with init
 		// returns undefined = do not proceed with init (e.g., already initialized)
-		details = wb.init( event, componentName, selector );
+		var details = wb.init( event, componentName, selector ),
+			$details;
 
 		if ( details ) {
+			$details = $( details );
 
 			key = "alert-collapsible-state-" + details.getAttribute( "id" );
 
@@ -58,7 +60,7 @@ var componentName = "wb-collapsible",
 			} catch ( e ) {}
 
 			// Identify that initialization has completed
-			wb.ready( $document, componentName );
+			wb.ready( $details, componentName );
 		}
 	};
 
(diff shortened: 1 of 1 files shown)


src/plugins/collapsible-alerts/collapsible-alerts.js
```diff
@@ -27,9 +27,11 @@ var componentName = "wb-collapsible",
 		// Start initialization
 		// returns DOM object = proceed with init
 		// returns undefined = do not proceed with init (e.g., already initialized)
-		details = wb.init( event, componentName, selector );
+		var details = wb.init( event, componentName, selector ),
```
duboisp: ```details``` variable is already declared at line 19. 

You might want to remove the declaration at line 19 and declare the variable again in the declaration block on line 75

duboisp (changes requested):
With how many collapsible alerts have you tested that is actually causing an issue?

Pacoup:
Nothing specific, just @chyne's code, which I've reproduced here. 1 is without the patch, 2 is with the performance patch.

1. https://pacoup.com/collapsible-perf-1/dist/demos/collapsible-alerts/collapsible-alerts-en.html
2. https://pacoup.com/collapsible-perf-2/dist/demos/collapsible-alerts/collapsible-alerts-en.html

Open IE11, the problem used to be present in Chrome and Firefox but the blocking code seems to have been multithreaded away in recent versions.

- Open number 1 and try scrolling, the page will block for a few seconds before actually starting to scroll.
- Open number 2 and try scrolling, the page only blocks momentarily.

duboisp:
This PR will require more testing to ensure it don't conflict.

Pacoup:
How do you mean?

duboisp:
Simply creating an independent working example and run a few test to ensure it works as expected. If there is no issue, this PR would be merged at the next Technical review meeting. I would sent you the meeting invite.

Pacoup:
Like a general test, or would you require a unit test to be added?

duboisp:
More general test, but if you want you can always add some unit testing too.

Pacoup:
Okay, I haven't created any unit tests, but I've made two test pages, one with the commit before this pull request (as I haven't rebased on top of the recent changes from February), and one with the contents of this pull request along with a couple of different plugins on the same page to demonstrate that they work the same and that there are no errors:

- e09259376 - [original](https://pacoup.com/collapsible-test-1/unmin/demos/collapsible-alerts/collapsible-alerts-en.html)
- e29270ba7 - [this pull request](https://pacoup.com/collapsible-test-2/unmin/demos/collapsible-alerts/collapsible-alerts-en.html)

duboisp (approved):

PR #8282, opened 2018-02-09
Theme: Reference SVG logos with img element instead of object.

### LINKED ISSUE #8276: All of WET 4.0's logo SVGs are implemented in a manner that fails HTML validation
A few days ago, validator/validator#422 was resolved. One of the commits that fixed it (validator/validator@1c02bc18e7a71d531fcd62e60b4dc2064c753cef) intentionally caused any elements containing ``tabindex`` attributes to trigger a validation error when situated within ``<a>``/``<button>`` elements.

Based on my research, validator/validator@1c02bc18e7a71d531fcd62e60b4dc2064c753cef's behaviour is in line with the [W3C's HTML spec](https://w3c.github.io/html/):

* Excerpt from the [spec's tabindex section](https://w3c.github.io/html/editing.html#element-attrdef-global-tabindex):
  * "An element with the [tabindex](https://w3c.github.io/html/editing.html#element-attrdef-global-tabindex) attribute specified is [interactive content](https://w3c.github.io/html/dom.html#interactive-content-2)."
* Excerpt from the [spec's interactive content section](https://w3c.github.io/html/dom.html#interactive-content-2):
  * "The [tabindex](https://w3c.github.io/html/editing.html#element-attrdef-global-tabindex) attribute can also make any element into [interactive content](https://w3c.github.io/html/dom.html#interactive-content-2)."
* Excerpt from the [spec's content model section for the ``a`` element](https://w3c.github.io/html/textlevel-semantics.html#elementdef-a):
  * "[**Content model**](https://w3c.github.io/html/dom.html#content-model): [Transparent](https://w3c.github.io/html/dom.html#transparent), but there must be no [interactive content](https://w3c.github.io/html/dom.html#interactive-content-2) or [<a>](https://w3c.github.io/html/textlevel-semantics.html#elementdef-a) element descendants."

So, what does this have to do with WET? It turns out that most of WET 4.0's themes use SVG site logos that are structured as links containing an ``<object>`` element with a ``tabindex="-1"`` attribute. In other words, WET is nesting interactive content (``<object tabindex="-1">``) within other interactive content (``<a>``), which goes against the HTML spec.

As a result, the following error is now appearing when [validating virtually any pages that use WET](https://validator.w3.org/nu/?doc=http%3A%2F%2Fwet-boew.github.io%2Fv4.0-ci%2Findex-en.html):
> **Error:** An element with the attribute tabindex must not appear as a descendant of the ``<a>`` element.
> 
> From line 61, column 1; to line 61, column 83
> 
> ``<object type="image/svg+xml" tabindex="-1" data="./theme-wet-boew/assets/logo.svg">``

IMO the best way of fixing this would be to replace the ``<object>`` element with ```` in WET's SVG images and remove the ``tabindex="-1"`` attributes. I believe it should work fine in all modern browsers and retain compatibility with the [SVG image replacer polyfill](https://github.com/wet-boew/wet-boew/blob/master/src/polyfills/svg/svg.js).

@duboisp @LaurentGoderre @nschonni @pjackson28 @shawnthompson Any thoughts/concerns?

**PS:**
The reason a ``tabindex="-1"`` attribute was used on SVG ``<object>`` elements in the first place was because of #1432 (which was fixed by wet-boew/wet-boew-legacy@cc8b81c27b92dc1b2caded92d5503f151393497f). Without that attribute, all browsers attempt to set keyboard tabbing focus inside the ``<object>`` element - which isn't desirable.

EricDunsworth:
* Fixes #8276.
* Adds a print CSS filter to change the logo's brightness to 0%:
  * Needed for Gecko, Blink and possibly Webkit due to their buggy handling of inline SVG media queries when the img element is used. When printing, they apply screen-only queries and ignore print-only queries.
  * Prevents the fallback PNG logo from being coloured white when printing in browsers that support CSS filters.
* Retains the SVG logo's inline screen media query to prevent issues in browsers that correctly handle them and lack support for CSS filters (i.e. Internet Explorer 10-11).
* Restores the splash page's "wb-sppe" class (which was inadvertently removed in #6113):
  * Allows the splash page's logo image to be specifically targeted by the print CSS filer.
  * Indirectly brings back #5293's fix for the splash page logo's size in Internet Explorer 10-11 (which also fixes the fallback PNG logo's size in all browsers).

### DIFF
--- a/theme/banner/_base.scss
+++ b/theme/banner/_base.scss
@@ -2,7 +2,10 @@
  Banner
  */
 
+// Needed for backwards compatibility with WET 4.0.27 and below's HTML markup.
+// Remove object references from all of this theme's logo selectors in WET 4.1+.
 .wb-sppe {
+	img,
 	object {
 		width: 100%;
 	}
--- a/theme/banner/_print.scss
+++ b/theme/banner/_print.scss
@@ -1,11 +1,25 @@
 /*
  Banner (print view)
  */
+%theme-wet-boew-banner-img-print-brightness-0 {
+	img {
+		filter: brightness(0);
+	}
+}
+
 #wb-bar {
 	@extend %theme-wet-boew-print-display-none-important;
 }
 
+.wb-sppe {
+	header {
+		@extend %theme-wet-boew-banner-img-print-brightness-0;
+	}
+}
+
 #wb-sttl {
+	@extend %theme-wet-boew-banner-img-print-brightness-0;
+
 	a {
 		&:after {
 			@extend %theme-wet-boew-print-display-none-important;
--- a/site/includes/brand.hbs
+++ b/site/includes/brand.hbs
@@ -1,6 +1,6 @@
 <div id="wb-sttl" class="col-md-8">
 	<a href="{{assets}}/index-{{#contains language "fr"}}{{language}}{{else}}en{{/contains}}.html">
-		<object type="image/svg+xml" tabindex="-1" data="{{assets}}/../{{site.theme}}/assets/logo.svg"></object>
+		<img src="{{assets}}/../{{site.theme}}/assets/logo.svg" alt="" />
 		<span>{{{i18n "site-title"}}}<span class="wb-inv">{{{i18n "comma-space"}}}</span><small>{{{i18n "site-tagline"}}}</small></span>
 	</a>
 </div>
--- a/site/layouts/splashpage.hbs
+++ b/site/layouts/splashpage.hbs
@@ -3,21 +3,23 @@
 	"layout": "core.hbs"
 }
 ---
-<header role="banner">
-	<div id="wb-bnr">
-		<div class="container">
-			<div class="row mrgn-tp-lg mrgn-bttm-lg">
-				<div class="col-md-8 col-md-offset-2">
-					<object type="image/svg+xml" tabindex="-1" role="img" data="{{assets}}/../{{site.theme}}/assets/logo.svg" aria-label="{{{i18n "site-title"}}}"></object>
+<div class="wb-sppe">
+	<header role="banner">
+		<div id="wb-bnr">
+			<div class="container">
+				<div class="row mrgn-tp-lg mrgn-bttm-lg">
+					<div id="wb-spllg" class="col-md-8 col-md-offset-2">
+						<img src="{{assets}}/../{{site.theme}}/assets/logo.svg" alt="{{{i18n "site-title"}}}" />
+					</div>
 				</div>
 			</div>
 		</div>
-	</div>
-</header>
-<main role="main" property="mainContentOfPage" class="container">
-	<div class="row mrgn-tp-lg">
-		<div class="col-md-12">
-			{{>body}}
+	</header>
+	<main role="main" property="mainContentOfPage" class="container">
+		<div class="row mrgn-tp-lg">
+			<div class="col-md-12">
+				{{>body}}
+			</div>
 		</div>
-	</div>
-</main>
+	</main>
+</div>


EricDunsworth:
Btw the Travis-CI issues seem to be related to a pre-existing issues with the repo:
* #8280 fixes the MathJax Bootlint errors.
* I believe the missing internal image errors are the result of longstanding issues in this repo that are only now appearing because extra checks are being performed on the ```` element's ``src`` attribute. Might be because the references to the SVG logo images aren't using relative paths properly. The logo image appears in the [minified version of GitHub pages' splash page](https://wet-boew.github.io/v4.0-ci/theme/splashpage-en.html), but doesn't appear in the [unminified version of it](https://wet-boew.github.io/v4.0-ci/unmin/theme/splashpage-en.html). In my local environment, the logo image didn't appear in either the unminified nor minified splash pages. None of this PR's changes modified any references to the logo image's path.

duboisp:
For the svg in 
```
- dist/index.html
  *  internal image ./assets/logo.svg does not exist (line 33)
- dist/theme/splashpage-en.html
  *  internal image ../assets/logo.svg does not exist (line 33)
- dist/theme/splashpage-fr.html
  *  internal image ../assets/logo.svg does not exist (line 33)
- dist/unmin/index.html
  *  internal image ./assets/logo.svg does not exist (line 40)
- dist/unmin/theme/splashpage-en.html
  *  internal image ../assets/logo.svg does not exist (line 40)
- dist/unmin/theme/splashpage-fr.html
  *  internal image ../assets/logo.svg does not exist (line 40)
```

Can you change the relative path for: ```../../theme-wet-boew/assets/logo.svg```. Like the link to the favicon.

LaurentGoderre:
This is a markup change though so it's not likely to be implemented by many WET4 sites.

duboisp:
* [ ] Test if the CSS change conflict for WET 4 sites without that markup change.

EricDunsworth:
@duboisp I'll send out a separate PR for the relative path adjustment.

Btw this PR's CSS changes should be backwards-compatible with the previous HTML markup. Although I haven't done any first-hand testing of that as of yet. The CSS changes are basically just adding CSS filters to improve support for modern browsers that don't correctly handle SVGs with inline media queries.

@LaurentGoderre I presume you're referring to this PR?

I agree it isn't ideal to be changing the template's HTML markup in a PR intended for a maintenance release. But the status quo [has technically gone against the W3C spec since HTML 5.1](https://www.w3.org/TR/html51/textlevel-semantics.html#elementdef-a). The status quo also doesn't meet [WCAG 2.0's parsing success criterion](https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-parses.html) because it isn't nested according to the HTML spec.

Due to the [W3C's Nu validator](https://validator.w3.org/nu/) throwing an error on that now, it's going to become much more obvious to implementers (some of which are required to comply with WCAG 2.0). With that in mind and since WET 4.1/5.0 still looks to be a long way off, IMO this PR would be the best stop-gap solution.

I'd of preferred making a more minimal change like just removing the ``tabindex="-1"`` attribute... but in every modern browser I've checked this in (IE11/Firefox/Opera), ``<object>`` elements that directly reference SVG images without a ``tabindex="-1"`` attribute will gain keyboard tabbing focus - which is undesirable. All of the browsers I tested in treat the ``<object>`` element as a [nested browsing context](https://w3c.github.io/html/browsers.html#nested-browsing-contexts) in this scenario. That's in line with what's described in the handler section of the [spec's description of the ``<object>`` element](https://w3c.github.io/html/semantics-embedded-content.html#the-object-element).

theme/banner/_print.scss
```diff
@@ -1,11 +1,25 @@
 /*
  Banner (print view)
  */
+%theme-wet-boew-banner-print-brightness-0 {
+	img {
+		filter: brightness(0);
```
duboisp: Is that fix is needed every time a SVG is used, or it is unique to wet-boew theme?

theme/banner/_base.scss
```diff
@@ -2,7 +2,10 @@
  Banner
  */
 
+// Needed for backwards compatibility with WET 4.0.27 and below's HTML markup.
+// Remove object references from all of this theme's logo selectors in WET 4.1+.
 .wb-sppe {
+	img,
```
duboisp: Where and when the CSS class ```.wb-sppe``` is used? I am not able to find any reference or description of it.

theme/banner/_print.scss
EricDunsworth: It's needed for SVGs that contain inline media queries that change their colours to white in screen viewports. It also benefits their fallback PNGs. WET's theme needs it.

I'm currently looking into porting this PR to other affected WET themes. I'm taking the behaviour of the inline media queries in the other themes's SVG images into account. So similar kinds of CSS filters might be present in some of my port PRs.

theme/banner/_base.scss
EricDunsworth: It was first introduced it in #5293 to prevent the splash page's logo from shrinking when printing in IE (including IE11). Then the HTML markup that referenced that class was apparently inadvertently removed in #6113. Since then, the SCSS selector for it has been present in the WET theme, but wasn't tied to any HTML markup to make it take effect.

As a part of this PR I needed a way of targeting the splash page's logo image. I was going to introduce my own class for that purpose, but then stumbled upon the ``wb-sppe`` class and looked into its history. Since it was trying to achieve the same purpose as I was and was accidentally removed, I figured the best way forward would be to restore it.

theme/banner/_print.scss
duboisp: Do you think the scope of that CSS should be made broader than only for some of the wet-boew theme elements.

EricDunsworth: I'm still going through the other themes, but based on what I've seen so far, I think this kind of thing would work best on a theme-by-theme basis. Some themes use SVGs with other kinds of media queries that'll need to be handled differently.

duboisp: ok, then this fix is only for our WET logo to be printable. Thanks @EricDunsworth for the clarification.

EricDunsworth:
FYI I just forced pushed an update to rename my placeholder selector from ``%theme-wet-boew-banner-print-brightness-0`` to ``%theme-wet-boew-banner-img-print-brightness-0``. Since it contains a nested ``img`` selector, I wanted it to be clear from its name that it only applies to images.

duboisp (approved):

Issue #7972, opened 2017-04-18
Multimedia Player - Video


anvistyle:
Hello 
On this page http://wet-boew.github.io/v4.0-ci/demos/multimedia/multimedia-en.html
You have the error below. 
When I add my video to a page I'm having the same error.
May you please take a look at this problem?
Thanks

wet-boew.min.js:8 Uncaught TypeError: Cannot read property 'defaults' of undefined
    at complete (wet-boew.min.js:8)
    at l (wet-boew.min.js:8)
    at wet-boew.min.js:8
    at h (wet-boew.min.js:8)
    at HTMLScriptElement.n.onreadystatechange.n.onload [as onreadystatechange] (wet-boew.min.js:8)

LaurentGoderre:
In which browser do you get that error?

anvistyle:
Hello
Chrome Version 57
Regards

duboisp:
Here a few additional info about the error

The error occur in the [lightbox plugin at line 251](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/lightbox/lightbox.js#L251) 

Here an extract from line 245 to 255
```javascript
// Load Magnific Popup dependency and bind the init event handler
Modernizr.load( {
	load: "site!deps/jquery.magnific-popup" + wb.getMode() + ".js",
	complete: function() {

		// Set the dependency i18nText only once
		$.extend( true, $.magnificPopup.defaults, i18nText );

		$document.trigger( dependenciesLoadedEvent );
	}
} );
```

StdGit:
@LaurentGoderre @duboisp @CalvinRodo @shawnthompson 
I had similar error in IE 11.0.9600.18860 update 11.049 (KB4052978) 
and firefox 580b15 (64 bits)
Unable to get property 'defaults' of undefined or null reference
At first it works, but after some refresh it crashes at  line 6764

$.extend( true, $.magnificPopup.defaults, i18nText );

After that jquery doesn't run anymore.

duboisp:
This issue seem to have been fixed via PR #8291

PR #8291, opened 2018-02-16
LightBox / dependency loading/executing racing issue

StdGit:
To resolve the bug (Jquery doesn't fire) when we got the following error in wet-boew.min.js

Unsupported format of the sourcemapUnhandled exception at line 8, column 73294 in https://www.canada.ca/etc/designs/canada/cdts/gcweb/v4_0_26/js/wet-boew.min.js

0x800a138f - JavaScript runtime error: Unable to get property 'defaults' of undefined or null reference

Content of wet-boew.min.js column 73294 is

,"lbx-title"),b.data=c}}),Modernizr.load({"load":"site!deps/jquery.magnific-popup"+d.getMode()+".js","complete":function(){a.extend(!0,a.magnificPopup.defaults,h),o.trigger(m)}})}

Need to be tested with <script src="https://www.canada.ca/etc/designs/canada/cdts/gcweb/v4_0_26/js/wet-boew.min.js"></script>

Could also resolve issue #7972

### DIFF
--- a/src/plugins/lightbox/lightbox.js
+++ b/src/plugins/lightbox/lightbox.js
@@ -251,6 +251,9 @@ var componentName = "wb-lbx",
 		// Load Magnific Popup dependency and bind the init event handler
 		Modernizr.load( {
 			load: "site!deps/jquery.magnific-popup" + wb.getMode() + ".js",
+			testReady: function() {
+				return $.magnificPopup;
+			},
 			complete: function() {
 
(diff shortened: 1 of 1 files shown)


nschonni (approved):

LaurentGoderre:
@StdGit this is great!! can you add this to all Modernizr load statements?

StdGit:
@LaurentGoderre I will have a look.

PR #8300, opened 2018-03-01
Fixed tasks not loading when installed in a theme

LaurentGoderre:


duboisp:
Is this would fix GCWeb build?

nschonni (approved):
Didn't run it, but reading the doc it makes sense https://github.com/sindresorhus/load-grunt-tasks#requireresolution
This would allow calling this grunt file without having to restore the node modules into the wet-boew dependency if I'm understanding it

LaurentGoderre:
@duboisp this wont fix the build by itself but is required in order to fix it.

LaurentGoderre:
@nschonni what this does is that if wet is installed via npm, it's dependencies will be in the parent project where grunt can't see them. This allows grunt to traverse up to the parent to find the tasks.

LaurentGoderre:
@duboisp can I merge?

PR #8341, opened 2018-04-06
chore: Fix NPM files

nschonni:
Lock file and spacing fixes missed when dependencies moved from devDependencies


duboisp (approved):
It's look good for me. I didn't tested to build with this update, but I assume with the type of edits it will continu to work.

Issue #8332, opened 2018-03-27
Adding French translation for  data-rule-pattern and data-rule-notEqualTo


biljanap:
We are using ```data-rule-pattern``` and ```data-rule-notEqualTo``` attributes. 
We know that French validation messages for all data-rule attributes are included in /18n/fr.js file using:
````html
$.extend( $.validator.messages, {
   ---
} );
````
Would it be possible to include translation for  ```data-rule-pattern``` and ```data-rule-notEqualTo``` as well?

RobJohnston:
The best fix would be to put it in the jQuery Validation project, specifically, in this file:
https://github.com/jquery-validation/jquery-validation/blob/master/src/localization/messages_fr.js

I already added the notEqualTo translation here:  https://github.com/jquery-validation/jquery-validation/pull/2033.

But then it's a wait for it to filter down into WET.

duboisp:
Yes as @RobJohnston mention, best fix would be to put it in the jQuery Validation project.

Regarding to filter down into WET, you can always submit a PR in WET to would update that dependency. You simply need to update the following line: https://github.com/wet-boew/wet-boew/blob/master/package.json#L42 However, updating the jQuery Validation plugin also mean that an integrity testing would be needed for that features.

I look at the jQuery validation plugin, and there is no way actually on how you can specify your own validation message.

PR #8370, opened 2018-04-26
feat(form-valid): Bump jQuery Validation to 1.17.0

nschonni:
Removes duplicate aria-required and adds a few additional validators

Full version diff can be seen here https://github.com/jquery-validation/jquery-validation/compare/1.15.1...1.17.0


duboisp (approved):

RobJohnston:
This would fix half of #8332 because the French translation for notEqualTo is included.

PR #8371, opened 2018-04-26
Bump jQuery to final 1.x and 2.x releases

### LINKED ISSUE #8047: JQuery version used in v4.0.25 contains security issues
Hi,

This v4.0.25 WET Toolkit is still using an old version of JQuery (v2.1.4)
which has security vulnerabilities 
 - specifically Ajax calls which do  not specify the dataType.

So can you tell me when the WET toolkit will be updated to use the most current JQuery version?

Thanks

Nitin Mistry
Bell Canada

nschonni:
A markup change might be required for those still supporting IE < 9 and those wanting to opt into the newer 2.x release

Closes #8047

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -310,7 +310,7 @@ module.exports = (grunt) ->
 		coreDist: "dist/wet-boew"
 		themeDist: "dist/theme-wet-boew"
 		jqueryVersion: "<%= pkg.dependencies.jquery %>"
-		jqueryOldIEVersion: "1.11.1"
+		jqueryOldIEVersion: "1.12.4"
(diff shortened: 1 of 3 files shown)


duboisp (approved):

duboisp:
We don't support IE < 9, so I don't think that would be an issue

nschonni:
Actually, I guess it's the case for the newer browser markup to since we generate out the link to jQuery in our templates
```html
<script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js"></script>
```
For implementers, they can do the update if they have the capacity, but since this was always a CDN link rather than a local include.
Might be something to add to the release notes about both

PR #8377, opened 2018-04-27
Form validation - Fix testReady major issue

### LINKED ISSUE #8376: Issues with Form validation example page
Is something wrong with Form validation example page?
[http://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-en.html](url)
Not sure if new version of WET 4 handles error messages differently but something is not quite right. Only required fields are being validated. Specialized validation (formats and data rules) are ignored and not validated.

duboisp:
fix #8376 

After this PR will be merge, I am going to release WET 4.0.28.1

### DIFF
--- a/src/plugins/formvalid/formvalid.js
+++ b/src/plugins/formvalid/formvalid.js
@@ -72,7 +72,7 @@ var componentName = "wb-frmvld",
 					"site!deps/additional-methods" + modeJS
 				],
 				testReady: function() {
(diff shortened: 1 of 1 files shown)


EricDunsworth (approved):
Built it locally and compared it against http://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-en.html. Seems to have fully resolved the issue from what I can tell.

PR #8380, opened 2018-04-27
Feeds: Allows correct xhtml feeds title parsing

### LINKED ISSUE #7976: Feeds widget does not respect Atom feed standard.
This is a followup to #7450, which improved it in certain cases but does not work for all.  It still displays as [object Object] for certain feeds.

In particular I'm having problem with the following.

As you can see in the following image, this code assumes that if title is an object, _content_ is a direct child of _title_
```
 if ( typeof( title ) === "object" && title.content ) {
	title = title.content;
}
```
image

Please see this section of the Atom standard: https://tools.ietf.org/html/rfc4287#section-3.1.1.3

Here is an excerpt: 

```
   <summary type="xhtml">
      <div xmlns="http://www.w3.org/1999/xhtml">
         This is <b>XHTML</b> content.
      </div>
   </summary>
```

As you can see _content_ can be under a div or other element type.

As a real life example of the broken behavior here is a feed used at StatCan: http://www.statcan.gc.ca/rss/dai-quo/10000-eng.atom

ptsimard:
Fixes #7976

Permits the following format:
```
<title type="xhtml">
      <div xmlns="http://www.w3.org/1999/xhtml">
         This is <b>XHTML</b> content.
      </div>
   </title>
```

### DIFF
--- a/src/plugins/feeds/feeds.js
+++ b/src/plugins/feeds/feeds.js
@@ -89,8 +89,12 @@ var componentName = "wb-feeds",
 		generic: function( data ) {
 			var title = data.title;
 
-			if ( typeof( title ) === "object" && title.content ) {
-				title = title.content;
+			if ( typeof( title ) === "object" ) {
+				if ( title.content ) {
+					title = title.content;
+				} else if ( title.type === "xhtml" && title.div ) {
+					title = title.div.content;
+				}
 			}
 			return "<li><a href='" + data.link + "'>" + title + "</a><br />" +
 				( data.publishedDate !== "" ? " <small class='feeds-date'><time>" +


ptsimard:
@LaurentGoderre could you take a look? Thanks

duboisp:
@ptsimard are you able to give us an URL to an ATOM feed that use that format?

Items to be considered for review:
* Is using a ```<div>``` elements as a container is mandatory according to ATOM feed spec? Can it be another XHTML element?
* How the ATOM feed spec define the ```type``` attribute?
* Is the ```type``` attribute are allowed for the ```<title>``` XML elements

ptsimard:
Here is a feed in this format to test: http://www.statcan.gc.ca/rss/dai-quo/10000-eng.atom

The following document outlines the atom format:

Type attribute: https://tools.ietf.org/html/rfc4287#section-3.1.1

> Text constructs MAY have a "type" attribute.  When present, the value
   MUST be one of **"text", "html", or "xhtml"**.  If the "type" attribute
   is not provided, Atom Processors MUST behave as though it were
   present with a value of "text".

This section is about the xhtml title format: https://tools.ietf.org/html/rfc4287#section-3.1.1.3

Referring to the **div** element, emphasis mine:

> **If the value of "type" is "xhtml", the content of the Text construct
   MUST be a single XHTML div element [XHTML] and SHOULD be suitable for
   handling as XHTML**.  The XHTML div element itself MUST NOT be
   considered part of the content.  Atom Processors that display the
   content MAY use the markup to aid in displaying it.  The escaped
   versions of characters such as "&" and ">" represent those
   characters, not markup.

duboisp (approved):

PR #8382, opened 2018-04-30
build(bootlint): Bump grunt-bootlinto to 0.10.2 and fix errors

nschonni:
.btn class isn't allowed on paragraph tags.
Doesn't really make sense but this might be a breaking change

### DIFF
--- a/site/includes/sign-on-off.hbs
+++ b/site/includes/sign-on-off.hbs
@@ -7,7 +7,7 @@
 				<button class="btn btn-default" type="button">{{i18n "tmpl-register"}}</button>
 				<button class="btn btn-primary" type="button">{{i18n "tmpl-signin"}}</button>
 	{{else}}
-				<p class="btn">{{i18n "tmpl-signed-in-as"}} <span class="wb-so-uname">John Doe</span></p>
+				<p>{{i18n "tmpl-signed-in-as"}} <span class="wb-so-uname">John Doe</span></p>
 				<button class="btn btn-default" type="button">{{i18n "tmpl-accountsettings"}}</button>
 				<button class="btn btn-primary" type="button">{{i18n "tmpl-signout"}}</button>
 	{{/is}}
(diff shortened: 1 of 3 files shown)


site/includes/sign-on-off.hbs
```diff
@@ -7,7 +7,7 @@
 				<button class="btn btn-default" type="button">{{i18n "tmpl-register"}}</button>
 				<button class="btn btn-primary" type="button">{{i18n "tmpl-signin"}}</button>
 	{{else}}
-				<p class="btn">{{i18n "tmpl-signed-in-as"}} <span class="wb-so-uname">John Doe</span></p>
+				<p>{{i18n "tmpl-signed-in-as"}} <span class="wb-so-uname">John Doe</span></p>
```
EricDunsworth: From what I can tell on http://wet-boew.github.io/v4.0-ci/theme/content-signedon-en.html, the main purpose of that misused ``btn`` class seems to be to set the "Signed in as John Doe" paragraph to ``display: inline-block`` in order to make it appear on the same line as the "Account settings"/"Sign out" buttons. Removing ``btn`` causes the paragraph to take up the full line and the real buttons to drop below it.

From what I can tell, replacing ``btn`` with ``visible-xs-inline-block visible-sm-inline-block visible-md-inline-block visible-lg-inline-block visible-print-inline-block mrgn-rght-md`` would produce a very similar layout without failing bootlint validation. Unfortunately it doesn't look like there's a single class available in WET 4.0 or Bootstrap 3 that can set ``display: inline-block`` across all viewports in one shot.

duboisp: @EricDunsworth can you submit that change and we will merge this PR.

This was also added in the release note.

EricDunsworth:
Just pushed an extra commit into this PR's branch. It turns out the combination of classes I proposed in my line comment didn't work properly. Went back to the drawing board and found a much simpler way of making it work.

duboisp (approved):

Issue #7146, opened 2015-10-15
Menu: When top level menus have no menu items what should happen when  these menus get and loose focus?.


dfarough:
Currently we are implementing WET 4 in one of our applications.

several of our menus have no menu items contained within the menu (IE no sub menus ). 

I am noticing that when arrowing across   these menus, that when focus moves to a menu which does contain menu items, that the menu items within that menu are displayed as expected, however if arrowing across to the next menu which does not contain any items, that the previously opened menu is still displayed on screen even though the focus has moved away from that top level menus structure.  I am not sure which version of wet 4 we are using, but the WET examples all seem to contain menus with menu items  so I can not determine if this issue occurs in the current version.

Another thing I am noticing is that if I move my mouse over one of these menus, that the colour changes but when I use tab to move focus there, that the effect is barely noticeable.

Do we have examples of different types of menus? or examples of mixed menus with some main menus with no menu items ETC?

andre-gh:
Which browser and WET version are you using?

dfarough:
I am using Internet explorer 10 and Firefox 40.0.2

I am not sure which version of WET we are using.  I know that it is version 4 but cannot be more specific at present.

andre-gh:
If you are able and know how, you can open the network panel in firefox and view the copy of wet-bow.min.js downloaded and the version number is in a comment at the top.  Are you able to do that?

dfarough:
We are using the most recent stable version.
4.0.18
Marc-André Dion (10:38:36 AM): The latest stable I believe.

dfarough:
@wet-boew/access-working-group 

I want to further describe what I am experiencing with version 4.0.18

our application menus include the following top level menus:
Home
job search
My account
Search archives 
help

Only the my account menu contains other menu items.

when I tab to and reach the home menu, the visible focus does not change to any great degree.  If you look carefully you will see a slight change to the outline of the menu.  However if you move the mouse to this menu, the background and forground colours changes much as it does when a menu with options gets focus.

If I arrow over to the my account menu, the menu expands and all items within the menu are displayed as they should be.  However if I use the right arrow to move the focus to the search archive menu, I do then focus on the search archive menu as expected, but the options within the my account menu will continue to be  displayed as if the accounts menu still had focus.

This will be confusing for users since if they do not realize that their focus has moved, they will not be able to use the up or down arrow keys to move through the empty menu with focus.

Can we add some additional menu samples which illustrate how wet 4 handles some of these possibilities?

[...]

PR #8417, opened 2018-06-28
Menu: Focus styling on menu items with no submenu

juleskuehn:
Resolves Issue #7146 by adding "active" class to focused menu items even if they don't have a submenu.
Adds menu item with no submenu to example menu.

### DIFF
--- a/src/plugins/menu/ajax/menu-include-en.html
+++ b/src/plugins/menu/ajax/menu-include-en.html
@@ -70,7 +70,7 @@
 					</li>
 				</ul>
 			</li>
-			<li><a href="#lorc" class="item">No Submenu</a></li>
+			<li><a href="#" class="item">No Submenu</a></li>
 		</ul>
 	</div>
 </div>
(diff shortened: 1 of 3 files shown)


src/plugins/menu/ajax/menu-include-en.html
```diff
@@ -70,6 +70,7 @@ <h2>Topics menu</h2>
 					</li>
 				</ul>
 			</li>
+			<li><a href="#" class="item">No Submenu</a></li>
```
duboisp: Can you replicate this new menu item in the french ajax file (menu-include-fr.html)

duboisp (changes requested):

src/plugins/menu/ajax/menu-include-en.html
juleskuehn: I have made the text "Sans Sous-Menu" on the french page. If this is not appropriate, please change.

duboisp:
Pre-approved at the technical review meeting if pass CI test.

duboisp (approved):

Issue #8418, opened 2018-06-28
Menu: Nested submenu (details/summary) has strange behaviour using enter key


juleskuehn:
To replicate:
http://wet-boew.github.io/wet-boew/demos/menu/menu-en.html

Open the "Section 3" submenu. The tree submenu, implemented with details/summary behaves as expected using the arrow keys. However, try pressing Enter to expand a submenu. It doesn't work, but does have some visual effect. Press the down key to navigate to focus another submenu. Press Enter again. The _previously_ selected submenu expands!

One solution for this is to disable the enter key within the submenus. Another is to force its behaviour to be identical to the spacebar - behaving like an accordion menu.

juleskuehn:
The spacebar actually expands the submenu but does not focus the first element - meaning you have to use the arrows to enter that submenu even after pressing space. So I think there are two options to make the behaviour consistent:
1. Enter and space behave the same as right arrow.
or
2. Disable enter and space.

duboisp:
What the ARIA Practice 1.1 keyboard interaction model said? https://www.w3.org/TR/wai-aria-practices-1.1/#menu

duboisp:
> Enter:
> * When focus is on a menuitem that has a submenu, opens the submenu and places focus on its first item.
> * Otherwise, activates the item and closes the menu.
>
> Space:
> * (Optional): When focus is on a menuitemcheckbox, changes the state without closing the menu.
> * (Optional): When focus is on a menuitemradio that is not checked, without closing the menu, checks the focused menuitemradio and unchecks any other checked menuitemradio element in the same group.
> * (Optional): When focus is on a menuitem that has a submenu, opens the submenu and places focus on its first item.
> * (Optional): When focus is on a menuitem that does not have a submenu, activates the menuitem and closes the menu.

juleskuehn:
Yes, I was referring to this spec, and that is the exact behaviour I have implemented.

But, the enter and spacebar do not work in detail/summary (nested) submenus in Internet Explorer, and I don't think there is anything I can do about that.

thekodester:
I don't think there much that can be done with Internet Explorer without some messy JS hacks. Considering the fact Microsoft has given up on IE I think it time to just let it die.

duboisp:
> The enter and spacebar do not work in detail/summary (nested) submenus in Internet Explorer

[...]

PR #8424, opened 2018-07-04
Menu: Fix enter and spacebar behaviour, focus by first letter, close nested submenus

juleskuehn:
Thoroughly tested in Chrome 67, Firefox 61, IE11.
Fixes Issue #8418 and a number of undocumented issues.

Fixes to behaviour:
- Spacebar and enter now behave as in ARIA recommendation (except in IE11 detail/summary).
 - As with right arrow key, enter or spacebar expands a submenu and focuses its first item.
 - If item has no submenu (ie. is a link), activates the item by emulating click.
- Nested submenus close on blur (except in mobile menu).
- Nested submenus and their items can be selected by letter from a focused top-level menuitem.

Code style improvements:
- symbolic variables for keycodes
- remove unneeded conditionals

Unresolved issues:
- Mobile menus do not collapse on blur.
- IE11 does not allow enter and spacebar to expand detail/summary with correct behaviour.

### DIFF
--- a/src/plugins/menu/menu.js
+++ b/src/plugins/menu/menu.js
@@ -421,7 +421,16 @@ var componentName = "wb-menu",
 				.attr( {
 					"aria-hidden": "true",
 					"aria-expanded": "false"
-				} );
+				} )
+
+				// Close nested submenus
+				.find( "details" )
+					.removeAttr( "open" )
+					.children( "ul" )
+						.attr( {
+							"aria-hidden": "true",
+							"aria-expanded": "false"
+						} );
 
 		if ( removeActive ) {
 			$elm.removeClass( "active" );
--- a/src/plugins/menu/menu.js
+++ b/src/plugins/menu/menu.js
@@ -676,7 +676,7 @@ $document.on( "keydown", selector + " [role=menuitem]", function( event ) {
 				);
 
 			// Enter sub-menu
-			} else if ( hasPopup && ( which === ENTER_KC || which === UP_KC || which === DOWN_KC ) ) {
+			} else if ( hasPopup && ( which === ENTER_KC || which === SPACE_KC || which === UP_KC || which === DOWN_KC ) ) {
 				event.preventDefault();
 				$parent = $menuItem.parent();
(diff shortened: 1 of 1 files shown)


duboisp (changes requested):
@juleskuehn, your PR seems to include unrelated files change, I will help you to cleaning up at the next code sprint.

juleskuehn:
I was unable to resolve the issue in IE11 without putting a check for role="menuitem" in the details.js polyfill. Looking forward to discussing this with you at the code sprint.

juleskuehn:
I was able to resolve the issue in IE11 by putting a check for role "menuitem" in the details.js polyfill.

`else if ( ( which === 13 || which === 32 ) && currentTarget.getAttribute( "role" ) !== "menuitem" )`

Since the polyfill is used in other plugins, I tested the other working examples with my changes in place:

- [x] Charts (all)
- [x] Collapsible alerts
- [x] Details/summary polyfill
- [x] Tabbed interface
- [x] Tabbed Interface - Carousel
- [x] Toggle
- [x] Multimedia Player - Video

src/polyfills/details/details.js
```diff
@@ -107,7 +107,9 @@ $document.on( "click keydown toggle." + componentName, selector, function( event
 		( event.type === "toggle" && event.namespace === componentName ) ) ) {
 		details = currentTarget.parentNode;
 		details.open = !details.open;
-	} else if ( which === 13 || which === 32 ) {
+
+	// Prevent double handling of enter / spacebar if within a menu (handled by menu.js)
+	} else if ( ( which === 13 || which === 32 ) && currentTarget.getAttribute( "role" ) !== "menuitem" ) {
```
nschonni: Is there another way to handle this? The polyfill shouldn't really care about other WET plug-ins

juleskuehn: I was able to fix it by adding `event.stopImmediatePropagation()` in menu.js. This depends on the menu.js handler being called before the details.js handler, which it is. Is this an acceptable solution?

duboisp (approved):

PR #8448, opened 2018-07-24
EditorConfig: Added global JSON/KML/CSV overrides.

EricDunsworth:
* Added JSON and KML as formats that should use a 2 space indentation style. Reasoning:
  * JSON: The vast majority of the repo's tracked JSON files are already using 2 spaces. Only "site/data/site.json" is currently using tabs.
  * KML: Its official documentation doesn't define a specific indentation style, but seems to unofficially use 2 spaces. The Geomap plugin's KML demo file ("src/plugins/geomap/demo/*.kml") is already using 2 spaces.
* Disabled whitespace trimming in CSV files. In WET's i18n CSV file, some lines end with strings that intentionally finish off with whitespace characters.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,6 +10,9 @@ insert_final_newline = true
 indent_style = tabs
 indent_size = 4
 
-[{package.json,bower.json,*.md}]
+[{*.json,*kml,*.md}]
 indent_style = space
 indent_size = 2
+
+[*.csv]
+trim_trailing_whitespace = false


.editorconfig
```diff
@@ -10,6 +10,9 @@ insert_final_newline = true
 indent_style = tabs
 indent_size = 4
 
-[{package.json,bower.json,*.md}]
+[*.csv]
```
nschonni: Minor, but I might suggest throwing this at the bottom to minimize the diff

.editorconfig
```diff
@@ -10,6 +10,9 @@ insert_final_newline = true
...
+trim_trailing_whitespace = false
+
+[{bower.json,package.json,package-lock.json,*kml,*.md}]
```
nschonni: We might be able to just say `*.json` unless there is some we want to keep. I don't mind the tabs in JSON since it's smaller on the wire.

nschonni (approved):

EricDunsworth:
@nschonni Regarding this:
> We might be able to just say ``*.json`` unless there is some we want to keep. I don't mind the tabs in JSON since it's smaller on the wire.

I didn't do anything with regards to that in the commit I force-pushed earlier, but...

What would your thoughts be on removing package.json and package-lock.json from .editorconfig? Only bower.json really needs 2 space indentation these days.

According to https://github.com/npm/npm/issues/5694#issuecomment-307142476, npm5 is more accommodating of alternate indentation styles in its .json files. Can't say the same about bower though (see bower/bower#1338).

FWIW, here's a quick rundown of which tabbing styles this repo's tracked .json files are currently using:
* 2 space indentation:
  * bower.json
  * package.json
  * package-lock.json
  * src/plugins/geomap/demo/sample.json
  * src/plugins/tables/ajax/datasource.json
* Tab indentation:
  * site/data/site.json

I guess 2 spaces wins out bigtime in pre-existing .json files. But there isn't really any good reason as to why the vast majority of them shouldn't be using tabs...

nschonni:
My vote would just to change the `site.json` to tabs if that is what the rest are using :)

EricDunsworth:
@nschonni I'm going to assume you meant the opposite :P lol.

I'll update this PR to add ``*.json`` to the 2 spaces override.

duboisp (approved):

PR #8450, opened 2018-08-02
EditorConfig: Added .*/sitemenureplace/unset overrides and removed KML one.

EricDunsworth:
* Added dot-prefixed files to 2 space indentation formats. Most already use 2 spaces. The few that aren't appear to be structured as JSON data.
* Removed KML from 2 space indentation formats. It turns out that the Geomap plugin's KML demo file doesn't only use 2 space indentation... It uses a mix of tabs and spaces (with tabs being much more prominent).
* Added a no final newline override for the "sitemenureplace" include. It's an HBS file that only contains a data attribute for use in a parent include's nav element.
* Added unset overrides for tracked third party JavaScript files. This should prevent any other overrides from affecting them if they're updated in the future.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,9 +10,20 @@ insert_final_newline = true
 indent_style = tab
 indent_size = 4
 
-[{*.json,*kml,*.md}]
+[{.*,*.json,*.md}]
 indent_style = space
 indent_size = 2
 
+[site/includes/sitemenureplace.hbs]
+insert_final_newline = false
+
 [*.csv]
 trim_trailing_whitespace = false
+
+# Tracked third party files
+[{dep/modernizr-custom.js,src/polyfills/events/mobile.js,src/polyfills/slider/slider.js}]
+charset = unset
+trim_trailing_whitespace = unset
+insert_final_newline = unset
+indent_style = unset
+indent_size = unset


.editorconfig
```diff
@@ -10,9 +10,14 @@ insert_final_newline = true
 indent_style = tab
 indent_size = 4
 
-[{*.json,*kml,*.md}]
+[{.*,*.json,*.md}]
 indent_style = space
 indent_size = 2
 
-[*.csv]
+[site/includes/sitemenureplace.hbs]
```
EricDunsworth: @nschonni Would it be a better practice to only specify the file's name (instead of its full path) in this kind of situation?

The repo doesn't contain any other files with the same names as the ones I'm targeting here and below.

nschonni: I think the indentation should be fixed in the file rather than an exception. The file https://github.com/wet-boew/wet-boew/blob/f5f33e0240649fe22fd12f28978113841762bffa/site/includes/sitemenu.hbs#L1 just needs to have that leading space instead

.editorconfig
```diff
@@ -10,9 +10,14 @@ insert_final_newline = true
...
+insert_final_newline = false
+indent_style = space
+indent_size = 1
+
+[src/i18n/i18n.csv]
```
nschonni: Think it makes sense to keep the general setting for `*.csv` since i guess it's true that trailing space can be significant in CSV files

.editorconfig
EricDunsworth: Ok. I'll remove the indent rules from that part (although the newline override will remain).

As far as modifying sitemenu.hbs goes... I'll do it in an incoming PR (whose goal is to bring the entire repo in line with this PR's .editorconfig rules). Should have it out later today or tomorrow.

.editorconfig
EricDunsworth: In that case - I'll revert this part back to its previous state.

FWIW, the second CSV file contains quotes around all of its strings and doesn't currently contain any trailing spaces.

EricDunsworth:
Just revised this PR to specifically exempt some of the third party JS files discussed in #8451.

Sorry for the mess of edits/force-pushes... :(.

.editorconfig
```diff
@@ -10,9 +10,19 @@ insert_final_newline = true
 indent_style = tab
 indent_size = 4
 
-[{*.json,*kml,*.md}]
+[{.*,*.json,*.md}]
 indent_style = space
 indent_size = 2
 
+[site/includes/sitemenureplace.hbs]
```
nschonni: I wonder if this should just be inlined again

.editorconfig
```diff
@@ -10,9 +10,19 @@ insert_final_newline = true
...
+insert_final_newline = false
+
 [*.csv]
 trim_trailing_whitespace = false
+
+[{dep/modernizr-custom.js,src/polyfills/events/mobile.js,src/polyfills/slider/slider.js}]
```
nschonni: Maybe just add a comment above that these are 3rd party scripts

nschonni (approved):

.editorconfig
EricDunsworth: You mean like by moving all of that file's contents into sitemenu.hbs?

.editorconfig
EricDunsworth: Will do :)!

.editorconfig
nschonni: Yeah since it's a single line that is used in a single file. Maybe there was a reason for extracting it though

EricDunsworth: Looks like @LaurentGoderre made that change in #7028.

Prior to that, ``site/layouts/default.hbs`` used to contain ``site/includes/sitemenureplace.hbs``' contents (as well as tons of other stuff that PR split into includes).

I presume ``sitemenureplace.hbs`` was just created to make ``sitemenu.hbs`` look less bloated. GCWeb doesn't reference nor override it.

I'm ok with either way. Although given that some other includes (like ``site/includes/pagetitle.hbs``) could be candidates for a similar rule in the future and since I'm juggling with 3 PRs atm, the status quo might be a better bet for the time being.

Issue #8357, opened 2018-04-19
Seek comments - Basic HTML mode and progressive enhancement


duboisp:
Latest version: https://wet-boew.github.io/wet-boew-documentation/decision/draft-2018-4.html

Your thoughts?

## Issue

There is no clear explanation of what the basic HTML mode should do and/or should behave. The progressive enhancement design principle are not or can not be always followed when using some WET feature.

A clear definition is needed for WET 5 and to solve various issue/question about basic HTML mode.

## History

Historically, the basic HTML mode was assumed to be the version equivalent of when the browser deactivate any javascript. WET4 currently apply that approach per his design because it would prevent WET feature to initiate. The funny fact is WET 4 rely on running some javascript code in order to be in Basic HTML mode.

WET 4 was kind of founded on the principle of progressive enhancement where author should create web pages that are in his basic form and then the web experience toolbox enhance the content into an interactive interface which have an high probability to meet WCAG 2.0 Level AA requirement. The documentation around WET feature do not clearly define what their basic HTML form should be. Even some WET feature became broken in the basic HTML mode.

## Rational

Conforming to the latest version of WCAG Level AA is the core driver for this project. WCAG 2.0 Level AA do not means to strictly following the progressive enhancement design and it do not require to check the conformity based on the principle that javascript are not executed. 

Some browser vendor do not have an easy way to prevent the execution of javacript. Like in Edge it is not possible to deactivate javascript via updating it's standard configuration GUI.

As per WCAG, the conformance check is evaluated only after the page are ready to be navigated and are in a state where the user can interact with.

The progressive enhancement design approach was used to ensure that WET feature would have a nice fall-back in the case the WET feature are broken. Usually that fall-back would provide a simpler and linear interface.

WET are designed and developed based on the capability of the supported browser.

Clarifying the interpretation of what means "Basic HTML" are going to support the design of the next major release of WET

## Proposition of interpretation

"Basic HTML" is a simpler and linear interface of a WET feature. It require to have the page content going be initialized in a basic HTML form. It may use some javascript and its usage are going to be limited to execute simple low impact task as per design of the Basic HTML mode.

When a WET feature are initialized in its standard form or any other pre-enhanced state and the page is requested to be in basic HTML mode one of the following should happen:
* The initial coded enhanced interface are transformed into its basic HTML mode.
* The initial coded enhanced interface remain but it do not interfere or prevent the user to consult all the content.
[...]

PR #8452, opened 2018-08-07
Fix #8419 Data Ajax: Not working in Basic HTML mode

### LINKED ISSUE #8419: Data Ajax: Not working in Basic HTML mode
When you turn on basic HTML mode, ajax content do not get pulled in.

http://wet-boew.github.io/wet-boew/demos/data-ajax/data-ajax-en.html?wbdisable=true

StdGit:
This fix allows the plugin Data-Ajax to run in Basic Html mode. I didn't implement a special init event.

### DIFF
--- a/src/core/wb.js
+++ b/src/core/wb.js
@@ -273,7 +273,7 @@ var getUrlParts = function( url ) {
 				i;
 
 			// Lets ensure we are not running if things are disabled
-			if ( wb.isDisabled && selector !== "#wb-tphp" ) {
+			if ( wb.isDisabled && ( selector !== "#wb-tphp" && $.inArray( selector, [ "[data-ajax-after]", "[data-ajax-append]", "[data-ajax-before]", "[data-ajax-prepend]", "[data-ajax-replace]", "[data-wb-ajax]" ] ) === -1 ) ) {
 				return 0;
 			}
 


nschonni:
I don't think the Ajax should be running in Basic http://wet-boew.github.io/wet-boew-documentation/decision/4.html

StdGit:
@nschonni  Do you see an impact with some technologies ?

nschonni:
> Support for basic HTML interface are kept in scope of what browser is supported by the WET core project.
> e.g.
> the tab widget becomes headings with the content of the panels
the mega menu becomes headings, links in a list

Basic HTML support shouldn't require any AJAX'ing in content. It's not a no-script view, but doesn't need anything beyond the regular HTML polifills and fallbacks

StdGit:
In this case, I think we should wait for the return of @shawnthompson to see what the issues are if we don't implement this fix.

N.B. We should keep in mind that we are doing all this for the users. Sometimes decisions are not easy to make...

duboisp:
> doesn't need anything beyond the regular HTML polifills and fallbacks

@nschonni do you have a reference or it is just an assumption?

I don't see an issue of running the data ajax plugin in Basic HTML mode. But the code currently in the PR should be modified and a "special" event should be triggered. Something like "wb-init-basic". And the ```wb``` object should contain a property like "mode" set to "basic".

duboisp (changes requested):
Use a new event to trigger the basic HTML mode. So the "basic html version" will be defined inside the plugin.

nschonni:
https://github.com/wet-boew/wet-boew/issues/8357#issuecomment-391024032
I think this is a change for v5 since this may have other unintended affects on live sites by changing how `wbdisabled` works now

duboisp:
@nschonni what kind of "unintended affects" could happen if the data-ajax is enabled in basic HTML mode? Do you have an example/link?

shawnthompson:
That's right, doing this will now make the menu work in **basic HTML**, hmmm. I'll let you folks fight it out.

duboisp:
The menu wont work, instead the full menu will be fully visible.

Issue #8091, opened 2017-07-18
Form validation: required, verbos?


shawnthompson:
The required attributes in the validation form seem to be a bit much. I'm wondering if we could eliminate a few of them.

## Source code

```html
<div class="form-group">
	<label for="fname1" class="required">
		<span class="field-name">First name</span> <strong class="required">(required)</strong>
	</label>
	<input class="form-control" id="fname1" name="fname1" type="text" required="required" pattern=".{2,}" data-rule-minlength="2" />
</div>
```
## Code after JS injection
``` html
<div class="form-group">
	<label for="fname1" class="required" aria-required="true">
		<span class="field-name">First name</span> <strong class="required" aria-required="true">(required)</strong>
	</label>
	<input class="form-control" id="fname1" name="fname1" required="required" data-rule-minlength="2" aria-required="true" type="text">
</div>
```

My suggestions would be to remove the `aria-required` attributes and only leave the HTML `required` attribute because, from my understanding, most devices support it.

I also added an `aria-hidden=true` attribute to the `<strong>` element in the label to stop screen readers from reading "required" as they announce it because of the `required` attribute anyways. @pbossley mentioned that this might need a `<span>`, he didn't think that `aria-hidden` can be put on a `<strong>`

## Suggestion after JS injection

```html
<div class="form-group">
	<label for="fname1" class="required">
		<span class="field-name">First name</span> <strong class="required" aria-hidden="true">(required)</strong>
	</label>
	<input class="form-control" id="fname1" name="fname1" required="required" data-rule-minlength="2" type="text">
</div>
```

I've created a working example of [suggested change](https://shawnthompson.github.io/wet-boew-aria-required/demos/formvalid/formvalid-en.html). This page was done with just the HTML markup required (no pun intended) and not the WET JS funtionality of the form validation. Which mean the form might not validate.

You can compare it to the WET working example of the [form validation page](http://wet-boew.github.io/v4.0-ci/unmin/demos/formvalid/formvalid-en.html)

pbossley:
It turns out that Safari desktop hasn't quite added support for required, so maybe leaving aria-required=true  makes sense. 
As I suspected, aria-hidden=true doesn't seem to work on a `<strong>` tag as it isn't a valid container. I recommend trying a span and seeing if that works.

shawnthompson:
Safari, really? And I thought Apple was all about the accessibility. 

Was there a need for `aria-required` in the `<label>`?

I made the change to my example 

[...]

PR #8456, opened 2018-08-10
breaking: Stop using `.required` for validation

nschonni:
jQuery validation removed support for these CSS based rules but we had our own hack that kept doing it.

Related to #8091


nschonni:
Changed in #9703, but there is still one chunk of code that seems to reference the old `.required` class `strong.required:not([aria-hidden='true'])`.
Closing this one though

PR #8462, opened 2018-08-16
.eslintignore: Removed out of scope files.

EricDunsworth:
* src/plugins/deps/tableparser.js isn't a third party script. #8461 fixes its ESLint issues.
* src/plugins/geomap/data/config-*.js don't exist in any capacity (not even as untracked files).
* Didn't add dep/modernizr-custom.js since ESLint doesn't currently check the "dep" folder.
* Related to #8451's discussion.

**PS:**
This PR will initially fail Travis-CI's build. To make it pass, #8461 should be merged in first, followed by a manual restart of this PR's build.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,2 @@
 src/polyfills/slider/slider.js
 src/polyfills/events/mobile.js
-src/plugins/deps/tableparser.js
-src/plugins/geomap/data/config-en.js
-src/plugins/geomap/data/config-fr.js


nschonni:
Might be better to just combine this as a separate commit to #8641

EricDunsworth:
Even though they kind of tie into each other, I didn't feel like they were intertwined-enough to merit combining into the same PR. I'd prefer keeping them separate this time around.

duboisp (approved):
Pre-approved, this PR will be merged after the release of WET 4.0.29

PR #8464, opened 2018-08-21
WIP - Content Filter - initial commit

duboisp:
See Research 2018-7 in the wet-boew project documentation for more information

### DIFF
--- a/site/pages/docs/ref/contentfilter/contentfilter-en.hbs
+++ b/site/pages/docs/ref/contentfilter/contentfilter-en.hbs
@@ -1,0 +1,90 @@
+---
+{
+	"title": "Content filter",
+	"language": "en",
+	"category": "Plugins",
+	"description": "Filters through content and show/hide content that match a certain tags.",
+	"tag": "filter",
+	"parentdir": "contentfilter",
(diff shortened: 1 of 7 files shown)


RobJohnston (commented):
Hello Pierre:  I hope you find this more useful than annoying.

duboisp:
@Garneauma What in this PR has been added in the tagFilter provisional plugin vs what feature it still need to be added.

Garneauma:
@duboisp I think the Tagfilter plugin eventually got further and further away from this PR as the use case changed. The Tagfilter plugin does not feature a "negative" select for tags, i.e. "!tag", and I don't think it ever will. Same for the "exclusive" select, i.e. "*tag". Also, the Tagfilter plugin does not feature a way to still keep the titles visible when items within its section are hidden, and I don't think we will add that feature.

Basically, I believe the Tagfilter eventually got to be its own plugin which pulled ideas from the Content Filter rather than being an updated version of it.

Garneauma:
Converting this to a draft PR as it will require some more work before it can be merged, if we ever decide it is worth it.

PR #8504, opened 2018-10-30
Release notes: Fixed issues in 4.0.28-4.0.29 pages.

### LINKED ISSUE #8503: v4.0.29 French release notes not updated
The [v4.0.29 - Notes de version](http://wet-boew.github.io/v4.0-ci/docs/versions/v4.0/v4.0.29-fr.html) were not updated

EricDunsworth:
* All
  * Added section elements around H3 headings and their associated content.
* 4.0.29:
  * Fixed GCWeb 4.0.29 commit links.
  * Removed leftover "4.0.28" heading from French version. Fixes #8503.
  * Added prettify all plugin to French version (for consistency with English variant).

### DIFF
--- a/site/pages/docs/versions/v4.0/v4.0.28-en.hbs
+++ b/site/pages/docs/versions/v4.0/v4.0.28-en.hbs
@@ -4,7 +4,7 @@
 	"language": "en",
 	"description": "v4.0.28.x release notes - Version history - Web Experience Toolkit (WET)",
 	"altLangPrefix": "v4.0.28",
-	"dateModified": "2018-04-27"
+	"dateModified": "2018-10-30"
 }
 ---
 <p>Version 4.0.28.1</p>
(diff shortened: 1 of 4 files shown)


duboisp:
@EricDunsworth Thanks for the PR. There is no need to add those explicit sub-section.

EricDunsworth:
I can remove the ``<section>`` elements if absolutely needed, but don't personally agree with it.

I always wrap every individual heading and their associated content into relevant sectioning elements (including sub-sections) and have done so for years. Many of my work colleagues do this too. WET has also been structuring sectioning elements in this manner for many years (ever since HTML5 page templates were first introduced in WET 2.x). I believe all of GCWeb's page templates also do it.

IMO it would be unwise to suddenly start breaking this pattern in WET 4.0.x maintenance releases. If there's any appetite to change the pattern, I think it'd be better to wait until WET 4.1/5.0 before introducing it.

On another note, I'd personally prefer not using ``<section>`` elements at all over using them only partially with high-level headings. The way I see it, I'd rather use sectioning elements for all their worth rather than only using them selectively.

Thoughts?

duboisp (approved):

PR #8524, opened 2018-12-05
Fix #8517

### LINKED ISSUE #8517: No mobile menu on non-AJAX menu
I'm using 4.0.29 with a non-AJAX menu. The "wb-init.wb-overlay" event isn't being called on the mobile menu because there's no AJAX happening to trigger the function passed to ajaxStop in:
https://github.com/wet-boew/wet-boew/blob/9e85c8b2099ae1de5b4d1c0b331a0a5df4ed6fb7/src/plugins/menu/menu.js#L322-L332

StdGit:


duboisp (changes requested):
Move the new function outside.

duboisp (changes requested):

duboisp (approved):

StdGit:
Keep in mind that this fix use an internal variable jQuery.active
Look for  `if ( $.active > 0 ) {` in the code
https://stackoverflow.com/questions/3148225/jquery-active-function

This means that if jquery decide to remove this variable, this fix will cease to work.

StdGit:
Si quelqu'un a une meilleure solution elle serait appréciée.

@delisma Could you send me your eMail

PR #8519, opened 2018-11-22
Update too DataTable Pagination

ghost:
Update DataTable pagination so only appears if more than one page. Fixes #8516.

PR #8544, opened 2019-01-22
DataTable Pagination Documentation

GormFrank:
Depends on PR #8519.


duboisp (changes requested):
Can you add it's French equivalent working example? Thanks

GormFrank:
> 
> 
> Can you add it's French equivalent working example? Thanks

@duboisp Ouais je ne sais pas pourquoi je n'y avais pas pensé.

duboisp (approved):

Issue #8546, opened 2019-01-25
DataTable filter with SELECT MULTIPLE field


michael-milette:
Has anyone managed to create a search field where users can select more than one item in a select dropdown field?

For example, I want to be able to filter on more than one province at a time.

Best regards,

Michael

duboisp:
Not in WET yet 

I know of some experiments that provide that feature but it is undocumented and not ready for being integrated to wet-boew.

thekodester:
I haven't seen this done with a dropdown, but I seen it with checkboxes. For example here: https://www.jobbank.gc.ca/jobsearch/.

<img width="298" alt="screen shot 2019-01-25 at 5 09 25 pm" src="https://user-images.githubusercontent.com/23484460/51775492-edb0a880-20c3-11e9-833d-f3466e80d7fc.png">

michael-milette:
Thanks @thekodester , I did not have that one.

If anyone is looking, I found several other examples:

- https://www150.statcan.gc.ca/n1/en/subjects/crime_and_justice/victimization
- http://w05.international.gc.ca/projectbrowser-banqueprojets/filter-filtre
- https://www.investcanada.ca/incentives?format=json&limit=10&fpur=1,2&naics=0
- https://www.ic.gc.ca/app/scr/innovation/list-liste/571c8f726c704f4c
- https://open.canada.ca/ckan/en/dataset

Unfortunately none of them seem to use native WET-BOEW framework for this functionality.

Still looking.

Best regards,

Michael

michael-milette:
Checkboxes would work for me too. Doesn't have to be a dropdown.

michael-milette:
Found another one which looks promising: https://www.canada.ca/en/privy-council/campaigns/mandate-tracker-results-canadians.html

thekodester:
I created a request to see if where able to add checkboxes to the filtering.

idekaresearch:
We are also looking for a multi-select filter option, as well as a Date range filter.

To get this working we simply had to roll our own and post back to the server when filtering the results.

thekodester:
@idekaresearch  Checkboxes filters where added (something as a multi-select). Also #8559 talks about expanding on the recent addition of checkboxes. Currently still no date range filter.

thekodester:
I have a working example of using a date range. See https://www.thekodester.ca/wet/tables/date-filter.

To have this apart of WET I could add the following code and there need to be some documentation. 
```
[...]

PR #8552, opened 2019-01-29
DataTable Filtering With Checkboxes

ghost:
For #8546...Allows the filtering with checkboxes using data-tables regex search.

### DIFF
--- a/src/plugins/tables/tables-en.hbs
+++ b/src/plugins/tables/tables-en.hbs
@@ -961,22 +961,22 @@
 						</select>
 					</div>
 					<div class="form-group">
-						<label for="dt_location">Location</label>
-						<select class="form-control" id="dt_location" name="dt_location" data-column="5">
-							<option value="">Anywhere</option>
-							<option value="Manitoba">Manitoba</option>
-							<option value="Canada">Canada</option>
-							<option value="Nova Scotia">Nova Scotia</option>
-							<option value="Alberta">Alberta</option>
-							<option value="Ontario">Ontario</option>
-							<option value="Newfoundland and Labrador">Newfoundland and Labrador</option>
-							<option value="New Brunswick">New Brunswick</option>
-							<option value="Saskatchewan">Saskatchewan</option>
-							<option value="British Columbia">British Columbia</option>
-							<option value="Quebec">Quebec</option>
-							<option value="Yukon Territory">Yukon Territory</option>
-							<option value="Prince Edward Island">Prince Edward Island</option>
-						</select>
+						<fieldset>
+							<legend><span class="field-name">Location</span></legend>
+							<div class="checkbox"><label for="dt_location1"><input type="checkbox" id="dt_location1" name="dt_location1" data-column="5" value=""/> Anywhere</label></div>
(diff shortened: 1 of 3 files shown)


thekodester:
See live demo here: https://www.thekodester.ca/wet/tables/dtfiltering

michael-milette:
@thekodester 

Definitely a +1 from me.

This is great!!! Thank you!

Michael

duboisp:
Thanks @thekodester for your contribution

We would need a E&F working example and the documentation before to merge this.

thekodester:
Did you want me to change the current example to include checkboxes for one of the dropdowns?

thekodester:
I added the update to the example pages.

duboisp (changes requested):
The "clear" don't work properly for checkboxes, you might need to add one JS line here: https://github.com/wet-boew/wet-boew/blob/master/src/plugins/tables/tables.js#L206

michael-milette:
Hi @thekodester,

I would like to try this out with our application. It looks like everything was finished about 10 days ago. Where can I get access to the documentation and final code? I am very excited to give it a try.

Best regards,

Michael

thekodester:
Hey @michael-milette,

Once this gets the final approval from @duboisp it should be merged and available when 4.0.30 come out that the hope anyways.

Thanks,

Kody

duboisp (approved):

michael-milette:
Thank you @duboisp and @thekodester,

It works great! I already have it setup with one of our datatables. Any chance it can be integrated into https://github.com/wet-boew/themes-dist.git soon?

Best regards,

Michael

duboisp:
@michael-milette you means when it will be included in the release of gcweb? I am currently working on that and it will be included in gcweb v5. 

An alternative if you need that fix asap in your gcweb theme implementation, you can simply replace the wet-boew.js file. You can follow the same process as to apply the [v4.0.29.1 patch](https://wet-boew.github.io/wet-boew/docs/versions/v4.0/patch-for-v4.0.21-to-v4.0.29-en.html)

michael-milette:
Thanks @duboisp

michael-milette:
Hi @thekodester ,

We did a little more extensive testing today and I believe we found a bug.

# Description

I have a datatable with two fields: Pillars and Provinces. Checking a box in the first field and none in the second field produces 0 results.

# How to Reproduce the issue

Create a form similar to your working example but with two fields. Both fields should be check boxes. The first will be Pillars and the second Provinces in this example.

- Test 1: Checking boxes in both fields - Success. Produces the expected results.
- Test 2: Checking a box ONLY in the first (Pillars) field - Failure. Produces 0 results.
- Test 3: Checking a box ONLY in the second field - Success. Produces the expected results.
- Test 4: Reverse the order of the fields by physically swapping the order of the HTML: Provinces first, Pillars second. Repeat Test 2. Same results except that, this time, it is still the first field (Provinces) field that fails - produces 0 results.

# What was Expected

Regardless of the order in which the fields appear, behavior should be consistent between the first and second field.  Checking one or more of the checkbox filters in the first field without any in the second should not result in 0 results.

Let me know if you have any questions. I can make a working example available upon request.

Best regards,

Michael

thekodester:
I post your own issue with a working example (or some code your using) as the example on my site has no issues.

duboisp:
@michael-milette Thanks for finding that bug, please open a new issue. 
Any fix for that issue should be submitted via a new pull request.

PR #8569, opened 2019-02-13
Feeds Plugin: Setting Limit

### LINKED ISSUE #8568: Feeds broken when limit is greater than the number of feed entries
When the wb-feeds limit-**#** is greater than the actual number of items in the feed, the plugin throws an error.

codejunkyard:
Fixes the feed limit when it is set to a number greater than the number of feed entries. Also fixes the case when there is only one item in the feed. Fixes #8568


duboisp:
Superseded by PR #8572

duboisp:
Thanks @gvautour for your help

PR #8572, opened 2019-02-14
Fix: When limit is greater than the number of feed entries

ipaksc:


duboisp:
May be a duplicate of #8569

gvautour:
I just wanted to point out that when the feed has only 1 item, the entry object is not an array, this breaks the feeds. There's also the case when there is no set limit, the default is 0, which actually sets a limit to 0 items when it should list all items instead. See #8568

duboisp:
@gvautour Thanks to reporting that edge use case.

@ipaksc Can you add an example in the feed page for those special use case?

duboisp (changes requested):
Hi Ilya,

There is something weird with your commit, when I rebase it remove all your changes. We will need to fix that prior to merge your PR.

duboisp (approved):

PR #7889, opened 2017-02-09
Plugin Geomap: Migration to OpenLayers v3.x

jvanulde:

PR #8578, opened 2019-02-19
Migration to OpenLayers v3.x

duboisp:
* Removed inline styles for interface skeletons.
* Added OpenLayers as a devDependency and pointed it to the NPM package.
* Removed protocol specification on service endpoints to fix mixed content issues.
* Removed Konqueror specific selector from _base.css
* Removed unused commented selectors from _base.css
* Added backward compatibility fix for non-standard WKT
* Updated documentation and added upgrade notes to samples.
* Geomap - Removing sprites
* Geomap - Lintspacing fix
* Geomap - Breaking change notices and add v4.0.x example
* Geomap - Supporting previous v4.0.30 geomap coding
* Geomap - Set the AOI to render by default in details/summary.

Close #7889


duboisp:
@jvanulde Can you help me how we can configure the draw box to show up optionally instead of by default?

Basically, I am expecting to have the first working example of the following page to render similar.
* http://universallabs.org/wet/labs/geomap/demos/geomap/geomap-v4.0.30-en.html
* https://wet-boew.github.io/wet-boew/demos/geomap/geomap-en.html

FYI - I implemented a work around for the previous coding that was used for Static map and OSM basemap. 

---

Working example: http://universallabs.org/wet/labs/geomap/demos/geomap/geomap-en.html
Working example as per the previous markup: http://universallabs.org/wet/labs/geomap/demos/geomap/geomap-v4.0.30-en.html
WET-BOEW build closely matching this PR: http://universallabs.org/wet/labs/geomap.zip

jvanulde:
@duboisp you can remove the `aoi` class from the geomap element and that will remove the draw box feature. It should not be on by default.

duboisp:
Thanks @jvanulde 

I did some additional change, like by default the AOI are going to be render in an expand/collapse but the author can change that for a panel by adding the CSS "aoi-open".

Is it possible for you to review my work to see if it still work as it should?

Here the test link:
* Working example: http://universallabs.org/wet/labs/geomap/demos/geomap/geomap-en.html
* Working example as per the previous markup: http://universallabs.org/wet/labs/geomap/demos/geomap/geomap-v4.0.30-en.html
* WET-BOEW build closely matching this PR: http://universallabs.org/wet/labs/geomap.zip

jvanulde (commented):
I think the default collapsed panel is a useful addition.

PR #8588, opened 2019-02-27
DataTables: Advanced Options

ghost:
Adds advanced options for checkboxes.

kodz.ca/projects/wet-boew/demos/tables/checkbox-filter

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -195,29 +195,73 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 	var $lastColumn = -1, $cbVal = "";
 	$form.find( "[name]" ).each( function() {
 		var $elm = $( this ),
-			$column = parseInt( $elm.attr( "data-column" ), 10 ),
-			$value = "";
+			$value = "",
+			$regex = "",
+			$isAopts = $elm.data( "aopts" ),
+			$column = parseInt( $elm.attr( "data-column" ), 10 );
 
+		// Ignore the advanced options fields
+		if ( $isAopts ) {
+			return;
+		}
+
+		// Filters based on input type
 		if ( $elm.is( "select" ) ) {
 			$value = $elm.find( "option:selected" ).val();
 		} else if ( $elm.is( ":checkbox" ) ) {
+
+			// Verifies if using same checkbox list
 			if ( $column !== $lastColumn || $lastColumn === -1 ) {
 				$cbVal = "";
 			}
 			$lastColumn = $column;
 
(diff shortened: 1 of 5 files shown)


thekodester:
Demo: https://www.thekodester.ca/wet/tables/checkbox-filter

src/plugins/tables/tables.js
```diff
@@ -185,7 +185,9 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 
 	event.preventDefault();
 
-	var $form = $( this ),
+	var $options = $( "input[name='advanced_options']:checked" ).val(),
```
duboisp: ````input[name='advanced_options']`` that seems very specific and not extendable or being configurable.

Aslo it will need to be documented

duboisp (dismissed):

src/plugins/tables/tables.js
thekodester: As stated above these are only for checkboxes? Also they have to be user visible so using data attributes not really an option.

thekodester: Also if you have suggestions provide them.

thekodester:
Reconfigured the code so more configurable and updated documentation.

thekodester:
https://github.com/wet-boew/wet-boew/pull/8592 should be merge first, then this updated to include the changes of that PR.

thekodester:
Updated to include the bug fix.

thekodester:
I made a changes to make the code more customizable, instead of one advanced options for all checkboxes, it can now be set to only affect the results of one checkbox filter.

duboisp:
@GormFrank Can you see if this configuration is aligned with the filter plugin and to do a local test of this new functionality? Thanks

GormFrank (commented):
Food for thoughts here. 
When the "Clear" button is pressed, should the Advanced options be set back to "Any"?
If not, I will approve right away because the rest looks good!

thekodester:
No that make sense I added `$form.find( "input:radio" ).prop( "checked", false );` to clear radio buttons.

GormFrank (approved):

idekaresearch:
What version is the checkbox filter expected to be in?

thekodester:
Checkbox Filters are in v4.0.30. See https://wet-boew.github.io/wet-boew/docs/versions/v4.0/v4.0.30-en.html#v4.0.30.

The advanced options wouldn't be until until v4.0.31. See https://wet-boew.github.io/v4.0-ci/docs/versions/rdmp-en.html for the estimated release date.

PR #8589, opened 2019-02-27
DataTables: Date Range Filter

ghost:
Adds date range filter.

kodz.ca/projects/wet-boew/demos/tables/custom-filter

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -192,7 +192,7 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 	$datatable.search( "" ).columns().search( "" );
 
 	// Lets loop throug all options
-	var $value = "", $lastColumn = -1;
+	var $value = "", $lastColumn = -1, $startDate = "1800-01-01", $endDate = "2100-01-31";
 	$form.find( "[name]" ).each( function() {
 		var $elm = $( this ),
 			$column = parseInt( $elm.attr( "data-column" ), 10 );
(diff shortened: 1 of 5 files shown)


src/plugins/tables/tables.js
```diff
@@ -209,6 +209,26 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 				$value += ( $value.length > 0 ) ? "|" : "";
 				$value += $elm.val();
 			}
+		} else if ( $elm.is( "input[type='date']" ) && $elm.val() ) {
+			if ( $elm.is( "input[name='dateFrom']" ) ) {
```
duboisp: ``` input[name='dateFrom'] ``` that don't see configurable.

It will also need to be documented.

duboisp (changes requested):

src/plugins/tables/tables.js
thekodester: If you have suggestion provide them.

thekodester:
Reconfigured the code so little more configurable and updated documentation.

To use make sure any date element contains `start` or `end` in the name attribute. Examples `name="start_date"` or `name="date_start"`.

thekodester:
I add code to prevent invalid date errors in Safari (IE might be affected from what I read online).

Safari see '2016-07-26 09:07:00' as invalid, but '2016-07-26' as valid.

screen shot 2019-03-01 at 10 11 16 am

thekodester:
This would fix Safari, but I feel like it could be over complicating things:

```
obj = obj.replace(/[0-9]{2}\s[0-9]{2}\:/g, function (e) {return e.replace( /\s/g, "T" );});
objDate = new Date( obj );
```

I we'll close the PR, if anyone has a better idea to work around Safari issues, is free to use the code from this.

Issue #8559, opened 2019-02-08
DataTable: Multiple filters for same field (AND)


michael-milette:
Does anyone know if there is a way to have multiple filters apply to the same field in a datatable?

For example, I have a province field which contains a commas delimited list of one or more provinces. I want to find all of the entries that include **BOTH Ontario AND Quebec**.

I tried the new Checkbox filter and, while it works great and we will be using it, it's filter returns results that include **Ontario OR Quebec**.

I also tried creating two province filters hoping that the second filter fields would filter the results produced by the first field. Unfortunately it only seems to take the second province filter into consideration, ignoring the first province filter.

Is there any way to have more than one filter apply to the same field?

Best regards,

Michael

duboisp:
The datatable third party plugin support regular expression, so yes it feasible. 

In order to do, when applying programmatically the filter on a column, the person would need to provide additional parameter where it activate the regular expression and deactivate the smart search.

See the docs here: 
https://datatables.net/reference/api/column().search()

RobJohnston:
Dot this relate to my issue https://github.com/wet-boew/wet-boew/issues/8500 where I found the search lacking the `regex`, `smart`, and `caseInsen`?

thekodester:
I enable regex for the filters when introducing checkboxes. It be really simple to update it so it not looking for exact string.

thekodester:
Also the regex I enable does not affect the search box. FYI.

thekodester:
Also @michael-milette what do you mean by?

> Is there any way to have more than one filter apply to the same field?

With the current design it searches using this code:

`$datatable.column( parseInt( $elm.attr( "data-column" ), 10 ) ).search( "(" + $value + ")", true ).draw();`

for checkboxes the value be `Ontario|Quebec`, so it searches the field looking for Ontario or Quebec.

Are you looking for it to search with Ontario AND Quebec instead of OR?

thekodester:
Maybe the checkbox filer could be updated in away that allows for **AND** or **OR**...should this be predetermined or up-to the user searching?

michael-milette:
Hi @thekodester ,

Ideally we would like to give the user the option of selecting AND or OR. The checkboxes already provide the OR which is great. I am definitely using this feature. Just looking for a way to provide the AND now. It can be AND or OR like the advanced filter option in excel offers. Looks like this:

[...]

PR #8592, opened 2019-03-01
Checkbox Filtering Bug

ghost:
Minor bug, rare edge case.

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -192,22 +192,24 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 	$datatable.search( "" ).columns().search( "" );
 
 	// Lets loop throug all options
-	var $value = "", $lastColumn = -1;
+	var $lastColumn = -1, $cbVal = "";
 	$form.find( "[name]" ).each( function() {
 		var $elm = $( this ),
-			$column = parseInt( $elm.attr( "data-column" ), 10 );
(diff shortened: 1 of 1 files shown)


thekodester:
I updated the PR, to prevent this or any feature edge case, the checkboxes use the own value variable, and the `$value` gets erased each time.

duboisp:
@thekodester How I can reproduce the bug?

Here a test example where I modified the last one to reproduce the bug but without success. Are you able to provide me what test step I should follow and what I should expect working and not working? 
https://universallabs.org/wet/labs/8592-replicate-issue/unmin/demos/tables/tables-en.html

Thank you

thekodester:
See https://github.com/wet-boew/wet-boew/issues/8559#issuecomment-468747868.

duboisp:
Is it only a bug when the code in #8588 is added to wet-boew?

thekodester:
No.

duboisp:
Are you able to reproduce the bug by using the last table in this test example page: https://universallabs.org/wet/labs/8592-replicate-issue/unmin/demos/tables/tables-en.html ?

thekodester:
It has nothing to do with the YEAR specifically...this order cause the bug:

screen shot 2019-03-07 at 1 02 50 pm

duboisp:
Thank @thekodester 

I was able to reproduce the bug. so it is when there is "select" filter followed by a "checkbox" filter.

duboisp (approved):

Issue #8232, opened 2017-12-08
Multimedia Player


fomag:
When the Multimedia player loads inside a Lightbox using Youtube as source the Overlay seems to stay active on top of the video (the big play icon stays present) and the seek bar doesn't move. I have tested this on all browsers and outside the our CRA network and it appears to happen in all cases. I have also seen cases where the overlay actually works correctly but I have no way to reproduce that. It seems to be random.

Examples:

https://www.canada.ca/en/revenue-agency/news/cra-multimedia-library/charities-video-gallery/charities-giving-series-gifting-receipting.html

Click Services: Are they receiptable? video.

https://www.canada.ca/en/services/health/campaigns/drug-prevention.html

Click the video player in the bottom right corner

duboisp:
Thanks @fomag for reporting the bug. Your issue would need to be tested with the latest version of wet-boew. 

I remarked that Canada.ca are using a development build from wet-boew released on July (* v4.0.26-development - 2017-07-17)

You may want to ask PP to update your pages with the latest version of wet-boew. FYI the version 4.0.27 will be released within few days.

fomag:
I tested this on a local version of 4.0.27-dev and it seems to be fine. I believe that PP will push 4.0.27 to Canada.ca as soon as it’s released so we should be good.
I’ll check it again at that point.

PR #8614, opened 2019-03-22
Multimedia Youtube player - YT event workaround when open in a modal

### LINKED ISSUE #7828: Multimedia Player: youtube video in a modal issue
## Issue description
Displaying a youtube video using the WET multimedia player in a lightbox (uses Magnific Popup plugin) causes an issue with the functionality of the player controls.

- The play image overlay does not fade away and remains over the video when the video starts to play.
- The play button does not switch to a pause image once the video starts to play.
- The timeline counter does not increase as the video plays. It stays at 00:00:00

See http://wb-example.ox.ca as a working example of this issue.

```
<a class="wb-lbx" title="Example of WET player with youtube in a modal" href="#popup1">WET Youtube video in modal</a>
<section class="mfp-hide modal-dialog modal-content overlay-def" id="popup1">
  <header class="modal-header"><h2 class="modal-title">Its broke!</h2></header>
  <div class="modal-body">
    <figure class="wb-mltmd">
      <video title="test youtube vid">
        <source type="video/youtube" src="https://www.youtube.com/watch?v=iT7vWd-nEoU">
      </video>
      <figcaption><p>Notice the controls don't function quite as they should and the play icon doesn't fade out from overtop the video. Counter doesn't count.</p>
      </figcaption>
    </figure>
  </div>
</section>
```

duboisp:
Working example: http://universallabs.org/wet/labs/multimedia-yt/demos/multimedia/multimedia-youtube-en.html

fix #7828, #8232

### DIFF
--- a/src/plugins/multimedia/multimedia.js
+++ b/src/plugins/multimedia/multimedia.js
@@ -718,7 +718,7 @@ $document.on( youtubeEvent, selector, function( event, data ) {
 			var elm = evt.currentTarget,
 				ds = elm.dataset;
 
-			// Do nothing on the first load and add a flag to indicate it is loaded a second times
+			// Do nothing on the first load and add a flag to indicate it is loaded a second time
 			if ( ds.L1 ) {
 				ds.L2 = true;
 			} else {
--- a/src/plugins/multimedia/multimedia.js
+++ b/src/plugins/multimedia/multimedia.js
@@ -66,7 +66,7 @@ var componentName = "wb-mltmd",
 				// the issue is when the iFrame is moved, it reloads and then it doesn't emit the right event to
 				// adjust the WET multimedia controller to represent its current state.
 				//
-				// This need to be executed only once that is why it is in the i18nText conditional
+				// This needs to be executed only once, that is why it is in the i18nText conditional
 				window.addEventListener( "message", function( e ) {
 					var data, frames, i, i_len, i_cache;
 
--- a/src/plugins/multimedia/multimedia.js
+++ b/src/plugins/multimedia/multimedia.js
@@ -64,7 +64,7 @@ var componentName = "wb-mltmd",
 				// YT workaround for when played inside a modal dialog, like with lightbox
 				//
 				// the issue is when the iFrame is moved, it reloads and then it doesn't emit the right event to
-				// adjust the WET multimedia controler to represent it's current state.
+				// adjust the WET multimedia controller to represent its current state.
 				//
 				// This need to be executed only once that is why it is in the i18nText conditional
 				window.addEventListener( "message", function( e ) {
--- a/src/plugins/multimedia/multimedia.js
+++ b/src/plugins/multimedia/multimedia.js
@@ -711,7 +711,7 @@ $document.on( youtubeEvent, selector, function( event, data ) {
 		data.ytPlayer = ytPlayer;
 
 		// Detect if the YT player reloads, like when magnific Popup show the modal, because it moves the iframe
-		// and then the iframe get refreshed and reloaded. So the issue is the iframe stop to emit the event
+		// and then the iframe gets refreshed and reloaded. So the issue is that the iframe stops emitting the event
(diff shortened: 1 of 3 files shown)


src/plugins/multimedia/multimedia.js
```diff
@@ -60,6 +60,47 @@ var componentName = "wb-mltmd",
 
 			// Only initialize the i18nText once
 			if ( !i18nText ) {
+
+				// YT workaround for when played inside a modal dialog, like with lightbox
+				//
+				// the issue is when the iFrame is moved, it reload and then it don't emit the right event to
```
GormFrank:
```suggestion
				// the issue is when the iFrame is moved, it reloads and then it doesn't emit the right event to
```

src/plugins/multimedia/multimedia.js
```diff
@@ -60,6 +60,47 @@ var componentName = "wb-mltmd",
...
+				// adjust the WET multimedia controler to represent it's current state.
```
GormFrank:
```suggestion
				// adjust the WET multimedia controller to represent its current state.
```

src/plugins/multimedia/multimedia.js
```diff
@@ -60,6 +60,47 @@ var componentName = "wb-mltmd",
...
+				//
+				// This need to be executed only once that is why it is in the i18nText conditional
```
GormFrank:
```suggestion
				// This needs to be executed only once, that is why it is in the i18nText conditional
```

src/plugins/multimedia/multimedia.js
```diff
@@ -60,6 +60,47 @@ var componentName = "wb-mltmd",
...
+					var data, frames, i, i_len, i_cache;
+
+					// Ensure we don't conflict with other postMessage listener
+					try {
+						data = JSON.parse( e.data );
+
+						// Only for a specific YT message
+						if ( data.event && data.event === "infoDelivery" && data.info && data.info.playerState ) {
+
+							// Find the iFrame and evaluate if it need to be repost
```
GormFrank:
```suggestion
							// Find the iFrame and evaluate if it needs to be reposted
```

src/plugins/multimedia/multimedia.js
```diff
@@ -662,6 +710,22 @@ $document.on( youtubeEvent, selector, function( event, data ) {
 		data.media = $media;
 		data.ytPlayer = ytPlayer;
 
+		// Detect if the YTplayer re-load like when magnific Popup show the modal because it move the iframe
```
GormFrank:
```suggestion
		// Detect if the YT player reloads, like when magnific Popup show the modal, because it moves the iframe
```

src/plugins/multimedia/multimedia.js
```diff
@@ -662,6 +710,22 @@ $document.on( youtubeEvent, selector, function( event, data ) {
...
+		// and then the iframe get refreshed and reloaded. So the issue is the iframe stop to emit the event
```
GormFrank:
```suggestion
		// and then the iframe gets refreshed and reloaded. So the issue is that the iframe stops emitting the event
```

src/plugins/multimedia/multimedia.js
```diff
@@ -662,6 +710,22 @@ $document.on( youtubeEvent, selector, function( event, data ) {
...
+		// needed to adjust the multimedia player controler, like the "onStateChange" event.
+		$media.on( "load", function( evt ) {
+
+			var elm = evt.currentTarget,
+				ds = elm.dataset;
+
+			// Do nothing on the first load and add a flag to indicate it is loaded a second times
```
GormFrank:
```suggestion
			// Do nothing on the first load and add a flag to indicate it is loaded a second time
```

GormFrank (changes requested):
Only typos and grammatical errors

GormFrank (approved):

Issue #8620, opened 2019-04-01
Global mega menu fails on click events on narrow displays with new Canada.ca look/feel


minorOffense:
When using the mega menus on narrow displays the sub menu items throw a javascript error on https://github.com/wet-boew/wet-boew/blob/1882dd3df9cefb0d8d64edb22783903ace778edd/src/plugins/wb-focus/focus.js#L90

E.g.
```js
	if ( testHref.charAt( 0 ) === "#" && !event.isDefaultPrevented() &&
		( $linkTarget = $( "#" + wb.jqEscape( testHref.substring( 1 ) ) ) ).length !== 0 ) {
		wb.ignoreHashChange = true;
		$linkTarget.trigger( setFocusEvent );
	}
```

You can see the error at https://nrc.canada.ca 

What happens is that if ``testHref`` is less than 2 characters long you end up with a jQuery selector of just "#" which is invalid in jQuery 3 (jQuery 2 incorrectly returns the main html document element). So the ``$linkTarget`` variable assignment causes an error.

Instead a check should be made to skip this whole logic block if the ``testhHref`` value is just "#"

Patch to come.

bnrkll:
<pre>
<aside class="list-group">
<a href="/en/example1.shtml" class="list-group-item list-group-item-success">Example 1</a>
<a href="/en/example2.shtml" class="list-group-item list-group-item-success active">Example 2</a><a href="http://www.external-example.com" class="list-group-item list-group-item-success">Example 3</a>
</pre>

duboisp:
PR merged

FYI - That fix are going to be packaged also with the next GCWeb 5.1 release

PR #8621, opened 2019-04-03
wb-focus - add extra logic condition for focus with empty href

ClassicCut:
https://github.com/wet-boew/wet-boew/issues/8620

### DIFF
--- a/src/plugins/wb-focus/focus.js
+++ b/src/plugins/wb-focus/focus.js
@@ -86,7 +86,7 @@ $document.on( clickEvents, linkSelector, function( event ) {
 	var testHref = event.currentTarget.getAttribute( "href" );
 
 	// Same page links only
-	if ( testHref.charAt( 0 ) === "#" && !event.isDefaultPrevented() &&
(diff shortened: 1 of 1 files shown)


src/plugins/wb-focus/focus.js
```diff
@@ -86,7 +86,7 @@ $document.on( clickEvents, linkSelector, function( event ) {
 	var testHref = event.currentTarget.getAttribute( "href" );
 
 	// Same page links only
-	if ( testHref.charAt( 0 ) === "#" && !event.isDefaultPrevented() &&
+	if ( wb.jqEscape( testHref.substring( 1 ) ).length !== 0 && testHref.charAt( 0 ) === "#" && !event.isDefaultPrevented() &&
```
duboisp: Did you tried to only test if "testHref" is longer than 1 character instead of calling the wb.jqEscape ?
Like: 
```suggestion
	if ( testHref.length > 1 && testHref.charAt( 0 ) === "#" && !event.isDefaultPrevented() &&
```

duboisp (changes requested):

duboisp (approved):

PR #8635, opened 2019-04-19
Plugin Share/Sprite: Fix build blocking condition (Fixes #3751)

Scott-D-Hodgson:
Switched the Grunt SASS component to enable a smooth build on a linux workstation.  Updated the configuration to account for the differences and added an optional element into the SASS file so the sprites would build.  (Fixes #3751)


duboisp:
Pre-approved at the technical review after @GormFrank testing and review pass.

src/plugins/share/_base.scss
```diff
@@ -33,7 +33,7 @@
 	@each $service in bitly, blogger, digg, diigo, facebook, feed, gmail, linkedin, myspace, pinterest, reddit, tumblr, twitter, yahoomail {
```
GormFrank: Remove this loop and replace it by:

```
@include sprites($spritesheet-sprites);
```

GormFrank (changes requested):
At [line 6](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/share/_base.scss#L6), insert the following mixin:
```
@mixin sprites($sprites) {
	@each $sprite in $sprites {
		$sprite-name: nth($sprite, 10);
		.#{$sprite-name} {
			&:before {
				@include sprite($sprite);
			}
		}
	}
}
```

Issue #8647, opened 2019-05-10
Toggle: issue with toggle all detail elements (expand/collapse)


tyhsimms:
I encounter an issue when using a toggle button to expand/collapse all detail elements. If every detail element is opened/closed manually before using the toggle button, clicking the toggle button does nothing on the first click (not the case if only some were open/closed manually). Clicking the button a 2nd time will then trigger the toggle. The issue can be reproduced with the working example.

https://wet-boew.github.io/v4.0-ci/demos/toggle/toggle-en.html#details

To reproduce, manually open the 3 detail elements (Example 1, Example 2, Example 3) and then click the toggle button to close them. First click is fruitless, second click is successful.

juleskuehn:
I'll see if I can fix this now. Thanks!

juleskuehn:
The issue you noted should be fixed, however, it reveals another problem:
The expected behaviour for a toggle on multiple elements, when some are expanded and some are collapsed, is not defined.

I suggest we define it as follows:
1. Toggle will expand all items if *any* are collapsed.
2. Toggle will collapse all items if and only if *all* are open.

Does this sound like a reasonable approach?t

duboisp:
@juleskuehn 

I am not sure. It could be more confusing to users.

What about :

1. Toggle will expand all items if and only if all are collapsed
2. Toggle will collapse all items if and only if all are expanded
3. Otherwise Toggle will follow is normal toggling cycle.

juleskuehn:
My preference was that the "state" of the toggle would be entirely determined by the state of the elements it is toggling. However, I see your point and I think it works just as well for the most common scenario:
- User expands one item, then realizes there are many and decides to "Toggle expand all". The normal toggling cycle will provide the desired behaviour.

I have no strong feelings about it really... I will wait for others to weigh in.

juleskuehn:
Just to clarify @duboisp 
On second thought, I disagree with your proposal.
- If the group toggle only expands all items when all are collapsed, that would not work for the common scenario I describe above.
"Normal toggling cycle" would then have to mean "collapse all expanded items and expand all closed items". I can't think of any case where a user would want this.
So the group toggle should, in my opinion, either expand all or collapse all.

Pacoup:
The problem is the Toggle button itself.

[...]

PR #8649, opened 2019-05-15
Toggle plugin: multiple details toggle fix

juleskuehn:
Resolves issue #8647

Changes behaviour **only** for toggling multiple detail/summary elements.

1. Toggle will expand all items if any are collapsed.
2. Toggle will collapse all items if and only if all are expanded.


GormFrank (approved):
This solution follows desired behaviour as part of the discussion.

PR #8658, opened 2019-06-05
Remove reset button

jkshapiro:
It's dangerous (too easy to hit by accident) and unnecessary

### DIFF
--- a/src/other/feedback/feedback-en.hbs
+++ b/src/other/feedback/feedback-en.hbs
@@ -179,7 +179,7 @@
 			</section>
 		</section>
 
-		<input type="submit" name="fbsbmt" id="fbsbmt" value="Submit feedback" class="btn btn-primary" /> <input type="reset" value="Reset" class="btn btn-default" />
+		<input type="submit" name="fbsbmt" id="fbsbmt" value="Submit feedback" class="btn btn-primary" />
 	</form>
 
 	<section>


duboisp:
Is it that based on research? Instead of remove it, may be we can style the reset button into a link so it will minimize it predominance compared to the submit styled with "primary" style.  

related to #8659

jkshapiro:
Not original research but yes. See, for instance:
https://www.nngroup.com/articles/reset-and-cancel-buttons/
https://www.nngroup.com/articles/web-form-design/

Also lots of best practice advice from UX thought leaders:
https://www.lukew.com/ff/entry.asp?571
https://www.uxmatters.com/mt/archives/2012/05/7-basic-best-practices-for-buttons.php
https://www.smashingmagazine.com/2017/06/designing-efficient-web-forms/

RobJohnston:
You don't have to go too far... it's mentioned in the style guide:
https://wet-boew.github.io/wet-boew-styleguide/design/buttons-en.html

But it's a little confusing, in that the text for the Clear button says "Do not use" and "Refer to 'Reset' button section".  Then the Reset button just says to avoid its usage.

thekodester:
As @RobJohnston mention the style guide doesn't encourage it, but it can be useful.

Maybe as @duboisp said we could restyle it to a link and maybe use the recommend text from the style guide of "Reset page to defaults".

jkshapiro:
This pull ew

@thekodester I agree that in some rare cases the button can be useful. But I don't think the case under consideration (https://wet-boew.github.io/v4.0-ci/demos/feedback/feedback-en.html) is one of them....

thekodester:
Even the article (https://www.nngroup.com/articles/reset-and-cancel-buttons/) you linked yourself states there should be a way for a user to undo there actions, but radio button and checkboxes can prevent this.

Also they can be useful when a form is pre-populated with data from the server and you want to clear that information.

So if where gonna to remove it all together we would first need to make sure that a clear way to undo actions that is also accessible.

thekodester:
Also as for your other comment I understand there multiple forums with the reset it not that difficult to restyle the reset button on each form in the demos.

jkshapiro:
Sorry for the confusion. I agree that reset buttons have their place. This pull request is only for one particular example.

thekodester:
Yes, what I'm stating is if where to restyle the reset button maybe create one pull request that restyles them all the same.

If that where you wanna go with this I can do that.

thekodester:
I opened a PR for a different approach here #8699.

thekodester:
Since https://github.com/wet-boew/wet-boew/pull/8699 was merged is this being closed?

jkshapiro:
#8699 is better than what we had before but it still doesn't meet the concerns described above.

PR #8659, opened 2019-06-05
Remove reset button

jkshapiro:
It's dangerous (too easy to hit by accident) and unnecessary

### DIFF
--- a/src/other/feedback/feedback-fr.hbs
+++ b/src/other/feedback/feedback-fr.hbs
@@ -179,7 +179,7 @@
 			</section>
 		</section>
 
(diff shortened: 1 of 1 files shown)


duboisp:
Is it that based on research? Instead of remove it, may be we can style the reset button into a link so it will minimize it predominance compared to the submit styled with "primary" style.  

related to #8658

jkshapiro:
See comments at #8658

thekodester:
Since https://github.com/wet-boew/wet-boew/pull/8699 was merged is this being closed?

jkshapiro:
See https://github.com/wet-boew/wet-boew/pull/8658#issuecomment-541073038

PR #8673, opened 2019-07-02
Datatables scroll to top when pages clicked from bottom (fix #7481)

### LINKED ISSUE #7481: Pagination on data table - "previous" button loses focus after pressing enter
Currently, if you tab through the pagination buttons using your keyboard and press enter on the "previous" button, the button loses focus, which makes it difficult to navigate to previous views of the table using your keyboard. If you do this with the "next" button, the button keeps focus and you are able to navigate through the pagination buttons easily. I believe this should also be the case with the "previous" button.

Can this functionality be added?

https://wet-boew.github.io/v4.0-ci/demos/tables/tables-en.html

ipaksc:


src/plugins/tables/tables.js
```diff
@@ -134,7 +134,9 @@ $document.on( "draw.dt", selector, function( event, settings ) {
 		paginate_buttons = $elm.next( ".bottom" ).find( ".paginate_button" ),
 		ol = document.createElement( "OL" ),
 		li = document.createElement( "LI" );
-
+		if ( !$elm.id ) {
```
duboisp: Move this block in the plugin initialization

src/plugins/tables/tables.js
```diff
@@ -158,8 +160,7 @@ $document.on( "draw.dt", selector, function( event, settings ) {
 		// Should be pushed upstream to DataTables
 		$elm.next( ".bottom" ).find( ".paginate_button" )
 			.attr( {
-				"role": "button",
-				"href": "javascript:;"
+				"href": "#" + $elm.context.id
 			} )
 			.not( ".previous, .next" )
 				.attr( "aria-pressed", "false" )
```
duboisp: Please update the markup as per the latest practice. As a reference see the pagination template documentation in the GCWeb search template.

src/plugins/tables/tables.js
```diff
@@ -173,7 +174,12 @@ $document.on( "draw.dt", selector, function( event, settings ) {
 	// Identify that the table has been updated
 	$elm.trigger( "wb-updated" + selector, [ settings ] );
 } );
-
+$document.on( "keypress", ".paginate_button", function( event ) {
+	var $elm = $( event.target );
+	if (event.keyCode == 13){
```
duboisp: The space key is also use for navigation, please add it.

Also, I know you need to do this override because of how the datatable manage the pagination, please add a comment before that function explaining the rational.

src/plugins/tables/tables.js
```diff
@@ -173,7 +174,12 @@ $document.on( "draw.dt", selector, function( event, settings ) {
...
+		location = $elm.context.href;
```
duboisp: The ```location``` global are undefined in this context because of the closure.

src/plugins/tables/tables.js
```diff
@@ -173,7 +174,12 @@ $document.on( "draw.dt", selector, function( event, settings ) {
 	// Identify that the table has been updated
 	$elm.trigger( "wb-updated" + selector, [ settings ] );
 } );
-
+$document.on( "keypress", ".paginate_button", function( event ) {
```
duboisp: use ```keyup``` or ```keydown```

duboisp: Be more specific for the CSS event selector.

duboisp (changes requested):

src/plugins/tables/tables.js
duboisp: Ignore my comment about the space key behaviour. It should behave as any other links, sorry for the confusion.

src/plugins/tables/tables.js
ipaksc: The block has been deleted. There is already a script that ensures, a unique id is assigned to the element

src/plugins/tables/tables.js
ipaksc: We have discussed this topic.

src/plugins/tables/tables.js
ipaksc: Done. Here is a new solution

// This is required to override the datatable.js (v1.10.13) behavior to cancel the event propagation on anchor element.
$elm.next( ".bottom" ).find( ".paginate_button" )
			.attr( {
				"href": "#" + $elm.context.id
			} )			
.on( "keypress", function( evn ){
				if ( evn.keyCode === 13 ){
					window.location = evn.target.href;
				}
			} )

src/plugins/tables/tables.js
ipaksc: It's fixed.

.on( "keypress", function( evn ){
				if ( evn.keyCode === 13 ){
					window.location = evn.target.href;
				}
			} )

duboisp (approved):

duboisp:
Before to have it merged, it will be preferable to update the i18n text for "page" to be more descriptive like "Jump to: Page 3" and "Aller à: Page 3".

GormFrank (approved):

duboisp:
i18n csv was updated.

PR #8674, opened 2019-07-03
Repo: Updated URLs of sites that support HTTPS.

EricDunsworth:
This pretty much covers all of this repo's URLs, but excludes the following:
- Sites that lack HTTPS support
- Sites with broken HTTPS support (e.g. site looks broken when accessed via HTTPS)
- Sites that no longer exist
- Proxy URL patterns (which generally don't play nicely with https: prefixes)
- XMLNS attributes (http:-prefixed values are literally the names of the namespaces)
- vocab="http://schema.org/" attributes (was asked to stick with HTTP for them)


site/layouts/core.hbs
```diff
@@ -11,7 +11,7 @@
 		{{>metadata}}
 		{{>headresources}}
 	</head>
-	<body {{#if pageclass}}class="{{pageclass}}" {{/if}}vocab="http://schema.org/" typeof="WebPage">
+	<body {{#if pageclass}}class="{{pageclass}}" {{/if}}vocab="https://schema.org/" typeof="WebPage">
```
duboisp: The schema.org vocabulary should use "http://" scheme, not the "https://" 

reference: https://www.w3.org/2011/rdfa-context/rdfa-1.1

site/layouts/servermessage.hbs
```diff
@@ -12,7 +12,7 @@
 		<meta name="robots" content="noindex, nofollow, noarchive" />
 		{{>headresources}}
 	</head>
-	<body vocab="http://schema.org/" typeof="WebPage">
+	<body vocab="https://schema.org/" typeof="WebPage">
```
duboisp: The schema.org vocabulary should use "http://" scheme, not the "https://" 

reference: https://www.w3.org/2011/rdfa-context/rdfa-1.1

duboisp (changes requested):

site/layouts/core.hbs
EricDunsworth: I changed the vocab attribute's protocol was due to the [question/answer about HTTPS on schema.org's FAQ](https://schema.org/docs/faq.html#19).

**Excerpt:**
> **Should we write ``https://schema.org`` or ``http://schema.org`` in our markup?**
> 
> There is a general trend towards using ``https`` more widely, and you can already write ``https://schema.org`` in your structured data. Over time we will migrate the schema.org site itself towards using https: as the default version of the site and our preferred form in examples. However ``http://schema.org`` -based URLs in structured data markup will remain widely understood for the forseeable future and there should be no urgency about migrating existing data. This is a lengthy way of saying that both ``https://schema.org`` and ``http://schema.org`` are fine.

Thoughts?

site/layouts/servermessage.hbs
EricDunsworth: See my other conversation comment.

EricDunsworth:
Just fixed a few issues I spotted:
- Reverted unrelated changes to package-lock.json
- Reverted an unrelated change in src/other/tablevalidator/demo/tablevalidator.js
- Reverted changes to src/polyfills/slider/assets/fd-slider-sprite.png (contains some XMLNS attributes I initially changed to HTTPS and backtracked on - but the file continued to be tracked as modified)
- Reverted some geogratis.gc.ca URLs back to HTTP (since they're tied to redirects that don't work via HTTPS)

GormFrank:
Pre-approved during weekely technical review as long as the schema.org link is switched back to "http".

site/layouts/core.hbs
EricDunsworth: Reverted this change as per our follow-up discussion.

site/layouts/servermessage.hbs
EricDunsworth: Reverted this change as per our follow-up discussion.

EricDunsworth:
@duboisp @GormFrank FYI I've reverted the schema.org link and rebased to resolved merge conflicts. Should be good to merge once Travis-CI has finished rebuilding.

duboisp (approved):

PR #8677, opened 2019-07-05
Boostrap overrides: Fix solid abbreviation underlines

EricDunsworth:
In Bootstrap 3.4.0 (via normalize.css 4.0.1-8.0.1), abbreviations with title attributes use "text-decoration: underline dotted;" to create a dotted underline effect in modern browsers.

"text-decoration: underline" is used as a fallback for older browsers such as IE11 and Edge... which causes them to render affected abbreviations with solid underlines. That isn't desirable since they end up looking too much like links.

This commit adjusts WET's CSS to continue using "text-decoration: underline dotted;" for abbreviations in browsers that support it and fallback on "border-bottom: 1px dotted;" in ones that don't.

It's a port of https://github.com/necolas/normalize.css/pull/738#issuecomment-387549760 (created by @mattbrundage).

### DIFF
--- a/src/base/bootstrap-overrides/_base.scss
+++ b/src/base/bootstrap-overrides/_base.scss
@@ -217,6 +217,27 @@ a {
 }
 
 /*
+ *	Abbreviations with title attributes
+ *	Workaround for abbreviations with titles using solid underlines in IE11/Edge in Bootstrap 3.4.0 (via normalize.css 4.0.1-8.0.1):
+ *	- Use a dotted underline in browsers that support it (e.g. Firefox, Chrome, etc...)
+ *	- Fallback on a dotted bottom border in browsers that don't support it (e.g. IE11, Edge, etc...)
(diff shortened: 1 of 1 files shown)


EricDunsworth:
FYI CSSLint is currently failing since the version WET depends on doesn't understand ``@supports``. I'll update that dependency in another PR.

Also, WET's Bootstrap dependency is in limbo atm:
- [WET's theme.css file](https://wet-boew.github.io/v4.0-ci/theme-wet-boew/css/theme.css) is on Bootstrap 3.3.7
- [GCWeb's theme.css file](https://wet-boew.github.io/themes-dist/GCWeb/GCWeb/css/theme.css) is on Bootstrap 3.4.1 (inherits its Bootstrap version from ``"bootstrap-sass": "^3.3.7"`` in WET's package.json file)

duboisp:
You can add an exception in the CSSlint for the ```@support``` rule

duboisp (approved):

PR #8679, opened 2019-07-08
Build: Update CSSLint/banner to support #8677's changes

EricDunsworth:
Related to #8677

### DIFF
--- a/.csslintrc
+++ b/.csslintrc
@@ -15,11 +15,14 @@
   "important": false,
   // Need due to use of "\9" hacks for oldIE
   "known-properties": false,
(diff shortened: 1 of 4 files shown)


duboisp (commented):
Do we need to deactivate the alphabetical order? And what about the other CSSlint rule?

duboisp:
Dismiss my comment, I just saw your extended description in your commit message.

Is those adjustment for the CSSlint is needed because of the change to Bootstrap 3.4 or because of the bumped version applied by this PR?

duboisp (approved):

PR #8690, opened 2019-07-25
Build: Update grunt to ~1.0.4

EricDunsworth:
Decreases build times due to an updated glob sub-dependency. Windows gets a drastic speed boost. Linux probably also benefits.

From a cold boot in Windows, grunt dist used to take ~19 minutes... now it only takes ~6.5 minutes (3x faster). The speed-up is mostly due to the cssmin tasks becoming much faster (14 minutes became 1 minute). Grunt dist is currently setup to automatically run at the end of npm install.

This PR's generated dist folder is virtually identical to before. Only difference is that /dist/wet-boew/payload.json's entries are now sorted alphabetically.

The newer version of CoffeeScript used in grunt 1.0.4 doesn't appear to introduce any breaking changes to Gruntfile.coffee.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1435,9 +1435,9 @@
       "dev": true
     },
     "colors": {
-      "version": "0.6.2",
-      "resolved": "https://registry.npmjs.org/colors/-/colors-0.6.2.tgz",
-      "integrity": "sha1-JCP+ZnisDF2uiFLl0OW+CMmXq8w=",
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/colors/-/colors-1.1.2.tgz",
+      "integrity": "sha1-FopHAXVran9RoSzgyXv6KMCE7WM=",
       "dev": true
     },
     "combine-source-map": {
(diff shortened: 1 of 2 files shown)


EricDunsworth:
@nschonni @LaurentGoderre Any concerns or other thoughts about the CoffeeScript update? AFAIK you two created most of what's in WET's Gruntfile.

According to jashkenas/coffeescript#3352, CoffeeScript doesn't use semver.

When I went through CoffeeScript's changelog to compare what's changed between [1.3.3](https://coffeescript.org/#1.3.3) and [1.10.0](https://coffeescript.org/#1.10.0), nothing really stood out to me as obvious red flags for WET/GCWeb's Gruntfiles. Plus the Travis-CI build is passing and producing what I'd expect it to.

nschonni:
I think it should be fine. You could also convert the CoffeeScript to JavaScript if it's easier to maintain.

EricDunsworth:
@nschonni Thanks :).

Built-in CoffeeScript support is [deprecated in grunt-cli 1.3.0](https://gruntjs.com/blog/2018-08-15-grunt-cli-1.3.0-released) and will cease in the next major version of grunt. So it'd potentially be an opportunity to eliminate a dependency.

I'll leave WET's Gruntfile as-is for the time being. If WET 4.0 is still supported by the time grunt 2 is released, maybe it'd be worthwhile to go in that direction.

GormFrank:
Pre-approved during technical review.

GormFrank (approved):

PR #8692, opened 2019-07-25
Forms: Add autocomplete attributes

EricDunsworth:
Improves compliance with WCAG 2.1 Success Criterion 1.3.5 ("Identify Input Purpose").

**Note:** Didn't add anything to the "Merge Server-Client Errors" page's "Beneficiary name" field since I wasn't sure about what kinds of corresponding changes would need to be made to [wet-boew/asp.net-formvalid-serverside](https://github.com/wet-boew/asp.net-formvalid-serverside).

### DIFF
--- a/src/polyfills/datalist/datalist-en.hbs
+++ b/src/polyfills/datalist/datalist-en.hbs
@@ -7,7 +7,7 @@
 	"tag": "datalist",
 	"parentdir": "datalist",
 	"altLangPrefix": "datalist",
-	"dateModified": "2014-08-06"
+	"dateModified": "2019-07-29"
 }
 ---
 <section>
@@ -20,7 +20,7 @@
 	<form action="#" method="get">
 		<div class="form-group">
 			<label for="city">City</label>
-			<input class="form-control" type="text" id="city" name="city" list="suggestions" />
+			<input class="form-control" type="text" autocomplete="address-level2" id="city" name="city" list="suggestions" />
 		</div>
 		<datalist id="suggestions">
 			<!--[if lte IE 9]><select><![endif]-->
--- a/src/polyfills/datalist/datalist-fr.hbs
+++ b/src/polyfills/datalist/datalist-fr.hbs
@@ -7,7 +7,7 @@
 	"tag": "datalist",
 	"parentdir": "datalist",
 	"altLangPrefix": "datalist",
-	"dateModified": "2014-08-06"
+	"dateModified": "2019-07-29"
 }
 ---
 <section>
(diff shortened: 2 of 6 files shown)


EricDunsworth:
@StdGit You might want to consider modifying the [Merge Server-Client Errors demo](https://wet-boew.github.io/v4.0-ci/unmin/demos/formvalid/formvalid-server-en.html)'s "Beneficiary name" field to use an ``autocomplete="name"`` attribute. A corresponding update would also need to be made to [wet-boew/asp.net-formvalid-serverside](https://github.com/wet-boew/asp.net-formvalid-serverside).

src/polyfills/datalist/datalist-en.hbs
```diff
@@ -20,7 +20,7 @@
 	<form action="#" method="get">
 		<div class="form-group">
 			<label for="city">City</label>
-			<input class="form-control" type="text" id="city" name="city" list="suggestions" />
+			<input class="form-control" type="text" autocomplete="address-level2" id="city" name="city" list="suggestions" />
```
GormFrank: In order to use "address-line1", "address-line2" or "address-line3", you must be using "street-address" also.

src/polyfills/datalist/datalist-fr.hbs
```diff
@@ -20,7 +20,7 @@
 	<form action="#" method="get">
 		<div class="form-group">
 			<label for="city">Ville</label>
-			<input class="form-control" type="text" id="city" name="city" list="suggestions">
+			<input class="form-control" type="text" autocomplete="address-level2" id="city" name="city" list="suggestions">
```
GormFrank: In order to use "address-line1", "address-line2" or "address-line3", you must be using "street-address" also.

GormFrank (changes requested):

src/polyfills/datalist/datalist-en.hbs
EricDunsworth: I just did a follow-up review. What's in use here is ``"address-level2"``, which is different than ``"address-line2"``. So the [``"street-address"`` requirement you saw on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#Values) doesn't apply.

Here's the [HTML spec's definition for ``address-level2``](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-address-level2):

> The [second administrative level](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#more-on-address-levels), in addresses with two or more administrative levels; in the countries with two administrative levels, this would typically be the city, town, village, or other locality within which the relevant street address is found

**PS:** The [HTML spec's definition for ``"address-line2"``](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-address-line2) doesn't contain any mention of ``"street-address"`` needing to be used in conjunction with it. So maybe [what you saw on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#Values) was just a strongly-worded suggestion or based on an outdated definition from the spec. IMO it would be strange to require both kinds of fields to be used together since it'd be pointless for the ``"address-line#"`` fields to exist at all if they were required to be used alongside a field that contains all of them combined (i.e. ``"street-address"``).

GormFrank (approved):

PR #8709, opened 2019-08-16
Minor Fix for Table Filters

### LINKED ISSUE #8708: DataTables filter options: Can't return to default view after filtering anything specific
We've noticed that in WET DataTables implementations that contain advanced filtering options (e.g. dropdowns, checkboxes, etc...), WET only triggers DataTables filtering if at least one field contains a non-empty value. If all fields contain empty values, nothing happens.

That's problematic in practice. If a user picks specific categories, triggers the filter, then switches back to the default categories and filters again, they'd expect DataTables to go back to showing all entries. But that isn't currently happening. As things stand, their only option would be to click the "Clear" button (whose functionality isn't very obvious - most people probably wouldn't expect that pressing it would instantly modify the results).

To address this, WET should be modified to always trigger DataTables filtering upon clicking the "Filter" button, regardless of whether or not all fields contain empty values.

**How to reproduce:**
1. Visit the [Tables page](https://wet-boew.github.io/v4.0-ci/unmin/demos/tables/tables-en.html)'s "Filtering Example" section.
2. The table will contain a total of 125 entries.
3. In the "Filter Options" panel, set the "Data Type" dropdown to "Speeches" and press the "Filter" button.
4. The table will show 4 out of 4 entries.
5. In the "Filter Options" panel, set the "Data Type" dropdown to "All news types" and press the "Filter" button.

**Expected outcome:**
The table should show 125 entries.

**Actual outcome:**
Nothing happened. The table continued to show 4 out of 4 entries.

**PS:**
Thanks to @fsnoddy and @donmcdill for having initially noticed this :).

ghost:
Fixes #8708

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -194,8 +194,8 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 	var $form = $( this ),
 		$datatable = $( "#" + $form.data( "bind-to" ) ).dataTable( { "retrieve": true } ).api();
 
-	// Lets reset the search;
-	$datatable.search( "" ).columns().search( "" );
+	// Lets reset the search
+	$datatable.search( "" ).columns().search( "" ).draw();
 
 	// Lets loop throug all options
 	var $lastColumn = -1, $cbVal = "";


EricDunsworth (approved):

EricDunsworth:
@thekodester FYI you might need to adjust your commit email address - doesn't look like this PR's commit is linked to your GitHub account.

thekodester:
Yeah that's always been broken (always forget about it). You can merge regardless doesn't really affect anything.

PR #8699, opened 2019-08-01
Reset Button Restyled

ghost:
Different approach to #8658 and #8659.

Also better follows https://wet-boew.github.io/wet-boew-styleguide/design/buttons-en.html.

PR #8738, opened 2019-10-11
Move bootstrap 4 supplemental style to its own file

duboisp:
Related to new style added by #8699.

Todo:
Need to be included in the release note and need to be documented in our style guide, the following new CSS class + quick description
* p-0
* pl-2, px-2
* pt-4, py-4

### DIFF
--- a/src/base/_bootstrap-4.scss
+++ b/src/base/_bootstrap-4.scss
@@ -1,0 +1,40 @@
+/*
+  WET-BOEW
+  @title: Bootstrap 4 suplemental style for WET-BOEW
+ */
+
+/* Padding */
+.p-0 {
+	padding: 0 !important;
+}
+
+.pl-2
+.px-2 {
+	padding-left: 5px !important;
+}
+.px-2 {
+	padding-right: 5px !important;
+}
+
+.pt-4,
+.py-4 {
+	padding-top: 30px !important;
+}
+.py-4 {
+	padding-bottom: 30px !important;
+}
+
+/* Stretched link */
+.stretched-link {
+	&:after {
+		background-color: rgba(0, 0, 0, 0);
+		bottom: 0;
+		content: "";
+		left: 0;
+		pointer-events: auto;
+		position: absolute;
+		right: 0;
+		top: 0;
+		z-index: 1;
+	}
+}
--- a/src/base/views/_bootstrap-4-screen-sm-min.scss
+++ b/src/base/views/_bootstrap-4-screen-sm-min.scss
@@ -1,0 +1,18 @@
+/*
+  WET-BOEW
+  @title: Bootstrap 4 suplemental style for WET-BOEW - Small view and over (screen only)
+ */
+
+/* Margin */
+.mb-sm-5 {
+	margin-bottom: 50px !important;
+}
+
+/* Padding */
+.p-sm-3 {
+	padding: 15px !important;
+}
+.px-sm-3 {
+	padding-left: 15px !important;
+	padding-right: 15px !important;
+}
--- a/src/base/_bootstrap.scss
+++ b/src/base/_bootstrap.scss
@@ -8,3 +8,4 @@ $icon-font-path: "#{$wb-core-path}/fonts/";
  */
 @import "bootstrap-sass/assets/stylesheets/bootstrap";
 @import "bootstrap-overrides/base";
+@import "bootstrap-4";
--- a/src/base/_wet-boew.scss
+++ b/src/base/_wet-boew.scss
@@ -188,6 +188,7 @@
 /* Small view and over */
 @media screen and (min-width: $screen-sm-min) {
 	@import "views/screen-sm-min";
+	@import "views/_bootstrap-4-screen-sm-min";
 	@import "proximity/screen-sm-min";
 }
 
--- a/src/base/bootstrap-overrides/_base.scss
+++ b/src/base/bootstrap-overrides/_base.scss
@@ -891,17 +891,3 @@ caption {
 	}
 	text-align: center;
 }
-
-.p-0 {
-	padding: 0 !important;
-}
-
-.pl-2
-.px-2 {
-	padding-left: .5rem !important;
-}
-
-.pt-4,
-.py-4 {
-	padding-top: 1.5rem !important;
-}


thekodester:
Would the PR need to remove where it currently exist?

https://github.com/wet-boew/wet-boew/blob/abf1ba2fcd47b8c788255dde811e50158becac4e/src/base/bootstrap-overrides/_base.scss#L895

duboisp:
> Would the PR need to remove where it currently exist?

Yes and I will update my PR.

src/base/_bootstrap-4.scss
```diff
@@ -0,0 +1,18 @@
...
+  @title: Bootstrap 4 suplemental style for WET-BOEW
+ */
+
+.p-0 {
+	padding: 0 !important;
+}
+
+.pl-2
+.px-2 {
+	padding-left: .5rem !important;
```
GormFrank: em not rem

GormFrank (changes requested):

duboisp:
Pre-approved after the change requested by Kodecount and GormFrank is completed. 

Also, this PR might include the additional Bootstrap 4 CSS that was introduced by GCweb, waiting for a discussion during today roadmap meeting.

duboisp:
I made the requested changes and we agreed at the last WET-BOEW roadmap to include the bootstrap 4 css added in GCWeb.

@GormFrank we will need to include in the release note the following CSS
* stretched-link
* mb-sm-5
* p-sm-3
* px-sm-3

src/base/views/_bootstrap-4-screen-sm-min.scss
```diff
@@ -0,0 +1,18 @@
+/*
+  WET-BOEW
+  @title: Bootstrap 4 suplemental style for WET-BOEW - Small view and over (screen only)
+ */
+
+/* Margin */
+.mb-sm-5 {
+	margin-bottom: 50px !important;
```
GormFrank: Shouldn't those be in `em` instead of `px`?

GormFrank (changes requested):

src/base/views/_bootstrap-4-screen-sm-min.scss
thekodester: Bootstrap 4 uses rem.

https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.css

GormFrank: > Bootstrap 4 uses `rem`.
> 
> https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.css

We know that Bootstrap 4 uses `rem`, however for minimal impact on overall WxT's layout right now (mostly GCWeb), the community agreed on using `em` temporarily.

duboisp: We will go with px because that is how it was coded in the utility CSS class: https://github.com/wet-boew/wet-boew/blob/master/src/base/proximity/_base.scss

So I am going to change the ```em``` into ```px```.

But in the next version, definitely we should use ```rem```. The challenge of using ```rem``` is we will need to re-caculate the measure for wet-boew + gcweb theme and re-test everything to see if there any visual impact.

src/base/_bootstrap-4.scss
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/* Padding */
+.p-0 {
+	padding: 0 !important;
+}
+
+.pl-2
+.px-2 {
+	padding-left: .5em !important;
```
GormFrank: Shouldn't it be `px` here too?

src/base/_bootstrap-4.scss
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.pt-4,
+.py-4 {
+	padding-top: 1.5em !important;
```
GormFrank: ... AND HERE

GormFrank (changes requested):

src/base/_bootstrap-4.scss
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.pl-2
+.px-2 {
+	padding-left: 5px !important;
+}
+
+.pt-4,
+.py-4 {
+	padding-top: 30px !important;
```
GormFrank: Doesn't the `.py-4` also have a `padding-bottom: 30px !important;`?

src/base/_bootstrap-4.scss
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/* Padding */
+.p-0 {
+	padding: 0 !important;
+}
+
+.pl-2
+.px-2 {
+	padding-left: 5px !important;
```
GormFrank: Doesn't the `.px-2` also have a `padding-right: 5px !important;`?

thekodester: It originally had `padding-left: 0.5rem !important;` so should be 8px.

src/base/_bootstrap-4.scss
thekodester: It originally had `padding-top: 1.5rem !important;` so should be 24px.

src/base/_bootstrap-4.scss
duboisp: I put the same as the value for ```mrgn-bttm-sm```

src/base/_bootstrap-4.scss
duboisp: The 30 px was calculated based on size of mrgn-tp-lg

duboisp:
In bootstrap 4, the spacing length is the same for margin and padding. 

I was assuming the following correspondence
* 5 = xl
* 4 = lg
* 3 = md
* 2 = sm
* 1 = ?
* 0 = 0 or none

https://getbootstrap.com/docs/4.3/utilities/spacing/

src/base/_bootstrap-4.scss
GormFrank: The real `rem` from WET is based on 10px unfortunately, that's the only reason stopping us from using the actual `rem`. Consequently, it is better (at least temporarily) to base our numbers on what we already have with the `mrgn` `-tp` `-bttm` etc.

duboisp:
Currently in wet-boew and in gcweb: ```1rem``` is ```10px```. I understand this could be improved and ```1 rem``` should match the font-size of the main content of the page which is ```16px``` for wet-boew and ```20px``` for gcweb. Historically, I think that override was made to ensure better compatibility with an old version of IE.
Changing the base font size from ```10 px``` to ```16 px``` or ```20px``` will require to do a regression testing of everything in wet-boew and gcweb which is mostly time consuming.

GormFrank (approved):

PR #8743, opened 2019-10-15
Data Inview: Fix accessibility issue for screen readers

xiayitan:
Issue: There is accessibility issue when using the Data Inview plugin with transitions. The sections with transition style are not readable by screen readers.

Run the test-1.html with screen reader without the data-inview.js changes, you will hear the screen reader skips the section with transition. 
I went back to check the history of the data-inview.js, here is the original code:
https://github.com/wet-boew/wet-boew/blame/d27369a81d79275941a853131b833753b42eb27d/src/plugins/data-inview/data-inview.js
I couldn’t figure out the reason why using the aria-hidden.

Here is my solution on this PR: deleting the attribute “aria-hidden” on the data-inview.js when using transitions with Data Inview plugin.


duboisp:
Here the test with the fix applied: http://universallabs.org/wet/labs/pr-8743/docs/ref/data-inview/test-1.html

Here the same test but without the fix: http://universallabs.org/wet/labs/pr-8743-withnofix/docs/ref/data-inview/test-1.html

@bsouster

duboisp:
@GormFrank as discussed, please test with a screen reader and report on your assessment

duboisp (approved):
Was tested by the accessibility working group

PR #8744, opened 2019-10-16
Inline inputs for checkbox and radio buttons

delisma:
CSS changes and coded documentation for inline checkboxes and radio buttons inputs in `<div>` elements and `<li>` elements

### DIFF
--- a/src/base/bootstrap-overrides/_base.scss
+++ b/src/base/bootstrap-overrides/_base.scss
@@ -590,6 +590,24 @@ input[type=button], input[type=reset], input[type=submit] {
 }
 
 /*
+ * Explicit labels for checkbox and radio button inputs
+ *
+ * Allow the use of horizontal explicit labels instead of the default implicit
+ * labelling pattern for checkbox and radio button inputs provided by Bootstrap.
+ */
+.form-inline {
+	.label-inline {
+		position: relative;
+		vertical-align: middle;
+		label {
+			font-weight: 400;
+			margin-bottom: 0px;
+			padding-left: 2px;
+		}
+	}
+}
+
+/*
  *	Fieldset/legend border position
  *
  *	Bootstrap puts the border below the legend which creates a visual
--- a/src/base/views/_screen-sm-min.scss
+++ b/src/base/views/_screen-sm-min.scss
@@ -6,3 +6,13 @@
 .col-sm-auto {
 	width: auto;
 }
+
+.form-inline {
+	.label-inline {
+		display: inline-block;
+		padding-right: 10px;
+		&:last-child {
+			padding-right: 0;
+		}
+	}
+}
--- a/src/plugins/formvalid/formvalid-en.hbs
+++ b/src/plugins/formvalid/formvalid-en.hbs
@@ -23,8 +23,8 @@
 	<p>Provides generic validation and error message handling for Web forms.</p>
 </section>
 
-<section>
-	<h2>Example</h2>
+<section class="wb-prettify all-pre">
+	<h2>Examples</h2>
 
 	<div class="wb-frmvld">
 		<form action="#" method="get" id="validation-example">
@@ -42,7 +42,8 @@
 				</div>
 				<details class="mrgn-bttm-md">
 					<summary>View code</summary>
-					<pre><code>&lt;div class="wb-frmvld"&gt;
+					<pre class="prettyprint prettyprinted"><code>
+&lt;div class="wb-frmvld"&gt;
 	&lt;form action="#" method="get" id="validation-example"&gt;
 		...
 		&lt;div class="form-group"&gt;
@@ -63,7 +64,8 @@
 				</div>
 				<details class="mrgn-bttm-md">
 					<summary>View code</summary>
-					<pre><code>&lt;div class="wb-frmvld"&gt;
+					<pre class="prettyprint prettyprinted"><code>
+&lt;div class="wb-frmvld"&gt;
 	&lt;form action="#" method="get" id="validation-example"&gt;
 		...
 		&lt;div class="form-group"&gt;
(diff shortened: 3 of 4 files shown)


src/plugins/formvalid/_base.scss
```diff
@@ -46,5 +46,31 @@
...
+		&.h5 {
+			float: none;
+			margin-bottom: 0;
+			.error {
+				margin-top: 1em;
+			}
+		}
 	}
 }
+ul.form-group {
```
duboisp: This style do not belong to the formvalid plugin, instead it should be more global, like the bootstrap override

Gruntfile.coffee
```diff
@@ -1009,6 +1009,7 @@ module.exports = (grunt) ->
 						"W010" # Using `.pull-left` or `.pull-right` as part of the media object component is deprecated as of Bootstrap v3.3.0. Use `.media-left` or `.media-right` instead.
 						"E013" # Only columns (`.col-*-*`) may be children of `.row`s
 						"E014" # Columns (`.col-*-*`) can only be children of `.row`s or `.form-group`s
+						"E017" # Support for checkboxes with explicit labels instead of implicit labels
```
duboisp: We will move this exception for the formvalid page only.

package.json
```diff
@@ -34,6 +34,7 @@
   "license": "MIT",
   "dependencies": {
     "bootstrap-sass": "3.4.1",
+    "bower": "^1.8.8",
```
GormFrank: Please do not include package.json and package-lock.json to your PR.

src/plugins/formvalid/_base.scss
GormFrank: Put those styles inside another file, do not encapsulate them inside formvalid

Gruntfile.coffee
GormFrank: We'll have to look into a new rule for this use case. (checkbox AND radio included)

GormFrank (changes requested):

duboisp (changes requested):
The update to package.json and package-lock.json would need to be removed from this PR.

Gruntfile.coffee
EricDunsworth: In case it helps: What @duboisp is suggesting would be very similar to what I recently did in wet-boew/GCWeb#1605

duboisp (commented):
Todo:
* Test accessibility - Check if the design pattern that use unordered list of "radio" or "checkbox" is acceptable
* Test accessibility - Check if the design pattern that use ```<ul>``` inside a ```fieldset``` is too much
* Test accessibility - Check if we only use ```div``` for a list of ```input[type=radio]``` and  ```input[type=radio]```
* Test accessibility - Is radio and checkbox must be always in a fieldset?
* Test accessibility - When radio and checkbox are not in a fieldset, can they be put in a list

Bootlint don't seems to support explicit labels. Explicit labels is needed to enable a improved visual for bigger and flat checkbox/radio that would improve UX.

duboisp:
@bsouster as discussed, you will find the pattern in the last example in this demo
https://universallabs.org/wet/labs/pr8744/unmin/demos/formvalid/formvalid-en.html

Also, see my [previous comments](https://github.com/wet-boew/wet-boew/pull/8744#pullrequestreview-305986836) regarding what we need to test and validate

EricDunsworth:
FWIW this is how Bootstrap 3.4's documentation recommends doing layouts for horizontal checkboxes/radio buttons:
https://getbootstrap.com/docs/3.4/css/#inline-checkboxes-and-radios

src/base/bootstrap-overrides/_base.scss
```diff
@@ -589,6 +589,24 @@ input[type=button], input[type=reset], input[type=submit] {
...
 
+/*
+ * Explicit labels for checkbox and radio button inputs
+ *
+ * Allow the use of horizontal explicit labels instead of the default implicit
+ * labelling pattern for checkbox and radio button inputs provided by Bootstrap.
+ */
+.form-inline {
+	.label-inline {
+		font-weight: 400;
```
GormFrank: There might be a variable available for this one.

src/base/bootstrap-overrides/_base.scss
```diff
@@ -589,6 +589,24 @@ input[type=button], input[type=reset], input[type=submit] {
...
+		margin-bottom: 0;
+		padding-left: 1px;
+		vertical-align: middle;
+	}
+	.form-group + .form-group {
+		padding-left: 10px;
```
GormFrank: The mrgn-lft-X class should be used instead.

src/base/bootstrap-overrides/_base.scss
```diff
@@ -648,6 +666,9 @@ fieldset {
 			font-weight: bold;
 			margin-bottom: 5px;
 		}
+		input + label {
+			padding: 0 1ch;
```
GormFrank: Update this one

GormFrank (changes requested):
Also don't forget to squash your commits.

src/base/bootstrap-overrides/_base.scss
delisma: The class mrgn-lft-x adds margin not padding.

src/base/bootstrap-overrides/_base.scss
delisma: I've checked all the scss files to find a variable that set a font-weight to normal but nothing seems to fit the needs.

src/base/bootstrap-overrides/_base.scss
delisma: Done

GormFrank:
Pre-approved during technical review conditionally to testing and @duboisp accepting changes.

GormFrank (approved):

src/base/bootstrap-overrides/_base.scss
```diff
@@ -589,6 +589,35 @@ input[type=button], input[type=reset], input[type=submit] {
...
+		position: relative;
+		vertical-align: middle;
+		label {
+			font-weight: 400;
+			margin-bottom: 0px;
+			padding-left: 2px;
+		}
+	}
+}
+@media (min-width: $screen-sm-min) {
```
duboisp: Move this in the view or create one for bootstrap and add it the main SCSS include file

duboisp (changes requested):

duboisp (approved):

duboisp:
Pre-approved waiting for Travis-CI biuld to pass.

PR #8768, opened 2019-11-22
Bootstrap overrides: Disable text-decoration-skip-ink on abbr[title]

### LINKED ISSUE #8767: Firefox sometimes renders dotted underlines as solid underlines - which impacts abbr[title]
I've noticed that recent versions of Firefox sometimes render dotted underlines as solid underlines. In practice this can affect the rendering of abbreviations that use ``title`` attributes (``abbr[title]``).

The issue seems to be caused by [a bug in Firefox's implementation of ``text-decoration-skip-ink``](https://bugzilla.mozilla.org/show_bug.cgi?id=1573723) (which is set to ``auto`` by default). Disabling it prevents the bug from manifesting itself.

``text-decoration-skip-ink`` was disabled in Bootstrap 4 via twbs/bootstrap#27673 due to rendering issues in Chrome. I tried backporting v4's fix into v3 via twbs/bootstrap#29707, but my PR was rejected since Bootstrap 3 has reached EOL.

**Reproduction steps**
1. In Firefox, visit the [WET theme's content page template](https://wet-boew.github.io/v4.0-ci/theme/content-en.html).
2. Open Firefox's DOM inspector.
3. Insert the following CSS selector:
```css
main * {
	text-decoration: dotted underline;
}
```
4. Observe the H2-H6 and paragraph samples. Some will show both solid and dotted underlines. Others will show dotted patterns that look like a mix of black and dark grey. Some will look like they're only using solid underlines.
5. Add ``text-decoration-skip-ink: none;`` to the selector.
6. Observe the H2-H6 and paragraph samples. Most of their dotted underline effects will look much better (except the H2 sample will have 2 sets of dotted underlines).

**Screenshots (Firefox Developer Edition 71.0b2 on Windows 10)...**

**Step 4 :**
step4-firefox-dotted-underline

**Step 6:**
step6-firefox-dotted-underline png

EricDunsworth:
Fixes #8767

### DIFF
--- a/src/base/bootstrap-overrides/_base.scss
+++ b/src/base/bootstrap-overrides/_base.scss
@@ -234,6 +234,8 @@ abbr[title] {
 	abbr[title] {
 		border-bottom: 0;
 		text-decoration: underline dotted;
+		//Next line ported from Bootstrap 4
+		text-decoration-skip-ink: none; // sass-lint:disable-line no-misspelled-properties
 	}
 }
 


EricDunsworth:
@duboisp @GormFrank I would highly recommend updating GCWeb's WET dependency to this commit once it gets merged in so it'll be in place the next time GCWeb gets updated on Canada.ca.

FWIW, the "GST/HST" abbreviation on https://www.canada.ca/en/services/benefits.html is affected by the bug this PR fixes. But the bug isn't showing up atm because the CSS statement that causes the bug was stripped out by Canada.ca's botched CSS minifier.

EricDunsworth:
I've just force-pushed an update to fix the sasslint failure.

WET's grunt-sass-lint dependency (^0.2.4) is currently up-to-date. But it doesn't recognize ``text-decoration-skip-ink`` due to sasstools/sass-lint#1281. My only option was to add a line comment with a disable rule.

EricDunsworth:
FYI the current Travis-CI failure has nothing to do with any of this PR's changes. I tried restarting it in case it was just a one-off, but it failed once again.

src/base/bootstrap-overrides/_base.scss
```diff
@@ -234,6 +234,7 @@ abbr[title] {
 	abbr[title] {
 		border-bottom: 0;
 		text-decoration: underline dotted;
+		text-decoration-skip-ink: none; // sass-lint:disable-line no-misspelled-properties
```
duboisp: Add a mention that CSS is already defined in Bootstrap 4

duboisp (changes requested):

duboisp:
Pre-approved at the technical review after the comment is added

src/base/bootstrap-overrides/_base.scss
EricDunsworth: IMO that's kind of weird, but I'll update this PR to add a comment along those lines.

EricDunsworth:
@duboisp Updated :)!

duboisp (approved):

PR #8772, opened 2019-11-27
Pagination Example (Bug Fix)

### LINKED ISSUE #8771: Datatables pagingType: simple is hidden..
I don't believe this has been addressed, but I'm wondering if it was intentional or an oversight (possibly an unintended error) that when using the datatables option "pagingType", for "simple" the script adds the class "hidden" to the pagination list which hides the Previous/Next buttons.   
None of the other pagination types are hidden, so I'm wondering if this was targeted, and if so, why?

If it was unintentional, I'm having difficulty finding the part of the script that adds the class specifically to that type and none others.  I'm hoping that could be fixed in a future update.  

Thanks.

ghost:
Fixes #8771 

I have reconfirmed all paging types work (with patch).

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -132,11 +132,28 @@ $document.on( "draw.dt", selector, function( event, settings ) {
 	var $elm = $( event.target ),
 		pagination = $elm.next( ".bottom" ).find( "div:first-child" ),
 		paginate_buttons = $elm.next( ".bottom" ).find( ".paginate_button" ),
+		pbLength = paginate_buttons.length,
+		pHasLF = pagination.find( ".last, .first" ).length === 2,
+		pHasPN = pagination.find( ".previous, .next" ).length === 2,
 		ol = document.createElement( "OL" ),
 		li = document.createElement( "LI" );
 
 	// Determine if Pagination required
-	if ( paginate_buttons.length === 1 || ( pagination.find( ".previous, .next" ).length === 2 && paginate_buttons.length < 4 ) ) {
+	if (
+		pbLength === 1 ||
+		(
+			pbLength === 3 &&
+			(
+				pHasLF ||
+				pHasPN
+			)
+		) ||
+		(
+			pbLength === 5 &&
+			pHasLF &&
+			pHasPN
+		)
+	) {
 		pagination.addClass( "hidden" );
 	} else {
 


duboisp:
Pre-approved after successful local testing.

src/plugins/tables/tables.js
```diff
@@ -136,7 +136,7 @@ $document.on( "draw.dt", selector, function( event, settings ) {
 		li = document.createElement( "LI" );
 
 	// Determine if Pagination required
-	if ( paginate_buttons.length === 1 || ( pagination.find( ".previous, .next" ).length === 2 && paginate_buttons.length < 4 ) ) {
+	if ( paginate_buttons.length === 1 || ( ( pagination.find( ".last, .first" ).length === 2 || pagination.find( ".previous, .next" ).length === 2 ) && paginate_buttons.length === 3 ) || ( pagination.find( ".last, .first" ).length === 2 && paginate_buttons.length === 5 ) ) {
```
duboisp: Is it possible to simplify that conditional?

May be doing some variable caching to avoid to reevaluate the following expression

* ``` paginate_buttons.length ```
* ``` pagination.find( ".last, .first" ).length === 2```

I am suggesting to start the evaluation with ``` paginate_buttons.length ``` first then the other conditional. something like:
```
	paginate_buttons.length === 1
		|| 
	( 
		paginate_buttons.length === 3  &&
		( 
			pagination.find( ".last, .first" ).length === 2 || 
			pagination.find( ".previous, .next" ).length === 2 
		)
		
	)
		|| 

	( 
		paginate_buttons.length === 5 &&
		pagination.find( ".last, .first" ).length === 2 
	)
```

duboisp (changes requested):

thekodester:
Probably won't be able to test and update till after the holidays, but we'll see.

thekodester:
Also pagingTypes can be found here: https://datatables.net/reference/option/pagingType.

src/plugins/tables/tables.js
thekodester: Fixed

thekodester:
All current changes have been updated.

GormFrank (changes requested):
@thekodester I think what we are trying to achieve here is optimizing the operation itself, making it more readable as a consequence, and not just making it more readable by adjusting the spacing. 

For instance, you are using  `paginate_buttons.length` thrice, so it may be a good idea to isolate it in a variable first and compare against that var. Also, it seems like you could probably use a "modulo 2" to check if the length is an odd number; that could simplify the `if` statement so you would need less conditions (I haven't tried it out, just throwing out ideas here).

thekodester:
It has to be a specific number. I can make `paginate_buttons.length` a variable, but the odd number thing we'll just break it.

GormFrank:
> It has to be a specific number. I can make `paginate_buttons.length` a variable, but the odd number thing we'll just break it.

No problem! The variable thing will make it more optimized and clear. There is also the `pagination.find( ".last, .first" ).length === 2` that is being evaluated twice; this could also be put in a variable.

thekodester:
Made requested changes, retested using `<table class="wb-tables table table-striped table-hover" data-page-length='5' data-wb-tables='{ "pagingType" : "full_numbers" }'>` (tested using page length of 5, 10, and 25 (with a list of 25 items).

GormFrank (approved):

duboisp:
Quick question: How I can reproduce the issue with the wet-boew working example?

thekodester:
Change the page length between 5, 10, and 25 plus add the pagingType `<table class="wb-tables table table-striped table-hover" data-page-length='5' data-wb-tables='{ "pagingType" : "full_numbers" }'>`. I used the 'Pagination Example'.

All pagingTypes found here: https://datatables.net/reference/option/pagingType

GormFrank:
@namjohn920 Can you reproduce?

thekodester:
Test code was posted in https://github.com/wet-boew/wet-boew/issues/8771.

namjohn920:
image
I tested the code that was posted on #8771 by @thekodester .
Yes the pagination buttons become hidden as soon as the "pagingType" becomes "simple"

duboisp:
We would need to update the logic, as per @namjohn920 screen shoot, we only see 10 items and there is 15 items and in that situation we should see the navigation button.

namjohn920:
I added the Js code @thekodester added in this pull request and the bug is now fixed.
Screenshot 2020-01-30 13 46 35

duboisp:
Pre-approved at the technical review.

duboisp (approved):

PR #8774, opened 2019-11-28
BROKEN: Add WhatsApp (Share Plugin)

ghost:
Allows WhatsApp as a share option.


duboisp:
We would need to add the logo in the shared plugin CSS

duboisp (changes requested):
We would need to add the logo in the shared plugin CSS

thekodester:
I could not update the shared plugin CSS, it was being ignored by git, and seem to auto generate when I ran grunt.

thekodester:
I ignored the git ignore and submitted the file.

thekodester:
Actually looking at https://github.com/wet-boew/wet-boew/tree/master/src/plugins/share that css file is not even listed.

PR #8775, opened 2019-11-28
Replace Bitly with TinyURL (Share Plugin)

ghost:
Allows short url without having to make an account (Bitly requires an account now).


duboisp:
Why we should replace bit.ly by tinyURL?

delisma:
@thekodester could we just add TinyURL as an option? That way we don't break the plugin for those who have a bit.ly account.
> Why we should replace bit.ly by tinyURL?

thekodester:
The bit.ly link doesn't do anything any more. It just takes you to the bit.ly website.

thekodester:
I created a free account and tried the url and it does nothing.

https://bitly.com/a/bitmarklet?u=https%3A%2F%2Fwet-boew.github.io%2Fwet-boew%2Fdemos%2Fshare%2Fshare-en.html

duboisp (approved):

duboisp:
Thanks for the explanation @thekodester

PR #8800, opened 2020-01-07
Proposed solution for "Twitter feed not loading in IE"

PierreCorriveau:
For those with AEM or another CMS that does not let you edit the code, they could simply update their wet-beow.js with that new twitter.js version.

Note: Since our current website does not have WET 4 and that we are not on Canada.ca, the language part starting at "var pathName" has not been tested but should be working. Please adapt the condition (if, else if) as needed to the URL that you have on your website.

### DIFF
--- a/src/plugins/twitter/twitter.js
+++ b/src/plugins/twitter/twitter.js
@@ -13,23 +13,21 @@
  * not once per instance of plugin on the page. So, this is a good place to define
  * variables that are common to all instances of the plugin on a page.
  */
+
 var componentName = "wb-twitter",
 	selector = "." + componentName,
 	initEvent = "wb-init" + selector,
 	$document = wb.doc,
-
 	/**
 	 * @method init
 	 * @param {jQuery Event} event Event that triggered the function call
 	 */
 	init = function( event ) {
-
 		// Start initialization
 		// returns DOM object = proceed with init
 		// returns undefined = do not proceed with init (e.g., already initialized)
 		var eventTarget = wb.init( event, componentName, selector ),
 			protocol = wb.pageUrlParts.protocol;
-
 		if ( eventTarget ) {
 			Modernizr.load( {
 				load: ( protocol.indexOf( "http" ) === -1 ? "http:" : protocol ) + "//platform.twitter.com/widgets.js",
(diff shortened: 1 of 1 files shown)


duboisp (changes requested):
I don't think that browser detection code is needed.

@PierreCorriveau Are you able to come at one of the weekly code sprint? We would be able to discuss about it.

PierreCorriveau:
> I don't think that browser detection code is needed.
> 
> @PierreCorriveau Are you able to come at one of the weekly code sprint? We would be able to discuss about it.

Hi Pierre D., over to you and your team now to make a decision on that request. This was just to answer a request from StdGit in hope that people could still use the Twitter feed to meet the accessibility req. in AEM.

Either way, it will not affect us at lobbying and my workload due to migration to a new CMS is keeping me busy.

Cheers!

GormFrank:
**The following is the best approach:**

If the browser is IE11 (using property `wb.ie11` in core), then hide the loader in CSS by changing a few rules related to the `.twitter-timeline` class, which are the `min-height` should be changed to 0, the `::before` pseudo-element should be display none and same with the `::after`.

We do not need to add a message for users to tell them that we do not support it. There will be the link to the Twitter feed on Twitter, and that is considered enough. You should then also include a brief rationale as a comment in your code about the fact that Twitter not supported for IE11 anymore.

Does that make sense?

PierreCorriveau:
> **The following is the best approach:**
> 
> If the browser is IE11 (using property `wb.ie11` in core), then hide the loader in CSS by changing a few rules related to the `.twitter-timeline` class, which are the `min-height` should be changed to 0, the `::before` pseudo-element should be display none and same with the `::after`.
> 
> We do not need to add a message for users to tell them that we do not support it. There will be the link to the Twitter feed on Twitter, and that is considered enough. You should then also include a brief rationale as a comment in your code about the fact that Twitter not supported for IE11 anymore.
> 
> Does that make sense?

Even if the WET doesn't not support <IE11 versions, it's up to you if you want to do so. IMO, you should check your Web analytics before making such a decision.

I agree that the Twitter link is enough but from a communication point of view, they think it's required at our place. That said, you can implement it the way you want to. Either way is fine for accessibility. You might want to consult with your comms team before.

Feel free to propose your changes to this code. This code was for people on AEM, not us.

EricDunsworth:
FYI #9529 sorta fulfilled this PR's goal. But not exactly as envisioned.

It snuck in a change that prevents modernizr from trying to load in Twitter's widget script if ``wb.ie11`` is truthy. Combined with other changes that changed when the loading icon appears, IE11 will now just revert to showing a plain old Twitter link (without an endless loading icon). If anyone's still using it in 2023 😇.

By comparison, this PR would've applied to any version of IE and explained why the widget hadn't loaded.

Garneauma:
Closing as IE is no longer a supported browser and #9529 fixed the Twitter timeline using another solution.

PR #8807, opened 2020-01-13
maxWidth function is added

### LINKED ISSUE #8586: Filter: Add max-width on filter input box
Re the filter script https://wet-boew.github.io/v4.0-ci/demos/filter/filter-en.html , the filter input box stretches the full width of the element it's inside.  In most cases, that would be the full size of the page 1170px.  This creates poor usability on the box, as inputs should be constrained the roughly the size of what the user would be able to put in the box (it would be like an input that is 1000px wide when all it wants in your phone number).  Can you restrict the size to something like max-width: 400px?

namjohn920:
maxWidth feature is now implemented. 
If user puts maxWidth attributes, the filter's width will change accordingly. 
fixes #8586

### DIFF
--- a/src/plugins/filter/base.scss
+++ b/src/plugins/filter/base.scss
@@ -1,3 +1,0 @@
-.wb-fltr-out {
-	display: none !important;
-}
--- a/src/plugins/filter/filter.js
+++ b/src/plugins/filter/filter.js
@@ -38,10 +38,11 @@ var componentName = "wb-filter",
 			$elm, elmTagName, filterUI, prependUI,
 			settings, setDefault,
 			inptId, totalEntries;
+
 		if ( elm ) {
 			$elm = $( elm );
-
 			elmTagName = elm.nodeName;
+
 			if ( [ "DIV", "SECTION", "ARTICLE" ].indexOf( elm.nodeName ) >= 0 ) {
 				setDefault = defaults.grp;
 				prependUI = true;
--- a/src/base/_wet-boew.scss
+++ b/src/base/_wet-boew.scss
@@ -190,6 +190,7 @@
 	@import "views/screen-sm-min";
 	@import "views/_bootstrap-4-screen-sm-min";
 	@import "proximity/screen-sm-min";
+	@import "../plugins/filter/screen-sm-min";
 }
 
 /* Medium view and over */
@@ -204,6 +205,7 @@
 @media screen and (min-width: $screen-lg-min) {
 	@import "views/screen-lg-min";
 	@import "proximity/screen-lg-min";
+	@import "../plugins/filter/screen-lg-min";
 }
 
 /* Extra-extra-small view */
--- a/src/plugins/filter/_base.scss
+++ b/src/plugins/filter/_base.scss
@@ -1,0 +1,9 @@
+.wb-fltr-out {
+	display: none !important;
+}
+
+.wb-filter {
+	.input-group {
+		max-width: 100%;
+	}
+}
--- a/src/plugins/filter/_screen-lg-min.scss
+++ b/src/plugins/filter/_screen-lg-min.scss
@@ -1,0 +1,6 @@
+//Screen size below medium
+.wb-filter {
+	.input-group {
+		max-width: 60%;
+	}
+}
--- a/src/plugins/filter/_screen-sm-min.scss
+++ b/src/plugins/filter/_screen-sm-min.scss
@@ -1,0 +1,6 @@
+//Screen size medium
+.wb-filter {
+	.input-group {
+		max-width: 80%;
+	}
+}


src/plugins/filter/filter.js
```diff
@@ -83,7 +91,8 @@ var componentName = "wb-filter",
 
 			totalEntries = $elm.find( ( settings.section || "" ) + " " + settings.selector ).length;
 
-			filterUI = "<div class=\"input-group\"><label for=\"" + inptId + "\" class=\"input-group-addon\"><span class=\"glyphicon glyphicon-filter\" aria-hidden=\"true\"></span> " + i18nText.filter_label + "</label><input id=\"" + inptId + "\" class=\"form-control " + inputClass + "\" data-" + dtNameFltrArea + "=\"" + elm.id + "\" type=\"search\"></div>" + "<p aria-live=\"polite\" id=\"" + elm.id + "-info\">" + infoFormater( totalEntries, totalEntries ) + "</p>";
+			filterUI = "<div style=\"max-width:" + maxWidth + ";\" class=\"input-group\"><label for=\"" + inptId +
```
GormFrank:
```suggestion
			filterUI = "<div " + (maxWidth ? "style=\"max-width:" + maxWidth + ";\" : "") + "class=\"input-group\"><label for=\"" + inptId +
```

GormFrank (changes requested):

src/plugins/filter/filter.js
```diff
@@ -37,11 +37,19 @@ var componentName = "wb-filter",
 		var elm = wb.init( event, componentName, selector ),
 			$elm, elmTagName, filterUI, prependUI,
 			settings, setDefault,
-			inptId, totalEntries;
+			inptId, totalEntries,
+			attr,
```
GormFrank: You do not need to have `attr` if it is only for `maxWidth`. `maxWidth` will do the trick just fine.

src/plugins/filter/filter.js
```diff
@@ -37,11 +37,19 @@ var componentName = "wb-filter",
...
+			maxWidth;
+
 		if ( elm ) {
 			$elm = $( elm );
-
 			elmTagName = elm.nodeName;
+			attr = elm.getAttribute( "maxWidth" );
+
+			if ( attr !== null ) {
```
GormFrank: Following that logic, you won't need to have this if

GormFrank (changes requested):

GormFrank (approved):

src/plugins/filter/filter.js
```diff
@@ -37,11 +37,14 @@ var componentName = "wb-filter",
...
 			settings, setDefault,
-			inptId, totalEntries;
+			inptId, totalEntries,
+			maxWidth;
+
 		if ( elm ) {
 			$elm = $( elm );
-
 			elmTagName = elm.nodeName;
+			maxWidth = elm.getAttribute( "maxWidth" );
```
duboisp: There is no attribute "maxWidth" for the host element that would.

src/plugins/filter/filter.js
```diff
@@ -83,7 +86,8 @@ var componentName = "wb-filter",
 
 			totalEntries = $elm.find( ( settings.section || "" ) + " " + settings.selector ).length;
 
-			filterUI = "<div class=\"input-group\"><label for=\"" + inptId + "\" class=\"input-group-addon\"><span class=\"glyphicon glyphicon-filter\" aria-hidden=\"true\"></span> " + i18nText.filter_label + "</label><input id=\"" + inptId + "\" class=\"form-control " + inputClass + "\" data-" + dtNameFltrArea + "=\"" + elm.id + "\" type=\"search\"></div>" + "<p aria-live=\"polite\" id=\"" + elm.id + "-info\">" + infoFormater( totalEntries, totalEntries ) + "</p>";
+			filterUI = "<div " +  ( maxWidth ? "style=\"max-width:" + maxWidth + ";\"" : "" ) + "class=\"input-group\"><label for=\"" + inptId +
```
duboisp: We should avoid inline style. Use the plugin SASS

https://github.com/wet-boew/wet-boew/blob/master/src/plugins/filter/base.scss

duboisp (changes requested):

duboisp:
Add 2 breaking point by default:
md device => 80%
lg & xl device => 60%

src/plugins/filter/base.scss
```diff
@@ -1,3 +1,28 @@
...
+
+//Screen size large/x-large
+.wb-filter {
+	.input-group {
+		max-width: 60%;
+	}
+}
+
+//Screen size medium
+@media screen and (max-width: 960px) {
```
GormFrank: Look at the Tabs plugin for how to deal with media queries in a plugin.

Also, your SASS file name should start with an underscore.

GormFrank (changes requested):

GormFrank (approved):

duboisp (approved):

PR #8809, opened 2020-01-16
geomap row added for better spacing of buttons and input fields

### LINKED ISSUE #8777: Geomap: WET 4.0.31 smaller map (grid col = 6) causes buttons to overlap
When the map is at a grid = 6, the buttons for the 'how to use map description' i.e., 'Draw box..., 'Clear' and 'Draw' overlap. Perhaps when the map is this size, the buttons should also adjust positions

namjohn920:
When grid size is reduced from 9 to 6, only tiny space is left for three buttons and it inevitably leads to unorganized spacing or overlapping between buttons. 
I instead inserted a row and it gives an organized look regardless of grid size. 
fixes #8777

### DIFF
--- a/src/plugins/deps/geomap-lib.js
+++ b/src/plugins/deps/geomap-lib.js
@@ -16,6 +16,7 @@ var componentName = "wb-geomap",
 	mapArray = [],
 	mobile = false,
 	i18n, i18nText, tooltip,
+	mapContainerWidth,
 
 	/*
 	 * Plugin users can override these defaults by setting attributes on the html elements that the
@@ -44,6 +45,8 @@ var componentName = "wb-geomap",
 			className = elm.className.split( /\s+/ ),
 			settings = {},
 			$elm, overrides;
+
+		mapContainerWidth = document.querySelector( ".wb-geomap-map" ).parentElement.clientWidth;
 
 		if ( /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test( navigator.userAgent ) ) {
 			mobile = true;
@@ -1122,7 +1125,12 @@ var componentName = "wb-geomap",
 
 		var interaction = new ol.interaction.DragBox(),
 			projLatLon = new ol.proj.Projection( { code: "EPSG:4326" } ),
-			projMap = geomap.map.getView().getProjection();
+			projMap = geomap.map.getView().getProjection(),
+			aoiInputLength,
+			aoiButtonLength;
+
+		aoiInputLength = mapContainerWidth > 600 ? 2 : 3;
+		aoiButtonLength = mapContainerWidth > 600 ? 4 : 12;
 
 		// Handle the draw end event
 		interaction.on( "boxend", function() {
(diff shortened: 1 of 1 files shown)


GormFrank (approved):
This will however put the controls on two lines, no matter what the view port is. I saw the result and personally think that is looks good. If everyone agrees, we'll move on and merge this request during technical review.

duboisp:
Need to create a working example for testing.

duboisp (changes requested):
For the input: Change `col-md-2` for `col-lg-2` and add `col-md-4`

namjohn920:
The code is now changed that only if the map container width goes below a certain size, it becomes two rows.

src/plugins/deps/geomap-lib.js
```diff
@@ -45,6 +46,8 @@ var componentName = "wb-geomap",
 			settings = {},
 			$elm, overrides;
 
+		mapContainerWidth = document.querySelector( ".wb-geomap-map" ).parentElement.clientWidth;
```
GormFrank: We'll have to discuss about this.

GormFrank (changes requested):

src/plugins/deps/geomap-lib.js
```diff
@@ -1162,7 +1165,8 @@ var componentName = "wb-geomap",
 		var aoiDiv = $( "#geomap-aoi-" + geomap.id ),
 			extent, left, bottom, right, top, geomProj;
 
-		aoiDiv.append( "<fieldset id='form-aoi-" + geomap.id + "'>" +
+		if ( mapContainerWidth > 600 ) {
```
GormFrank: Just put the numbers into variables instead of duplicating the append statement.

GormFrank (changes requested):

GormFrank (approved):

duboisp (changes requested):
We would need to discuss how it is done. we would need to find a way to not rely on a hard coded fixed width value.

duboisp:
For this one, as we can't agree. An alternative solution would be to add some documentation to prevent the situation. Like mention to remove the draw box feature when the map is not used in full width screen.

duboisp:
https://github.com/wet-boew/wet-boew/blob/master/site/pages/docs/ref/geomap/geomap-en.hbs

duboisp:
Just add a note that for the AOI (Area of Interest) feature, we recommend the geomap plugin to use the full width on the page.

namjohn920:
> Just add a note that for the AOI (Area of Interest) feature, we recommend the geomap plugin to use the full width on the page.

Should i discard the changes done up to this point?

duboisp:
@namjohn920 you can create a new PR or change the commit in this one. The PR will contain a modification to:
* doc/ref/geomap/geomap-en.hbs
* doc/ref/geomap/geomap-fr.hbs
with the recommendation that AOI should be use when the map take that full width of the content page area.

GormFrank:
> 
> 
> @namjohn920 you can create a new PR or change the commit in this one. The PR will contain a modification to:
> 
>     * doc/ref/geomap/geomap-en.hbs
> 
>     * doc/ref/geomap/geomap-fr.hbs
>       with the recommendation that AOI should be use when the map take that full width of the content page area.

@namjohn920 Need some guidance on this?

Issue #8688, opened 2019-07-18
Form Validation on Required Fields: Displaying required message too early for some controls


sproules:
For file selection (type="file") that is set as required (required="required"), the "The form could not be submitted"/"This field is required" validation messages appears when clicking the Browse (Edge)/Choose File (Chrome) button. Similarly, for dropdown list (select tag) that is set as required, the same validation messages appears when the user switches from a valid option to the "header" option like "Please select an option".  The required validation messages should be displayed once the user submits their information, not as they're trying to enter their information.

Looking at the code in GitHub, the problem appears to be in "formvalid.js" with the following block of code:

/* fixes validation issue (see PR #7913) */
$form.on( "change", "input[type=date], input[type=file], select", function() {
	$form.validate().element( this );
} );

thekodester:
I would argue it not really an issue, as when you click on a file input the browser becomes disabled until you select a file, once you select a file the error is cleared, if you cancel the file request you be informed that a file is required for that form.

juleskuehn:
I agree that the validation is too eager and negatively impacts usability. In my order of preference, validation should occur on:
- submission
- field blur
- field change, with delay

EricDunsworth:
I agree with @juleskuehn.

I feel like the status quo is way too aggressive to the point of making it annoying to fill out a typical form.

I personally prefer the way in which form validation originally worked in WET 4.0. Errors would only appear immediately if a user inserted something that was outright wrong (like an invalid email address) and focused out. Required fields were only validated upon submission.

In other words, users could fill out a form in whatever order they felt like, tab over empty required fields, temporarily change required fields to be empty, etc without randomly triggering required field validation. It was assumed that by the time a user actually tried to submit their form, they'd have filled out all required fields (and only bug them if needed at that point).

Currently, all of the aforementioned situations would trigger required validation in realtime and cause the errors to constantly shift form fields downwards (and upwards once corrected). IMO that makes it really, really annoying to fill out and navigate through WET's forms.

juleskuehn:
Good point about the shifting around @EricDunsworth 
In the current implementation of error messages (without changing to floating call-outs etc) the only solution is to change the behaviour to validate only on form submission.

grifare:
As of now, there are 61 open issues discussing this "form validation" topic! An alternative validation system that would be accepted might close them all. Perhaps we should create a workgroup and resolve this matter once and for all.
[...]

Issue #8160, opened 2017-10-02
Form validation: validating onfocusout, WCAG Compliant?


Karomar:
Hi, 

I am wondering if the functionality of triggering the validation on the onfocusout event is WCAG compliant and if so where I could find some information or the specific WCAG rules specifying that this is permitted?

The reason I am asking is in our department we have an in-house WCAG group that has indicated to us that this is not WCAG compliant and that all such validation should only be triggered when the user hits the Save button, or navigates away from the page. I am trying to challenge this but am having a hard time finding WCAG documentation to support my argument.

Is there an easy way to turn off the onfocusevent validation trigger?

```javascript

// The jQuery validation plug-in in action
validator = $form.validate( {
    ...
    ...
    /* adds on tab validation */
    onfocusout: function( element ) {
        this.element( element );
    }
} ); /* end of validate() */
```

Thanks!

duboisp:
@Karomar can you describe what specific WCAG Success Criteria it is failling and which techniques and faillures, if any, are applicable to that accessibility issue?

rlambert27:
It would be great to ask your WCAG group to cite a particular success criteria that fails or even a common failure that has been achieved.

WCAG actually does encourage automatic error checking through SC 3.3.1 and this includes a couple of sufficient techniques such as:

G83 - https://www.w3.org/TR/WCAG20-TECHS/G83.html
G85 - https://www.w3.org/TR/WCAG20-TECHS/G85.html
SCR18 - https://www.w3.org/TR/WCAG20-TECHS/SCR18.html
SCR32 - https://www.w3.org/TR/WCAG20-TECHS/SCR32.html

IMO it's not up to you to challenge this requirement but up to your WCAG group to provide more information on why this fails accessibility. The SC 3.3.1 is pretty darn clear that automatic error checking is perfectly acceptable if done properly.

Karomar:
Thanks @rlambert27! , this is exactly the kind of information I was looking for to go back to them with. I agree that it shouldn't be up to me to tell them what is and isn't compliant but unfortunately they are a roadblock in our department for getting applications approved.

I will let you know what they come back with as it seems like this is perfectly acceptable behavior.

Karomar:
Based on the information cited from rlambert27 our WCAG group determined that this is perfectly valid functionality and is indeed WCAG so thanks for you help!

Issue #8111, opened 2017-08-11
Form Validation Issue


NoisyMonk:
Hello,

I'm having some issues with form validation.

Essentially, the "required field" form validation doesn't kick in the first time that I switch focus away to another field. The only way it kicks in is with the following pattern;

1.  Set focus on the input field "First Name" (currently empty)
2.  Enter in some text
3.  **Leave the "First Name" field (put focus on some other field ("Last Name" for example))**
4.  Go back to the "First Name" field, and delete the text
5.  Switch away to another field again
6.  "Required" validation error activates

Under any other scenario, the validation doesn't work. The important step is 3 - i.e. I must leave the field after entering something, then come back to the field, then remove the text, and then leave the field again to get the "required" error.

My assumption from the examples I've seen is that even if I set focus on an empty field and leave without entering anything, the "required" error should be displayed. But for whatever reason I can't get that to work. Any ideas?

**WET version**: 2.0.24
**Browser**: IE 11.0.96 (unfortunately I don't have access to any other browsers to test on)

kodecount:
I would recommend updating your WET. Their been many improvements to form validation and security vulnerabilities fixed.

NoisyMonk:
Fixed when upgraded to 2.0.25.

This can be closed.

Thanks!

EricDunsworth:
@NoisyMonk FYI you should have the ability to close this issue directly by clicking the "Close issue" button at the bottom of this page.

Issue #8066, opened 2017-06-21
Problem with required field form - no alert


marieeved:
Hi,

Can someone help me find the mistake that I've done in this code?
When you click on a field form like "titre" and you write nothing in this field and you click outside this field, there is no alert to said that this field is required. It was working before. I saw it. It was like in the examples on this [wet page](http://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-fr.html). but now, ther eis no alert. Other problem: for the date field. If I enter 2017-15-31, I have no alert to tell me that is a date invalide (15e month doesn't exist) and again, this morning it was working too.
Thank you!!!

```
<div class="wb-frmvld">
    

<form enctype="application/x-www-form-urlencoded" method="post">
<p class="mrgn-bttm-lg">Texte explicatif du formulaire...</p>


   <div class="form-group">
			<label for="title1" class="required"><span class="field-name">Bureau de première responsabilité</span> <strong class="required">(obligatoire)</strong></label>
			<select class="form-control" id="bpr" name="bureau de première responsabilité" required="required">
				<option label="Veuillez choisir un BPR"></option>
                        <option value="dosi1">EXT</option>
                    <option value="dosi2">ACEN</option>
                    	<option value="desi">ATL</option>
                    <option value="deosi">ONT</option>
                    <option value="dosi1">OUEST</option>
                    
				<option value="dscp">DSCP</option>
                	<option value="dgscp">DGSCP</option>
                    <option value="deosc">DEOSC</option>
				    <option value="deop">DEOP</option>
                    <option value="deoppt">DEOPPT</option>
                     <option value="dgscp">BSLGICN</option>
                    <option value="deosc">LATON</option>
				    <option value="deop">ESTRIEM</option>
                    <option value="deoppt">MTLAVAL</option>
                     <option value="deoppt">QCSAGLJ</option>
                    <option value="dgscp">CQML</option>
                    <option value="deosc">PROGDEC</option>
				    <option value="deop">PROGCEN</option>
                    <option value="deoppt">CTPG</option>
                     <option value="dgscp">DOPBRML</option>
                    <option value="deosc">DOPBRQ</option>
				    <option value="deop">DSFPPT</option>
                  
                <option value="dsvp">DSVP</option>
                	    <option value="deop">DGSVP</option>
                    <option value="deoppt">DEOSVP</option>
                     <option value="deoppt">DEOSV</option>
                    <option value="dgscp">DEOAE</option>
[...]

PR #8815, opened 2020-01-24
Form validation: Reduce aggressiveness of onfocusout validation

### LINKED ISSUE #8353: Validating onfocusout usability concerns
This has been already discussed here  #8160 but from WCAG compliance perspective.
This behavior seems to be causing a lot of issues and confusion for many people.

Automatically modifying the label to include error messages is causing the page to shift vertically. It’s very annoying and a lot of people complain that they ended up completing the wrong fields as a result…which in return created more validation errors. 
Also we received many comments that this behavior is distracting and people lose the focus especially on longer forms. 
We believe this became a usability issue and we wonder if the negatives outweigh positives.

We received the same comments from different people involved in our latest project (testers, clients, developers).
Testers and clients raised it as an issue and filed it as a bug multiple times. We spent a lot of times explaining and justifying the feature but it keeps coming back.

I also think that heading text “The form could not be submitted because (#) error was found.” is misleading and confusing. The text makes sense only if the user is trying to submit a form with invalid or missing mandatory fields. The text is misleading if the user did not click on Submit yet. It could be argued that this is usability issue as well.

I don’t mean to challenge the validity of this feature but I am wondering if this automatic validating onfocusout could be made optional somehow.

EricDunsworth:
This partially-reverts #7921.

One of its changes caused form validation to always get triggered when focusing out of fields. The end result was that required field validation became drastically more aggressive, arguably to the point of impeding usability.

For example:
* Tabbing through required inputs, checkboxes or radio buttons that had never received any user input would immediately trigger required field errors
* Ticking and then unticking any required checkbox would immediately trigger a required field error

Examples of situations where that would annoy users:
* Keyboard user getting bombarded with required field errors while tabbing through a form
* Mouse/Touch screen user accidentally clicking or touching a random required field, then having to deal with fields visually-shifting downwards

According to the jQuery Validation Plugin's reference documentation (https://jqueryvalidation.org/reference/#link-validation-event), the plugin is meant to provide quick contextual error feedback, without going so far as to annoy users. It considers displaying error messages without having received user input to cross that line.

I personally find the behaviour I'm reverting to be really annoying and aren't the only one that dislikes it... That behaviour was also brought up as a concern by a usability subject matter expert that reviewed a form project I'm currently helping out with.

Fixes #8353 and partially-addresses #8688.

Related to #8111, #8160 and #8066.

CC @biljanap @RobJohnston @duboisp @thekodester @sproules @juleskuehn @lisafast @NoisyMonk @marieeved @rlambert27 @andrewnordlund @shawnthompson

### DIFF
--- a/src/plugins/formvalid/formvalid.js
+++ b/src/plugins/formvalid/formvalid.js
@@ -336,11 +336,6 @@ var componentName = "wb-frmvld",
 
 						invalidHandler: function() {
 							submitted = true;
-						},
-
-						/* adds on tab validation */
-						onfocusout: function( element ) {
-							this.element( element );
 						}
 
 					} ); /* end of validate() */


RobJohnston:
Makes sense to me.  Also related to #7394.

biljanap:
I am looking forward for this being resolved. It's still annoying me and everybody else involved. It's constantly being flagged as a bug.

lisafast:
Agree - very confusing and annoying for users to get errors before finishing the form.

EricDunsworth:
Btw I just wanted to mention that even with this PR in place, there will still be some circumstances in which fields will be validated before a user attempts to submit a form. But those will only be triggered when a user really has messed up - like if they've inserted an invalid value.

For example:
* If a user types a letter into a number field, then focuses away, an error will immediately appear
* If a field is already highlighted as having an error, the field will be re-validated in realtime each time anything is typed into it


Here's an excerpt of the [part I linked to in the OP](https://jqueryvalidation.org/reference/#link-validation-event) that covers how the plugin is intended to work:
> **Validation event**
> 
> By default, forms are validated on submit, triggered by the user clicking the submit button or pressing enter when a form input is focused (option ``onsubmit``). In addition, once a field was highlighted as being invalid, it is validated whenever the user types something in the field (option ``onkeyup``). When the user enters something invalid into a valid field, it is also validated when the field loses focus (option ``onfocusout``).
> 
> The goal of these interactions is to provide feedback as early as possible, whilst avoiding user annoyance. Displaying error messages before the user had the chance to even type something is not helpful.

juleskuehn (approved):

EricDunsworth:
Just went over this with @delisma and @GormFrank. They agree that this PR should be the form validation plugin's default behaviour in WET. But they want to add a configuration option to optionally make it aggressive (i.e. #7921's behaviour).

So I'll either revise this PR to do that or open a separate PR to add the extra option.

GormFrank:
@bsouster ,
@duboisp would like to discuss with you on this topic.

EricDunsworth:
I tried looking into adding an extra configuration option earlier, but haven't been able to figure out how to properly implement it to my liking. jQuery validation has [its own built-in ``onfocusout`` method](https://jqueryvalidation.org/validate/#onfocusout) that only kicks in in specific scenarios. Not sure how to go about modifying it to retain its standard behaviour by default and replace it with aggressive behaviour if the extra config option is enabled.

I'm personally against adding this config option since IMO its sole purpose would be to annoy users and impede usability. Pretty much everyone agrees that the status quo isn't ideal, Plus jQuery Validation was never intended to validate this aggressively and its documentation even advocates against what WET is currently doing.

IMO the config option shouldn't be a show-stopper to merge in this PR. I believe the usability benefits on all (or the vast majority of) WET forms would far outweigh the need to add an extra configuration option that activates undesired behaviour.

@GormFrank @delisma Would one of you be willing to add the custom configuration option if you strongly believe in having it available?

duboisp:
We need to evaluate the accessibility impact (evidence) and state formally what is the best default behaviour.

Initially that feature was inserted to improve accessibility for screen reader and it was supported by @neoinsight in 2017. But seeing this to be a reverse opinion now, that means we are missing some evidence. This change suggestion would need to be supported by UX research or/and expertise. After those document published, we would be able to discuss what behaviour is the best for the default and what behaviour should be optional.

duboisp:
@lisafast can you provide the evidence of your testing.

bsouster:
I have no evidence to have a preference for one or the other. I do think the validation error should wait until you leave a set of options (checkbox or radio) to trigger instead of after tabbing out of the first option element of a set of options

GormFrank:
Waiting on Eric's design decision on validation aggressiveness for this one.

EricDunsworth:
Finally got around to sending wet-boew/wet-boew-documentation#262 :).

juleskuehn:
> 
> 
> @GormFrank Finally got around to sending [wet-boew/wet-boew-documentation#262](https://github.com/wet-boew/wet-boew-documentation/pull/262).

I couldn't find the decision document itself? (13.html)?

EricDunsworth:
@juleskuehn Looks like I forgot to add it to my working tree before committing lol. It got buried among a bunch of other untracked files I had there. I'll sneak it in a moment.

EricDunsworth:
@juleskuehn Added :)! Btw thanks for having spotted that.

duboisp:
La [décision de conception DD-13](https://wet-boew.github.io/wet-boew-documentation/decision/13.html) a été présenté au dernier WET roadmap et est en bonne voie pour une approbation à la prochaine réunion. Si nécessaire on fera des ajustements ultérieurements.

duboisp (approved):

PR #8817, opened 2020-01-29
Release note links added in download page

namjohn920:
links for Release note for latest releases are added in a new column.


GormFrank (dismissed):
I'll show you where to add it, but it is not in the Latest releases table.

duboisp (approved):

PR #8829, opened 2020-02-24
wb5-click plugin altered and added

namjohn920:
wb5-click plugin altered and added.
Both documentation and working examples are created.

### DIFF
--- a/site/pages/docs/ref/provisional-en.hbs
+++ b/site/pages/docs/ref/provisional-en.hbs
@@ -44,6 +44,12 @@
 </thead>
 <tbody>
 	<tr>
+		<th>Plugin <code>.wb5-AJAX-search</code></th>
+		<td>This plugin makes an AJAX request when a form is submitted</td>
+		<td>v4.0.34</td>
+		<td><a href="/docs/ref/wb-onpage-submit/wb-onpage-submit-en.html">Detailed documentation</a></td>
+	</tr>
+	<tr>
 		<th>CSS <code>.bg-cover</code></th>
 		<td>Background size cover</td>
 		<td>v4.0.34</td>
--- a/src/plugins/wb-onpage-submit/wb-onpage-submit.js
+++ b/src/plugins/wb-onpage-submit/wb-onpage-submit.js
@@ -1,0 +1,48 @@
+/**
+ * @title WET-BOEW wb-onpage-submit
+ * @overview This plugin implements AJAX request for form data
+ * @license wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
+ * @author @masterbee, @namjohn920
+ **/
+
+( function( $, wb ) {
+"use strict";
+
+var $document = wb.doc,
+	componentName = "wb-onpage-submit",
+	selector = ".provisional." + componentName,
+	initEvent = "wb-init" + selector,
+
+	init = function( event ) {
+		var elm = wb.init( event, componentName, selector );
+
+		elm.addEventListener( "submit", function( e ) {
+			e.preventDefault();
+			var $elm = $( this );
+			try {
+				$.ajax( {
+					type: $elm.attr( "method" ),
+					url: $elm.attr( "action" ),
+					data: $elm.serialize() // serializes the form's elements.
+				} );
+			} catch ( error ) {
+				$( this.querySelector( ".formContents" ) ).toggleClass( "hidden" );
+				$( this.querySelector( ".failureMessage" ) ).toggleClass( "hidden" );
+				throw error;
+			}
+			$( this.querySelector( ".formContents" ) ).toggleClass( "hidden" );
+			$( this.querySelector( ".successMessage" ) ).toggleClass( "hidden" );
+
+			return true;
+		} );
+
+		wb.ready( $( elm ), componentName );
+	};
+
+// Bind the init event of the plugin
+$document.on( "timerpoke.wb " + initEvent, selector, init );
+
+// Add the timer poke to initialize the plugin
+wb.add( selector );
+
+} )( jQuery, wb );
--- a/site/pages/docs/ref/provisional-fr.hbs
+++ b/site/pages/docs/ref/provisional-fr.hbs
@@ -46,6 +46,12 @@
 </thead>
 <tbody>
 	<tr>
+		<th>Plugin <code>.wb5-AJAX-search</code></th>
+		<td>This plugin makes an AJAX request when a form is submitted</td>
+		<td>v4.0.34</td>
+		<td><a href="/docs/ref/wb-onpage-submit/wb-onpage-submit-fr.html">Detailed documentation</a></td>
+	</tr>
+	<tr>
 		<th>CSS <code>.bg-cover</code></th>
 		<td>Background size cover</td>
 		<td>v4.0.34</td>
--- a/site/pages/docs/ref/wb-onpage-submit/wb-onpage-submit-en.hbs
+++ b/site/pages/docs/ref/wb-onpage-submit/wb-onpage-submit-en.hbs
@@ -1,0 +1,97 @@
+---
+{
+	"title": "Onpage Submit",
+	"language": "en",
+	"category": "Plugins",
+	"categoryfile": "plugins",
+	"description": "Makes AJAX request when submitting a form.",
+	"altLangPrefix": "wb-onpage-submit",
+	"dateModified": "2029-02-21"
+}
+---
+<span class="wb-prettify all-pre hide"></span>
+
+<div class="alert alert-warning">
+	<h2>Unstable feature</h2>
+	<p>To be used at <strong>your own risk</strong>. This feature described below can be removed in any subsequent minor/major release</p>
+	<p><small><a href="https://wet-boew.github.io/wet-boew-documentation/decision/7.html">Learn more about the design decision around provisional features.</a></small></p>
+	<p><small><a href="/docs/ref/provisional-en.html">Check other provisional features available.</a></small></p>
+</div>
+
+<section>
+	<h2>Purpose</h2>
+	<p>This plugin makes an AJAX request when a form is submitted.</p>
+</section>
+
+<section>
+	<h2>Use when</h2>
+	<ul>
+		<li>When making AJAX request for submitting a form</li>
+	</ul>
+</section>
+
+<section>
+	<h2>Working example</h2>
+	<ul>
+		<li><a href="../../../demos/wb-onpage-submit/wb-onpage-submit-en.html">English example</a></li>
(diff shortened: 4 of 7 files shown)


site/pages/docs/ref/provisional-en.hbs
```diff
@@ -111,3 +117,136 @@
...
+<div class="col-sm-6 col-md-5 col-lg-4">
+	<details class="brdr-0">
+		<label class="btn btn-default text-center">Report a problem or mistake on this page</label>
+		<div class="clearfix"></div>
+			<div class="well row">
+				<div>
+					<div class="gc-rprt-prblm">
+						<div class="gc-rprt-prblm-frm gc-rprt-prblm-tggl">
+							<form id="gc-rprt-prblm-form" action="/gc/services/notify" method="post">
+								<input type="hidden" name="externalReferer" value="externalref">
```
duboisp: Use a more simple form. Like 2 input field

site/pages/docs/ref/wb5-AJAX-submit/wb5-AJAX-submit-en.hbs
```diff
@@ -0,0 +1,97 @@
...
+	<h2>Use when</h2>
+	<ul>
+		<li>When making AJAX request for submitting a form</li>
+	</ul>
+</section>
+
+<section>
+	<h2>Working example</h2>
+	<ul>
+		<li><a href="../../../demos/wb5-AJAX-submit/wb5-AJAX-submit-en.html">English example</a></li>
```
duboisp: remove the mention of "wb5"

src/plugins/wb5-AJAX-submit/_base.scss
```diff
@@ -0,0 +1,33 @@
+/*
+* Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)
+* wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
+*/
+.wb-steps {
```
duboisp: This looklike CSS for step form, please remove it.

duboisp (changes requested):
Should this functionality be bind to the ```form``` or to the ```button```?
Me I think it should be bind to the form element.

This plugin should be able to show "Success" or "Fail" message via it's configuration or how it can be done in the documentation.. May be leverage the ```<output>```? https://html.spec.whatwg.org/multipage/form-elements.html#the-output-element

As mentioned by Eric, We will need to ensure the screen reader would announce it when the form is submit. We will need to test it with NVDA or/and JAWS.

src/plugins/wb5-AJAX-submit/wb5-AJAX-submit.js
```diff
@@ -0,0 +1,51 @@
...
+ * @overview This plugin implements AJAX request for form data
+ * @license wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
+ * @author @masterbee, @namjohn920
+ **/
+
+( function( $, wb ) {
+"use strict";
+
+var $document = wb.doc,
+	componentName = "AJAX-submit",
```
duboisp, line 12: The component name need to start with "wb-". Probably name it like: "wb-submit-ajax"

src/plugins/wb5-AJAX-submit/wb5-AJAX-submit.js
```diff
@@ -0,0 +1,51 @@
...
+	selector = ".provisional.AJAX-submit",
```
duboisp, line 13: Please use the component name variable instead of duplicating the string

src/plugins/wb5-AJAX-submit/wb5-AJAX-submit.js
```diff
@@ -0,0 +1,51 @@
...
+
+	init = function( event ) {
+		var elm = wb.init( event, componentName, selector );
+
+		elm.addEventListener( "submit", function( e ) {
+			e.preventDefault();
+
+			var $form = $( this );
+
+			console.log( "action", $form.attr( "action" ) );
```
duboisp, line 24: To remove

src/plugins/wb5-AJAX-submit/wb5-AJAX-submit.js
```diff
@@ -0,0 +1,51 @@
...
+
+			try {
+				$.ajax( {
+					type: $form.attr( "method" ),
+					url: $form.attr( "action" ),
+					data: $form.serialize() // serializes the form's elements.
+				} );
+			} catch ( error ) {
+				console.error( error );
```
duboisp, line 33: Use throw

src/plugins/wb5-AJAX-submit/wb5-AJAX-submit.js
```diff
@@ -0,0 +1,51 @@
...
+			}
+
+			//Insert an appropriate sucess message within quotation mark
+			$form.html( "<h1>Thank you! It is successfully submitted</h1>" );
```
duboisp, line 37: Use a plugin config instead. That config could be a jQuery selector that you will happen the text next to the form element.

src/plugins/wb5-AJAX-submit/wb5-AJAX-submit.js
```diff
@@ -0,0 +1,51 @@
...
+	selector = ".provisional.AJAX-submit",
+	initEvent = "wb-init" + selector,
+
+	init = function( event ) {
+		var elm = wb.init( event, componentName, selector );
+
+		elm.addEventListener( "submit", function( e ) {
+			e.preventDefault();
+
+			var $form = $( this );
```
duboisp, line 22: Change `$form` for `$elm`

src/plugins/wb5-AJAX-submit/wb5-AJAX-submit.js
```diff
@@ -0,0 +1,51 @@
...
+				console.error( error );
+			}
+
+			//Insert an appropriate sucess message within quotation mark
+			$form.html( "<h1>Thank you! It is successfully submitted</h1>" );
+
+			return true;
+		} );
+
+		wb.ready( $( elm ), componentName );
```
duboisp, line 42: use `$elm`

duboisp (changes requested):

src/plugins/wb5-AJAX-submit/wb5-AJAX-submit.js
GormFrank: This is also to deal with bilingualism

GormFrank:
Can you please call it wb-onpage-submit instead of AJAX-submit?

src/plugins/wb5-AJAX-submit/wb5-AJAX-submit.js
GormFrank: What you should do instead is to have a class named .wb-onpage-thnk on a hidden element underneath the form, and make it show on submit (take the Report a problem on Canada.ca as an example). Also, it would be nice to also have the possibility to show a failure message the same way as the thank yo message for when the AJAX goes wrong. The default could still be the thank you message.

GormFrank (changes requested):

GormFrank:
> Can you please call it wb-onpage-submit instead of AJAX-submit?

@duboisp Actually change the name to `wb-postback`

src/plugins/wb-onpage-submit/wb-onpage-submit.js
```diff
@@ -0,0 +1,48 @@
...
+		var elm = wb.init( event, componentName, selector );
+
+		elm.addEventListener( "submit", function( e ) {
+			e.preventDefault();
+			var $elm = $( this );
+			try {
+				$.ajax( {
+					type: $elm.attr( "method" ),
+					url: $elm.attr( "action" ),
+					data: $elm.serialize() // serializes the form's elements.
```
duboisp, line 26: we will need to manage the serialize separatell because the button value used to submit is out of scope of the serialization because jQuery said the button are out of scope which is true unless that button is used to submit. So we need to detect if "elm" was the button used to do the submission and if so, we need to serialize it's value.

src/plugins/wb-onpage-submit/wb-onpage-submit.js
```diff
@@ -0,0 +1,48 @@
...
+"use strict";
+
+var $document = wb.doc,
+	componentName = "wb-onpage-submit",
+	selector = ".provisional." + componentName,
+	initEvent = "wb-init" + selector,
+
+	init = function( event ) {
+		var elm = wb.init( event, componentName, selector );
+
```
duboisp, line 18: need to ensure "elm" is valid by 
```
if ( elm ) {

}
```
like in the helloworld plugin example.

src/plugins/wb-onpage-submit/wb-onpage-submit.js
```diff
@@ -0,0 +1,48 @@
...
+			e.preventDefault();
+			var $elm = $( this );
+			try {
+				$.ajax( {
+					type: $elm.attr( "method" ),
+					url: $elm.attr( "action" ),
+					data: $elm.serialize() // serializes the form's elements.
+				} );
+			} catch ( error ) {
+				$( this.querySelector( ".formContents" ) ).toggleClass( "hidden" );
```
duboisp, line 29: Do this as a config and only once, and hidding the form should be just a single boolean parameter

src/plugins/wb-onpage-submit/wb-onpage-submit.js
```diff
@@ -0,0 +1,48 @@
...
+	initEvent = "wb-init" + selector,
+
+	init = function( event ) {
+		var elm = wb.init( event, componentName, selector );
+
+		elm.addEventListener( "submit", function( e ) {
+			e.preventDefault();
+			var $elm = $( this );
+			try {
+				$.ajax( {
```
duboisp, line 23: Use Promise instead of try catch because it is more common with `ajax`.

src/plugins/wb-onpage-submit/wb-onpage-submit.js
```diff
@@ -0,0 +1,48 @@
...
+					url: $elm.attr( "action" ),
+					data: $elm.serialize() // serializes the form's elements.
+				} );
+			} catch ( error ) {
+				$( this.querySelector( ".formContents" ) ).toggleClass( "hidden" );
+				$( this.querySelector( ".failureMessage" ) ).toggleClass( "hidden" );
+				throw error;
+			}
+			$( this.querySelector( ".formContents" ) ).toggleClass( "hidden" );
+			$( this.querySelector( ".successMessage" ) ).toggleClass( "hidden" );
```
duboisp, line 34: use a Jquery selector or a CSS selector, no need to swapping between both JQuery and VanillaJS

duboisp (changes requested):
Use `hide` instead of `hidden`

duboisp:
superseeded by #8850

PR #8836, opened 2020-03-09
[Do not merge] Toggle: Add checkbox support and example

EricDunsworth:
@fsnoddy FYI.

### DIFF
--- a/src/plugins/toggle/toggle-en.hbs
+++ b/src/plugins/toggle/toggle-en.hbs
@@ -8,7 +8,7 @@
 	"parentdir": "toggle",
 	"altLangPrefix": "toggle",
 	"css": ["demo/toggle"],
-	"dateModified": "2015-04-22"
+	"dateModified": "2020-03-10"
 }
 ---
 <section>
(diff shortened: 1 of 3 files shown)


EricDunsworth:
Just force-pushed an update to remove the ``lang-html`` class from the new code sample (to be in line with #8837's incoming changes).

duboisp (commented):
Need testing to see if it is working when clicking on a explicit label, when the label don't wrap the checkbox.

Should we support radio button too? May be we can do that later.

EricDunsworth:
@duboisp I can confirm that as things stand, toggling works properly even when the ``<label>`` isn't wrapped around ``<input type="checkbox">``.

Gonna look into radio buttons... Not yet sure if I'll roll it into this PR or create a separate one.

EricDunsworth:
Forgot to mention this in my last post, but my toggling tests with un-nested checkbox inputs were performed in Firefox/IE11/Edge (Trident)/Opera on Windows 10.

Btw feel free to merge in this PR as-is. IMO radio button support shouldn't be considered a show-stopper. If worst comes to worst, that can be brought in via a separate PR.

EricDunsworth:
I don't think this PR is merge-worthy yet. It doesn't seem to work properly with pre-checked checkboxes (ones that use the``checked`` attribute) and ``aria-controls`` doesn't seem to be added to the checkbox (it's present in all the other examples).

duboisp:
`$elm.prop( attrname, content );`

PR #8843, opened 2020-03-23
Polyfill/datepicker: Fix for clear button on date picker fields

esand:
The clear ("x") button on input fields doesn't appear on date picker inputs because the width is not big enough at 10em's (the button is basically hidden in overflow). By setting the width to 11em, there's enough space for the "x" to appear (which now allows people to clear a date) and the field width is still small enough to fit inside a .col-md-2 as it did before.

### DIFF
--- a/src/polyfills/datepicker/datepicker.scss
+++ b/src/polyfills/datepicker/datepicker.scss
@@ -10,7 +10,7 @@
 
 .wb-date-wrap {
 	&.input-group {
-		width: 10em;
+		width: 11em;
 	}
 }
 


duboisp:
In what circumstance it don't appear? Is it in IE11?
In IE 11 and even when zoomed it is still there.

GormFrank:
@esand Can you provide more information in reply to @duboisp?

esand:
In IE11, go to https://wet-boew.github.io/v4.0-ci/demos/datepicker/datepicker-en.html and enter a date in either date box. Click your cursor *in* to the text box and you'll see no "X" to clear the input field.
Now, in dev tools change the width for the input-group of the parent span to 11em and then click in to the input box again - you'll see the "X" appear and can be used to clear the input field.

GormFrank (approved):

Issue #8854, opened 2020-04-21
Overlay: cannot keep the focus trapped in a modal


dorina1:
When we Shift-TAB in any modals, the focus does not stay in the modal, but it goes behind into the browser top elements and rotates until you press TAB and you go back into the modal: [Overlay](https://wet-boew.github.io/v4.0-ci/demos/overlay/overlay-en.html)

Similar behavior when we TAB, the focus does not stay in the modal : it goes behind and after a few more tabs it comes back in. 

This occurs in multiple browsers (Edge, IE11, Chrome).

Keyboard-access expected behavior for modals should be:
**TAB**
- Moves focus to next focusable element inside the modal.
- When focus is on the last focusable element in the modal, moves focus to the first focusable element in the modal.
**SHIFT-TAB**
- Moves focus to previous focusable element inside the modal.
When focus is on the first focusable element in the dialog, moves focus to the last focusable element in the modal.
**ESC**
- Closes the modal (this works)

EricDunsworth:
This will finally be fixed by #10153.

Looks like it was previously addressed in the lightbox plugin in #8855 (which also benefited the [overlay demo page](https://wet-boew.github.io/v4.0-ci/demos/overlay/overlay-en.html)'s lightbox examples).

**PS:**
I was unable to replicate this behaviour:
> Similar behavior when we TAB, the focus does not stay in the modal : it goes behind and after a few more tabs it comes back in.

Maybe it got fixed at some point between when this issue was originally posted and now?

The overlay plugin previously did a poor job of keeping track of interactive ("tabbable") elements within its container due to buggy logic it borrowed from a super old version of jQuery UI. It caused problems when trying to forward-tab beyond the overlay's X close button _if_ the first interactive element in the container wasn't reachable. That'll be fixed by #10152.

**PS2:**
The OP's expected behaviours correspond to what I'm getting with both of the aforementioned PRs in place 🤓.

Issue #8737, opened 2019-10-10
Overlay: Lightbox and Modal: Tab focus lost


shawnthompson:
When you open a modal / lightbox and use tab to navigate, you lose focus after the last focused item. It returns to the first item after a few more tab presses.

juleskuehn:
Seems like the hidden, focused items are
```html
<span class="lbx-end wb-inv" tabindex="0">
```
One of these is added each time the modal is invoked.

duboisp:
After looking at the git blame for the [line 167](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/lightbox/lightbox.js#L167), that line was to [add focus redirection to prevent Jaws from exiting the Lightbox prematurely](https://github.com/wet-boew/wet-boew/commit/22f7540de619466b4968779641047aef5ab89021). Related to #5009

Next step would be to remove that fix and see if the issue described in #5009 is still accurate today and if this issue get fixed.

PR #8855, opened 2020-04-22
lightbox: trap tabbing within modal while active

Logan-Lim:
tabbing through a modal does not capture the tabbable focus of the user. When a user tabs to the end of a modal, they should be returned to the beginning, and vice-versa.

#8854

### DIFF
--- a/src/plugins/lightbox/lightbox.js
+++ b/src/plugins/lightbox/lightbox.js
@@ -164,11 +164,9 @@ var componentName = "wb-lbx",
 						$content.attr( "role", "document" );
 					}
 
-					$wrap.append( "<span tabindex='0' class='lbx-end wb-inv'></span>" )
-						.find( ".activate-open" )
-						.trigger( "wb-activate" );
-
 					this.contentContainer.attr( "data-pgtitle", document.getElementsByTagName( "H1" )[ 0 ].textContent );
+
+					trapTabbing( $wrap );
 				},
 				close: function() {
 					$document.find( "body" ).removeClass( "wb-modal" );
@@ -245,6 +243,9 @@ var componentName = "wb-lbx",
 								.attr( "id", "lbx-title" );
 
 					mfpResponse.data = $response;
+				},
+				ajaxContentAdded: function() {
+					trapTabbing( this.wrap );
 				}
 			};
 		}
@@ -287,10 +288,29 @@ var componentName = "wb-lbx",
 
 				$( footer ).append( overlayCloseFtr );
 				if ( !hasFooter ) {
-					$modal.append( footer );
+					$( footer ).insertAfter( $modal.find( ".modal-body" ) );
 				}
 			}
 		}
+	},
+	trapTabbing = function( $wrap ) {
+
+		$wrap.on( "keydown", function( e ) {
+			if ( e.which === 9 ) {
+				var tabbable = $wrap.find( ".mfp-container :tabbable:visible" ),
+					firstTabbable = tabbable.first()[ 0 ],
+					lastTabbable = tabbable.last()[ 0 ],
+					currentFocus = $( document.activeElement )[ 0 ];
+
+				if ( !e.shiftKey && currentFocus === lastTabbable ) {
+					e.preventDefault();
+					firstTabbable.focus();
+				} else if ( e.shiftKey && ( currentFocus === firstTabbable || currentFocus === $wrap[ 0 ] ) ) {
+					e.preventDefault();
+					lastTabbable.focus();
+				}
+			}
+		} );
 	};
 
 // Bind the init event of the plugin


duboisp:
Thanks for your contribution, We are going to test your fix for the next week technical review.

GormFrank:
Code works as expected for the lightbox modal popup, as shown on the [overlay working examples page](https://wet-boew.github.io/v4.0-ci/demos/overlay/overlay-en.html), which are those two: "Centred popup (Lightbox)" and "Centred popup - Modal (Lightbox + Modal)".

However, note that it doesn't apply to lightboxes related to images and galleries, but this is probably out of scope here.

Also, more importantly it doesn't work with AJAX popup and seems to be breaking the inline popup when opened through Javascript; both available for testing in the ["Opening popups through JavaScript" section](https://wet-boew.github.io/v4.0-ci/demos/lightbox/lightbox-en.html#examples5) of the Lightbox working examples page.

By breaking, here is what I mean: using the same example with the dropdown to open popups with Javascript, if you select the Modal popup from the dropdown, open it and play with the tabbing, then close it and select the Inline content one, open it and start tabbing, you will get stuck on the **Close** button even though the **X** is also present. A reset of the tabbing sequence at popup close may help prevent that.

GormFrank (changes requested):
To recap on my comment, we need a consistent experience with the tabbing sequence in the different "popup" types.

Logan-Lim:
I encountered another issue as well with having dynamic modal content. For example I have a modal with some validation inside, and when an error summary is triggered it's typically prepended before the first tabbable element. So it ends up not being reachable. What do you think of making the modal header a tabbable element, so there's a consistent start to all modals. 

For the last tabbable element, it seems like we can't necessarily rely on the X being available either. Going to think a bit more about some good options for that one.

Needless to say I'll also look into fixing those bugs you mentioned.

Logan-Lim:
Alright should be ready for another review.

I've made the modal header a tabbable element. This will give a consistent first tabbable element for all modals. Modals should also always have a control for closing the modal, so I don't think we need to do anything to ensure a consistent last tabbable element.

I also fixed a bug causing inconsistent tab order in modal close controls. The createCloseButton callback function was appending itself to the end of the modal dialog, which would sometimes put itself after the mfp-close button (top right X). This behavior can be viewed at https://wet-boew.github.io/v4.0-ci/demos/lightbox/lightbox-en.html when activating the 'inline content' modal on first load. The footer close button will be the last tabbable element, but will change if you open an AJAX modal then switch back. The solution was to insert after the modal body instead of appending to the end.

duboisp:
@GormFrank can you re-test

GormFrank:
Technically it works like a charm, but then I questioned the accessibility/feasability of that solution and rightfully so.

The goal of the lightbox and overlay is to eventually be replaced by the dialog element, meaning that we should get as close as possible to its behaviour. 

That said, the dialog will focus on the first focusable element on "open". A focusable element should be an interactive element like a button, a link, etc. Headings are not meant to be focusable per say, however nothing prevents you (firmly) from adding focus on it with tabindex=0 to get the focus in the popup. But at the very least, it should then lose its focus-ability as soon as you jump to the next element, making it focusable only once and then out of the focus cycle.

For instance, if you look here: https://www.w3.org/TR/html53/editing.html#the-tabindex-attribute there is a clear red flag on using tabindex=0 on non-interactive elements, as it mentions: ...authors should only make elements focusable if they act as interactive controls or widgets...

All that to say, @Logan-Lim can you try declaring the tabbing sequence later on in the opening instead of adding focus to the title to set your first tabbable element? You should wait after the AJAX gets loaded so that you you can look for a first tabbable element then.

Logan-Lim:
Binding the event to an element within the modal body might cause issues with dynamic content. For example if an error summary is created it could be created prior to the first tabbable element. Could bind an event to the entire modal and check if focus should be redirected on each key press. Only issue I can see is it will be less efficient, but it seems like a decent option. I'll implement it and see how it goes.

Logan-Lim:
I've updated the commit with the different approach. An event is now bound to the whole modal. If a tab is detected it will find first and last tabbable elements and check if we are navigating out of the modal then redirecting focus back to the other side if so.

I noticed something strange within the JavaScript. There's an lbx-end span being appended to the end of modals on the open callback (line:167). This is being given a tabindex=0 and is continuously being appended on each open. Not an issue with this PR, and couldn't really figure out what it's purpose is, so I just left it.

GormFrank:
Love the approach. Only issue I can see is if you start off with "Shift + Tab" instead of "Tab" alone; then the focus exits the popup and goes to the browser's URL bar.

Logan-Lim:
Alright should be good to go now. I made the parameter of the trapTabbing function use the modal wrapper. So the initial focused element will be covered in our keydown event. I then added some logic to the shift-tab event, which includes whether the current focus was the wrapper (inital focus).

I saw this [issue](https://github.com/wet-boew/wet-boew/issues/8737) that mentions the lbx-end spans. This PR should remove any use for those, so I've removed the code that appends these.

GormFrank:
Thank you @Logan-Lim. @duboisp and @bsouster What do you think of the approach? It doesn't act the same way as a dialog element would, however it seems to be working fine on every point (usability, operability and accessibility wise). @bsouster I am interested in knowing how that would behave using JAWS, especially the auto-focus on the container. Here is the link to the test:
http://universallabs.org/wet/labs/pr8855/demos/lightbox/lightbox-en.html#examples5

bsouster:
@GormFrank I sent this to a handful of testers... here is one so far, relying on Jaws and keyboard-only navigation:

* When you clicked the button, what happened?
The box opened up as indicated and told me there was an image.  Not sure what the image is, but does show as indicated. 

* Do you know where you are?
Not sure of the function, but yes the speech responds well. 

* What is being shown to you? (it is a test, so it should say ‘image 1’ as the description)
That is exactly what it says, so test successful? 

* Are you able to get out of it?
Definitely able to get out, “esc” works beautifully.

duboisp (approved):

GormFrank (approved):

GormFrank:
Thank you very much @Logan-Lim for this contribution!

Issue #8875, opened 2020-06-02
Links and form controls may not have proper focus outline/ring as of Chromium/Chrome/Edge 83+


michael-milette:
On May 23rd, 2020, Google launched Chrome 83 which includes changes to the default styles we have all known for years, especially for links and form elements.

Unfortunately this has had a negative impact on the WET-BOEW project. For example, the focus outline no longer displays correctly in breadcrumbs on Government of Canada Websites with the GCWeb theme:

![image](https://user-images.githubusercontent.com/3808579/83394497-b42e3280-a3c6-11ea-945d-8c8c3862ebac.png)

It seems that the new thicker black outline no longer fits in the tight space permitted in the GCWeb theme's breadcrumb items.

The following link mentions the change Google recently implemented:
* https://developers.google.com/web/updates/2020/05/nic83#forms
* https://blog.chromium.org/2020/03/updates-to-form-controls-and-focus.html

You can still go back and see what it used to look like using the following experimental settings, but only until Chrome 84:
chrome://flags/#form-controls-refresh

Time to check all your links and UI throughout your websites folks!

Best regards,

Michael Milette

GormFrank:
Merged, but will only be visible when a new version of GCWeb comes out.

StdGit:
@michael-milette @GormFrank 

Hi, I'm not sure to correctly understand the issue here.
Is it the dotted vertical bar between the breadcrumb that are missing?

Does this site https://ssl-templates.services.gc.ca/app/cls/WET/gcweb/v4_0_32/cdts/samples/breadcrumbs-en.shtml have the same issue?

Thx

thekodester:
This is what it use to look like when selected using tab.
<img width="416" alt="Screen Shot 2020-06-08 at 5 23 09 PM" src="https://user-images.githubusercontent.com/23484460/84081886-d0c10080-a9ac-11ea-8a52-f760779655d4.png">

michael-milette:
@StdGit - Yes, the page you mention definitely has the issue:
https://ssl-templates.services.gc.ca/app/cls/WET/gcweb/v4_0_32/cdts/samples/breadcrumbs-en.shtml

![image](https://user-images.githubusercontent.com/3808579/84088536-b2620180-a9ba-11ea-9209-b3d97141a5ca.png)

That is what it currently looks like in Chrome 83 when the CDTS link has focus. There should be a box all the way around. If I give focus to the home link, there isn't even a piece of the outline border at all.

PR #8876, opened 2020-06-02
Breadcrumb: fix issues related to new version of Chrome - focus outline

GormFrank:
As described in #8875

Give a little bit more breathing space for the breadcrumb elements to show the outline that was hidden because of the overflow hidden. Adjusted the spacing around the breadcrumb to keep the same outside spacing as before. However, this adds a bit more space horizontally on the breadcrumb items without causing any layout issue (from what I tested).

### DIFF
--- a/src/base/breadcrumb/_base.scss
+++ b/src/base/breadcrumb/_base.scss
@@ -10,11 +10,13 @@
 		ol {
 			border-radius: 0;
 			margin-bottom: 0;
+			padding: 3px 13px;
 		}
 
 		li {
 			max-width: 100%;
 			overflow: hidden;
+			padding: 5px 2px;
 			text-overflow: ellipsis;
 			white-space: nowrap;
 


duboisp:
We will need to re-test gcweb after the wet-boew version would be updated inside GCWeb.

duboisp (approved):

michael-milette:
Hi @duboisp and @GormFrank  ,

Thanks for contributing the commit. However, I tried your solution on Canada.ca:

image

as well as on the CDTS Application template v4.0.32 (GCWeb theme):

image

As you can see, in both cases I found that the bottom is still slightly cut off (1 of the 2 pixels) and the top is still missing completely.

I would like to offer an alternative solution which is presented here in a format that people can just add to their own CSS while waiting for the next release of GCWeb. It is a minor adjustment to your proposed solution. The difference is in bold:

#wb-bc ol {
      padding: 3px 13px;
}
#wb-bc li {
      padding: **7px** 2px; /* Your version has 5px. */
}

The minor change resolves the issue on every Canada.ca (GCWeb) site that I tried it on:

image

image

Let me know if you want me to submit a pull request for this.

Best regards,

Michael Milette

PR #8879, opened 2020-06-02
Canonical and Hreflang link tags

### LINKED ISSUE #8877: Have the ability to add a canonical link tag, on any selected page, pointing to themselves like in the example above, or to another internal page URL or even an external URL
Canonical is useful for crawlers and good for SEO in general.

The canonical link tag (referred in the text as rel canonical) is a way of telling robots and search engines that a specific URL represents the original or master copy of a page. The use of the rel canonical prevents problems caused by duplicate content appearing on multiple URL.

Even on one page, you agree with me that all of those are the same page:

- https://www.canada.ca/en.html
- http://www.canada.ca/en.html
- https://canada.ca/en.html
- https://canada.ca/en/
- https://canada.ca/en.html?utm_campaign...

Well for crawlers they are all considered as different pages, and canonical sets one as the main one.

Also, rel canonical can point to URLs on other websites. For instance, sometimes news sites post articles that are duplicates or very similar to other ones on other websites, and so they can refer the original article there through the rel canonical.

It is good practice to have every single page with a rel canonical, even if it is pointing to itself. So this should be the default on every single page on canada.ca; to have the rel canonical point to itself. 

For example, the following would be the rel canonical for the english home page:
<link rel="canonical" href="https://www.canada.ca/en.html" />

Language alternatives are not considered as duplication so the same page in French can have the rel canonical on itself too. Again, for the French home page we would have:
<link rel="canonical" href="https://www.canada.ca/fr.html" />

**NOTE:** However, there is a way to tag languages alternatives, as explained in issue #8878...

To recap, the minimum would be to have the ability to add a rel canonical, on any selected page, pointing to themselves like in the example above, or to another internal page URL or even an external URL. 

Resources:
https://yoast.com/academy/technical-seo-training/
https://moz.com/learn/seo/canonicalization

### LINKED ISSUE #8878: Have alternate hreflang link tags on every pages (EN & FR in scope)
Hreflang is important for crawlers and for SEO in general.

On every pages, there should be link tags with a rel="alternate" attribute and a hreflang="lang" to showcase all the languages alternatives of the same page there are out there, including current alternative.

This means that the home page would look like so: 

<link rel="alternate" hreflang="en" href="https://www.canada.ca/en.html" />
<link rel="alternate" hreflang="fr" href="https://www.canada.ca/fr.html" />

French would have the same.

The output hreflang link tag should preferably have an absolute URL in its href attribute, but a relative path works too.

Resources:
https://html.spec.whatwg.org/#rel-alternate:rel-alternate-6
https://support.google.com/webmasters/answer/189077

GormFrank:
As explained in the issues below 
Fix #8877 
Fix #8878

### DIFF
--- a/site/includes/metadata.hbs
+++ b/site/includes/metadata.hbs
@@ -1,1 +1,16 @@
-<meta name="description" content="{{#if page.description }}{{page.description}}{{else}}{{{i18n "site-description"}}}{{/if}}" />
+<meta name="description" content="{{#if page.description}}{{page.description}}{{else}}{{{i18n "site-description"}}}{{/if}}" />
+
+{{#if canonical}}<link rel="canonical" href="{{canonical}}" />{{/if}}
+
+{{#if altLangPrefix}}
+
+{{#is language "en"}}
+<link rel="alternate" hreflang="fr" href="{{altLangPrefix}}-fr.html" />
+{{else}}
+
+{{#is language "fr"}}
+<link rel="alternate" hreflang="en" href="{{altLangPrefix}}-en.html" />
+{{/is}}
+{{/is}}
+
+{{/if}}
--- a/site/pages/index-en.hbs
+++ b/site/pages/index-en.hbs
@@ -3,6 +3,7 @@
 	"title": "Web Experience Toolkit (WET)",
 	"language": "en",
 	"altLangPrefix": "index",
+	"canonical": "https://wet-boew.github.io/wet-boew/index-en.html",
 	"dateModified": "2020-01-22",
 	"breadcrumb": "false"
 }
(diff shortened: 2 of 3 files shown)


site/includes/metadata.hbs
```diff
@@ -1 +1,18 @@
-<meta name="description" content="{{#if page.description }}{{page.description}}{{else}}{{{i18n "site-description"}}}{{/if}}" />
+<meta name="description" content="{{#if page.description}}{{page.description}}{{else}}{{{i18n "site-description"}}}{{/if}}" />
+
+{{#if canonical}}<link rel="canonical" href="{{canonical}}" />{{/if}}
+
+{{#if altLangPrefix}}
+
+{{#is language "en"}}
+<link rel="alternate" hreflang="en" href="{{pagename}}" />
```
duboisp, line 8: this seems redundent information compared to the canonical if the canonical is english

duboisp (changes requested):

duboisp (approved):

PR #8882, opened 2020-06-04
Stepsform Focus Order Fix

### LINKED ISSUE #8871: WCAG 2.1 - 2.4.3 Focus Order - WET "Steps Form"
Description: When the user presses the "Previous" button in any given step of the form, the focus should subsequently go to the legend element in the previous step of the form.

WCAG 2.1: https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html 

FocusOrder_StepsForm

FocusOrder_StepsForm_2

namjohn920:
When next or prev button is clicked, the legend element will be focused.
closes #8871

### DIFF
--- a/src/plugins/stepsform/stepsform.js
+++ b/src/plugins/stepsform/stepsform.js
@@ -204,6 +204,9 @@ var componentName = "wb-steps",
 				buttonGroup = fieldset.querySelector( "div.buttons" );
 				if ( legend ) {
 					legend.classList.add( "wb-steps-active" );
+					legend.tabIndex = 0;
+					legend.focus();
+					legend.tabIndex = -1;
 				}
 				if ( elm ) {
 					elm.classList.remove( "hidden" );


GormFrank (changes requested):
The `<legend>` should get the focus with the `tabindex="0"` only when the **Previous** button is clicked, then it should lose it's focusability since it isn't an interactive element, so you should set the `tabindex="-1"` to achieve that.

GormFrank:
Ensure that the "Previous" experience is consistent with the "Next" experience; so if you put the focus on the legend of the previous step when you hit the previous button then you should pu the focus on the legend of the next step when you hit the next button.

namjohn920:
> Ensure that the "Previous" experience is consistent with the "Next" experience; so if you put the focus on the legend of the previous step when you hit the previous button then you should pu the focus on the legend of the next step when you hit the next button.

Yes it does that. 😄

GormFrank (approved):

GormFrank:
@bsouster You can test here: http://universallabs.org/wet/labs/pr-8882/demos/stepsform/stepsform-en.html

bsouster:
NVDA/Firefox
* The date format is not read aloud when I tab through them, I suppose because it's a spinner input field?
* When I submit a question, the next legend is not read aloud.

bsouster:
User tester relying on Jaws, keyboard-only, IE11:

"I had difficulty  with the date of birth field. The indicators were that they were combo boxes, but the arrow would not work correctly.  In the year box for example I was finally able to enter the numbers directly, and use the arrows to a degree in the month and day.  However, it was after playing with it for about 30 minutes.   

All other info including the check boxes etc. were accepted well enough.  If had been a real form however, the date of birth would have been a problem. Headings, buttons and links all seemed to be recognized."

She will try it with another browser. I will update this comment when I get more feedback.

GormFrank:
> "I had difficulty with the date of birth field. The indicators were that they were combo boxes, but the arrow would not work correctly. In the year box for example I was finally able to enter the numbers directly, and use the arrows to a degree in the month and day. However, it was after playing with it for about 30 minutes.

@bsouster From what I can take from this, the "Go to" previous and next step behaviour is working as expected. The problems are more related to types of inputs and form validation. You could create an issue for this since it is out of scope in this PR. Did I get it right?

GormFrank (approved):

PR #8884, opened 2020-06-05
Breadcrumb: fix issues related to new version of Chrome - focus outline

GormFrank:
(take 2)

PR was previously #8876 

As described in #8875

### DIFF
--- a/src/base/breadcrumb/_base.scss
+++ b/src/base/breadcrumb/_base.scss
@@ -10,13 +10,13 @@
 		ol {
 			border-radius: 0;
 			margin-bottom: 0;
-			padding: 3px 13px;
+			padding: 1px 13px;
 		}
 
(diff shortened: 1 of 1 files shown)


GormFrank:
Visually, the most major impact is the vertical space between each element of the breadcrumb when on mobile view. 

**Before:** _(Ignore orange line)_
bc-before

**After:**
bc-after

I consider it acceptable, even better than the status quo, but @delisma can you give me your 2 cents on this?

delisma:
@GormFrank So the vertical spacing is pushing down the main content further on mobile view.

@jmealing, @pcwsmith do we have any objections against it?

delisma:
Ok it's approved

cc @jmealing, @pcwsmith

duboisp (approved):

PR #8894, opened 2020-06-19
exitscript plugin

ipaksc:


src/plugins/exitscript/exitscript-en.hbs
```diff
@@ -0,0 +1,51 @@
...
+	"description": "Redirects users to non secure sites.",
+	"tag": "exitscript",
+	"parentdir": "exitscript",
+	"altLangPrefix": "exitscript",
+	"dateModified": "2020-06-17"
+}
+---
+<p>The script redirects users to non secure site.</p>
+<p>Plugin default settings:</p>
+<ul class="list-unstyled">
```
duboisp, line 15: This is definition list, `<dl>`

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+			"<ul class='list-inline text-center'>" +
+			"<li><button class='btn btn-primary popup-modal-dismiss pull-left' id='eCancel'>" + settings.cancelMsg + "</button></li>" +
+			"<li><a class='btn btn-default pull-right popup-modal-close' " + settings.targetBlank + " href='" + settings.exitURL + exitParam + "' type='button'>" + settings.yesMsg + "</a></li>" +
+			"</ul></div></section>";
+				var moDal = document.createDocumentFragment();
+				moDal.appendChild( tpl );
+				document.body.appendChild( moDal );
+
+		//Unwrap div id="tm"
+				var wrapper = document.querySelector( "#tm" );
```
duboisp, line 91: you can use the "tpl" object directly instead of going through `document.querySelector`. A such id "tm".
Try to find a solution without an id.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+			self.location.href = ( settings.exitURL +  exitParam );
+		} else {
+			window.open( settings.exitURL + exitParam );
+		}
+
+	}
+
+} );
+
+$( document ).on( "click", ".popup-modal-close", function() {
```
duboisp, line 134: Why, the modal dialog already manage that?

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+	ev.preventDefault();
+
+	var settings = $.extend(
+	defaults, {}
+);
+
+	var exitParam = "?" + encodeURIComponent( "targetUrl=" + String( this.href ) );
+
+	if ( settings.targetBlank ) {
+		exitParam = exitParam + encodeURIComponent( "&&targetBlank=true" );
```
duboisp, line 110: What the double "&&" do?

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+			settings;
+		if ( elm ) {
+			$elm = $( elm );
+			settings = $.extend(
+		true,
+		defaults, {},
+		window[ componentName ],
+		wb.getData( $elm, componentName )
+	);
+			i18nDict = i18nDict[ $( "html" ).attr( "lang" ) || "en" ];
```
GormFrank, line 45: You should be using the following instead of scrapping the DOM
```suggestion
			i18nDict = i18nDict[ wb.lang || "en" ];
```

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+				}
+				if ( !settings.yesMsg ) {
+					settings.yesMsg = i18nDict.yesMsg;
+				}
+				if ( !settings.cancelMsg ) {
+					settings.cancelMsg = i18nDict.cancelMsg;
+				}
+				var tpl = document.createElement( "div" );
+				tpl.id = "tm";
+				tpl.innerHTML = "<section id='sExitModal' class='mfp-hide modal-dialog modal-content overlay-def'>" +
```
duboisp, line 78: is the `id='sExitModal'` needed?

The prefer format for custom ID is to use `wb.getId()`

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+			"<header class='modal-header'><h2 class='modal-title'>" + i18nDict.warning + "</h2></header>" +
+			"<div class='modal-body'>" +
+			"<p>" + settings.exitMsg + "</p>" +
+			"<ul class='list-inline text-center'>" +
```
duboisp, line 82: This will be a modal footer. See the working example in the overlay working example

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+			var exitParam = "?" + encodeURIComponent( "targetUrl=" + String( this.href ) );
+
+			if ( settings.targetBlank ) {
+				exitParam = exitParam + encodeURIComponent( "&&targetBlank=true" );
+			}
+
+			// Set default propertie
+
+			if ( settings.displayModal !== false ) {
+				if ( !settings.targetBlank ) {
```
duboisp, line 62: This conditional could be rewritten like this:
```
targetBlankAttribute = !!settings.targetBlank || "target='_blank'"
```

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+			$elm.trigger( "setModaldialog", settings );
+			var exitParam = "?" + encodeURIComponent( "targetUrl=" + String( this.href ) );
+
+			if ( settings.targetBlank ) {
+				exitParam = exitParam + encodeURIComponent( "&&targetBlank=true" );
+			}
+
+			// Set default propertie
+
+			if ( settings.displayModal !== false ) {
```
duboisp, line 61: This one can be written
```
if ( settings.displayModal ) {
```

GormFrank (changes requested):

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+	};
+
+// Add your plugin event handler
+$document.on( "click", selector, function( ev ) {
+	ev.preventDefault();
+
+	var settings = $.extend(
+	defaults, {}
```
duboisp, line 104: There is spacing issue here?

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+			[ {
+				src: "#sExitModal",
+				type: "inline"
+			} ],
+			true
+		] );
+
+	} else {
+
+		if ( settings.displayModal === false && settings.targetBlank !== true ) {
```
duboisp, line 124: I think for this case, the only thing the script will add value is to append query parameter to the href attribute. If the plugin require the author to set it on every single link, then will be preferable to show them to use the attribute `target='_blank'`

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+
+// Add your plugin event handler
+$document.on( "click", selector, function( ev ) {
+	ev.preventDefault();
+
+	var settings = $.extend(
+	defaults, {}
+);
+
+	var exitParam = "?" + encodeURIComponent( "targetUrl=" + String( this.href ) );
```
duboisp, line 107: What happen if `href ` already contains "?" character

You can look to use `wb.getUrlParts( url )` in order to rebuild and know if there already a query string and or an anchor

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,144 @@
...
+			i18nDict = {
+				warning: i18nDict[ "exitscript-warning" ],
+				exitMsg: i18nDict[ "exitscript-exitMsg" ],
+				yesMsg: i18nDict[ "exitscript-yesMsg" ],
+				cancelMsg: i18nDict[ "exitscript-cancelMsg" ]
+			};
+			$elm.trigger( "setModaldialog", settings );
+			var exitParam = "?" + encodeURIComponent( "targetUrl=" + String( this.href ) );
+
+			if ( settings.targetBlank ) {
```
duboisp, line 55: Why this is evaluate twice? here and in teh click function?

duboisp (changes requested):
General comment: we try to have all variable declaration immediatly to where the "function" start, see the other plugin for code example.

Other question:
Is this use of this plugin to be applied 1 one in the page and then it will hook on all links or is the developper/author will need to declare it on every "anchor" element. 
For an example that use "global" config, see the pretify plugin.

StdGit:
https://github.com/wet-boew/wet-boew/pull/8894#pullrequestreview-437543034

> Other question:
> Is this use of this plugin to be applied 1 one in the page and then it will hook on all links or is the developper/author will need to declare it on every "anchor" element.
> For an example that use "global" config, see the pretify plugin.

Le but d'exitScript est d'avertir l'utilisateur qu'il navigue sur un site étranger. Donc tous les liens externes (Qui ne sont pas dans une liste d'exclusions (Voir le exitScript du SGDC-CDTS) devraient être activés automatiquement afin de faciliter le travail des développeurs.

src/plugins/exitscript/exitscript.js
ipaksc: It's fixed

GormFrank:
Closes https://github.com/wet-boew/GCWeb/pull/1308

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+		defaults, {},
+		window[ componentName ],
+		wb.getData( $elm, componentName )
+	);
+			i18nDict = i18nDict[ wb.lang || "en" ];
+			i18nDict = {
+				warning: i18nDict[ "exitscript-warning" ],
+				exitMsg: i18nDict[ "exitscript-exitMsg" ],
+				yesMsg: i18nDict[ "exitscript-yesMsg" ],
+				cancelMsg: i18nDict[ "exitscript-cancelMsg" ]
```
duboisp, line 50: You can use the generic one,
ex:
```
wb.i18n("cancel")
```

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+		var elm = event.currentTarget,
+				$elm = $( elm ),
+				moDalId,
+				wrapper,
+				exitURLsearchParam,
+				exitURL,
+				exitParam = encodeURIComponent( "targetUrl=" + String( this.href ) );
+		var moDal = document.createDocumentFragment();
+		var tpl = document.createElement( "div" );
+		var settings = $.extend(
```
duboisp, line 70: This only provide default setting which ignore the custom setting defined by the user.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+		defaults, {}
+		);
+
+		if(this.hasAttribute("target")){ 
+		  exitParam = exitParam + encodeURIComponent( "&targetBlank=true" );
```
duboisp, line 75: Can you mention in the docs or in the working example a description of the new query parameter added to the URL?

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+		}
+			 
+		exitURL = settings.exitURL;
```
duboisp, line 78: This return "undefined" value because `exitURL` is not defined in this plugin.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+			
+		if( !exitURL.includes( "targetUrl" ) ){
```
duboisp, line 80: What is the purpose of this conditional? What is "targetUrl" means, is it another inner setting?

If this condition means no code processing should continue to be executed, simply return "true" instead of closing the conditional at the end of this JS function. ex:
```
if (condition) {
    return true;
}
```

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+
+		if(this.hasAttribute("target")){ 
+		  exitParam = exitParam + encodeURIComponent( "&targetBlank=true" );
+		}
+			 
+		exitURL = settings.exitURL;
+			
+		if( !exitURL.includes( "targetUrl" ) ){ 
+	
+		if( !wb.getUrlParts( exitURL ).search ){
```
duboisp, line 82: Call the function `wb.getUrlParts( exitURL )` once and cache it into a variable instead of calling that function over and over.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+			exitParam = exitParam.split( '%26' ).join( '&' );
+	
+		} else { 
+				exitURLsearchParam = String(wb.getUrlParts( exitURL ).search );
+				exitURLsearchParam = exitURLsearchParam.replace("?","");		
+				exitParam =  exitURLsearchParam + "&" + exitParam;
+				exitParam = exitParam.split('%26').join('&');
+				//Build exitURL
+				exitURL = wb.getUrlParts( exitURL ).protocol + 
+				"//" +
```
duboisp, line 93: Add indentation, 1 tab more

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+						type: "inline"
+					}],
+				 
+				true
+			 
+		]);					 
+ 
+                } else {
+					 
+				if ( this.hasAttribute("target") ) {
```
duboisp, line 144: Merge this conditional with the previous `else`.  So you will use `else if ( condition ) {` structure

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+				//Build exitURL
+				exitURL = wb.getUrlParts( exitURL ).protocol + 
+				"//" + 
+				wb.getUrlParts( exitURL ).host + 
+				wb.getUrlParts( exitURL ).pathname  + 
+				"?" + 
+				exitParam;
+			} 
+		}
+			var defaultModadisplay = settings.displayModal !== false;
```
duboisp, line 100: There is no need to have conditional like `!== false` except of rare exception which it don't apply here.

To convert the setting value into a real boolean by using double negation, like this: `!!settings.displayModal`

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+			
+			// Set default properties
+			if ( defaultModadisplay ) {
+			   
+				if(this.hasAttribute("target")){
+					settings.targetBlank = "target='_blank'"
+				} 
+				if ( !settings.exitMsg ) {
```
duboisp, line 108: Is those message is defined per link or it is a global setting?

Consider a instruction like this instead of the whole conditional block for one (1) instruction:
`var exitMsg = settings.exitMsg || i18nDict.exitMsg`

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+				}
+				if ( !settings.yesMsg ) {
+					settings.yesMsg = i18nDict.yesMsg
+				}
+				if ( !settings.cancelMsg ) {
+					settings.cancelMsg = i18nDict.cancelMsg
+				}
+				 
+				tpl.innerHTML = "<section class='mfp-hide modal-dialog modal-content overlay-def'>" +
+					"<header class='modal-header'><h2 class='modal-title'>" + i18nDict.warning + "</h2></header>" +
```
duboisp, line 119: `i18nDict` is not consistent with the rest of other customizable string.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+				wrapper.firstChild.id = wb.getId();
+				moDalId = wrapper.firstChild.id;
+				document.body.appendChild( wrapper.firstChild );
+				$document.trigger("open.wb-lbx", [
+					[{
+						src: "#" + moDalId,
+						type: "inline"
+					}],
+				 
+				true
```
duboisp, line 138: Fix indentation

src/plugins/exitscript/exitscript.js
duboisp, line 144: Why to not consider the value of the attribute `target`? I think this case you are only looking for `target="_blank"`, is it?

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+				  
+				} else {
+					
+					self.location.href = ( settings.exitURL );
+				}
+			}
+				 
+    });
+
+    $( document ).on( "click", ".popup-modal-close", function( event ) {
```
duboisp, line 155: Use the close button that was integrated into the lightbox. See the configuration option `popup-modal-dismiss` documented here: https://wet-boew.github.io/wet-boew/docs/ref/lightbox/lightbox-en.html

duboisp (changes requested):

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+*/
+	init = function( event ) {
+		var elm = wb.init( event, componentName, selector ),
+			$elm,
+			settings;
+		if ( elm ) {
+			$elm = $( elm );
+			settings = $.extend(
+		true,
+		defaults, {},
```
duboisp, line 41: Currently it is the variable `defaults` that get extended, reverse the order of "defaults" and "{}"

see: https://api.jquery.com/jQuery.extend/

duboisp (changes requested):

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,166 @@
...
+		window[ componentName ],
+		wb.getData( $elm, componentName )
+	);
+			i18nDict = i18nDict[ wb.lang || "en" ];
+			i18nDict = {
+				warning: i18nDict[ "exitscript-warning" ],
```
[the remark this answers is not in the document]
duboisp: Something like
```
i18nDict = $.extend( {}, i18nDict, settings.i18n );
```

And the user configuration would look like: 
```
data-wb-exitscript='{"exitURL": "http://www.canada.ca", "i18n": { "exitMsg": "New exit script message" } }'
```

src/plugins/exitscript/exitscript.js
ipaksc: @duboisp Pierre: I am using the class popup-modal-close for "Yes" button. When users click "Yes" button and redirected to the middle/exit page, I want to make sure the modal window will be closed. The class popup-modal-dismiss is not working for me. When target attribute is undefined the $.magnificPopup.close() closes the window too fast the anchor/exit url is not triggered.

src/plugins/exitscript/exitscript.js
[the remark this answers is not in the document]
ipaksc: @duboisp Here is an example how users can change the warning message. data-wb-exitscript='{"exitURL": "http://www.canada.ca", "exitMsg": "New exit script message"  }'

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,135 @@
...
+
+	var elm = event.currentTarget,
+		$elm = $( elm ),
+		moDalId,
+		wrapper,
+		moDal = document.createDocumentFragment(),
+		tpl = document.createElement( "div" ),
+		settings =  $elm.data( componentName ),
+		modalWindow = settings.displayModal !== false,
+		msgboxHeader = settings.msgboxHeader || i18nDict.msgboxHeader,
```
duboisp, line 73: It will be preferable to do that initialization (setting up the default of the variable "settings") in the init function, instead of when the anchor is clicked.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,135 @@
...
+
+			);
+
+			$elm.data( componentName, settings );
+
+			i18nDict = i18nDict[ wb.lang || "en" ];
+
+			if ( $elm.hasClass( exiturlparam ) ) {
+				this.outerHTML = "<a href='" + sessionStorage.getItem( "originalURL" ) + "'>" + sessionStorage.getItem( "originalURL" ) + "</a>";
+				if ( document.querySelector( selector + "-targetparam" ) ) {
```
duboisp, line 54: It is preferable to keep it in scope of the element instead of getting the first instance in the page.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	if ( modalWindow ) {
+		tpl.innerHTML = "<section class='mfp-hide modal-dialog modal-content overlay-def'>" +
+			"<header class='modal-header'><h2 class='modal-title'>" + msgboxHeader + "</h2></header>" +
+			"<div class='modal-body'>" +
+			"<p>" + exitMsg + "</p>" +
+			"<ul class='list-inline text-center'>" +
+			"<li><button class='btn btn-primary popup-modal-dismiss pull-left' id='eCancel'>" + cancelBtn + "</button></li>" +
+			"<li><a class='btn btn-default pull-right popup-modal-close'" + targetAttribute + " href='" + exitURL + "' type='button'>" + yesBtn + "</a></li>" +
```
duboisp, line 95: To make the "popup-modal-dismiss" working, it probably need to be defined in the dialog footer

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,135 @@
...
+			"</ul></div></section>";
+		moDal.appendChild( tpl );
+		wrapper = moDal.firstChild;
+		wrapper.firstChild.id = wb.getId();
+		moDalId = wrapper.firstChild.id;
```
duboisp, line 100: You don't need to define an id here

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,135 @@
...
+		document.body.appendChild( wrapper.firstChild );
+
+		$document.trigger( "open.wb-lbx", [
```
duboisp, line 103: You can do instead of using the global event trigger
```
$( wrapper.firstChild ).trigger( ...... )
```

duboisp (changes requested):

src/plugins/exitscript/exitscript.js
ipaksc: @duboisp Hi Pierre: When the modal window should be defined? In init function or when the anchor is clicked?

duboisp: It can be created when the anchor is clicked and re-use the same dialog when the link is clicked again.

src/plugins/exitscript/exitscript.js
ipaksc: @duboisp Please review...

src/plugins/exitscript/exitscript.js
ipaksc: @duboisp It's fixed.

src/plugins/exitscript/exitscript.js
ipaksc: @duboisp I need to discuss this with Stéphane. Maybe I will remove   if ( document.querySelector( selector + "-targetparam" ) ) {  from the script.

src/plugins/exitscript/exitscript.js
ipaksc: @duboisp Pierre. I had put back popup-modal-close, popup-modal-dismiss is no working for me.

duboisp (changes requested):
Can we use a simpler version for the plugin, like

Open in modal
`<a href="link" class="wb-exitscript">...</a>`

Open in an intermediate page
`<a href="link" class="wb-exitscript" data-wb-exitscript="URL of the intermediate page">...</a>`

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,141 @@
...
+			if ( settings.exitURL ) {
+				$( this ).attr( "href", settings.exitURL + "?exturl=" +  encodeURIComponent( this.getAttribute( "href" ) ) );
+			}
+
+			if ( !i18n ) {
+				i18n = i18nDict[ wb.lang || "en" ];
+			}
+
+			if ( $elm.hasClass( exiturlparam ) ) {
+				this.outerHTML = "<a href='" + originalURL + "'>" + originalURL + "</a>";
```
duboisp, line 64: Be more clear about what is happening here, add a comment before saying this conditional is for the middle static exit page .

src/plugins/exitscript/exitscript.js
duboisp: what is the status of this

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,141 @@
...
+	if ( !settings.exitURL ) {
+
+		event.preventDefault();
+	}
+
+	if ( this.getAttribute( "target" ) !== null ) {
+		targetAttribute = "target='" + this.getAttribute( "target" ) + "'";
+		exitMsg = exitMsg  + " " + i18n.targetWarning;
+	} else {
+		targetAttribute = "";
```
duboisp, line 93: Leave the default in the variable declaration, like at line 75. This else condition is not needed

duboisp (changes requested):
Todo:
* [ ] Simplify the configuration "exitURL" for "url". That configuration is already in the context of exit script.
* [ ] Adjust the working example to the recent changes. I can not test with the modal popup with the example provided.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,140 @@
...
+			);
+
+			$elm.data( componentName, settings );
+
+			if ( settings.url ) {
+				$( this ).attr( "href", settings.url + "?exturl=" +  encodeURIComponent( this.getAttribute( "href" ) ) );
+			}
+
+			if ( !i18n ) {
+				i18n = $.extend( true, i18nDict[ wb.lang || "en" ], settings.i18n || {} );
```
duboisp, line 60: Change your docs accordonly, now every string need to be declared inside the i18n object

duboisp (changes requested):

src/plugins/exitscript/exitscript.js
ipaksc: popup-modal-dismiss will not work when target is undefined.

$( document ).on( "click", ".popup-modal-dismiss", function( event ) {
	if ( !this.hasAttribute( "target" ) ) {
		event.preventDefault();
	}

	$.magnificPopup.close();
} );

src/plugins/exitscript/exitscript.js
duboisp: This click function need to be removed or move in the right plugin

duboisp (changes requested):

src/plugins/exitscript/exitscript.js
ipaksc: Hi @duboisp,

Here is the markup that you can add in a wet-boew test page.
Please test when you get a chance.
Scenario – 1: The class popup-modal-dismiss is not working for “Yes” button. The popup-modal-dismiss close the modal dialog but not triggered the link 
Scenario – 2: The anchor has target attribute, the “Yes” button is working as expected.

Thank you
------------
`<p>Scenario – 1 <a class="wb-lbx lbx-modal" title="Example of a modal popup" href="#inline_content_modal">Modal popup</a></p>
		<section id="inline_content_modal" class="mfp-hide modal-dialog modal-content overlay-def">
			<header class="modal-header">
				<h2 class="modal-title">Title</h2>
			</header>
			<div class="modal-body">
				<p>You are about to leave a secure site, do you wish to continue?</p>	
				
			</div>
			<div class='modal-footer'>
			<ul class='list-inline text-center'>
			<li><button class='btn btn-primary popup-modal-dismiss pull-left' id='eCancel'>Cancel</button></li>
			<li><a class='btn btn-default pull-right popup-modal-dismiss' href='http://www.google.ca' type='button'>Yes</a></li>
			</ul>
			
		</section>
<p>Scenario – 2 <a class="wb-lbx lbx-modal" target="_blank" title="Example of a modal popup" href="#inline_content_modal-1">Modal popup -2</a></p>
		<section id="inline_content_modal-1" class="mfp-hide modal-dialog modal-content overlay-def">
			<header class="modal-header">
				<h2 class="modal-title">Title</h2>
			</header>
			<div class="modal-body">
				<p>You are about to leave a secure site, do you wish to continue?</p>	
			</div>
			<div class='modal-footer'>
			<ul class='list-inline text-center'>
			<li><button class='btn btn-primary popup-modal-dismiss pull-left' id='eCancel'>Cancel</button></li>
			<li><a class='btn btn-default pull-right popup-modal-dismiss' target="_blank"  href='http://www.google.ca' type='button'>Yes</a></li>
			</ul>
			</div>
		</section>`

ipaksc: @duboisp 

```
<p>Scenario – 1 <a class="wb-lbx lbx-modal" title="Example of a modal popup" href="#inline_content_modal">Modal popup</a></p>
		<section id="inline_content_modal" class="mfp-hide modal-dialog modal-content overlay-def">
			<header class="modal-header">
				<h2 class="modal-title">Title</h2>
			</header>
			<div class="modal-body">
				<p>You are about to leave a secure site, do you wish to continue?</p>	
				
			</div>
			<div class="modal-footer">
			<ul class="list-inline text-center">
			<li><button class="btn btn-primary popup-modal-dismiss pull-left" id="eCancel">Cancel</button></li>
			<li><a class="btn btn-default pull-right popup-modal-dismiss" href="http://www.google.ca" type="button">Yes</a></li>
			</ul>
			
		</section>
<p>Scenario – 2 <a class="wb-lbx lbx-modal" target="_blank" title="Example of a modal popup" href="#inline_content_modal-1">Modal popup -2</a></p>
		<section id="inline_content_modal-1" class="mfp-hide modal-dialog modal-content overlay-def">
			<header class="modal-header">
				<h2 class="modal-title">Title</h2>
			</header>
			<div class="modal-body">
				<p>You are about to leave a secure site, do you wish to continue?</p>	
			</div>
			<div class="modal-footer">
			<ul class="list-inline text-center">
			<li><button class="btn btn-primary popup-modal-dismiss pull-left" id="eCancel">Cancel</button></li>
			<li><a class="btn btn-default pull-right popup-modal-dismiss" target="_blank"  href="http://www.google.ca" type="button">Yes</a></li>
			</ul>
			</div>
		</section>
```

src/plugins/exitscript/exitscript-en.hbs
```diff
@@ -0,0 +1,64 @@
...
+	"description": "Redirects users to non secure sites.",
+	"tag": "exitscript",
+	"parentdir": "exitscript",
+	"altLangPrefix": "exitscript",
+	"dateModified": "2020-06-17"
+}
+---
+<p>The script redirects users to non secure site.</p>
+<p>Plugin default settings:</p>
+<dl>
```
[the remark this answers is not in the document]
duboisp: Please update the default setting variable content and name.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,149 @@
...
+		targetAttribute = "",
+		moDal = document.createDocumentFragment(),
+		tpl = document.createElement( "div" ),
+		settings =  $elm.data( componentName ),
+		msgboxHeader = i18n.msgboxHeader,
+		yesBtn = i18n.yesBtn,
+		cancelBtn = i18n.cancelBtn,
+		exitMsg = i18n.exitMsg;
+
+	if ( settings.i18n ) {
```
duboisp, line 87: I think the i18n variable (set at line 62) already contain the final wording. Use the `i18n` variable global at this plugin instead this conditional.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	if ( this.getAttribute( "target" ) !== null ) {
+		targetAttribute = "target='" + this.getAttribute( "target" ) + "'";
+		exitMsg = exitMsg  + " " + i18n.targetWarning;
+	}
+
+	if ( document.getElementById( moDalId ) ) {
+		var tplModal = document.getElementById( moDalId );
+		tplModal.remove();
```
duboisp, line 106: You can do it in one call like
```
document.getElementById( moDalId ).remove();
```

src/plugins/exitscript/exitscript.js
duboisp: Did you still need this "click" function? even after you added into the modal footer?

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,149 @@
...
+	if ( !settings.url ) {
+		tpl.innerHTML = "<section id='" + moDalId + "' " + "class='mfp-hide modal-dialog modal-content overlay-def'>" +
+			"<header class='modal-header'><h2 class='modal-title'>" + msgboxHeader + "</h2></header>" +
+			"<div class='modal-body'>" +
+			"<p>" + exitMsg + "</p>" +
+			"</div>" +
+			"<div class='modal-footer'>" +
+			"<ul class='list-inline text-center'>" +
+			"<li><button class='btn btn-primary popup-modal-dismiss pull-left' id='eCancel'>" + cancelBtn + "</button></li>" +
+			"<li><a class='btn btn-default pull-right popup-modal-close'" + targetAttribute + " href='" + this.getAttribute( "href" ) + "' type='button'>" + yesBtn + "</a></li>" +
```
duboisp, line 118: remove `type="button"`

duboisp, line 118: We have discussed about this pattern and PP and WAWG agree that

Reverse the order and change the style.
```
<ul class="list-inline text-center">
<li><a href="http://csszengarden.com/219" class="btn btn-primary pull-right popup-modal-close">Yes</a></li>
<li><button class="btn btn-default popup-modal-dismiss pull-left">Cancel</button></li>
</ul>
```

"Yes" option should be primary, it's the natural order and common pattern order for modal like this. Also, the user has already click on a link to navigate away, so it is high risk they will navigate away anyway.

Remove the ID of the button `id="eCancel"`

duboisp (changes requested):
@ipaksc it's almost ready, please see the inline comment I did.

Great work.

src/plugins/exitscript/exitscript.js
ipaksc: It's fixed

src/plugins/exitscript/exitscript.js
ipaksc: done

ipaksc: done

src/plugins/exitscript/exitscript.js
ipaksc: @duboisp Hi Pierre,

I had to rollback i18n = $.extend( true, i18nDict[ wb.lang || "en" ], settings.i18n || {} ); to i18n = i18nDict[ wb.lang || "en" ] and delet if ( !i18n ).

i18n = $.extend( true, i18nDict[ wb.lang || "en" ], settings.i18n || {} ) is not merging the contents of two objects. It overrides the default variables values

src/plugins/exitscript/exitscript.js
ipaksc: @duboisp Yes. I need the "click" function. I can move the popup-modal-close to lightbox plugin or edit popup-modal-dismiss.

src/plugins/exitscript/exitscript.js
ipaksc: I have updated the default setting variable content and name in src/plugins/exitscript/exitscript-en.hbs

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,147 @@
...
+		yesBtn = i18n.yesBtn,
+		cancelBtn = i18n.cancelBtn,
+		exitMsg = i18n.exitMsg;
+
+	if ( settings.i18n ) {
+		msgboxHeader =  settings.i18n.msgboxHeader || i18n.msgboxHeader;
+		yesBtn = settings.i18n.yesBtn || i18n.yesBtn;
+		cancelBtn = settings.i18n.cancelBtn || i18n.cancelBtn;
+		exitMsg = settings.i18n.exitMsg || i18n.exitMsg;
+	}
```
duboisp, line 90: The jQuery extend should already did the work, you shouldn't need those individual OR operator.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,147 @@
...
+
+$document.on( "click", selector, function( event ) {
+
+	var elm = event.currentTarget,
+		$elm = $( elm ),
+		wrapper,
+		targetAttribute = "",
+		moDal = document.createDocumentFragment(),
+		tpl = document.createElement( "div" ),
+		settings =  $elm.data( componentName ),
```
duboisp, line 79: You already did the setting extension at line 48. Either do it in this click function or during the init(), but not at both places. And use the jQuery extend to set defaults value.

src/plugins/exitscript/exitscript.js
duboisp: I tried just to overwrite the close button by adding my own modal-footer in the lightbox working example, and I see your issue. However, that fix must be done in lightbox and somehow the lightbox plugin overwrite the modal footer set by the author. 

That fix must be submitted in a different PR because it is not related directly to the exit script plugin but to a bug in lightbox plugin.

duboisp (changes requested):
If you want, we can chat about it during the weekly afternoon code sprint. See the information here: https://github.com/wet-boew/wet-boew/wiki/WET-BOEW-Code-sprint

duboisp (changes requested):

src/plugins/exitscript/exitscript.js
ipaksc: @duboisp The class is popup-modal-dismiss closes the modal window but the anchor is not working.

duboisp: When the button is changed for an anchor? but you only have an anchor when the person would navigate to the other page. Closing the dialog is useless when the navigation happen in the same tab, no? 

The click event handler support the anchor element.
I saw that it prevent default if it don't contains a target attribute.
ref. https://github.com/wet-boew/wet-boew/blob/master/src/plugins/lightbox/lightbox.js#L348
That behaviour was added by you - https://github.com/wet-boew/wet-boew/commit/5288ca529de6a40e55d061119da5aca449e69f26

The solution will be to define a target "_self" to open the link in the page, but to close the dialog.

The implementation of the dismiss button in the lightbox do the same as the code that was added.

duboisp (changes requested):

src/plugins/exitscript/exitscript.js
ipaksc: @duboisp Pierre.  You asked me to add this behaviour (prevent default if it don't contains a target attribute.) to the lightbox plugin.

ipaksc: @duboisp Can we just remove this behaviour (prevent default if it don't contains a target attribute)? It will fix the issue.

GormFrank: @ipaksc Do you mean removing the event handler from line 137 to 139? Yes because it doesn't belong in here.

GormFrank (changes requested):

GormFrank:
@ipaksc If needed, we can still invite you to the code sprint.

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,147 @@
...
+		cancelBtn = settings.i18n.cancelBtn || i18n.cancelBtn;
+		exitMsg = settings.i18n.exitMsg || i18n.exitMsg;
+	}
+
+	if ( !settings.url ) {
+
+		event.preventDefault();
+	}
+
+	if ( this.getAttribute( "target" ) !== null ) {
```
duboisp, line 97: You can just test if it is true
```suggestion
	if ( this.getAttribute( "target" ) ) {
```

src/plugins/exitscript/exitscript.js
```diff
@@ -0,0 +1,147 @@
...
+			wb.ready( $elm, componentName );
+		}
+	};
+
+$document.on( "click", selector, function( event ) {
+
+	var elm = event.currentTarget,
+		$elm = $( elm ),
+		wrapper,
+		targetAttribute = "",
```
duboisp, line 76: Set it by default to "_self"

duboisp (changes requested):

duboisp (approved):

GormFrank (approved):

StdGit:
So it is done Good!

What is the link for samples and documentation?

GormFrank:
> What is the link for samples and documentation?

@StdGit It'll be: https://wet-boew.github.io/wet-boew/docs/ref/exitscript/exitscript-en.html
It will be available as soon as Travis-CI finishes building it.

StdGit:
I saw that Travis finished...Strange
continuous-integration/travis-ci/pr The Travis CI build passed

StdGit:
from https://www.timeanddate.com/worldclock/timezone/utc it is 20:07:00 and in montreal it is 15:07

image

GormFrank:
@StdGit C'est dispo maintenant. C'est que après que Travis-CI ait terminé le build, il doit le refaire pour déployer le site GitHub de WET-BOEW.

Par contre, j'ai remarqué que la doc et les exemples de l'exitscript plugin ne sont pas accessible à partir de l'index. J'ai ouvert un issue à cette intention https://github.com/wet-boew/wet-boew/issues/8977.

StdGit:
There is a typo Exit**sct**ipt

PR #8903, opened 2020-07-10
Loading icon not cleared after seek event in Edge

esand:
In Edge (tested in v16 and v17), if you start playing a video and then seek, the playing event is not re-fired after seeked event, so the waiting/loading state is never cleared from the player. By listening to the seeked event and treating it the same as canplay, it clears the waiting/loading state so the loading icon is removed and doesn't obstruct the player as the video continues to play.

### DIFF
--- a/src/plugins/multimedia/multimedia.js
+++ b/src/plugins/multimedia/multimedia.js
@@ -36,6 +36,7 @@ var componentName = "wb-mltmd",
 		"timeupdate",
 		"waiting",
 		"canplay",
+		"seeked",
 		"progress",
 		captionsLoadedEvent,
 		captionsLoadFailedEvent,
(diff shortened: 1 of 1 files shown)


GormFrank:
Can you please provide a link to the page where you encountered the issue, and also can you provide a screenshot of the issue. Lastly, does this fixes any opened issue? But also, do you mean EdgeHTML version 16 and 17?

esand:
You can use your demos - https://wet-boew.github.io/v4.0-ci/demos/multimedia/multimedia-en.html. Sorry I can't do screenshots or a video (due to current work set up) to show it in action.
Yes, referring to EdgeHTML 16/17 - I used that version number because WET4 currently mentions "Edge 16" as your lowest supported Edge version.

To reproduce, simply hit play, then seek the video and you'll see the loading animation stay on screen. You can pause, play, seek etc... and it won't go away (even when you reach the end of the video).

duboisp (approved):

GormFrank:
Approved during technical review.

GormFrank:
Thank you @esand for this contribution!

PR #8923, opened 2020-08-18
Bootstrap 4 - Added position-relative to improve stretched-link

### LINKED ISSUE #1706: stretched-link overflowing inner container in grid
I really would love to use the stretched-link feature rather than wrapping an entire component that is in a grid with an <a> tag (to save on screen readers having to hear all the content within the tag).  However one small issue is that when there is a grid row of 2/3/4/etc containers with the stretched link, and they are all in panels or wells, that the clickable area extends beyond the container to the edge of the col-* which seems a bit odd when hovering (as the clickable area should be within borders).

The simple fix would be to have an added class that can add 15px to the left/right styles of the stretched-link class.  Or have the stretched-link class have those margins by default and must use "full-width" to extend beyond the inner container.  

Or is there a feature/class/code structure I'm not aware of?

GormFrank:
This is to be used with `.stretched-link` to prevent clickable area from propagating all the way up to `.col-x` margins.
Closes https://github.com/wet-boew/GCWeb/issues/1706

### DIFF
--- a/site/pages/docs/ref/provisional-en.hbs
+++ b/site/pages/docs/ref/provisional-en.hbs
@@ -3,7 +3,7 @@
 	"title": "Provisional features",
 	"language": "en",
 	"altLangPrefix": "provisional",
-	"dateModified": "2020-06-25",
+	"dateModified": "2020-08-18",
 	"share": "true"
 }
 ---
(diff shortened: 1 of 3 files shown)


GormFrank:
Closes https://github.com/wet-boew/GCWeb/issues/1706

duboisp (approved):

PR #8924, opened 2020-08-18
Postback - Prevent multiple submissions by double-clicking on submit

GormFrank:


src/plugins/wb-postback/wb-postback.js
```diff
@@ -41,42 +41,45 @@ var $document = wb.doc,
...
+					.done( function() {
+						$selectorFailure.addClass( classToggle );
+						$selectorSuccess.removeClass( classToggle );
+					} )
+					.fail( function() {
+						$selectorSuccess.addClass( classToggle );
+						$selectorFailure.removeClass( classToggle );
+					} )
+					.always( function() {
+						if ( selectorContent ) {
```
duboisp, line 78: Remove the double click prevention flag. Or if we only want to be click once, that option should be turn on by the author via the button or the form.

duboisp (changes requested):

src/plugins/wb-postback/wb-postback.js
```diff
@@ -24,59 +24,60 @@ var $document = wb.doc,
 					defaults,
 					wb.getData( $elm, componentName )
 				),
-				attrClick = "data-wb-clicked",
+				attrEngaged = "data-wb-engaged",
 				$buttons = $( "[type=submit]", $elm ),
+				multiple = typeof $elm.data( componentName + "-multiple" ) !== "undefined",
```
duboisp, line 29: Just add one liner about this feature with the documentation/working example

duboisp (changes requested):

duboisp (approved):

PR #8929, opened 2020-09-11
Update DD details and add 12 & 13

GormFrank:


duboisp (approved):

duboisp:
Merging it right away as it only concern the wet-boew website.

PR #8937, opened 2020-09-18
Fix clicking on child elements of anchor

esand:
If tabs are used in a modal window and the tab summary element has child elements (span, abbr, etc...) then clicks to the tab anchor (which is generated) throw an error on line 328. The error is thrown because the event assumes that event.target is an anchor. Since it is not (it is the child element, inside the anchor), it has no href attribute and the code fails. By changing to event.currentTarget, it ensures that the eventTarget element is the anchor from the event filter.

### DIFF
--- a/src/plugins/lightbox/lightbox.js
+++ b/src/plugins/lightbox/lightbox.js
@@ -319,7 +319,7 @@ $document.on( "timerpoke.wb " + initEvent, selector, init );
 // Handler for clicking on a same page link within the overlay to outside the overlay
 $document.on( "click vclick", ".mfp-wrap a[href^='#']", function( event ) {
 	var which = event.which,
-		eventTarget = event.target,
+		eventTarget = event.currentTarget,
 		$lightbox, linkTarget;
 
 	// Ignore middle/right mouse buttons


esand:
If you create an example page with the following code, you can see the error by trying to click on either of the first two tabs (specifically, their text - the padding around the text works). Both the first two tabs have a child element within the anchor; the first is an abbr, the second is a span.

`<section>
                <a href="#poppy" class="wb-lbx" title="Phasers go brrr">Activate phasers</a>
</section>
<section id="poppy" class="mfp-hide modal-dialog modal-content overlay-def">
                <header class="modal-header">
                                <h3 class="modal-title">Starship Enterprise</h3>
                </header>
                <div class="modal-body">
                                <div class="wb-tabs">
                                                <div class="tabpanels">
                                                                <details id="pop1">
                                                                                <summary><abbr title="Starship Enterprise">NCC-1701</abbr></summary>
                                                                                <p>Pew pew! You killed the Klingons</p>
                                                                </details>
                                                                <details id="pop2">
                                                                                <summary><span>USS Reliant</span></summary>
                                                                                <p>Prrrt! Oh no, you killed a whale!</p>
                                                                </details>
                                                                <details id="pop3" open="open">
                                                                                <summary>Borg Cube</summary>
                                                                                <p>All your base are belong to us!</p>
                                                                </details>
                                                </div>
                                </div>
                </div>
</section>`

GormFrank (approved):

PR #8940, opened 2020-09-21
Update v8.0.2 doc to include STR and deprecate dwnld page

GormFrank:


GormFrank:
Pre-approved during technical review.

duboisp (approved):

PR #8947, opened 2020-09-30
Multimedia player: Add hidden volume slider label

### LINKED ISSUE #8939: Multimedia Player - Video: Only title used to generate label for form element
Multimedia Player - The label for the Volume control
https://wet-boew.github.io/v4.0-ci/demos/multimedia/multimedia-en.html

`<input type="range" class="volume" aria-controls="wb-auto-2-md" aria-labelledby="wb-auto-2-md" title="Volume" min="0" max="100" value="100" step="5">`

The title attribute is only supposed to provide extra or fallback information, not essential information.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title

Suggested Solution
Add an visually hidden label element "Volume"

EricDunsworth:
Doesn't function like a typical label... aria-describedby is being used to programmatically-associate it to the volume slider.

The volume slider was already using an aria-labelledby attribute to programmatically-associate it to the video's title - just like all the other controls. That causes the video's title to function as every control's label. So for each control, screen readers announce the video's title as a prefix, then the control's attributes, then the control's name (i.e. "Play", "Mute", etc...).

It would've been possible to add the label's ID at the end of the volume slider's aria-labelledby attribute, but that caused "Volume" to be announced before the slider control's attributes in NVDA + Firefox/Chrome. So it produced "Looking for a Job Volume slider 100" instead of "Looking for a Job slider 100 Volume". Not ideal.

By comparison, aria-describedby gets announced after the volume slider's attributes and supersedes the title attribute. That relegates the label to secondary information, but is better overall since it's consistent with how all the other controls are announced.

Fixes #8939.

### DIFF
--- a/src/plugins/multimedia/assets/mediacontrols.html
+++ b/src/plugins/multimedia/assets/mediacontrols.html
@@ -9,7 +9,8 @@ wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licenc
 		<div class="btn-group">
 			<button type="button" class="btn btn-default playpause" aria-controls="{{mId}}"  aria-labelledby="{{mId}}" title="{{play}}" data-state-on="{{play}}" data-state-off="{{pause}}"><span class="glyphicon glyphicon-play"><span class="wb-inv">{{play}}</span></span></button>
 			<button type="button" class="btn btn-default mute" aria-controls="{{mId}}" aria-labelledby="{{mId}}" title="{{mute_on}}" data-state-on="{{mute_on}}" data-state-off="{{mute_off}}" aria-pressed="false"><span class="glyphicon glyphicon-volume-up"><span class="wb-inv">{{mute_on}}</span></span></button>
-			<input type="range" class="volume" aria-controls="{{mId}}" aria-labelledby="{{mId}}" title="{{volume}}" min="0" max="100" value="100" step="5" />
+			<p id="{{mId}}-vlm-lbl" hidden>{{volume}}</p>
+			<input type="range" class="volume" aria-controls="{{mId}}" aria-labelledby="{{mId}}" aria-describedby="{{mId}}-vlm-lbl" title="{{volume}}" min="0" max="100" value="100" step="5" />
 		</div>
 	</div>
 	<div class="tline">


EricDunsworth:
@GormFrank @duboisp Could a demo site based on this PR's branch be created?

@szet0018 @bsouster @andrewnordlund Any chance this could be tested in an army of screen readers :)?

I've tried NVDA 2020.2 in Firefox/Chrome/IE11. Firefox/Chrome were fine, but IE11 wasn't so great. But it wasn't any worse in IE11 than it was before. So it hasn't regressed.

Unfortunately I don't have any other screen readers or OSes to test with beyond Windows 10 + NVDA.

Btw for the time being I've set the label's value to end with a 1 ("Volume 1") to distinguish it from the volume slider's ``title`` attribute ("Volume"). If testing works out, I'll remove the 1 and promote this PR out of draft.

src/plugins/multimedia/assets/mediacontrols.html
```diff
@@ -9,7 +9,8 @@
 		<div class="btn-group">
 			<button type="button" class="btn btn-default playpause" aria-controls="{{mId}}"  aria-labelledby="{{mId}}" title="{{play}}" data-state-on="{{play}}" data-state-off="{{pause}}"><span class="glyphicon glyphicon-play"><span class="wb-inv">{{play}}</span></span></button>
 			<button type="button" class="btn btn-default mute" aria-controls="{{mId}}" aria-labelledby="{{mId}}" title="{{mute_on}}" data-state-on="{{mute_on}}" data-state-off="{{mute_off}}" aria-pressed="false"><span class="glyphicon glyphicon-volume-up"><span class="wb-inv">{{mute_on}}</span></span></button>
-			<input type="range" class="volume" aria-controls="{{mId}}" aria-labelledby="{{mId}}" title="{{volume}}" min="0" max="100" value="100" step="5" />
+			<label id="{{mId}}-vlm-lbl" for="{{mId}}-vlm" hidden>{{volume}} 1</label>
```
duboisp, line 12: Because this `<label>` element is not used as being the main label for the input and it wont be. Please change it for a paragraph and keep it hidden.

duboisp (changes requested):

src/plugins/multimedia/assets/mediacontrols.html
EricDunsworth: I had considered doing something along those lines, but in the end felt that a label might still be a better bet overall (even if it's not currently functioning like one in practice) since that's more in line with how that kind of relationship should traditionally be coded. But I guess you're right. It adds extra complexity for nothing. I'll change it.

IMO the way all of the media player controls are currently labelled is kind of strange. I find it inappropriate to be programmatically-identifying the video's title as every control's label. It's bloated/repetitive and relegates useful details about the controls to secondary information. At best it could maybe be argued that the video's title could be associated to the controls with ``aria-describedby`` (to relegate it to secondary information announced at the end), but even that might be a bit of a stretch.

EricDunsworth: Updated - went with ``<span>`` instead of ``<p>``.

duboisp: `<span>` don't provide any semantic and `<span>` is a text level.

src/plugins/multimedia/assets/mediacontrols.html
```diff
@@ -9,7 +9,8 @@
 		<div class="btn-group">
 			<button type="button" class="btn btn-default playpause" aria-controls="{{mId}}"  aria-labelledby="{{mId}}" title="{{play}}" data-state-on="{{play}}" data-state-off="{{pause}}"><span class="glyphicon glyphicon-play"><span class="wb-inv">{{play}}</span></span></button>
 			<button type="button" class="btn btn-default mute" aria-controls="{{mId}}" aria-labelledby="{{mId}}" title="{{mute_on}}" data-state-on="{{mute_on}}" data-state-off="{{mute_off}}" aria-pressed="false"><span class="glyphicon glyphicon-volume-up"><span class="wb-inv">{{mute_on}}</span></span></button>
-			<input type="range" class="volume" aria-controls="{{mId}}" aria-labelledby="{{mId}}" title="{{volume}}" min="0" max="100" value="100" step="5" />
+			<span id="{{mId}}-vlm-lbl" hidden>{{volume}} 1</span>
```
duboisp, line 12: > The `span` element doesn't mean anything on its own

ref. https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-span-element

EricDunsworth: What about the [HTML spec's definition of paragraphs](https://html.spec.whatwg.org/multipage/dom.html#paragraphs)?

The ``p`` element is not the only means by which a paragraph can be represented (although it's obviously the most explicit way of denoting it).

From my understanding, what's currently being done in mediacontrols.html (**without** my PR's changes) technically represents an "implied" paragraph of buttons and inputs (since they're adjacent elements whose content model is [phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2)).

Using using a ``span`` element to label the input would fit in with that structure since it's also phrasing content. By comparison, using a ``p`` element would technically represent 3 paragraphs (since the "real" paragraph in-between the "implied" ones would break the sequence of adjacent phrasing elements).

So semantically, IMO the way I'm proposing (``span``) would be more semantically-correct than the way you're proposing (``p``). But I don't mind amending this PR to use ``p`` if you insist. Either way is better than the status quo in any case :).

duboisp: The ask here is to be explicit and not relying on implicit semantic.

EricDunsworth: I still personally disagree with using ``p`` over ``span`` in this scenario since the group of buttons/inputs were already an implicit paragraph. My intent was to modify something inside of it - not introduce a new paragraph.

Anyway, I've updated this PR to use ``p`` to bring this PR back to life :).

EricDunsworth:
Gonna bring this out of draft now since everything's good to go on my end. The remaining sticking point is that it'll need to be tested in a variety of screen readers (like JAWS, VoiceOver, etc) that I don't have at my disposal.

GormFrank (commented):
Approved as long as screen reader tests all pass.

EricDunsworth:
@shawnthompson Just wondering if you'd be able to get this tested in other common screen readers to ensure it's consistent with the results I got in NVDA? @GormFrank suggested I bug you :P lol.

shawnthompson:
Can we get a version of this PR hosted somewhere and I can send it to the AWG for testing with different screen readers?

EricDunsworth:
@GormFrank Would it be possible to get a copy of it up on Universal labs?

If not, I guess I could deploy a copy of this PR's built pages to my fork's ``gh-pages`` branch, but that would only be temporary. Not sure how long it'd take to conduct testing on it...

shawnthompson:
Have you heard of https://raw.githack.com? I've been looking to try it out but haven't had the chance. W3C uses it for PRs for WCAG 3.0.

GormFrank:
@EricDunsworth @duboisp @shawnthompson Here it is: https://universallabs.org/wet/labs/pr8947/demos/multimedia/multimedia-en.html

eburggraaf:
Hello, I'm Erik, and I've been working with Ben and Shawn on testing the proposed controls for this player using a screen reader.  From earlier posts in this thread, it looks like aria-labelledby is being used to override the actual button values.  Please reconsider this.  It's technically incorrect and It's creating a lot of problems from an accessibility standpoint.
		
Because of aria-labelledby, the name of the video appears to be the control’s name from a screen reader point of view.  This isn’t allowed. The buttons and controls all need to be self-aware so they can be found in a screen search.  For example, if I go to the top of the webpage and do a jaws find for the word play, jaws finds nothing, because there isn’t a control named play on the screen.  When I do an insert f5 to get the controls list, all of the player controls are called “looking for a job”.  In both of these cases, we need to see “play”, “mute”, “volume”, and “Closed captions”.  This is also the case when I use item chooser on the mac.

As you can see, the handling of the tooltips that contain the actual names of the buttons in various screen reader and browser combinations varies wildly, with some tips on some configurations being detected while others not, or only when tabbing or only when arrowing.  Ideally, we want the tooltip information displayed in the control name and that currently displayed in the control name in the tooltip.  The tooltips appear to come from title attributes, an approach that I personally like, but I suppose only because it largely works in my favourite screen reader/browser combinations. It can't be used anyhow because of the below requirement for more than one player on a page.

If this is the way the page looks to dictation software, the player controls as they are will thwart that as well.  I can see that the actual name of each button does properly reflect it's action.  IE, the inner html of the buttons read, "Play", "Mute", etc.  This will need to be tested if you decide to keep the aria-labelledby for some reason.

This lead to a further discussion among the testers because we have a requirement to support multiple instances of a player on a page.  In order to maintain self-awareness, the buttons can't simply be called their actions as would be preferred.  If that were the case, then we would have two or more "play", "mute"...  And self-awareness goes out the window.  Thus, we need to eligantly incorporate the name of the video into the button name.

We currently store the name of the video in the ID of the main player container.  So clearly we have this variable that we can use.  Here is a list of places we might be allowed to use that variable.
In the innerHTML of the button,
In the alt tag of an image in the innerHTML of the button,
In the title attribute of the button that has no innerHTML.

In other words, in order to have multiple players on screen at once, the name of the video can not be missing from the buttons, but can in no way superceed the description of what action that button does.  Examples of this would be:
The title attribute of a button that has innerHTML,
Aria-describedby,
aria-labelledby.

Any of these will result in the button losing self-awareness.

We also discussed using regions, landmarks, and other grouping mechanisms to automatically speak the name of the video when a user enters a group.  Unfortunately, that has the same inconsistancies as demonstrated by our spreadsheet of screen reader feedback, and still leaves the buttons unaware of how to describe themselves.  So, no screen search, no clicking with dictation.

So to conclude on this, the variable containing the title of the video must be emplaced into either the innerHTML of the button alongside the button's action, or into an alt tag attached to an image in the innerHTML of the button alongside the button's action, or in a title attribute of a button that has no innerHTML alongside the button's action.  Thus, the text of the button will visually and screenreaderically read "play video title" button, "mute video title" button...  The buttons will be fully self-aware and detectable by access technologies accounting for more than one player on a page.

If we discard the requirement to have more than one player on a page, then we get a lot more flexibility in what we can do.  That must be a conversation for another post though.

shawnthompson:
Thanks @eburggraaf. I will be sharing the data we found for the inconsistencies when using ARIA in the different screen reader / browser combinations shortly.

shawnthompson:
I created this page with the data we collected for what is read the screen reader combinations:

[WET issue: Multimedia player: Add hidden volume slider label #8947: data](https://a11y.canada.ca/8947/en/)

duboisp:
@EricDunsworth it's up to you to create a new PR to address the suggestion and your need in this PR. Or we can move ahead with this PR and have a second PR to address the suggestion.

duboisp (approved):
The "volume" information don't seem to be spoke twice. So I am ok with this PR. @EricDunsworth i will wait for your confirmation before to merge it in the case you do a new PR.

EricDunsworth:
@eburggraaf @shawnthompson What would you think about merging this PR "as-is" to get a quick win for #8939, then look into a bigger rework of the media player's control labels via a separate PR? As @duboisp mentioned, it doesn't look like this PR's change is degrading anything compared to what currently exists. All it does in practice is improve semantics.

FWIW I agree that the status quo with video names acting as prefixes for every media player control is super annoying and want to try sorting it out. But I wanted to start off by fixing the volume slider issue in what already exists before trying to rework things as a whole.

@eburggraaf I also had a few additional questions for you :).

**1.**
> The buttons and controls all need to be self-aware so they can be found in a screen search. For example, if I go to the top of the webpage and do a jaws find for the word play, jaws finds nothing, because there isn’t a control named play on the screen. When I do an insert f5 to get the controls list, all of the player controls are called “looking for a job”. In both of these cases, we need to see “play”, “mute”, “volume”, and “Closed captions”. This is also the case when I use item chooser on the mac.

Just wondering - do those screen reader features try appending ``aria-describedby`` values to the end of the control names they announce in the scenarios you mentioned (e.g. JAWS find/control list, VoiceOver item chooser, etc)?

For example, with the volume slider control (which uses a void ``input`` element):
* Most screen readers currently announce the volume slider control in this order:
  * "[``aria-labelledby``] [slider description] [silder value] [``aria-describedby``] [JAWS: min/max values]"
* Which comes out like this in some of your JAWS samples:
  * "[Looking for a Job] [left right slider] [95] [Volume] [min 0 max 100]"

I believe the new kind of labelling structure you're proposing would cause the volume control to be announced like this:
* "[Volume Looking for a Job] [left right slider] [95] [min 0 max 100]"

Since users would probably prefer to hear the volume level before the video's name, I was thinking that maybe ``aria-describedby`` could come in handy to relegate the video's name to the end of the announcement. But only if that doesn't end up causing other issues.

**2.**
> In order to maintain self-awareness, the buttons can't simply be called their actions as would be preferred. If that were the case, then we would have two or more "play", "mute"... And self-awareness goes out the window. Thus, we need to eligantly incorporate the name of the video into the button name.

How would this apply to pages that only contain one instance of the multimedia player? Would you recommend trying to make the plugin capable of excluding the video name from control labels if it determines that there's only one video in the current page?

duboisp:
@EricDunsworth please move the conversation into a new issue and include the test from @eburggraaf and your last comments.

Once that is completed, the changes in this PR will be ready to be merged.

Issue #8918, opened 2020-08-07
Toggle: Screen readers cannot read text inside accordion panels


szet0018:
https://wet-boew.github.io/v4.0-ci/demos/toggle/toggle-en.html#accordion

In the example above, neither JAWS 2020/NVDA on Chrome 84.0.4147.105 nor NVDA on Firefox 79 can read the text inside the opened panels. Neither ARROW keys nor TAB allows the screen reader to access the text inside the panel when the panel does not have interactive elements like form fields that can be TABed to.

This may be due to the div wrapping around the accordion given role="tablist"
`<div class="accordion" role="tablist">`

role="tablist" automatically triggers application mode in most screen readers.
Regular text in p, headings, div, span, lists, tables, etc. are not available to screen readers in application regions unless it is keyboard focusable.

Tried adding role="document" around the panel, but this did not solve the problem. Not sure what the solution is.

RubyDo:
CC @duboisp  to stay on the loop for the issue

GormFrank:
Relates to #8922

piopi:
@duboisp @GormFrank I think this issue can also be closed because of the improvement done in https://github.com/wet-boew/wet-boew/pull/8948#issuecomment-703368337 , Screen readers can read text inside accordion panels.

GormFrank:
Thank you for your contribution @piopi! @szet0018 Do you want us to close this issue right now or do you want to wait and test when it gets to Canada.ca in a few weeks first?

PR #8948, opened 2020-10-01
Removed the Accordion "summary" role="tab" to improve accessibility

### LINKED ISSUE #8919: Toggle: Accordion <summary> has role="tab" which is not allowed
https://wet-boew.github.io/v4.0-ci/demos/toggle/toggle-en.html#accordion

aXe accessibility checker is reporting the following issue:

**ARIA role must be appropriate for the element**

**Issue description**
Ensures role attribute has an appropriate value for the element
Impact: minor

**Element location**
`#wb-auto-11`

**Element source**
`<summary class="wb-toggle tgl-tab wb-init wb-toggle-inited" data-toggle="{&quot;parent&quot;: &quot;.accordion&quot;, &quot;group&quot;: &quot;.acc-group&quot;}" id="wb-auto-11" role="tab" aria-selected="false" tabindex="0" aria-posinset="1" aria-setsize="3">Example 1</summary>`

To solve this violation, you need to:
Fix the following:
ARIA role tab is not allowed for given element

No role is permitted on summary, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary
- Default role for summary is button

piopi:
Removed the Accordion "summary" role="tab" to improve the accessibility, since ARIA role tab is not allowed for summary tags. Please see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary for more details.  Fixes #8919.

### DIFF
--- a/src/plugins/toggle/test.js
+++ b/src/plugins/toggle/test.js
@@ -114,10 +114,6 @@ describe( "Toggle test suite", function() {
 				data = $( this ).data( "toggle" );
 				$parent = $( data.parent );
 				expect( $parent.attr( "role" ) ).to.equal( "tablist" );
-
-				$parent.find( ".tgl-tab" ).each( function() {
-					expect( this.getAttribute( "role" ) ).to.equal( "tab" );
-				} );
 				$parent.find( ".tgl-panel" ).each( function() {
 					expect( this.getAttribute( "role" ) ).to.equal( "tabpanel" );
 				} );
--- a/src/plugins/toggle/test.js
+++ b/src/plugins/toggle/test.js
@@ -114,6 +114,9 @@ describe( "Toggle test suite", function() {
 				data = $( this ).data( "toggle" );
 				$parent = $( data.parent );
 				expect( $parent.attr( "role" ) ).to.equal( "tablist" );
+				$parent.find( "div.tgl-tab" ).each( function() {
+					expect( this.getAttribute( "role" ) ).to.equal( "tab" );
+				} );
 				$parent.find( ".tgl-panel" ).each( function() {
 					expect( this.getAttribute( "role" ) ).to.equal( "tabpanel" );
 				} );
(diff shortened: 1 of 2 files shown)


duboisp (dismissed):
The solution would need to impact the tabs interface too because it use that pattern.

piopi:
@duboisp thank you for your feedback. I checked the WCAG documentation https://www.w3.org/TR/html-aria/, and we can see that the summary and details tags cannot have defined roles because they have an implicit role of a button and a group respectively and also have some additional aria-* attributes restrictions. As for the wrapping, they are wrapped actually in a tablist elements: https://github.com/wet-boew/wet-boew/blob/0ccb73de71e284cd2b486264f237972a2071b660/src/plugins/toggle/toggle.js#L97
The issue with a tablist elements is the requirement that all of its children need to have a role of tab and since we can't give roles to the summary/details tags, I can see two solutions for this problem. 
1. Wrap the details/summary in a div that has a tab role and the appropriate aria-* attributes while keeping the tablist role on the parent

or

2. Remove the tablist role from the parents and make the appropriate changes on the tabs interface

I will start with the first solution and do another PR.

piopi:
Added the div wrapping, removed the incompatible aria role/attributes from the summary/details accordion, and made sure to give the div the appropriate aria-roles and attributes.

As we can see from the screenshots below, these changes removed the accessibility issues found before:

**Before**
<details><summary>Click to expand </summary>

image

image

</details>

**After**
<details><summary>Click to expand </summary>

image

image

</details>

piopi:
After testing with NVDA screen reader v.2020.2 on Firefox and Chrome, it seems that the issue #8918 is also solved with this solution.
image

duboisp:
Change approved upon successful testing.

GormFrank (approved):

duboisp:
@piopi and @GormFrank we will need to revert those change, it prevent to initialize the subsequent tabbed interface.

You can see the issue here: https://wet-boew.github.io/wet-boew/demos/tabs/tabs-en.html and the error console show an error "Uncaught DOMException: Node.replaceChild: Child to be replaced is not a child of this node" for line 131 - https://github.com/wet-boew/wet-boew/blame/master/src/plugins/toggle/toggle.js#L131

piopi:
@duboisp @GormFrank Issue was solved here https://github.com/wet-boew/wet-boew/pull/8986

PR #8951, opened 2020-10-02
Multimedia player: Fix presentation of closed captioning errors

EricDunsworth:
The media player previously had strange, buggy behaviour when failing to load closed captions. It put the CC button into a disabled state and inserted an error message into the collapsed captions panel.

But that setup had several problems:
* The captions panel isn't open by default and therefore couldn't be expanded since the CC button was disabled.
* The same error message was repeated twice in a row.
* The errors overflowed on top of content below the media player.
* The errors were almost invisible since they use a white colour.
* The errors were hidden via display: none; while the media player was playing... and unveiled when paused.
* Hovering over the disabled CC button caused its background colour to brighten and contrast poorly with the CC icon.
* Clicking the disabled CC button caused a flicker effect.
* The disabled CC button had an aria-pressed="false" attribute... which got announced by screen readers despite being irrelevant.

This commit resolves the problems as follows:
* Automatically opens the captions panel if an error occurs.
* Intentionally keeps the panel open to make the error message stand out. It can't be collapsed.
* Only shows one instance of the error.
* Doesn't hide the error during playback.
* Fixes the disabled CC button's hover/flicker issues.
* Removes the CC button's aria-pressed="false" attribute.

**Screenshots...**

**Before (dotted white pixels in the "View code" summary's top border):**
multimedia-cc-error-before-1

**Before (highlighted the white error messages):**
multimedia-cc-error-before-2

**After:**
multimedia-cc-error-after

### DIFF
--- a/src/plugins/multimedia/_base.scss
+++ b/src/plugins/multimedia/_base.scss
@@ -87,13 +87,6 @@
 			@include multimedia_overlay_play;
 		}
 
-		&.playing {
-
-			.errmsg {
-				display: none;
-			}
-		}
-
 		&.waiting {
 			@include multimedia_overlay_loading;
 		}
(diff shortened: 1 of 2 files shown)


duboisp:

duboisp (dismissed):
There is weird behaviour where the close caption button still work when there is an error. May be add a plain paragraph `$elm.after()` the close caption block

EricDunsworth:
@duboisp @GormFrank I managed to fix the captions toggle issues. Also snuck in a change to remove the ``aria-pressed`` attribute (it's pointless if the button can't be pressed and gets announced by screen readers even if the button is disabled).

Btw the issue that caused clicking the edges of the captions panel to cause it to collapse will be fixed by #8960. It conflicts with this PR. I'd recommend merging that one first.

GormFrank:
Approved upon testing.

GormFrank (approved):

PR #8952, opened 2020-10-02
Multimedia player: Hide closed captions in print view

EricDunsworth:


duboisp:
To be tested at the same time of #8951

GormFrank (approved):

PR #8954, opened 2020-10-08
Multimedia player: Make YouTube play button background opaque

EricDunsworth:
The multimedia's play button uses a translucent (70% opacity) background colour. That looks fine in normal videos.

But in YouTube videos... the translucency causes the YouTube player's play button to be visible behind the WET player's play button. It's especially noticeable when hovering over the video's viewport (which brightens YouTube's play button).

This commit resolves it by using an opaque colour for the play button's background. But only for YouTube videos. Normal videos will still continue to use translucent backgrounds.

**Screenshots...**

**Before (default):**
multimedia-youtube-before-default

**Before (hover - bright YouTube play button is visible behind the WET player's play button):**
multimedia-youtube-before-hover

**After:**
multimedia-youtube-after

### DIFF
--- a/src/plugins/multimedia/_base.scss
+++ b/src/plugins/multimedia/_base.scss
@@ -144,6 +144,16 @@
 				@extend %multimedia-display-none;
 			}
 		}
+
+		&.video {
+			&:not(.playing):not(.waiting) {
+				.display {
+					&:before {
+						background: $mm-ctrl-bg-color;
+					}
+				}
+			}
+		}
 	}
 
 	/*


EricDunsworth:
@duboisp @GormFrank Btw I could change the background to black if you'd prefer. I'm currently using dark grey to for consistency with the media player's controls. But I'd personally prefer black.

duboisp:
To be tested at the same time of #8951 and #8952

EricDunsworth:
Just wanted to point out something else about colours.

The play button's "normal" translucent background colour comes from [``%global-centered-overlay``](https://github.com/wet-boew/wet-boew/blob/master/src/base/_wet-boew.scss#L55-L66) and [``%global-loading``](https://github.com/wet-boew/wet-boew/blob/master/src/base/_wet-boew.scss#L78-L80). The translucent colour is used for play/loading overlays in the media player's dark+light themes and the geomap/twitter timeline plugins.

So in other words, that translucent background colour doesn't normally have any sort of connection to the multimedia player player's controls. In fact, the translucency is derived from black. So it could potentially be argued that it'd make more sense to use black for this PR.

@duboisp Thoughts?

GormFrank (approved):

PR #8960, opened 2020-10-14
Multimedia player: Fix captions collapse bug

EricDunsworth:
Clicking certain areas of the captions panel was causing it to collapse.

Specifically:
* Anywhere inside the panel if no caption text is visible
* The edges of the panel (i.e. its padding)
* The ::before pseudo-element's space character

That behaviour was due to a click event that used a flawed regex to check for the CC button's "cc" class. Its looked for word boundaries around "cc"... but dashes count as boundary separators. So the captions panel's "wb-mm-cc" class was getting matched.

This fixes it by replacing the regex's first boundary with an alternation operator.

Related to #8951.

**Screenshot (clicking anywhere in the highlighted purple area used to cause the captions panel to collapse):**
multimedia-cc-panel-highlight


GormFrank:
Approved upon testing.

GormFrank (approved):

PR #8962, opened 2020-10-15
Multimedia player: Show caption errors if parsing fails

EricDunsworth:
Caption errors currently only appear if external captions fail to load (e.g. if a 404 error occurs while trying to load an external XML/HTML file). That only covers one scenario.

This change takes things a step further. It triggers caption errors if no caption items are returned after parsing.

Examples of situations that benefit:
* External XML/HTML files that exists... but don't actually contain any captions
* Missing container ID or "tmtxt" classes in HTML captions


GormFrank:
Approved upon testing.

GormFrank (approved):

PR #8965, opened 2020-10-17
Number Range Table Filter (New Feature)

ghost:
Concept for a new feature that been discussed/requested in a few issue trackers. Well need testing.


thekodester:
Files for testing purposes:
[employees.json.txt](https://github.com/wet-boew/wet-boew/files/5395099/employees.json.txt)
[tables-test.html.txt](https://github.com/wet-boew/wet-boew/files/5395100/tables-test.html.txt)

src/plugins/tables/tables.js
```diff
@@ -228,16 +228,43 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
...
 		if ( $elm.is( "select" ) ) {
 			$value = $elm.find( "option:selected" ).val();
-		} else if ( $elm.is( ":checkbox" ) ) {
+		} else if ( $elm.is( "input[type='number']" ) ) {
+			var $minNum, $maxNum, $fData;
 
-			// Verifies if using same checkbox list
-			if ( $column !== $lastColumn || $lastColumn === -1 ) {
-				$cbVal = "";
+			// Retain minium number (always the first number input)
```
RobJohnston, line 243: minium -> minimum

src/plugins/tables/tables.js
```diff
@@ -228,16 +228,43 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
...
+			if ( $cachedVal === "" ) {
+				$cachedVal = $elm.val();
 			}
-			$lastColumn = $column;
+			$minNum = $cachedVal;
+
+			// Maxium number is always the current selected number
```
RobJohnston, line 249: Maxium -> maximum

thekodester:
FYI: I only have done basic testing with Safari.

I would recommend testing:
- [ ] Combined with additional filters (selects, checkboxes)
- [ ] Additional web browsers (Firefox, IE, Chrome)

thekodester:
Did some further testing with different number formats. That seems to be corrected.

I separated the filter examples to their own page as the table example page was becoming long.

src/plugins/tables/tables-en.hbs
```diff
@@ -972,193 +972,5 @@
...
-					<th>Minister</th>
-				</tr>
-				</thead>
-			</table>
-		</div>
-	</div>
-
-
-
+	<p>View example <a href="../../demos/tables/tables-fltrs-en.html">here</a>.</p>
```
duboisp, line 975: Instead of this, can you add a table of content with anchor at the top of the page?

Kind of like fieldflow working example: https://wet-boew.github.io/themes-dist/GCWeb/demos/fieldflow/fieldflow-en.html

src/plugins/tables/tables-fltrs-en.hbs
```diff
@@ -0,0 +1,292 @@
...
+	"language": "en",
+	"category": "Plugins",
+	"description": "Integrates the DataTables plugin into WET providing searching, sorting, filtering, pagination and other advanced features for tables.",
+	"tag": "tables filtering example",
+	"parentdir": "tables filtering example",
+	"altLangPrefix": "tables filtering example",
+	"dateModified": "2020-10-22"
+}
+---
+<section>
```
duboisp, line 13: Can you add a section that lists what filtering option is available, and include notes for author as needed on how to use that options.

duboisp (changes requested):
Thanks for adding those features, 
It looks good to merge, we will need to test it locally first. 

The request changes is about editorial changes, let me know if you need assistance. You can always join us during our weekly code sprint, see the info here: https://github.com/wet-boew/wet-boew/wiki/WET-BOEW-Code-sprint

thekodester:
Did you want the information about filters on the page with the examples or on https://wet-boew.github.io/v4.0-ci/docs/ref/tables/tables-en.html ?

thekodester:
The links are now:
Screen Shot 2020-10-22 at 2 30 40 PM

thekodester:
The links are now:
Screen Shot 2020-10-22 at 2 36 21 PM

thekodester:
Made some updates. I decided to move the how to information from https://wet-boew.github.io/wet-boew/docs/ref/tables/tables-en.html to the example page dedicated to filters.

thekodester:
Screenshot of changes:
Screen Shot 2020-10-22 at 4 46 55 PM
Screen Shot 2020-10-22 at 4 47 11 PM
Screen Shot 2020-10-22 at 4 47 21 PM

GormFrank (changes requested):
Alright so I tested and everything seemed to work as expected until I got to try out French numbers. Things like this: 

- `2 400`
- `2 400,50 $`
- `2&nbsp;400`
- `2&nbsp;400&nbsp;$`
- `2&nbsp;400,50&nbsp;$`

Do not work at all. French numbers have a space for thousands and for currencies more specifically the sign comes at the end with a space before. Therefore, it is also likely to see the numbers written with non-breaking spaces symbols (&nbsp;) to keep everything on one line. Also, the decimal separator is a comma instead of a period in French.

The input of type number obviously doesn't support spaces, but this is a browser issue. 

I noticed that if you put a number as a minimum and no number as a maximum, it doesn't show results based on the minimum provided. I am not against it but is it the behaviour that we want? Should we make the maximum required when the minimum is provided and vice-versa? I am just starting a discussion here.

src/plugins/tables/tables.js
```diff
@@ -307,6 +335,18 @@ $document.on( "click", ".wb-tables-filter [type='reset']", function( event ) {
...
 } );
 
+// converts a number string to actual number, see https://stackoverflow.com/a/42213804
+function parseLocaleNumber( stringNumber ) {
+	stringNumber = stringNumber.replace( /[^0-9\,\.\-]+/g, "" );
+	var thousandSeparator = ( 1111 ).toLocaleString().replace( /1/g, "" );
+	var decimalSeparator = ( 1.1 ).toLocaleString().replace( /1/g, "" );
+
+	return parseFloat( stringNumber
+		.replace( new RegExp( "\\" + thousandSeparator, "g" ), "" )
```
thekodester, line 345: The issue with french numbers is this line.

What this issue suppose to do is convert 25,000.00 to 25000.00, but it also changing 2 500,50 to 250050.

Any thoughts?

thekodester: This function should be removed, after further testing it causes lot of issue, maybe like:
```
var $form = $( this ),
....
$toNumber = function stringToNumber( number ) {
number = number.replace( /[^0-9\,\.\-]+/g, "" );
console.log( "number: " + number );
return parseFloat( number );
};
```
Console results are:
````
number: 38198.01
number: 25400,50
number: 34259,68
number: 47322.96
number: 25400,50
number: 35,017.12
````
Now last thing to do is to find away to deal with 35,017.12 and 34259,68.

**Comment by thekodester** [comment]
I found a way to make this work there is a caveat though. We'll only work with numbers with 2 decimal points. We could add this as a rule to the documentation.

Sadly there no simple way to do this otherwise, as JavaScript was not design to take Canadian French Numbers/Currency Formats (but it well create currency formats ironically).

File: src/plugins/tables/tables.js [hunk 6]
```diff
@@ -228,16 +236,44 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
...
+
+			// Maximum number is always the current selected number
+			$maxNum = parseFloat( $elm.val() );
+
+			// Generates a list of numbers (within the min and max number)
+			if ( !isNaN( $minNum ) && !isNaN( $maxNum ) ) {
+				$fData = $datatable.column( $column ).data().filter( function( obj ) {
+					var $num = $toNumber( obj.toString() );
+
+					if ( !isNaN( $num ) && $num >= $minNum && $num <= $maxNum ) {
```
thekodester, line 265: As for the other thing @GormFrank that would be here in the code, currently it must have both.

It could be `!isNaN( $num ) && ( $num >= $minNum || $num <= $maxNum )`.

Now if you wanted both the AND and OR options I need to code advance options similar to the checkboxes (already exist).

thekodester:
Most recent commit adds:
```
Advanced Options For Number Range
Allows the user to change how number range is filtered (default is EITHER).
Add data-aopts='{"type": "", "column": ""}' arrtibute to any radio element to apply the advanced options.

Available filter types:
EITHER - This option will return all entries that are either equal to/greater then A OR/AND equal to/less then B.
AND - This option will only return entries that are equal to/greater then A AND equal to/less then B.
Column is the number range that the filter well apply too (must match there 'data-column').
<label for="dt_age_aopts1"><input type="radio" name="dt_age_aopts" id="dt_age_aopts1" data-aopts='{"type": "either", "column": "5"}' checked />Either</label>
```

GormFrank:
I will test as soon as I can and let you know.

GormFrank (approved):
Honestly, love the approach and the documentation!

GormFrank:
Approved during technical review as long as Pierre's editorial change was addressed. @duboisp Are you ok with what has been done?

duboisp (approved):

Issue #8959, opened 2020-10-14
Lightbox: Navigation arrows do not pass colour contrast when overlayed on certain images


nfrenette:
Depending on the image dimensions and viewport, the arrows can be overlayed on the image.
![image](https://user-images.githubusercontent.com/15619641/96011930-90b5d280-0e11-11eb-82ef-1a55b6783ebf.png)

The arrows do not always provide enough contrast with the image.
![image](https://user-images.githubusercontent.com/15619641/96012669-56006a00-0e12-11eb-8b88-019d44f396e4.png)

This issue is to investigate solutions for providing enough contrast, for example:
![image](https://user-images.githubusercontent.com/15619641/96012704-6284c280-0e12-11eb-9d1e-6db22c71f204.png)

duboisp:
You can add the CSS here: https://github.com/wet-boew/wet-boew/blob/master/src/plugins/lightbox/_base.scss

xiayitan:
@duboisp 
I found the comment on the CSS file that you indicated:
https://github.com/wet-boew/wet-boew/blob/master/src/plugins/lightbox/_base.scss#L88

/* Should fix upstream in Magnific Popup rather than overriding in WET */

Please confirm that I can still add the CSS there. Thank you!

duboisp:
Yes you can still add CSS there.

PR #8967, opened 2020-10-22
Lightbox: Fix color contrast for navigation arrows

xiayitan:
To solve the issue:
https://github.com/wet-boew/wet-boew/issues/8959


duboisp:
Approved at the technical upon successful testing.

duboisp (approved):

GormFrank:
That works. Here is a little sneak peak of what it looks like: 
lightbox-prev-next

GormFrank (approved):

PR #8970, opened 2020-10-28
Update Share Plugin Docs and Remove Bit.ly Image

ghost:
Updates the share docs and remove unused icon.


GormFrank:
@thekodester Can you please rename your PR so that it primarily reflects the fact that you are removing an image? and then mention the documentation update in the description. A title like: "Share plugin - Removed bit.ly image and update doc". Because that's what's more important to the API.

thekodester:
Bit.ly image should of been removed here (same with doc updates my mistake): https://github.com/wet-boew/wet-boew/pull/8775

GormFrank:
@thekodester Thank you for renaming your PR! I have no problem with all the social medias you have removed in the doc, but are we sure that Bitly isn't a thing anymore? 

If it is still alive I would prefer keeping the image in there because people could still want to use it, otherwise we would cause a major version by removing it. We could still remove it from the doc however since we are not strongly supporting its use (then your PR would go back to being called "Update Share plugin Docs"). 

If it is deprecated or is by subscription payments only, then we could get rid of it no problem and it would be considered a patch.

thekodester:
It is deprecated. See https://github.com/wet-boew/wet-boew/pull/8775#issuecomment-565099187 for more info.

GormFrank (approved):

PR #8975, opened 2020-11-03
Lightbox: example for Full width inline content

xiayitan:
The new example is called "Full width inline content" which is placed under Examples-Single items. 

It's similar as the example of Inline content but the popup window is in full width.

### DIFF
--- a/site/pages/docs/ref/lightbox/lightbox-en.hbs
+++ b/site/pages/docs/ref/lightbox/lightbox-en.hbs
@@ -59,6 +59,18 @@
 		</thead>
 		<tbody>
 			<tr>
+				<td><code>mfp-hide</code></td>
+				<td>Hides inline popup content.</td>
+				<td>Add <code>mfp-hide</code> to the <code>class</code> attribute of the popup's <code>section</code> element.</td>
+				<td>n/a</td>
+			</tr>
+			<tr>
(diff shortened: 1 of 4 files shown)


GormFrank:
@xiayitan I will test. At first glance I am surprised that there is no CSS on JS change, but I will see the result while testing.

Also, is this PR related to an issue opened on GCWeb? If so, an you tag it?


Thank you!

xiayitan:
@GormFrank @nfrenette Yes, please test it.

It's just another option for the lightbox, very similar as the "Inline content", we just removed the class "modal-dialog" for the popup section.

I don't think this PR is related to an issue on GCWeb.

Thank you!

duboisp:
@GormFrank what was your test result?

GormFrank:
This works as expected but I am questioning the choice of having a table as the working example, because as shown in the attached image, on mobile the table overflows to the right, which looks technically broken but is not specific to this case because all the other examples have the same problem. I would use the list again, or lorem ipsum paragraphs or even columns layout. @duboisp and @xiayitan what do you think?
Screen Shot 2020-11-25 at 1 58 13 PM

GormFrank:
I changed the example to a list with column counts.

duboisp (approved):

Issue #7891, opened 2017-02-10
Email validation question


crochefort:
Hi client just came to me saying that the validation is not working...

I did the test here : http://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-en.html#email1

start typing test
Error 1: Please enter a valid email address.
add the @ 
still Error 1: Please enter a valid email address.
add 1 letter in my case is a s ...

it's now validating...

From the source code on the unmin version... 
js/deps/jquery.validate.js

```js
// http://jqueryvalidation.org/email-method/
		email: function( value, element ) {

			// From https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
			// Retrieved 2014-01-14
			// If you have a problem with this implementation, report a bug against the above spec
			// Or use custom methods to implement your own email validation
			return this.optional( element ) || /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/.test( value );
		},
```
Went to the site whatwg.com and saw something different for the email validation... 
https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address

> A valid e-mail address is a string that matches the email production of the following ABNF, the character set for which is Unicode. This ABNF implements the extensions described in RFC 1123. [ABNF] [RFC5322] [RFC1034] [RFC1123]
> 
> email         = 1*( atext / "." ) "@" label *( "." label )

So why an email is validating when i think i'm seeing a . after the @ ...  ?

Do we need to change what we have and replace it with the line for email??

nschonni:
Note that the RegEx on the WHATWG is the same as used here

crochefort:
thanks @nschonni but why are we saying it's a valid email when it's not ... can someone fix the regex in order to have zz@zz.zz as a valid email

kodecount:
That is a request you have to make with: https://github.com/jquery-validation/jquery-validation

WET should never change third party scripts. The only other option is for you to overwrite the method with your own custom method.

kodecount:
If you want to know how to customize the plugin/override an existing method look at: https://github.com/wet-boew/wet-boew/issues/7690#issuecomment-253100290

crochefort:
Why bother ...
[...]

PR #8983, opened 2020-11-13
Form validation: Allow pattern attribute

xiayitan:
Removed the restriction code for the pattern attribute of the email input.

Related issue: https://github.com/wet-boew/wet-boew/issues/7891

### DIFF
--- a/src/plugins/formvalid/formvalid-en.hbs
+++ b/src/plugins/formvalid/formvalid-en.hbs
@@ -7,7 +7,7 @@
 	"tag": "formvalid",
 	"parentdir": "formvalid",
 	"altLangPrefix": "formvalid",
-	"dateModified": "2020-09-15"
+	"dateModified": "2020-12-23"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
@@ -25,7 +25,7 @@
 </section>
 
 <section>
-	<h2>Examples</h2>
+	<h2>Examples using regular or specialized validation</h2>
 
 	<div class="wb-frmvld">
 		<form action="#" method="get" id="validation-example">
@@ -60,7 +60,7 @@
 
 				<div class="form-group">
 					<label for="fname1" class="required"><span class="field-name">First name</span> <strong class="required">(required)</strong></label>
-					<input class="form-control" id="fname1" name="fname1" type="text" autocomplete="given-name" required="required" pattern=".{2,}" data-rule-minlength="2" />
+					<input class="form-control" id="fname1" name="fname1" type="text" autocomplete="given-name" required="required" data-rule-minlength="2" />
 				</div>
 				<details class="mrgn-bttm-md">
 					<summary>View code</summary>
@@ -69,13 +69,13 @@
 		...
 		&lt;div class="form-group"&gt;
 			&lt;label for="fname1" class="required"&gt;&lt;span class="field-name"&gt;First name&lt;/span&gt; &lt;strong class="required"&gt;(required)&lt;/strong&gt;&lt;/label&gt;
-			&lt;input class="form-control" id="fname1" name="fname1" type="text" autocomplete="given-name" required="required" pattern=".{2,}" data-rule-minlength="2" /&gt;
+			&lt;input class="form-control" id="fname1" name="fname1" type="text" autocomplete="given-name" required="required" data-rule-minlength="2" /&gt;
 		&lt;/div&gt;</code></pre>
 				</details>
 
 				<div class="form-group">
 					<label for="lname1" class="required"><span class="field-name">Last name</span> <strong class="required">(required)</strong></label>
-					<input class="form-control" id="lname1" name="lname1" type="text" autocomplete="family-name" required="required" pattern=".{2,}" data-rule-minlength="2" />
+					<input class="form-control" id="lname1" name="lname1" type="text" autocomplete="family-name" required="required" data-rule-minlength="2" />
 				</div>
 				<details class="mrgn-bttm-md">
 					<summary>View code</summary>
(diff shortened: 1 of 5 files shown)


EricDunsworth:
I'd have concerns with this kind of change:
1. No [French i18n translation](https://github.com/jquery-validation/jquery-validation/blob/master/src/localization/messages_fr.js) currently exists for jQuery validation's pattern error messages. So if they were to start appearing, the English error ("Invalid format.") error would appear as-is in French pages that use pattern attributes. They'd probably also lack ``lang`` attributes.
2. I found [the backstory](https://github.com/wet-boew/wet-boew/pull/1626/files#diff-3cce1734e8ae04998cf8a8a7659648c2883fd8798502ad8926a7f06beea99ed8L70) behind why it was decided to strip out ``pattern`` attributes in the first place. Apparently jQuery validation used to serve the really vague pattern error in place of more useful error messages back in 2013. Has anything changed since then?
3. @nschonni mentioned in https://github.com/wet-boew/wet-boew/issues/7891#issuecomment-278957027 that the jQuery validation plugin uses the same regex as the WHATWG's HTML spec to determine valid email addresses. IMO it wouldn't be wise to deviate from the spec for something like that. Especially since in noscript scenarios, a user's browser would fall back to the WHATWG's regex.

duboisp:
@xiayitan I am ok with that change, but I think we would benefit to also include an example on how we can override the default message from jQuery. I think we can do it through the attribute `data-msg-pattern` or `data-msg`. With a such example, we will be able to know if we can override it and contextualize the error message.

Do you want to give a try?

GormFrank (dismissed):
Works as intended, however as discussed in old thread regarding the `pattern` subject, the error messages get overwritten by a default one when the `pattern` attribute is used with the jQuery validation plugin. 

But as @duboisp brought up, there is a `data-msg` attribute (and I confirm it is `data-msg`) that allows the implementer to put a custom message to over-overwrite the default one. 

That said, we could move forward with the `pattern` unlock but we would need an example of using the `data-msg` attribute to set a proper error message when the `pattern` attribute is in use. 

For instance, on the [Formvalid plugin page](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/formvalid/formvalid-en.hbs), the **Letters and punctuation only** field could look something like that:

```
<input class="form-control" id="text1" name="text1" type="text" pattern="[A-Za-z-.,()'&#34;\s]" data-rule-letterswithbasicpunc="true" data-msg="Only letters and punctuation are accepted" />
```

@xiayitan Can you add an example of some sort on the page? Thanks!

xiayitan:
@duboisp @GormFrank Thank you for the suggestion! Yes, I will add an example on the page.

EricDunsworth:
Wouldn't it be better to properly address the issues outlined in https://github.com/wet-boew/wet-boew/pull/8983#issuecomment-730535663 before merging in this kind of PR? Like sending upstream changes to add a French i18n translation for pattern error messages and making contextual error messages take precedence over the generic pattern error?

I'm all for adding documentation about ``data-msg``/``data-msg-*`` into the form validation plugin's documentation. But I don't think it'd be ideal to make the documentation encourage implementers to use workarounds for what should be upstream fixes.

For example, let's say the documentation encouraged implementers to do something like this:
```html
<div class="form-group">
	<label for="email2"><span class="field-name">Canada.ca email address</span> (yourname@canada.ca)</label>
	<input class="form-control" id="email2" name="email2" type="email" autocomplete="work email" pattern="[a-z\.]+@canada.ca" data-msg-email="Please enter a valid email address." />
</div>
```

Now let's say [jQuery validation's ``email`` error message](https://github.com/jquery-validation/jquery-validation/blob/79bed393e6f4bf6876d0e917baed7ef6447efe6a/src/core.js#L356) gets changed somewhere down the line and a future version of WET adopts it.

Anyone that hardcoded the old error message via ``data-msg-email`` would be left in the dust. They'd need to manually update their code to stay up-to-date (if they even realize WET's message has changed).

That situation could be entirely avoided if jQuery Validation's contextual error messages took precedence over pattern ones. That way, no one would have to go down the path of hardcoding redundant messages.

nfrenette:
Hi Eric, I completely agree that workarounds are not a best practice. However, issue #7891 has been opened Feb 10, 2017 with no updates since.

Can we add a warning in the documentation stating that the onus is on the implementer to keep their code up to date and inline with future jQuery Validation development?

Thank you

@duboisp, @xiayitan, @EricDunsworth

GormFrank:
Added the documentation for `pattern` and `data-msg`, plus a working example and as a bonus I fixed the Reset button since it was broken big time.

RobJohnston:
FYI:  I know this doesn't go far to resolve this issue, but the French translation of "Invalid format" for the pattern attribute is now in the latest release of jQuery Validation:  https://github.com/jquery-validation/jquery-validation/releases/tag/1.19.3

More importantly, that release fixes a ReDoS vulerability.

src/plugins/formvalid/formvalid-en.hbs
```diff
@@ -425,7 +425,31 @@
...
 
-			<input type="submit" value="Submit" class="btn btn-primary" /> <input type="reset" value="Reset page to defaults" class="btn btn-link btn-sm show p-0 pt-4" />
+			<input type="submit" value="Submit" class="btn btn-primary" /> <input type="reset" value="Reset page to defaults" class="btn btn-link btn-sm show p-0 mrgn-tp-md" />
+		</form>
+	</div>
+</section>
+<hr>
+<section>
+	<h2>Example using custom pattern validation</h2>
+	<div class="wb-frmvld">
```
duboisp, line 435: Can you add a note like:

If the pattern attribute is used within the data-rule, the pattern validation will considerated before the data-rule. So at the end you get only it's generic error validation message.
It's recommended to use "data-msg" to customize the error message displayed by the pattern when the field is in error state.

src/plugins/formvalid/formvalid.js
```diff
@@ -106,12 +105,6 @@ var componentName = "wb-frmvld",
 						labels[ i ].innerHTML += " ";
 					}
 
-					// Remove the pattern attribute until it is safe to use with jQuery Validation
-					len = $pattern.length;
-					for ( i = 0; i !== len; i += 1 ) {
-						$pattern.eq( i ).removeAttr( "pattern" );
-					}
-
 					// Change form attributes and values that interfere with validation in IE7/8
```
duboisp, line 108: Can you remove the code block only applicable for IE8 and lower

duboisp (changes requested):

src/plugins/formvalid/formvalid.js
EricDunsworth: FWIW I have an upcoming PR whose goal will be to kill off old IE support. It'll remove that block.

duboisp (approved):

duboisp:
Considering the work for this feature is completed and we are doing a release before the next technical review, it was decided to include it.

PR #8990, opened 2020-11-30
chore: Bump dev deps and linting setup

nschonni:

PR #8995, opened 2020-12-03
chore: Bump node-sass for node v14 support

nschonni:
This is the last version of node-sass to support the old Node 8 used by the Travis setup. This will support up-to Node 14, but Node 15 and above will require node-sass v5. v5 no longer supports EOL versions of Node like 8


GormFrank:
@nschonni Can you please rebase since we merged https://github.com/wet-boew/wet-boew/pull/8990? Thank you :)

nschonni:
Not needed since node-sass got bumped to v5

PR #9006, opened 2020-12-15
Editorial - Change GCWeb v8.2.0 to v9.0.0

duboisp:


duboisp:
This change is required because of departmental priority to deploying this minor release on Canada.ca by the end of this week.

I tested the build locally and it was successfully building with no errors.

PR #9018, opened 2021-01-12
WET-140 Spacing Issue Fixed

namjohn920:
* According to WET-140 Ticket, the spacing for <p> element in Alert component has margin inherited from Bootstrap. So I overrode it with a value of 11.5px which is used for our <p> elements everywhere.
* For panel component, I haven't found an example that is exhibiting the same pattern.

### DIFF
--- a/src/base/bootstrap-overrides/_base.scss
+++ b/src/base/bootstrap-overrides/_base.scss
@@ -265,6 +265,10 @@ abbr[title] {
 			}
 		}
 
+		p {
+			margin-bottom: 11.5px;
+		}
+
 		strong,
 		em,
 		span {


src/base/bootstrap-overrides/_base.scss
```diff
@@ -265,6 +265,10 @@ abbr[title] {
 			}
 		}
 
+		p {
```
duboisp, line 268: The bottom spacing should be applied only to the last paragraph in the alert.

duboisp (changes requested):
The bottom spacing of the last paragraph should be the same as the top spacing of the first paragraph.

Please check if the issue is limited to gcweb with the enlarged font size made during the 2019 visual update (since GCWeb v5.0).

duboisp:
@namjohn920 this change should be applied to GCWeb, not WET-BOEW.

This is problematic CSS is:
```
.alert > :last-child {
    padding-bottom: 25px;
}
```

GormFrank (changes requested):
As @duboisp mentioned, this PR should be open in the GCWeb project and not WET-BOEW.

Here is the faulty line: https://github.com/wet-boew/GCWeb/blob/d0daa950aabdf2299668ab9808124b2f13bd591d/src/components/_alert.scss#L56

namjohn920:
@duboisp @GormFrank Will Do!

duboisp:
Cancelled, we are not able to properly reproduce the issue

PR #9045, opened 2021-02-19
Details/Summary: Improve support for heading classes

EricDunsworth:
Fixes layout issues when using heading classes (e.g. "``h2``", etc...) on ``summary`` elements - which is needed to make disclosure widgets match the font size of child headings.

Specific changes:
* Use pixels for left/right margins (based on Bootstrap's grid gutter width... wasn't able to use ``$grid-gutter-width`` directly)
* Remove top/bottom margins on summaries that use heading classes (isn't needed for real heading elements)
* Rename the ``$details-identation`` variable to ``$details-indentation`` (typo fix)
* Use the default link colour on summaries that are styled as headings (except for collapsible alerts)
* Use the ``$link-color`` and ``$link-hover-color`` variables for summaries instead of duplicating their HEX values
* Update the third example in the details/summary polyfill's demo page to demonstrate a summary styled as a heading

### DIFF
--- a/src/base/details/_base.scss
+++ b/src/base/details/_base.scss
@@ -26,13 +26,22 @@ summary {
 }
 
 details {
-	padding-left: $details-identation;
-	padding-right: $details-identation;
+	padding-left: $details-indentation;
+	padding-right: $details-indentation;
 
 	> {
 		summary {
-			margin-left: -$details-identation;
-			margin-right: -$details-identation;
+			margin-left: -$details-indentation;
+			margin-right: -$details-indentation;
+
+			&.h2,
+			&.h3,
+			&.h4,
+			&.h5,
+			&.h6 {
+				margin-bottom: 0;
+				margin-top: 0;
+			}
 		}
 	}
 
--- a/src/base/details/_base.scss
+++ b/src/base/details/_base.scss
@@ -33,15 +33,6 @@ details {
 		summary {
 			margin-left: -$details-indentation;
 			margin-right: -$details-indentation;
-
-			&.h2,
-			&.h3,
-			&.h4,
-			&.h5,
-			&.h6 {
-				margin-bottom: 0;
-				margin-top: 0;
-			}
 		}
 	}
 
--- a/src/polyfills/details/_base.scss
+++ b/src/polyfills/details/_base.scss
@@ -19,10 +19,27 @@ summary {
 	}
 }
 
+%summary-goodies {
+	color: $link-color;
+	text-decoration: inherit;
+}
+
+%summary-heading-goodies {
+	margin-bottom: 0;
+	margin-top: 0;
+}
+
+/*
+TODOS:
+-Give the placeholder selectors better names
+-Are any of the extends below unnecessary? Not sure if I might be overdoing it and duplicating stuff at lower levels than what's actually needed
+*/
+
 details {
 	margin-bottom: .25em;
 
 	summary {
+		@extend %summary-goodies;
 		border: 1px solid #ddd;
 		border-radius: 4px;
 		padding: 5px 15px;
@@ -33,7 +50,22 @@ details {
 		&.h4,
 		&.h5,
 		&.h6 {
-			color: $link-color;
+			@extend %summary-goodies;
+			@extend %summary-heading-goodies;
+		}
+
+		&.summary-title {
+			@extend %summary-goodies;
+			@extend h3;
+			@extend %summary-heading-goodies;
+
+			h2,
+			h4,
+			h5,
+			h6 {
+				@extend h3;
+				@extend %summary-goodies;
+			}
 		}
 
 		&:focus,
--- a/src/polyfills/details/_base.scss
+++ b/src/polyfills/details/_base.scss
@@ -25,13 +25,21 @@ details {
 	summary {
 		border: 1px solid #ddd;
 		border-radius: 4px;
-		color: #295376;
 		padding: 5px 15px;
+
+		&,
+		&.h2,
+		&.h3,
+		&.h4,
+		&.h5,
+		&.h6 {
+			color: $link-color;
+		}
 
 		&:focus,
 		&:hover {
 			background-color: transparent;
-			color: #0535d2;
+			color: $link-hover-color;
 			text-decoration: underline;
 		}
 
--- a/src/polyfills/details/details-en.hbs
+++ b/src/polyfills/details/details-en.hbs
@@ -7,7 +7,7 @@
 	"tag": "details",
 	"parentdir": "details",
 	"altLangPrefix": "details",
-	"dateModified": "2021-02-25"
+	"dateModified": "2021-04-25"
 }
 ---
 <p>The HTML5 <code>details</code> and <code>summary</code> elements are used to provide expandable/collapsible content. This component adds support for these elements in browsers that do not already have native support.</p>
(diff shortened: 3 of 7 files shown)


EricDunsworth:
Switched this back into a draft state since IMO it isn't where it needs to be yet:

* Headings aren't currently adopting the summary's colour scheme
  * Some other properties might be in the same boat
* Nesting might mess up heading line heights (need to double-check)
* Haven't added working examples to the details/summary demo page yet
* Need to finish prepping up a sister PR for GCWeb

src/base/details/_base.scss
```diff
@@ -26,13 +26,23 @@ summary {
...
+	padding-right: $details-indentation;
 
 	> {
 		summary {
-			margin-left: -$details-identation;
-			margin-right: -$details-identation;
+			margin-left: -$details-indentation;
+			margin-right: -$details-indentation;
+
+			&.h1,
```
GormFrank, line 37: We don't need to support .h1, let's just support our use cases (h2 to h6) knowing that .h1 shouldn't be used at all.

src/polyfills/details/_base.scss
```diff
@@ -25,9 +25,18 @@ details {
...
 		padding: 5px 15px;
 
+		&,
+		&.h1,
+		&.h2,
+		&.h3,
+		&.h4,
+		&.h5,
+		&.h6 {
+			color: #295376;
```
GormFrank, line 37: Please use the variable meant for this color.

GormFrank (dismissed):
Don't forget to include a working example of simply just one of the .hX classes.

GormFrank:
This should be considered as being part of the upcoming WET-BOEW release.

EricDunsworth:
@GormFrank I think I finally have it in a finalized state now :)!

I went a bit back and forth between draft and final at the end when I noticed the collapsible alerts plugin normally uses the default heading colour (black or dark grey) + the summary's focus/hover colour. So I tinkered around with things to get that combo working again.

The only other thing I noticed is that as a result of the aforementioned adjustment, the collapsible alert plugin is now generating some selectors that contains this kind of pattern (pseudo-class before class name):

```css
details.alert summary:focus.h2 {
	...
}
```

But from what I can tell, it's valid CSS and all the browsers I checked (FF/Edge Chromium/IE11) seemed to be ok with it. Its SASS code also looks cleaner. So I think I'll stick with it.

I still have a PR to send to GCWeb that ties into this, but I don't have any concerns about merging WET's one first and tackling GCWeb another day. Although having said that, I might be able to send my GCWeb PR later today.

EricDunsworth:
@GormFrank FYI it looks like GCWeb re-uses the ``$details-identation`` variable for the report a problem feature's wells... So my incoming GCWeb PR will probably need to be merged alongside this one. Otherwise GCWeb's build will run into an undefined variable error.

duboisp (changes requested):
Is the CSS class "h2 ... h6" are only use to enlarge the size of the arrow? May be we should consider to normalize the summary size/style when it contains an heading. Like how panels and dialog do. May be for the use case when a heading is contained in a summary element, we could add a new and unique CSS class to enlarge/re-style only the arrow instead of changing the font-size.

We can discuss that during the next WET-BOEW code sprint.

Todo:

* [ ] Rework the CSS to remove the direct dependence between applying the CSS class heading on the `<summary>` element that must be the same of the header element contained in the `<summary>` element.
* [ ] Separate the working example, a section/series of details/summary with no header and a section/series of details/summary with with header.


**IMPORTANT NOTE for GCWeb maintainer:** Once merged, we will need to update the following 2 line in GCWeb when updating the wet-boew dependance: https://github.com/wet-boew/GCWeb/blob/master/src/_defaults.scss#L542-L543

\cc @GormFrank

src/polyfills/details/details-en.hbs
```diff
@@ -80,7 +80,7 @@
 		</li>
 		<li>
 			<details>
-				<summary>Example 3</summary>
+				<summary class="h3"><h3>Example 3</h3></summary>
```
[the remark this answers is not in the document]
EricDunsworth: @duboisp To properly achieve something like this:
```html
<summary><h3>Example 3</h3></summary>
```

You'd need to resort to the [:has() pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:has)... which is dead in the water in terms of browser support.

Another way I can think of would be to use a ``::before`` pseudo-element to add a fake disclosure widget to child heading elements and hope the fake one hides the real one. But IMO that probably wouldn't work in practice. If the heading's line height is substantially taller than the summary's line height, the heading's fake disclosure widget probably wouldn't be able to hide the summary's native one.

If you're thinking about disabling the ``<summary>`` element's native disclosure widget and using ``::before`` to insert a fake one into its first child element (regardless of if it's a heading or normal text)... that wouldn't work out either since  summaries don't typically contain child elements.

Plus according to the HTML spec, the [``<summary>`` element's content model](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-summary-element:concept-element-content-model) allows it to contain intermixed phrasing/heading content. So any CSS solution that doesn't resort to ``:has()`` would be opinionated and go against the spec's content model.

**TL;DR**
I don't think there's a better way of doing this than using a ``h#`` class and an equivalent ``<h#>`` element. IMO the other methods would have big downsides. I also don't think it's worthwhile to add extra CSS selectors to enforce using  equivalent classes/heading elements since that'd result in a ton of extra selectors (25 by my count).

EricDunsworth:
@duboisp 

> Is the CSS class "h2 ... h6" are only use to enlarge the size of the arrow?

Yes.

> May be we should consider to normalize the summary size/style when it contains an heading. Like how panels and dialog do.

FWIW [Bootstrap 3.4's panels](https://getbootstrap.com/docs/3.4/components/#panels) don't force a specific heading font size. It's up to implementers to decide if they want to place ``panel-title`` classes on their heading elements. In other words, Bootstrap gives them the flexibility to decide what's best for their situation instead of forcing a one-size-fits-all solution.

I wouldn't be against adding an optional class that works similarly to ``panel-title``. As long as implementers still have the freedom to use ``h#`` classes if they want to base themselves on WET's standard heading sizes.

> May be for the use case when a heading is contained in a summary element, we could add a new and unique CSS class to enlarge/re-style only the arrow instead of changing the font-size.

I suppose that could be done, but going down that path would add extra selectors for something that can already be achieved with WET's pre-existing ``h#`` classes. GCWeb is in a different boat though due to its its heading classes using em units instead of pixels.

duboisp:
Let go through this PR at the next code sprint.

GormFrank:
I love where you are going with this, which is similar to the panels approach.

duboisp:
@EricDunsworth as discussed during today code sprint, let's try the CSS selector `:where`

duboisp:
Is a selector like: 
```
summary:where( h1 ) {
   font-size: 00px;
}

```

EricDunsworth:
@duboisp Just tried it to no avail. Your code sample is exactly how ``:has()`` would work if any browsers actually had native support for it.

So, we've got 3 options:
1. Keep this PR's status quo (i.e. leave it up to implementers to manually add an ``h#`` class to ``summary`` elements that contain headings)
2. Add a WET plugin whose sole purpose is to auto-add ``h#`` classes to ``summary`` elements that contain headings
3. Add a ``:has()`` polyfill to WET (if we want to go down that road... could be messy if it eventually leads to an over-reliance on the polyfill for other WET/GCWeb CSS selectors)

Thoughts?

PR #8955, opened 2020-10-09
Bootstrap overrides: Darken panel borders

EricDunsworth:
This increases the contrast between panel headings and borders to >= 3:1. The resulting effect looks very stark.

The upside of this change is that it makes the shapes of panels more perceptible to users with poor vision. It may also benefit WCAG 2.1 success criterion 1.4.11 (Non-text Contrast) - especially when panels are used with stretched links.

PR #9048, opened 2021-03-01
Multimedia player: Disable top border radii on details/summary

EricDunsworth:
When WET's media player uses an embedded transcript via expandable/collapsible content, it's intended to be situated directly below the media player's controls.

Previously, the expandable/collapsible content looked slightly-disconnected from the player due to its rounded top border. This change addresses it by disabling the top border radii.

**Screenshots...**

**Before:**
multimedia-transcript-border-before

**After:**
multimedia-transcript-border-after

### DIFF
--- a/src/plugins/multimedia/_base.scss
+++ b/src/plugins/multimedia/_base.scss
@@ -182,6 +182,11 @@
 		text-align: right;
 	}
 
+	summary,
+	details[open] {
+		border-top-left-radius: 0;
+		border-top-right-radius: 0;
+	}
 }
 
 .wb-mm-cc {


EricDunsworth:
Btw the benefits of this change might look almost imperceptible at first glance, but it's a bit more obvious in GCWeb (due to its bigger font size making summaries taller) and will become much more obvious if details/summary borders get darkened in the future to be more on par with what was done with panels in #8955.

src/plugins/multimedia/_base.scss
```diff
@@ -74,6 +74,11 @@
 	vertical-align: middle;
 }
 
+%multmedia-details-summary-border-top-radius-0 {
+	border-top-left-radius: 0;
+	border-top-right-radius: 0;
+}
+
```
duboisp, line 81: This became a mixin very very specify, only for 1 use case and don't seems to be re-usable in other context.

Please just repeat those 2 line in your CSS code at line 191 and line 196.

duboisp (changes requested):
Other than removing the mixin,

 it seems good to go after testing.

src/plugins/multimedia/_base.scss
EricDunsworth: The placeholder selector is being extended twice. Are you sure you want me to eliminate it?

src/plugins/multimedia/_base.scss
```diff
@@ -182,6 +187,15 @@
 		text-align: right;
 	}
 
+	summary {
```
duboisp, line 190: Suggested changes:

```
summary, details[open] { }
```

duboisp (changes requested):

src/plugins/multimedia/_base.scss
EricDunsworth: That's not in line with how WET has traditionally dealt with this kind of situation. Selector rules are usually placed on their own lines and separated as much as possible. That's why I had resorted to ``&[open]`` and a placeholder selector instead of ``details[open]``).

In any case, I'll go with the way you're suggesting to simplify the code and since I don't feel super strongly about it.

GormFrank:
Approved as long as @duboisp approves.

duboisp (approved):

PR #9049, opened 2021-03-02
Build - Fix the minified files to keeps comments and stay compressed

duboisp:

Keep the comment on the wet-boew minified files and stay compress.


GormFrank (approved):
**Works great!** Minified file size is pretty similar to what we had before, even a little bit lighter ~140 kb compared to ~150 kb.

duboisp:
Great, let merge this and start the releasing process as discussed last week.

Issue #9054, opened 2021-03-08
[OTHER] Datatables serverSide: true, filtering on one column, with select dropdown forces regex in post


mercury64:
I am not sure if this is a bug, or if I have misconfigured something.
I have a Datatable that has server side processing, global search works great.
Attempting to add single column filtering using a select drop down list.

The problem is, in the POST request, the following is submitted:
`columns[1][search][value]: (searchValue)`
What I need is:
`columns[1][search][value]: searchValue`

After a bit of digging, I found the following code in wet-boew.js:
`// Verifies if regex was preset, if not preset use 'contains value' as default
if ( !$regex ) {
	$value = $value.replace( /\s/g, "\\s*" );
	$regex = "(" + $value + ")";
}
$datatable.column( $column ).search( $regex, true ).draw();`

that always wraps $value in parenthesis. 
I can fix my issue by removing the parenthesis, forcing the $regex = $value;

So, am I doing something wrong, or is this a bug?

mercury64:
Here are the original commits:
DataTable filter with SELECT MULTIPLE field #8546
DataTable Filtering With Checkboxes #8552

thekodester:
That code is only called when using the filter plugin.

thekodester:
We also have these filter plugin options:
https://wet-boew.github.io/themes-dist/GCWeb/demos/doaction/tblfilter-en.html
https://wet-boew.github.io/v4.0-ci/demos/tables/tables-fltrs-en.html

The Do Action is for GCWeb only.

mercury64:
My end goal is to have a drop down select, that filters a single column in the table when clicking filter. Then show all after clicking  "reset to defaults". The filter plugin options, on the Table Filters page: https://wet-boew.github.io/v4.0-ci/demos/tables/tables-fltrs-en.html is exactly what I am doing.

To replicate the problem, use the example Tables Filters page above, select the Filter Options -> Data Type drop down and select any data type, then click Filter. 

Zero results show. I will assume this is a bug, since it doesn't work on the demo site either.

Clicking reset to defaults work.

thekodester:
Thanks for the clarification.

For sure a bug, must be due to a more recent release as older versions of WET-BOEW work fine. If I have the time I might try to debug.

###### I am just a volunteer on this GitHub project.

mercury64:
Ok, I will override the filter function in the server side processing.
Thank you.

duboisp:
[...]

PR #9058, opened 2021-03-10
Date Filter Does Not Check If Empty

ghost:
Fixes issue in title

N/A

https://github.com/wet-boew/wet-boew/issues/9054

N/A

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -293,7 +293,7 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 				$value = ( $fData ) ? $fData : "-0";
 				$regex = "(" + $value.replace( /&nbsp;|\s/g, "\\s" ).replace( /\$/g, "\\$" ) + ")";
 			}
-		} else if ( $elm.is( "input[type='date']" ) ) {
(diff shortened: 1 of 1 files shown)


src/plugins/tables/tables.js
```diff
@@ -293,7 +293,7 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 				$value = ( $fData ) ? $fData : "-0";
 				$regex = "(" + $value.replace( /&nbsp;|\s/g, "\\s" ).replace( /\$/g, "\\$" ) + ")";
 			}
-		} else if ( $elm.is( "input[type='date']" ) ) {
+		} else if ( $elm.is( "input[type='date']" ) && $elm.val() ) {
```
duboisp, line 296: Should we add the same logic for "select" (line 243), "number" (line 255) and "Checkbox" (line 329) ? Do we will experience the same issue as it was reported?

thekodester: Currently the only effect this issue has is it makes the date required for any filter to work if the date filter is used

As for the select and checkbox they well only fail if HTML is coded incorrectly

I can add fail safes for that if you wish

GormFrank:
Approved as long as @duboisp is OK with just affecting the date filter.

duboisp (approved):

PR #9088, opened 2021-04-29
Bootstrap overrides: Make [hidden] !important

EricDunsworth:
Bootstrap 3.4.1's ``hidden`` attribute selector sets it to ``display: none;``... But that isn't powerful-enough for its purpose. Unrelated selectors can inadvertently unveil it. For example, placing an element with a ``hidden`` attribute at the beginning of a ``summary`` doesn't hide anything. Why? Because another more specific selector sets the summary's first child to ``display: inline;``.

Btw the ``"hide"`` and ``"hidden"`` classes are already set to ``display: none !important``. Latter versions of Bootstrap also updated the ``hidden`` attribute's selector to use ``!important``.

### DIFF
--- a/src/base/bootstrap-overrides/_base.scss
+++ b/src/base/bootstrap-overrides/_base.scss
@@ -918,3 +918,7 @@ mark {
 	color: $txthl-color;
 	font-weight: $txthl-font-weight;
 }
+
+[hidden] {
+	display: none !important;
+}


GormFrank (dismissed):

duboisp:
@EricDunsworth can you provide a design pattern where the attribute `hidden` don't work?

EricDunsworth:
@duboisp The OP and the commit message already describe a scenario in which the status quo would cause a [hidden] element to be unexpectedly unveiled due to the lack of ``!important`` in its CSS selector. In last week's WET technical review meeting I also provided a link to a non-public page that demonstrated the scenario in action (although I had already switched to using ``class="hidden"`` as a workaround).

Here's a specific example of that scenario...

**Code sample:**
```html
<details>
	<summary><span hidden>Hidden prefix for analytics tracking: </span>Summary that's meaningful to users but overly vague for analytics tracking</summary>
	<p>...</p>
</details>
```

**Expected behaviour:**
Hides ``<span hidden>Hidden prefix for analytics tracking: </span>``.

**Current behaviour in WET/GCWeb:**
Shows ``<span hidden>Hidden prefix for analytics tracking: </span>``.

Caused by another selector that has more specificity (``summary > :first-child {display: inline;}``).

**This PR's behaviour:**
Same as expected behaviour.

**PS:**
The [HTML spec's ``hidden`` attribute definition](https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute) contains this:
> User agents should not render elements that have the hidden attribute specified.

duboisp:
@EricDunsworth I think there is a data attribute you can use for that. ` data-gc-analytics`. If not, then I suggest you to submit a request to our analytic team to add that feature. 

You issue more likely to be a content issue related to unsupported context by the analytic software.

EricDunsworth:
I don't really know what more I can do to prove the usefulness of this PR :S. But I'll keep trying lol.

Regarding ``data-gc-analytics``, if that data attribute does in fact exist, it could potentially be a viable option for my use case. But then I'd need to repeat every summary's content twice: once in the summaries themselves and again in their data attributes. That would increase the page's code complexity and make it more difficult to maintain over the longterm (if the content of the summaries get changed throughout the page's lifec). So I still prefer the method I'm using since it reduces duplication and is easier to maintain.

@duboisp In any case, even if you don't personally agree with my example, it's still valid HTML markup and that clearly shows something is wrong with the status quo. Plus my example doesn't even need any classes. It just uses plain HTML elements.

Here's another potential use case I can think of (except this time it's theoretical):
 * WET has 321 selectors that contain ``display`` properties. 71 of them are ``display: none;``. That means there are 249 selectors out there that can potentially unhide elements that use the ``hidden`` attribute. The vast majority of them are probably tied to classes (as opposed to elements).
 * If any WET implementation (such as a web app, CMS, other JS framework, custom JS, etc) were to use the ``hidden`` attribute to toggle between hiding/unveiling an element, there's a chance the element that's being toggled could end up in a state where it has both a ``hidden`` attribute and a ``class`` attribute at the same time. If any of the ``class`` attribute's classes are covered by one of the 249 selectors I mentioned earlier, they'd unhide the hidden element. That would be undesirable behaviour.

All of the following factors support my position:
* The HTML spec indicates that the ``hidden`` attribute's purpose is to hide things. Anything, under any scenario.
* The [HTML spec's definition of the ``hidden`` attribute](https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute) contains this note that puts the burden on authors to be careful about how they write their style sheets to avoid unveiling hidden elements:
  > Because this attribute is typically implemented using CSS, it's also possible to override it using CSS. For instance, a rule that applies 'display: block' to all elements will cancel the effects of the hidden attribute. **Authors therefore have to take care when writing their style sheets to make sure that the attribute is still styled as expected.**
  
  So as far as the HTML spec is concerned, WET's CSS is at fault.
* Bootstrap's 3.4.x's ``hide``/``hidden`` classes already use ``!important`` in their CSS selectors. I don't see any reasoning as to why the ``hidden`` attribute (the semantic way of doing it) shouldn't be treated the same way.
* Bootstrap (which is where WET's ``[hidden]`` selector came from) realized the same thing at some point. [Bootstrap 4.x changed its ``[hidden]`` selector to use ``!important``](https://getbootstrap.com/docs/4.0/content/reboot/#html5-hidden-attribute). If WET were to update its Bootstrap dependency in the future, it'd inherit that ``!important`` selector.
* IIRC @GormFrank had previously approved this PR. I presume he did so because he also agreed with the premise of this change.

duboisp (approved):
I understand that in some circumstance the attribute "hidden" don't work as expected and as per the HTML spec. But the working example provided is not valid.

PR #9107, opened 2021-05-19
Footnote Demo Page Bug

### LINKED ISSUE #9085: Footnotes: Footnotes using symbols such as asterix * are broken/not linking
Tested here: https://wet-boew.github.io/v4.0-ci/demos/footnotes/footnotes-en.html

image

ghost:
Removed

### DIFF
--- a/src/plugins/footnotes/footnotes-en.hbs
+++ b/src/plugins/footnotes/footnotes-en.hbs
@@ -29,7 +29,7 @@
 <section>
 	<h2>Recommended usage</h2>
 	<ul>
-		<li>Implementing footnotes in Web pages<sup id="fn*-rf"><a class="fn-lnk" href="#fn*"><span class="wb-inv">Footnote </span>*</a></sup></li>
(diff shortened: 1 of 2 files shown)


EricDunsworth (changes requested):
Wouldn't it be better to fix the bug that's preventing the footnotes plugin from correctly handling links to/from footnotes that use wildcards in their IDs?

FWIW that example was originally created specifically for that purpose (to show off that the footnotes plugin supports non-numerical footnotes, including special characters in IDs). The example also used to work in the past. Looks like it stopped working at some point and no one noticed.

EricDunsworth:
Btw these would probably be decent starting points to troubleshoot the bug:

https://github.com/wet-boew/wet-boew/blob/98f8a130cf910ceef603962920c5a97f3e4bd6b8/src/plugins/footnotes/footnotes.js#L59

https://github.com/wet-boew/wet-boew/blob/98f8a130cf910ceef603962920c5a97f3e4bd6b8/src/plugins/footnotes/footnotes.js#L83

EricDunsworth:
@thekodester Just wondering if you were planning on sending a replacement PR to fix the underlying issue?

If not, I could try looking into it. Just wanted to be 100% sure in any case to avoid any duplication of work.

Thanks :).

thekodester:
No time unfortunately.

PR #9115, opened 2021-05-26
Core: Fix jqEscape() regression

### LINKED ISSUE #9085: Footnotes: Footnotes using symbols such as asterix * are broken/not linking

EricDunsworth:
#8990 removed some of that helper method's backslashes, which in turn made it unable to escape certain special characters.

This resolves it by reintroducing the aforementioned backslashes. Also adds a configuration comment to prevent ESLint's ``no-useless-escape`` rule from treating them as errors.

* Fixes #9085
* Related to #9107

CC @nschonni @thekodester

### DIFF
--- a/src/core/helpers.js
+++ b/src/core/helpers.js
@@ -172,7 +172,8 @@ wb.addSkipLink = function( text, attr, isBtn, isLast ) {
 // Escapes the characters in a string for use in a jQuery selector
 // Based on https://totaldev.com/content/escaping-characters-get-valid-jquery-id
 wb.jqEscape = function( selector ) {
-	return selector.replace( /([;&,.+*~':"\\!^/#$%@[]()=>\|])/g, "\\$1" );
+	// eslint-disable-next-line no-useless-escape
+	return selector.replace( /([;&,\.\+\*\~':"\\\!\^\/#$%@\[\]\(\)=>\|])/g, "\\$1" );
 };
 
 // RegEx used by formattedNumCompare


src/core/helpers.js
```diff
@@ -172,7 +172,8 @@ wb.addSkipLink = function( text, attr, isBtn, isLast ) {
 // Escapes the characters in a string for use in a jQuery selector
 // Based on https://totaldev.com/content/escaping-characters-get-valid-jquery-id
 wb.jqEscape = function( selector ) {
-	return selector.replace( /([;&,.+*~':"\\!^/#$%@[]()=>\|])/g, "\\$1" );
+	// eslint-disable-next-line no-useless-escape
+	return selector.replace( /([;&,\.\+\*\~':"\\\!\^\/#$%@\[\]\(\)=>\|])/g, "\\$1" );
```
EricDunsworth, line 176: FWIW jqEscape()'s functionality is built-into [modern-day JavaScript via ``CSS.escape()``](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape) and [jQuery 3.x via ``$.escapeSelector()``](https://api.jquery.com/jQuery.escapeSelector/).

So... I was thinking of putting ``CSS.escape()`` into use here, but unfortunately IE11 doesn't support it and I didn't feel like it'd be wise to intentionally break it.

duboisp (approved):
Approved based on successful testing. Like with the footnote plugin.

Aside, it might worth to explore or add an escape function similar to this one: https://github.com/jquery/jquery/blob/a684e6ba836f7c553968d7d026ed7941e1a612d8/src/selector/escapeSelector.js

duboisp (approved):
I tested locally with the footnote component, the link with the asterisk, and it is working as expected.

duboisp:
This is a PATCH

PR #9116, opened 2021-05-28
A11y name landmarks

### LINKED ISSUE #9072: Landmarks: accessible names
All landmarks should have accessible names to them. Just having a heading in them is not enough.

Multiple instances of the same type of landmark SHOULD be distinguishable by different programmatically determinable labels (`aria-label` or `aria-labelledby`).

shawnthompson:
This relates the accessible names to both the breadcrumbs and footer sections of the template. This is needed when you have more than one of the same landmark elements. Hidden heading aren't enough read by users who navigate by landmarks. 

Related #9072

### DIFF
--- a/site/includes/breadcrumbs.hbs
+++ b/site/includes/breadcrumbs.hbs
@@ -1,5 +1,5 @@
-<nav role="navigation" id="wb-bc" property="breadcrumb">
-	<h2>{{{i18n "you-are-here"}}}</h2>
+<nav role="navigation" id="wb-bc" property="breadcrumb" aria-labelledby="wb-bc-nm">
+	<h2 id="wb-bc-nm">{{{i18n "you-are-here"}}}</h2>
 	<div class="container">
 		<div class="row">
 			<ol class="breadcrumb">
--- a/site/includes/footer.hbs
+++ b/site/includes/footer.hbs
@@ -1,7 +1,7 @@
 <footer role="contentinfo" id="wb-info" {{>footerreplace}} class="visible-sm visible-md visible-lg wb-navcurr">
 	<div class="container">
-		<nav role="navigation" class="row">
-			<h2>{{{i18n "tmpl-about-site"}}}</h2>
+		<nav role="navigation" class="row" aria-labelledby="wb-info-a11y-nm">
+			<h2 id="wb-info-a11y-nm">{{{i18n "tmpl-about-site"}}}</h2>
 			{{>footercontent}}
 		</nav>
 	</div>


duboisp:
Ok, do we need to add that to all sectionning content?

When or what software this cause an issue? what test has been run?

shawnthompson:
`section` isn't a landmark role so there is no **need** for the accessible name for navigating by landmarks, if there is multiple of the same elements.

Having said it could be a good thing to add an accessible name to all the section elements. I've seen some add it to all `ul` elements.

duboisp:
@shawnthompson in what situation or with what AT software that is needed. When this is beneficial vs not using ARIA. The first rule of WAI-ARIA is to not use WAI-ARIA. ;-)

shawnthompson:
If you read the MDN documentation on [ARIA: Navigation Role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role) it states the following:

> If a page includes more than one navigation landmark, each should have a unique label. If two or more navigational landmark on a page have the same set of links, use the same label for each.

This was brought up by a developer from Deque, that a landmark that's used more than once, like we are doing in with the `nav` element should have a unique accessible name associated with it.  The heading isn't considered as an accessible name. 

The first rule of ARIA is not to use it **if** there is already something in HTML that does the same thing. In this case there isn't a way to give the `nav` elements their own accessible name without using `WAI-ARIA`.

duboisp:
Thanks for the information @shawnthompson but I still don't know how to test and what is the expected result with Aria vs without Aria. Adding a aria-label to the native HTML `nav` element here might illustrate an unsupported browser feature where applying an aria-label is simply a work-around. I would like the why documented explaining why those aria-label is needed on nav element and what kind of test I can do in the future remove that interim solution. Or is it because of a specific AT software which don't support the HTML navigation element?

When I look at the HTML5 spec, they are using the heading structure to provide more context to those navigation element. ref. https://html.spec.whatwg.org/multipage/sections.html#the-nav-element
When I look at WAI-ARIA specification, the accessible name with an aria-label attribute is optional for the navigation element. ref. https://w3c.github.io/aria/#namefromauthor

I suspect the issue is probably because either the browser don't support the document outline or the AT software don't support the document outline. That algorithm is defined here: https://html.spec.whatwg.org/multipage/sections.html#outlines

What ever the result, the technique to name navigational element should be applied consistently. I am unsure if this should be done automatically or manually. To get more clarity at that question, I need a clear test case with the expected result with and without aria, official reference of when it's going to be fixed in the HTML or in browser or in AT software.

shawnthompson:
So I found this article that explains why you need to add an accessible name when using multiple elements with the same landmark roles.

Does this demonstrate what you're looking for?

[Multiple Navigation Landmarks](https://www.aditus.io/patterns/multiple-navigation-landmarks/).

shawnthompson:
As @duboisp suggested during a phone conversation, would should be creating a new plugin that sets the **accessible names** elements using their first heading in the elements.

Will open an issue to document it.

Stay tuned

PR #9122, opened 2021-06-15
Overlay: Add bottom-right popup pattern

Ricokola:
Creation of a new bottom-right popup pattern to the existing overlay WET component.

This Pull request is adding four things basically:
- An overlay that displays at the bottom right or the screen (and fullscreen when on mobile)… mostly for a chat interface.
- A config that allows to remove the close button in the footer whil keeping the “X” button obviously.
- A config that prevents the overlay from closing when clicking outside of its window.
- A class that keeps the header and footer of the overlay fixed so that only its body can scroll.

A second commit has been created to fix the bug with adding custom close buttons to overlays/modals functionality where the close function was generating an error in the console and the default close button was still appearing along with the custom button.

This will support the common GC chat experience: wet-136

### DIFF
--- a/src/plugins/overlay/_base.scss
+++ b/src/plugins/overlay/_base.scss
@@ -67,6 +67,29 @@
 	bottom: 0;
 }
 
+.wb-popup-br {
+	border-radius: 6px;
+	bottom: 20px;
+	max-height: 500px;
+	min-height: 200px;
+	right: 20px;
+	width: 35%;
+}
+
+.wb-popup-flex {
+	flex-direction: column;
+	flex-wrap: nowrap;
+
+	.modal-body {
+		overflow: auto;
+	}
+
+	&.open {
+		display: inline-flex;
+		position: fixed;
+	}
+}
+
 .wb-popup-mid {
 	@extend %overlay-popup;
 	border-radius: 6px;
@@ -128,6 +151,7 @@
 
 .hidden-hd {
 	.modal-body {
+		overflow: auto;
 		padding-top: 50px;
 	}
 
--- a/src/plugins/overlay/overlay-en.hbs
+++ b/src/plugins/overlay/overlay-en.hbs
@@ -9,7 +9,7 @@
 	"altLangPrefix": "overlay",
 	"js": ["demo/overlay"],
 	"css": ["demo/overlay"],
-	"dateModified": "2014-07-22"
+	"dateModified": "2021-09-24"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
@@ -27,15 +27,14 @@
 			<details class="mrgn-bttm-lg">
 				<summary>View code</summary>
 				<pre><code>&lt;p&gt;&lt;a href=&quot;#left-panel&quot; aria-controls=&quot;left-panel&quot; class=&quot;overlay-lnk&quot; role=&quot;button&quot;&gt;Left-panel&lt;/a&gt;&lt;/p&gt;
-
-&lt;section id=&quot;left-panel&quot; class=&quot;wb-overlay modal-content overlay-def wb-panel-l&quot;&gt;
-	&lt;header class=&quot;modal-header&quot;&gt;
-		&lt;h2 class=&quot;modal-title&quot;&gt;Left panel&lt;/h2&gt;
-	&lt;/header&gt;
-	&lt;div class=&quot;modal-body&quot;&gt;
-		...
-	&lt;/div&gt;
-&lt;/section&gt;</code></pre>
+				&lt;section id=&quot;left-panel&quot; class=&quot;wb-overlay modal-content overlay-def wb-panel-l&quot;&gt;
+					&lt;header class=&quot;modal-header&quot;&gt;
+						&lt;h2 class=&quot;modal-title&quot;&gt;Left panel&lt;/h2&gt;
+					&lt;/header&gt;
+					&lt;div class=&quot;modal-body&quot;&gt;
+						...
+					&lt;/div&gt;
+				&lt;/section&gt;</code></pre>
 			</details>
 		</li>
 		<li>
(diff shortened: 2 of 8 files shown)


Ricokola:
- [ ] Create settings to disable footer

theme/skip-links/_base.scss
```diff
@@ -9,3 +9,9 @@
 		text-decoration: none;
 	}
 }
+
+button.wb-sl:focus {
+	border: none;
+	margin: 0;
+	width: auto;
+}
```
duboisp, line 17: I think this change is not related at this PR subject. It will probably be best to either move this in a separate commit or do a different pull request

src/plugins/overlay/_base.scss
```diff
@@ -21,7 +23,7 @@
 	}
 
 	&.open {
-		display: inline-block;
+		display: inline-flex;
```
duboisp, line 26: Todo: test, this change might impact the other overlay type

src/plugins/overlay/overlay.js
```diff
@@ -22,6 +22,7 @@ var componentName = "wb-overlay",
 	OverlayOpenFlag = "wb-overlay-dlg",
 	initialized = false,
 	sourceLinks = {},
+	defaults = { noFooter: false },
```
duboisp, line 25: What this is used for? And it is not documented.

duboisp (changes requested):
Working example are not there yet.

src/plugins/overlay/_base.scss
```diff
@@ -78,6 +80,16 @@
 	width: 90%;
 }
 
+.wb-popup-br {
+	@extend %overlay-popup;
+	border-radius: 6px;
+	bottom: 20px;
+	font-size: .9em;
```
GormFrank, line 87: Please remove this `font-size` property, it shouldn't belong to the overlay itself, but instead to whatever goes in the overlay if font size is needed to be different i.e. chat wizard in GCWeb.

GormFrank (changes requested):
I will have to test the noFooter feature in-depth.

GormFrank (changes requested):
@Ricokola Voici pour débogguer le full-width: http://universallabs.org/wet/labs/pr9122/demos/overlay/overlay-en.html

src/plugins/overlay/overlay-en.hbs
```diff
@@ -250,6 +261,63 @@ wb.doc.on( "click vclick", "#overlay-open-btn", function( event ) {
 &lt;/section&gt;</code></pre>
 		</details>
 	</section>
+
+	<section>
+		<h3>Fix header and footer to overlays</h3>
```
GormFrank, line 266:
```suggestion
		<h3>Fixed header and footer</h3>
```

src/plugins/overlay/overlay-en.hbs
```diff
@@ -250,6 +261,63 @@ wb.doc.on( "click vclick", "#overlay-open-btn", function( event ) {
...
+		&lt;/section&gt;</code></pre>
+		</details>
+	</section>
+
+	<section>
+		<h3>Remove default footer or disable the close event from overlays</h3>
+		<p>Add <code>data-wb-overlay</code> to customize footer and or the close event.</p>
+		<ul>
+			<li><code>{ &quot;noFooter&quot;: true }</code> this will remove the default footer from overlays so a customized footer can be added.</li>
+			<li>><code>{ &quot;noClose&quot;: true }</code> this will prevent overlays to close on mouse right-click event or when pressing the escape key.</li>
```
GormFrank, line 288: Small typo here:
```suggestion
			<li><code>{ &quot;noClose&quot;: true }</code> this will prevent overlays to close on mouse right-click event or when pressing the escape key.</li>
```

src/plugins/overlay/overlay-en.hbs
```diff
@@ -474,3 +564,69 @@ wb.doc.on( "click vclick", "#overlay-open-btn", function( event ) {
...
+					</li>
+				</ol>
+			</li>
+		</ol>
+	</div>
+</section>
+
+<section id="mid-screen" class="wb-overlay modal-content overlay-def wb-popup-mid" data-wb-overlay='{ "noClose": true }'>
+	<header class="modal-header">
+		<h2 class="modal-title">Middle screen overlay preventing close event</h2>
```
GormFrank, line 614: Changer pour ceci : 
```suggestion
		<h2 class="modal-title">Persistent middle screen overlay</h2>
		<p>Remains open after a click outside of the overlay</p>
```
Ajouter aussi cet exemple en français.

GormFrank (changes requested):
@Ricokola Can you make sure that we can put the focus back in the content of the overlay when in "Fixed header and footer" mode so that we can scroll through the content using the arrows keys on the keyboard? Right now, there is now way to see the rest of the content inside the overlay because the focus goes from "Close" to "X" and never on the actual content.

Aussi, faire en sorte que l'option `noClose` fonctionne tel que prévu et puis la renommer quelque chose comme `persistent`.

duboisp (changes requested):

src/plugins/overlay/overlay.js
```diff
@@ -307,7 +325,7 @@ $document.on( "click vclick touchstart focusin", "body", function( event ) {
 			if ( overlay && overlay.getAttribute( "aria-hidden" ) === "false" &&
 				eventTarget.id !== overlayId &&
 				overlay.className.indexOf( ignoreOutsideClass ) === -1 &&
-				!$.contains( overlay, eventTarget ) ) {
+				!$.contains( overlay, eventTarget )  && !$( "#" + overlayId ).data( "wb-overlay" ).persistent ) {
```
GormFrank, line 328:
```suggestion
				!$.contains( overlay, eventTarget )  && !wb.getData( $( overlay ), componentName ).persistent ) {
```

GormFrank (changes requested):

GormFrank (changes requested):
@Ricokola When testing, I noticed that you can't scroll down or up with the keyboard in the content of the overlay when it is in the bottom right.

duboisp (changes requested):
**Popup persistence**
Should not the bottom right popup be persistently open?

When I open it via a button and click away, the popup close automatically.

**Overflow content**
I am not able to scroll in the content by leveraging the keyboard only.  
Should we follow the Grid popup keyboard interaction? https://www.w3.org/TR/wai-aria-practices-1.1/#grid-popup-keyboard-interaction

**Accessing the overlay**
Considering this pattern is intended to be immersive into the user experience (opening on page load) but not intended to take over immediately the full user attention, like other overlay, we will need a soft way to get in and out of the bottom right overlay. Like how we did with the chat bubble button.

Issue #8425, opened 2018-07-05
Grunt: Very slow copy:js task on watch


juleskuehn:
The copy:js task is greatly slowing down the build each time I make a change to a js file. (On a Surface Book i7, 16GB RAM, NVME SSD) Can this be sped up by keeping a cache of these 802 directories and 1384 files, and only updating the ones that have changed?

```
Running "copy:js" (copy) task
Created 802 directories, copied 1384 files
...
Execution Time (2018-07-05 13:58:51 UTC)
i18n_csv:js            3.1s  ██ 3%
copy:js            1m 29.4s  ███████████████████████████████████████ 90%
concat:core            1.8s  █ 2%
concat:coreIE8         1.3s  █ 1%
concat:pluginsIE8      1.3s  █ 1%
concat:i18n            2.4s  ██ 2%
Total 1m 39.5s
```

LaurentGoderre:
This shocked me because it wasn't like this last I checked. This is caused by the fact that we are now bundling MathJax as a backup. Perhaps MathJax should be excluded from `copy:js` and only included on deploy.

PR #9143, opened 2021-07-23
MathML polyfill: Update MathJax to ^3.2.0

### LINKED ISSUE #8769: Mathjax slows down file system
Mathjax slows down file system because of all the files/folders

EricDunsworth:
Drastically reduces MathJax's burden on the build:
* Brings a huge speed boost to Grunt's "copy:js" task (from ~13 to ~2 seconds)
* Reduces MathJax's compiled size down from ~25 MB (1,347 files/809 folders) to ~1 MB (24 files/5 folders)
* Almost halves the production dist folder's size (from ~55 MB to ~31 MB)

Related changes:
* Disable MathJax's context menu
* Add an ES6 polyfill for IE11 support
* Add extra line comments to give more context about how the polyfills are loaded in
* Change the MathJax fallback test to look for a configuration option we're not customizing
* Uses the CDN's latest version of MathJax 3
  * Falls back on WET's local version of MathJax 3.2.0 if the CDN is unreachable
  * Tries loading a local version of MathJax 2 if all else fails
* Add a temporary CSS workaround for mathjax/MathJax#2765 (thanks @dpvc!)
* Remove the build system's string-replace:inline task (was only being used to insert a specific version number in the CDN's URL)
* Update the documentation's known issues section:
  * Change it into an unordered list
  * Replace mentions of IE7-9 support with Chromium and IE11
  * Remove mention of SVG support
  * Add an item about IE11 only being supported in internet-connected networks (due to its ES6 polyfill)
  * Remove items specific to IE7-8

Notes:
* Disabling MathJax's context menu has the following impacts:
  * Works around MathJax 3.x's missing i18n support
  * Makes MathJax more closely mimic native MathML implementations (i.e. scraps bells and whistles that go beyond being a polyfill)
  * Removes some accessibility/usability enhancements that were present in WET's MathJax 2.x implementation (e.g. focusing on expressions, double-click to zoom). But it's not a huge loss since those enhancements don't exist in native MathML implementations. Visually-hidden MathML markup is still provided for ATs that support it.
  * Prevents users from accessing options that are excluded from WET's local copy of MathJax (e.g. "Math Settings - Math Renderer - SVG")
* IE considerations:
  * Breaks compatibility with older versions of IE (older than IE11)
  * IE11 will only work if internet access is available
  * Letter capitalization issues exist due to mathjax/MathJax#2440
* Somewhat alleviates #8425 (could still be improved by removing MathJax from the "copy:js" task)
* Fixes #8769

CC @juleskuehn @zzucco @AtomicNumber28 @nathanv

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -126,7 +126,6 @@ module.exports = (grunt) ->
 			"copy:js"
 			"concat:core"
 			"concat:coreIE8"
-			"string-replace"
 			"concat:pluginsIE8"
 			"concat:i18n"
 		]
@@ -313,7 +312,6 @@ module.exports = (grunt) ->
 		themeDist: "dist/theme-wet-boew"
 		jqueryVersion: "<%= pkg.dependencies.jquery %>"
 		jqueryOldIEVersion: "1.12.4"
-		MathJaxVersion: "<%= pkg.dependencies.mathjax %>"
 		banner: "/*!\n * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)\n * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html\n" +
 				" * v<%= pkg.version %> - " + "<%= grunt.template.today('yyyy-mm-dd') %>\n *\n */"
 		modernizrBanner: "/*! Modernizr (Custom Build) | MIT & BSD */\n"
@@ -1141,15 +1139,10 @@ module.exports = (grunt) ->
 					expand: true
 					flatten: true
 				,
-					cwd: "node_modules/mathjax"
+					cwd: "node_modules/mathjax/es5"
 					src: [
-						"MathJax.js"
-						"config/**"
-						"extensions/**"
-						"jax/**"
-						"localization/**"
-						"fonts/HTML-CSS/TeX/woff/**"
-						"fonts/HTML-CSS/TeX/otf/**"
+						"mml-chtml.js"
+						"output/chtml/fonts/woff-v2/**"
 					]
 					dest: "<%= coreDist %>/js/MathJax/"
 					expand: true
@@ -1390,16 +1383,6 @@ module.exports = (grunt) ->
 				dest: "<%= coreDist %>/js/i18n/"
 			assemble:
 				dest: 'site/data/i18n'
-
-		"string-replace":
-			inline:
-				files:
-					"dist/wet-boew/js/": "dist/wet-boew/js/*.js"
-				options:
-					replacements: [
-						pattern: "WET_BOEW_VERSION_MATHJAX"
-						replacement: "<%= MathJaxVersion %>"
-					]
 
 		"gh-pages":
 			options:
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "jquery": "2.2.4",
     "jquery-validation": "1.19.3",
     "magnific-popup": "git+https://github.com/wet-boew/Magnific-Popup.git#1.0.0+keyboard_trap_fix",
-    "mathjax": "2.7.4",
+    "mathjax": "^3.2.0",
     "proj4": "2.3.3",
     "unorm": "^1.6.0"
   },
--- a/src/core/wb.js
+++ b/src/core/wb.js
@@ -564,6 +564,7 @@ Modernizr.load( [
 
 			// Defer loading the polyfill till an element is detected due to the size
 			if ( !Modernizr.mathml ) {
+				let isTrident = new Boolean( window.navigator.msSaveOrOpenBlob );
 
 				// Bind the init event of the plugin
 				$document.one( "timerpoke.wb wb-init." + componentName, selector, function() {
@@ -571,18 +572,67 @@ Modernizr.load( [
 					// Start initialization
 					wb.init( document, componentName, selector );
 
+					// Disable MathJax's context menu to more closely mimic native MathML implementations
+					window.MathJax = {
+						options: {
+							enableMenu: false
+						}
+					};
+
+					// Extra tasks for IE11
+					if ( isTrident ) {
+
+						// Load an ES6 polyfill
+						Modernizr.load( "timeout=500!https://polyfill.io/v3/polyfill.min.js?features=es6" );
+
+						// Specify the CDN's font URL
+						// Note: IE11 is unable to resolve this on its own
+						window.MathJax.chtml = {
+							fontURL: "https://cdn.jsdelivr.net/npm/mathjax@3/es5/output/chtml/fonts/woff-v2"
+						};
+					}
+
 					// Load the MathML dependency. Since the polyfill is only loaded
 					// when !Modernizr.mathml, we can skip the test here.
 					Modernizr.load( [ {
-						load: "timeout=500!https://cdn.jsdelivr.net/npm/mathjax@WET_BOEW_VERSION_MATHJAX/MathJax.js?config=Accessible",
-						complete: function() {
-							Modernizr.load( [ {
-								test: window.MathJax === undefined,
-								yep: "mthjx!MathJax.js?config=Accessible"
-							} ] );
 
-							// Identify that initialization has completed
-							wb.ready( $document, componentName );
+						// Load latest version of MathJax 3 from a CDN
+						// Also load a CSS workaround for a MathJax 3.2.0 bug (refer to CSS file for details)
+						load: [
+							"timeout=500!https://cdn.jsdelivr.net/npm/mathjax@3/es5/mml-chtml.js",
+							"plyfll!mathml.min.css"
+						],
+						complete: function() {
+
+							// Wait a moment to reduce the risk of a race condition
+							setTimeout( function() {
+
+								// Specify a font URL for a local copy of MathJax 3 for IE11
+								// Note: Useful if IE11 has internet access but fails to reach the CDN
+								if ( isTrident && !window.MathJax.startup ) {
+									window.MathJax.chtml.fontURL = paths.js + "/MathJax/output/chtml/fonts/woff-v2";
+								}
+
+								// Fall back on a local copy of MathJax 3 if the CDN is unreachable
+								// Note: Won't work with IE11 in isolated networks (ES6 polyfill has no local fallback)
+								Modernizr.load( [ {
+									test: window.MathJax.startup,
+									nope: "mthjx!mml-chtml.js",
+									complete: function() {
+
+										// Try loading a local copy of MathJax 2 as a last dith effort
+										Modernizr.load( [ {
+											test: window.MathJax.startup,
+											nope: "mthjx!MathJax.js?config=Accessible",
+											complete: function() {
+
+												// Identify that initialization has completed
+												wb.ready( $document, componentName );
+											}
+										} ] );
+									}
+								} ] );
+							}, 100 );
 						}
 					} ] );
 				} );
--- a/package-lock.json
+++ b/package-lock.json
@@ -8562,9 +8562,9 @@
       "dev": true
     },
     "mathjax": {
-      "version": "2.7.4",
-      "resolved": "https://registry.npmjs.org/mathjax/-/mathjax-2.7.4.tgz",
-      "integrity": "sha512-B6SLtoETmV08kqTepRyz8eqMRGFzllPy2QASKFdFl0g0YB6SNgQVnWae11RaF3/y8ZKLTJNJ7hLbMIqmBEkLzw=="
+      "version": "3.2.0",
+      "resolved": "https://registry.npmjs.org/mathjax/-/mathjax-3.2.0.tgz",
+      "integrity": "sha512-PL+rdYRK4Wxif+SQ94zP/L0sv6/oW/1WdQiIx0Jvn9FZaU5W9E6nlIv8liYAXBNPL2Fw/i+o/mZ1212eSzn0Cw=="
     },
     "maxmin": {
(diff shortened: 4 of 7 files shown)


GormFrank (changes requested):
@EricDunsworth My biggest concern with this PR is the way it affects powers in equations. See the difference for yourself and notice that the power (sup) becomes a sub. For instance, see the "2" below:

**Before PR:** 
Screen Shot 2021-07-29 at 11 28 38 AM

**After PR:**
Screen Shot 2021-07-29 at 11 28 28 AM

This needs to be fixed. Thanks!

EricDunsworth:
I took a quick look at it yesterday. Looks like an upstream bug.

Here's an excerpt of the problematic part of that equation from the [MathML demo page](https://wet-boew.github.io/v4.0-ci/unmin/demos/mathml/mathml-en.html):
```html
<msubsup>
	<mrow>
		<mo>(</mo>
		<munderover>
			<mo>∑</mo>
			<mrow>
				<mi>k</mi>
				<mo>=</mo>
				<mn>1</mn>
			</mrow>
			<mi>n</mi>
		</munderover>
		<msubsup>
			<mi>a</mi>
			<mi>k</mi>
			<mrow></mrow>
		</msubsup>
		<msubsup>
			<mi>b</mi>
			<mi>k</mi>
			<mrow></mrow>
		</msubsup>
		<mo>)</mo>
	</mrow>
	<mrow></mrow>
	<mn>2</mn>
</msubsup>
```

[MDN's ``<msubsup>`` element page](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/msubsup) states that ``<msubsup>`` is designed to support 3 children.

The order of the children affect how they're presented. The second child represents superscripted content, whereas the third represents subscripted content.

I think what's happening is that MathJax 3.2.0 is ignoring the empty ``<mrow></mrow>`` element in our problematic equation. That in turn is causing the ``<mn>2</mn>`` element to get treated as the second child (instead of the third). End result is that it gets presented as a sub instead of sup.

FWIW Firefox's native MathML implementation and WET's MathJax 2.7.4 polyfill don't have that issue.

Gruntfile.coffee
```diff
@@ -1141,17 +1141,12 @@ module.exports = (grunt) ->
...
-						"extensions/**"
-						"jax/**"
-						"localization/**"
-						"fonts/HTML-CSS/TeX/woff/**"
-						"fonts/HTML-CSS/TeX/otf/**"
+						"mml-chtml.js"
+						"output/chtml/fonts/woff-v2/**"
 					]
-					dest: "<%= coreDist %>/js/MathJax/"
+					dest: "<%= coreDist %>/js/MathJax/es5/"
```
duboisp, line 1149: just put in the root of "MathJax" folder

src/core/wb.js
```diff
@@ -571,14 +571,29 @@ Modernizr.load( [
...
+						],
 						complete: function() {
+
+							// Fall back on a local copy MathJax if CDN loading fails
+							// Note: Won't work with IE11 in isolated networks (ES6 polyfill has no local fallback)
 							Modernizr.load( [ {
-								test: window.MathJax === undefined,
-								yep: "mthjx!MathJax.js?config=Accessible"
+								test: window.MathJax.startup === undefined,
+								yep: "mthjx!mml-chtml.js"
```
duboisp, line 596: Can we try to load the old MathJax if it don't work with 3

src/core/wb.js
```diff
@@ -571,14 +571,29 @@ Modernizr.load( [
...
+							"timeout=500!https://cdn.jsdelivr.net/npm/mathjax@WET_BOEW_VERSION_MATHJAX/es5/mml-chtml.js"
+						],
 						complete: function() {
+
+							// Fall back on a local copy MathJax if CDN loading fails
+							// Note: Won't work with IE11 in isolated networks (ES6 polyfill has no local fallback)
 							Modernizr.load( [ {
-								test: window.MathJax === undefined,
-								yep: "mthjx!MathJax.js?config=Accessible"
+								test: window.MathJax.startup === undefined,
```
duboisp, line 595: You might need to use the trick like "IsReady" and implement a little delay before to check again if it is loaded or not.

src/core/wb.js
```diff
@@ -571,14 +571,29 @@ Modernizr.load( [
...
+
 					// Load the MathML dependency. Since the polyfill is only loaded
 					// when !Modernizr.mathml, we can skip the test here.
 					Modernizr.load( [ {
-						load: "timeout=500!https://cdn.jsdelivr.net/npm/mathjax@WET_BOEW_VERSION_MATHJAX/MathJax.js?config=Accessible",
+
+						// Load ES6 polyfill (for IE11) and MathJax from CDNs
+						load: [
+							"timeout=500!https://polyfill.io/v3/polyfill.min.js?features=es6",
+							"timeout=500!https://cdn.jsdelivr.net/npm/mathjax@WET_BOEW_VERSION_MATHJAX/es5/mml-chtml.js"
```
duboisp, line 588: Considering the polyfill is required to run the mml-chtml.js, ensure the polyfill is loaded and ready to be used before to load the mathJax.

duboisp (changes requested):

src/core/wb.js
```diff
@@ -571,14 +571,29 @@ Modernizr.load( [
...
+					};
+
 					// Load the MathML dependency. Since the polyfill is only loaded
 					// when !Modernizr.mathml, we can skip the test here.
 					Modernizr.load( [ {
-						load: "timeout=500!https://cdn.jsdelivr.net/npm/mathjax@WET_BOEW_VERSION_MATHJAX/MathJax.js?config=Accessible",
+
+						// Load ES6 polyfill (for IE11) and MathJax from CDNs
+						load: [
+							"timeout=500!https://polyfill.io/v3/polyfill.min.js?features=es6",
```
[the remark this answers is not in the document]
duboisp: You can use a technique like this: https://github.com/wet-boew/wet-boew/blob/master/src/core/helpers.js#L46

src/core/wb.js
```diff
@@ -481,7 +481,7 @@ yepnope.addPrefix( "i18n", function( resourceObj ) {
  * @prefix: mthjx! - adds the root directory of MathJax resources
  */
 yepnope.addPrefix( "mthjx", function( resourceObj ) {
-	resourceObj.url = paths.js + "/MathJax/" + resourceObj.url;
+	resourceObj.url = paths.js + "/MathJax/es5/" + resourceObj.url;
```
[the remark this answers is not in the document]
EricDunsworth: Done.

Gruntfile.coffee
EricDunsworth: Done.

src/core/wb.js
EricDunsworth: Just added this in... but should note that it isn't really ideal and will take a while to load in. I've implemented it as a last ditch effort to load in *some* kind of MathJax flavour.

It won't attempt to load in at the same time as MathJax 3 (since there's a risk of conflicts if both exist). It'll only try to load in if both MathJax 3 attempts (for the CDN and local versions) fail.

One way of speeding this up a bit would be to stop trying to load MathJax 3 from a CDN and just go straight to checking the local MathJax 3 dependency from the get-go. That'd skip one "step" in the event that MathJax 2 needs to be fallen back on.

EricDunsworth:
Haven't forced-pushed any updates yet, but it looks like I won't need to switch over to the SVG renderer to get around the CHTML renderer's ``msubsup`` issue.

mathjax/MathJax#2765 was accepted and @dpvc sent a PR to fix it. In the meantime, he suggested a CSS workaround fo 3.2.0. I've revised this PR to use it :). We'll probably be able to remove it post-3.2.0.

src/core/wb.js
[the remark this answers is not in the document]
EricDunsworth: Done via your suggestion :).

FWIW there could potentially be an opportunity to set this up as a global ES6 polyfill for IE11. Aside from MathJax 3, the geomap plugin currently uses ES6 in ways that are incompatible with IE11.

src/core/wb.js
EricDunsworth: My latest changes bring it in prior to MathJax 3, although there aren't any extra checks in place atm to be completely certain that it's ready to rock and roll.

duboisp (changes requested):

duboisp (commented):
For the **release note**:

New file:
* mathml.min.css
* js/MathJax/mml-chtml.js
* js/MathJax/output/chtml/fonts/woff-v2/**  (All the file/folder contained in this location)
* wet-boew/css/polyfills/mathml.min.css

File to remove:
* MathJax.js
* All the file/folder contained in those location:
   * config/**
   * extensions/**
   * jax/**
   * localization/**
   * fonts/HTML-CSS/TeX/woff/**
   * fonts/HTML-CSS/TeX/otf/**

Modified file:
* wet-boew.min.js

src/core/wb.js
```diff
@@ -571,18 +571,49 @@ Modernizr.load( [
...
+						test: window.navigator.msSaveOrOpenBlob,
+						yep: "timeout=500!https://polyfill.io/v3/polyfill.min.js?features=es6"
+					} ] );
+
 					// Load the MathML dependency. Since the polyfill is only loaded
 					// when !Modernizr.mathml, we can skip the test here.
 					Modernizr.load( [ {
-						load: "timeout=500!https://cdn.jsdelivr.net/npm/mathjax@WET_BOEW_VERSION_MATHJAX/MathJax.js?config=Accessible",
+
+						// Load latest version of MathJax 3 from a CDN
```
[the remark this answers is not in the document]
EricDunsworth: Done :).

package.json
```diff
@@ -41,7 +41,7 @@
     "jquery": "2.2.4",
     "jquery-validation": "1.19.3",
     "magnific-popup": "git+https://github.com/wet-boew/Magnific-Popup.git#1.0.0+keyboard_trap_fix",
-    "mathjax": "2.7.4",
+    "mathjax": "^3.2.0",
```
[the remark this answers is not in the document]
EricDunsworth: Wouldn't the status quo be better since it sets the minimum version to 3.2.0, while also allowing newer minor versions (3.x)? Try throwing ``^3.2.0`` into the [npm semver version calculator](https://semver.npmjs.com/)'s range field to visualize what it'd represent in practice.

Setting this to the latest major version (e.g. ``3``, ``3.x``, etc) would be kind of misleading since it'd technically indicate that WET's MathJax dependency is compatible with versions 3.0.0-3.1.4 (which we can't guarantee since those older versions have never been tested in WET).

Gruntfile.coffee
```diff
@@ -1391,16 +1385,6 @@ module.exports = (grunt) ->
...
 				dest: 'site/data/i18n'
 
-		"string-replace":
-			inline:
-				files:
-					"dist/wet-boew/js/": "dist/wet-boew/js/*.js"
-				options:
-					replacements: [
-						pattern: "WET_BOEW_VERSION_MATHJAX"
-						replacement: "<%= MathJaxVersion %>"
```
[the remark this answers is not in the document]
EricDunsworth: Done :).

src/core/wb.js
```diff
@@ -571,18 +571,49 @@ Modernizr.load( [
...
+						load: [
+							"timeout=500!https://cdn.jsdelivr.net/npm/mathjax@3/es5/mml-chtml.js",
+							"plyfll!mathml.min.css"
+						],
 						complete: function() {
+
+							// Fall back on a local copy of MathJax 3 if CDN loading fails
+							// Note: Won't work with IE11 in isolated networks (ES6 polyfill has no local fallback)
```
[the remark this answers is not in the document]
EricDunsworth: Done :).

src/core/wb.js
EricDunsworth: Gonna mark this as resolved too since I don't think this is still relevant (due to ``setTimeout()``).

EricDunsworth:
Gonna have to hold off on this for a little longer...

IE11 is failing to load in MathJax 3's fonts. Seems to be unique to this PR's MathJax 3 implementation. Looks like the cause is invalid ``@font-face`` source URLs in a ``<style>`` element. I don't believe it's related to a race condition (threw in some long ``setTimeout()`` functions to no avail). FWIW Edge Chromium is fine.

By comparison, [MathJax 3's official MML-CHTML demo page](https://mathjax.github.io/MathJax-demos-web/mml-chtml.html) looks fine in IE11. WET's demo page also looks fine in it if I hardcode these into the page:
```
<script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
<script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/mml-chtml.js"></script>
```

EricDunsworth:
Forced-pushed some new updates that resolve IE11's font loading issues.

Capitalization is still kind of messed up in IE11, but that's a known issue and might not have an easy fix (it's due to Trident mishandling pseudo-element selectors).

duboisp (changes requested):
See the small optimization suggestion for the IE11 detection.

We would need to quickly test it on IE11, FF and Chromium

duboisp (changes requested):
I tested in IE11 and it work as expected. Both fallback work, except the typo that need to be fixed.

When the fallback JS filename is fixed, we will be good to merge this PR.

Test page: https://universallabs.org/wet/labs/pr-9143/demos/mathml/mathml-en.html 

Thanks @EricDunsworth for your work on this.

duboisp (approved):

GormFrank (approved):

PR #9146, opened 2021-08-03
Form validation: Hide "required" label text from screen readers

### LINKED ISSUE #6966: Form validation: "Required" is spoken twice
The word "required" is being spoken twice for each form element example here.  Why not use the ARIA=required true?

### LINKED ISSUE #7002: Form validation: "required" spoken twice for screen readers
Why is the word "required" identified and spoken twice?  For example using JAWS 16 and/or NVDA (latest build), the word "required" is spoken by the screen reader before and after the required edit field.
@wet-boew/access-working-group

### LINKED ISSUE #8091: Form validation: required, verbos?

### LINKED ISSUE #9132: "required" label seems  redundant Form validation:
All form field working examples announce "required" twice for screen readers. Is this necessary? IMO hearing "required" twice is diconfusing and slightly distracts from the content. 
It is, of course, important to know if a form field is "required" but does it need to be offered twice for each "required" form field?

EricDunsworth:
* Makes the plugin automatically add ``aria-hidden="true"`` to required indicators in the labels of semantically-required fields (i.e. any fields that use ``required`` or ``aria-required="true"`` attributes)
  * Useful for older/pre-existing forms that might not otherwise benefit from this change
* Adds info alerts to recommend hardcoding aria-hidden="true" near required field labels in demo pages
* Adds a new section to the plugin's documentation that re-uses the include's content
* Fixes #6966
* Fixes #7002
* Fixes #8091
* Fixes #9132
* Depends on #9166

CC @StacyBleeks @shawnthompson

### DIFF
--- a/site/includes/alertariahidden.hbs
+++ b/site/includes/alertariahidden.hbs
@@ -1,0 +1,8 @@
+{{! Page content include}}
+{{#contains this.alert "false"}}{{else}}<div class="alert alert-info">{{/contains}}
+{{#contains language "en"}}
+	<p>Consider adding an <code>aria-hidden="true"</code> attribute to <code>&lt;strong class="required"&gt;</code> (i.e. <code>&lt;strong class="required" aria-hidden="true"&gt;</code>) in required field labels. Hardcoding it will prevent screen readers from announcing "required" twice in the noscript and basic HTML versions of the page. The form validation plugin automatically adds the attribute to the JavaScript version. Don't use the attribute in required <code>&lt;legend&gt;</code> elements.</p>
+{{else}}
+	<p>Considérez ajouter un attribut <code lang="en">aria-hidden="true"</code> à <code lang="en">&lt;strong class="required"&gt;</code> (c.-à-d. <code lang="en">&lt;strong class="required" aria-hidden="true"&gt;</code>) dans les libellés des champs obligatoires. Le code ajouté de façon statique préviendra les lecteurs d’écran d’annoncer «&#160;obligatoire&#160;» deux fois dans les versions sans script et HTML simplifiée de la page. Le plugiciel de validation de formulaires ajoute l’attribut automatiquement à la version JavaScript. N’utilisez pas l’attribut dans les éléments <code lang="en">&lt;legend&gt;</code> obligatoires.</p>
+{{/contains}}
+{{#contains this.alert "false"}}{{else}}</div>{{/contains}}
--- a/site/pages/docs/ref/formvalid/formvalid-en.hbs
+++ b/site/pages/docs/ref/formvalid/formvalid-en.hbs
@@ -6,7 +6,7 @@
 	"categoryfile": "plugins",
 	"description": "Provides generic validation and error message handling for Web forms.",
 	"altLangPrefix": "formvalid",
-	"dateModified": "2022-01-25"
+	"dateModified": "2022-04-13"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
@@ -278,6 +278,10 @@
 		<dd>Overwrites the default generic message that comes with the use of the pattern attribute by the one given through this attribute.</dd>
 	</dl>
 	<p>See <a href="https://html.spec.whatwg.org/#the-pattern-attribute" rel="external">WHATWG HTML Living Standards</a> for more details about the pattern attribute.</p>
+</section>
+<section>
+	<h3 id="required">Prevent screen readers from announcing "required" twice</h3>
+	{{>alertariahidden alert="false"}}
 </section>
 <section>
 	<h3 id="MergeSCE">Merge Server-Client errors functionality</h3>
--- a/src/plugins/formvalid/formvalid.js
+++ b/src/plugins/formvalid/formvalid.js
@@ -102,6 +102,26 @@ var componentName = "wb-frmvld",
 						labels[ i ].innerHTML += " ";
 					}
 
+					// Hide "required" label text in older forms from screen readers
+					// Prevents redundant "required" announcements on semantically-required fields whose labels mention they're required
+					$form.find( "strong.required:not([aria-hidden='true'])" ).each( function() {
+						const $requiredText = $( this ),
+							$label = $requiredText.closest( "label" ),
+							fieldId = $label.attr( "for" );
+						let $field = fieldId ? $( "#" + fieldId ) : $label.find( ":input" ).first();
+
+						// If the label's field has yet to be found, look for fields that refer to the label via aria-labelledby
+						if ( !$field.length ) {
+							const labelId = $label.attr( "id" ) || $requiredText.closest( "id" );
+							$field = $form.find( "[aria-labelledby~='" + labelId + "']:input" ).first();
+						}
+
+						// Hide the "required" text if its field is semantically-required
+						if ( $field.is( "[required], [aria-required='true']" ) ) {
+							$requiredText.attr( "aria-hidden", "true" );
+						}
+					} );
+
 					// The jQuery validation plug-in in action
 					validator = $form.validate( {
 						meta: "validate",
--- a/site/pages/docs/ref/formvalid/formvalid-fr.hbs
+++ b/site/pages/docs/ref/formvalid/formvalid-fr.hbs
@@ -6,7 +6,7 @@
 	"categoryfile": "plugins",
 	"description": "Effectue la validation de formulaires Web selon un ensemble de règles de base avant qu'ils soient soumis.",
 	"altLangPrefix": "formvalid",
-	"dateModified": "2022-01-25"
+	"dateModified": "2022-04-13"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
@@ -281,6 +281,10 @@
 		<dd>Overwrites the default generic message that comes with the use of the pattern attribute by the one given through this attribute.</dd>
 	</dl>
 	<p>See <a href="https://html.spec.whatwg.org/#the-pattern-attribute" rel="external">WHATWG HTML Living Standards</a> for more details about the pattern attribute.</p>
+</section>
+<section lang="fr">
+	<h3 id="required">Préviens les lecteurs d’écran d’annoncer «&#160;obligatoire&#160;» deux fois</h3>
+	{{>alertariahidden alert="false"}}
 </section>
 <section>
 	<h3 id="MergeSCE">Merge Server-Client errors functionality</h3>
--- a/src/other/feedback/feedback-en.hbs
+++ b/src/other/feedback/feedback-en.hbs
@@ -9,9 +9,10 @@
 	"altLangPrefix": "feedback",
 	"css": ["demo/feedback"],
 	"js": ["demo/feedback"],
-	"dateModified": "2015-03-06"
+	"dateModified": "2022-04-13"
 }
 ---
+{{>alertariahidden}}
 <div class="wb-frmvld wb-fdbck nojs-hide">
 	<p>If you have difficulty with the following form, you can use any of our <a href="#osc-avs">other service channels</a> to contact us.</p>
 	<form action="#" method="get" id="fdbck-frm">
--- a/src/other/feedback/feedback-fr.hbs
+++ b/src/other/feedback/feedback-fr.hbs
@@ -9,9 +9,10 @@
 	"altLangPrefix": "feedback",
 	"css": ["demo/feedback"],
 	"js": ["demo/feedback"],
-	"dateModified": "2014-02-28"
+	"dateModified": "2022-04-13"
 }
 ---
+{{>alertariahidden}}
 <div class="wb-frmvld wb-fdbck nojs-hide">
 	<p>Si vous avez des problèmes avec le formulaire suivant, vous pouvez utiliser une des <a href="#osc-avs">autres voies de service</a> pour communiquer avec nous.</p>
 	<form action="#" method="get" id="fdbck-frm">
--- a/src/plugins/formvalid/formvalid-en.hbs
+++ b/src/plugins/formvalid/formvalid-en.hbs
@@ -7,7 +7,7 @@
 	"tag": "formvalid",
 	"parentdir": "formvalid",
 	"altLangPrefix": "formvalid",
-	"dateModified": "2022-01-25"
+	"dateModified": "2022-04-13"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
@@ -43,6 +43,7 @@
 				</div>
 				<details class="mrgn-bttm-md">
 					<summary>View code</summary>
+					{{>alertariahidden}}
 					<pre><code>&lt;div class="wb-frmvld"&gt;
 	&lt;form action="#" method="get" id="validation-example"&gt;
 		...
@@ -64,6 +65,7 @@
 				</div>
 				<details class="mrgn-bttm-md">
 					<summary>View code</summary>
+					{{>alertariahidden}}
 					<pre><code>&lt;div class="wb-frmvld"&gt;
 	&lt;form action="#" method="get" id="validation-example"&gt;
 		...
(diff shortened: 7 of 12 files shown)


StacyBleeks:
@EricDunsworth  thanks for following up. Is there something here to test? The description sounds brilliant! I am not sure what to do in order to test the fix.

EricDunsworth:
To test it out either I or @GormFrank will need to host this PR's dist folder somewhere.

I don't know how useful this is, but I also created a temporary demo page at "src/plugins/formvalid/formvalid-test-en.hbs" to show off the plugin enhancements injecting ``aria-hidden="true"`` attributes onto "(required)" text in in an outdated form.  I expect to remove that test page by the time this PR gets out of a draft state.

EricDunsworth:
@StacyBleeks I've setup some demo pages on GitHub pages :).

**Links:**
* Form validation plugin:
  * [Old form example](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-test-en.html)
  * [Normal demo page](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-en.html)
  * [Server demo page](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-server-en.html)
* [Feedback form demo](https://ericdunsworth.github.io/wet-boew/unmin/demos/feedback/feedback-en.html)

@GormFrank We're going to need to discuss the build failure at some point. ESLint is failing because it doesn't like my ``const``/``let`` keywords - even though IE11 seems to be fine with how I'm using them...

There doesn't seem to be a good way of adjusting the ESLint configuration to account for IE11's partial support for those variable types. No configuration options exist specifically for ``const``/``let`` support and setting the environment to ES6 would be overkill for IE11.

StacyBleeks:
@EricDunsworth  Awesome! Testing the "required" fields using various screen reader browser combinations is showing excellent results. Some fields are not identified as "required" (email, web site, date picker) is this intentional? Most of the form elements are testing well, with only one mention of "required" 😂

EricDunsworth:
> Some fields are not identified as "required" (email, web site, date picker) is this intentional?

The fields you mentioned (i.e. email, web site and date picker) are intended to be optional.

I suspect that some mandatory fieldsets might still have repetitive "required" announcements. #9091 modified some of them to use ``data-rule-required="true"`` (instead of using a ``required`` attribute on their first checkbox/radio button). But others were left out (like the [normal demo](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-en.html)'s stacked checkbox/radio button fieldsets and all of the [server demo](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-server-en.html)'s required fieldsets).

Here's an overview of which fields are supposed to be required in each demo page:
* [Old form example](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-test-en.html)
  * All fields
* [Normal demo](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-en.html)
  * Title
  * First name
  * Last name
  * Postal code
  * File
  * Favourite pets (fieldset with checkboxes)
  * Citizenship status (fieldset with radio buttons)
  * Application number
  * All the stacked checkbox/radio button fieldsets at the end of the content
* [Server demo](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-server-en.html)
  * All fields are required... but the form looks really messy
* [Feedback form demo](https://ericdunsworth.github.io/wet-boew/unmin/demos/feedback/feedback-en.html)
  * Reason for contacting us
  * Message

> Most of the form elements are testing well, with only one mention of "required" 😂

Which field was that?

EricDunsworth:
@StacyBleeks @shawnthompson @dfarough Just wondering if any of you wanted to perform any further testing on this?

If no one has any lingering concerns about this PR's changes, I'll bring it out of draft and remove the [old form example (a test page I created to test my JavaScript logic)](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-test-en.html) 😃.

Btw #9159 will need to be merged in before this PR (to prevent my code from causing linting errors).

shawnthompson:
I might be late in the game to share this, but here goes anyways.

@andrewnordlund and I were talking about this the other day.

The [Identifying a required field with the aria-required property](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA2.html) technique in WCAG has this as an example:

```html
  <p>
    <label for="zip_post">Zip / Postal code: </label>
    <input type="text" id="zip_post" size="6" aria-required="true" />
    [required]
  </p>
```

Which moves the visible required text out of the label. We'd need to test but I think this might stop the repeat.

I'm assuming the same goes if the `required` attribute is used instead of `aria`.

dfarough:
This might work, but if a screen reader user was arrowing through the form, I suspect that the user would still hear the required text even though it was outside of the label. It would probably be necessary to use aria-hidden to prevent this.

shawnthompson:
Great point @dfarough.

EricDunsworth:
I believe what this pull request is already doing is functionally-identical to what [@shawnthompson's code sample](https://github.com/wet-boew/wet-boew/pull/9146#issuecomment-909181219) would be like if its "required" text used an ``aria-hidden="true"`` attribute.

I think it'd be best to keep the "required" text inside the labels for the following reasons:
* Having the hardcoded "required" text inside the labels expands the labels' "linked" areas. That might be helpful to sighted users that use a mouse/touch screen/etc, since clicking on a label usually shifts focus to the label's associated field (or toggles checkboxes/radio buttons).
* There are probably almost a decade's-worth of WET 4.x forms out there that are based on the [form validation demo page](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-en.html). I don't think it'd be wise to do anything too drastic with the "required" text's position this late in the game since it'd require CSS changes that might risk breaking visual layouts in unexpected ways.
* It might be tricky (if not impossible) to use CSS to visually-position the "required" text right next to the label in the HTML code sample from [@shawnthompson's comment](https://github.com/wet-boew/wet-boew/pull/9146#issuecomment-909181219). It'd probably necessitate absolute positioning or flexbox. Those might not work out very well for more elaborate form designs.

As far as the ``required`` attribute vs ``aria-required="true"`` goes... WET uses the ``required`` attribute since it's a native HTML attribute and [has the same implicit ARIA semantics as ``aria-required="true"``](https://www.w3.org/TR/html-aria/#att-required).

In any case, the changes this pull request makes to the form validation plugin use both kinds of attributes to detect required fields. It doesn't attempt to change one kind of required attribute into another.

@shawnthompson @dfarough @andrewnordlund The reason as to why "required" usually gets repeated twice in quick succession by screen readers in WET's forms is because labels contain hardcoded "required" text and the fields themselves use a ``required`` (or ``aria-required=true``) attribute.

Screen readers always announce the label's hardcoded "required" text no matter what. Modern-day screen readers also announce "required" upon encountering a field that uses a ``required`` attribute or ``aria-required=true``. So the end result is a stutter effect.

This pull request addresses it by adding ``aria-hidden="true"`` to the hardcoded "required" text. That removes one of the two announcements without disadvantaging sighted users.

duboisp:
@EricDunsworth just add a single exception for this file

src/plugins/formvalid/formvalid.js
```diff
@@ -102,6 +102,26 @@ var componentName = "wb-frmvld",
 						labels[ i ].innerHTML += " ";
 					}
 
+					// Hide "required" label text in older forms from screen readers
+					// Prevents redundant "required" announcements on semantically-required fields whose labels mention they're required
```
duboisp, line 106: Can you also add this note in the plugin documentation, thanks

duboisp (changes requested):

src/plugins/formvalid/formvalid.js
EricDunsworth: Done :).

EricDunsworth:
I've just force-pushed an update that does the following:
* Removes my [temporary test page](https://ericdunsworth.github.io/wet-boew/unmin/demos/formvalid/formvalid-test-en.html)
* Adds ``aria-hidden="true"`` into one of the examples in the documentation's how to section
* Addresses @duboisp's requested change

Also bumped it up from draft to ready for review since I think it's all set to be merged now (so long as #9166 is brought in first).

duboisp:
Please rebase

site/pages/docs/ref/formvalid/formvalid-en.hbs
```diff
@@ -276,6 +276,10 @@
 	</dl>
 	<p>See <a href="https://html.spec.whatwg.org/#the-pattern-attribute" rel="external">WHATWG HTML Living Standards</a> for more details about the pattern attribute.</p>
 </section>
+<section>
```
duboisp, line 279: Please replicate this block in French.

duboisp (changes requested):

site/pages/docs/ref/formvalid/formvalid-en.hbs
EricDunsworth: Fixed :).

duboisp (changes requested):
We should check to implement a "basic mode" init where it will add the `aria-hidden="true"` instead of requiring the web author/developper to manually add it in their page.
The first rule of ARIA is not to use ARIA.

@GormFrank can you look to implement an init event when basic HTML is activated, like when the page is in Basic Mode.

EricDunsworth:
> We should check to implement a "basic mode" init where it will add the aria-hidden="true"...

Just wondering if it'd be possible to provide an estimate as to when that feature might become available to use?

From my end, I believe this PR is ready to go. All that's holding it back seems to be whether the ``aria-hidden="true"`` attributes should be hardcoded and workarounds for basic HTML mode.

> ...instead of requiring the web author/developper to manually add it in their page.

Is the purpose of WET's code samples to convey (minimum?) requirements? Or to demonstrate best practices?
* If it's minimum requirements, then I guess it'd make sense to remove the hardcoded ``aria-hidden="true"`` attributes.
* If the intent is to show off best practices, I think it'd be best to keep the attributes since omitting them would degrade things in noscript.

In any case, no one would need to start hardcoding ``aria-hidden="true"`` attributes if they don't want to.

>  The first rule of ARIA is not to use ARIA.

I agree with that premise, but what this PR is doing with hardcoding ``aria-hidden`` attributes isn't one of those cases.

ARIA isn't being used to mimic something that [could've been done using native HTML semantics](https://www.w3.org/TR/using-aria/#rule1). It's fulfilling a purpose that'd otherwise be impossible. Specifically - instructing screen readers to ignore content that conveys semantic information (that they can already pick up on) to sighted users.

Btw it's not unheard of for ARIA attributes to be hardcoded into static web content. Even WET's Lightbox plugin uses an ``aria-describedby`` attribute in [one of its examples](https://wet-boew.github.io/v4.0-ci/demos/lightbox/lightbox-en.html#examples2).

StacyBleeks:
@EricDunsworth  thanks for the update and insight regarding the rationale behind using ARIA in this particular case.

GormFrank:
@duboisp Do you think it could be worth a shot accepting this PR and going back to the basic HTML eventually? We would need to document it as a work item (issue and/or JIRA ticket).

EricDunsworth:
Just fixed the merge conflicts and updated modified dates.

GormFrank:
As discussed during the technical review, let's meet in the middle. @EricDunsworth Can you please remove the hardcoded `aria-hidden="true"` from the examples, but **add a notice blurb** recommending authors to hardcode them in this specific context if they have the knowledge required to play with aria attributes.

On a separate note, @Ricokola we will look into the Basic HTML init previously mentioned by @duboisp that could be beneficial for this PR.

EricDunsworth:
I've just force-pushed an updated commit with the following updates:
* Removed all hardcoded instances of ``aria-hidden="true"``
* Added an info alert near all examples that use required field labels:
  * Set it up as an include for maintainability
  * Placed it in the "View code" dropdowns of related code samples (same idea as the ``data-rule-required="true"`` alerts)
  * Placed it at the top of examples that lack code samples
* Updated documentation section to match the include's content (didn't re-use the include since it's structured differently)
* Updated page modified dates

GormFrank:
Looks good. To do:

- [ ] Test locally

site/includes/alertariahidden.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="alert alert-info">
```
duboisp, line 1: Please add a comment regarding this include is a content include

duboisp (dismissed):
Approved upon successful local testing by another dev

EricDunsworth:
Just force-pushed a few more updates:
* Added a handlebars comment to the include to indicate it's for page content
* Added extra logic to the alert to allow its alert div to be omitted via an ``alert="false"`` parameter
* Improved the include's French translation
* Revised the new documentation section to source its content from the include
* Added ``lang="fr"`` to the French version of the new documentation section
* Translated the H3 heading in the French version of the new documentation section

duboisp:
Approve upon successful local testing

GormFrank:
**Tested and works like a charm. Thank you @EricDunsworth!**

One thing I noticed when testing is the fact that the fields in the two forms of the Postback working examples page share the same `id`s which creates duplication (semantic error) and therefore prevents your script from working properly. Not your fault @EricDunsworth. Can I ask you to fix the `id`s duplication on the forms if you don't mind? Otherwise I don't see it as a blocker, I can open a following PR fixing those later on.

cc: @duboisp

GormFrank (approved):

EricDunsworth:
@GormFrank Would you recommend that I try revising the plugin to do a better job of coping with duplicate IDs in botched pages?

As for the postback demos, I'll send in a separate PR to fix them since HTML validation errors in pre-existing pages go beyond this PR's scope. Some other pages also have their own errors.

Btw I just tried running ``grunt htmllint`` in my fork of wet-boew/wet-boew and it detected a few dozen hits. Shouldn't GitHub Actions run that task automatically on all incoming PRs? The only "combined" task that currently calls ``htmllint`` is ``dist`` (which GitHub Actions doesn't run). It isn't part of the ``test`` task. There's currently nothing in place to automatically catch HTML validation errors in net-new PRs.

GormFrank:
@EricDunsworth We can't foresee all botched semantics authors will code and so we should focus on solutions based on right semantics; therefore I was not requesting any work item on that aspect. 

No problem if you are making a different PR for the working examples, was just pointing out the fact that it affected our ability to test. Thank you for agreeing to do it, I was not imposing anything on you! Very appreciated.

Let's discuss on the other item with @duboisp during technical review :)

PR #9147, opened 2021-08-04
Form SCSS: Remove invalid selector and associated scss-lint rule

EricDunsworth:
That selector's sub-selectors are designed to place a red wildcard in a ``::before`` pseudo-element and red text in a inner ``strong`` element on required fields. Which makes sense on labels, legends, etc... but not on void ``input`` elements.

This also removes an inline scss-lint rule whose only purpose was to prevent the checkbox input's selector from causing a linting error. WET replaced scss-lint with sass-lint long ago.

### DIFF
--- a/src/base/forms/_base.scss
+++ b/src/base/forms/_base.scss
@@ -10,11 +10,9 @@ $asterisk-width: .665em;
 	font-weight: 700;
 }
 
-// scss-lint:disable QualifyingElement
 label,
 legend,
-.checkbox,
-input[type="checkbox"] {
+.checkbox {
 	&.required {
 		&:before {
 			@extend %forms-color-red-bold;
@@ -29,7 +27,6 @@ input[type="checkbox"] {
 		}
 	}
 }
-// scss-lint:enable QualifyingElement
 
 [dir=rtl] {
 	label,


GormFrank (approved):

EricDunsworth:
Btw just so it's noted somewhere...

When we first discussed this PR a few weeks ago, @GormFrank noticed that placing a ``class="required"`` attribute on ``<input type="checkbox">`` causes a red wildcard to appear before the checkbox in Blink (and maybe WebKit?). That's really surprising since it doesn't make any sense :S. Btw Gecko and Trident don't try to render it.

The ``:before`` pseudo-element is supposed to be the first child of the selected element. What's the first child of ``<input>``? Nothing, since it's a void element. If Blink's behaviour had been coded as HTML, it would've been the same as doing ``<input type="checkbox">*</input>`` (which goes against the HTML spec and fails validation).

duboisp:
@EricDunsworth thank for explaining the issue, we will investigate the impact.

Here a code sample, it render differently in Chrome but not in FF
`<input type="checkbox" name="animal" value="1" id="animal1" data-rule-required="true" required="" class="required">`

duboisp:
FYI - I submitted the request for gather some data to let us evaluating the impact. (MWS-2357). They are going to search for pages containing match with the CSS selector: `input[type="checkbox"].required`

duboisp:
We have completed the search, and we didn't found any instance. So we will be able to merge this PR at our next technical review.

PR #9150, opened 2021-08-10
Form validation: Improve standalone checkboxes

EricDunsworth:
* Add an example of a required standalone checkbox
* Move validation error messages to the end of standalone checkbox labels instead of the beginning (pre-#8397 behaviour for fields that don't use the "wb-server-error"/"css-implicite-input" classes)
* Add a "checkbox-standalone" class to allow standalone checkbox labels to match those of other kinds of standalone fields (e.g. text inputs, etc...)
* Cover the "checkbox-standalone" class in the documentation's how to implement section.
* Enhance the look of required standalone checkboxes:
  * Prevent wildcard symbols from shifting downwards if a validation error appears
  * Prevent multi-line labels from causing the checkbox/label to appear on a different line than the wildcard
  * Don't change the label to dark red if a validation error appears

CC @StdGit @delisma

**Screenshots...**

**Before #8397:**
checkbox-required-before-pr-8397

**After #8397:**
checkbox-required-after-pr-8397

**After this PR (without bolding):**
checkbox-required-after-pr-9150-v2

**After this PR (with bolding):**
checkbox-required-after-pr-9150-bold-v2

### DIFF
--- a/src/base/forms/_base.scss
+++ b/src/base/forms/_base.scss
@@ -3,7 +3,7 @@
   @title: Form additions to WET-BOEW
  */
 
-$asterisk-width: .665em;
+$asterisk-width: .67em;
 
 %forms-color-red-bold {
 	color: #d3080c;
@@ -20,6 +20,7 @@ input[type="checkbox"] {
 			@extend %forms-color-red-bold;
 			content: "* ";
 			margin-left: -$asterisk-width;
+			vertical-align: top;
 		}
 
 		strong {
@@ -30,6 +31,22 @@ input[type="checkbox"] {
 	}
 }
 // scss-lint:enable QualifyingElement
+
+.form-group {
+	&.has-error {
+		.checkbox {
+			color: $text-color;
+		}
+	}
+
+	.checkbox {
+		&.checkbox-standalone {
+			label {
+				font-weight: 700;
+			}
+		}
+	}
+}
 
 [dir=rtl] {
 	label,
--- a/src/plugins/formvalid/formvalid-en.hbs
+++ b/src/plugins/formvalid/formvalid-en.hbs
@@ -7,7 +7,7 @@
 	"tag": "formvalid",
 	"parentdir": "formvalid",
 	"altLangPrefix": "formvalid",
-	"dateModified": "2021-08-11"
+	"dateModified": "2022-01-25"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
@@ -350,6 +350,23 @@
 		&lt;/div&gt;</code></pre>
 				</details>
 			</fieldset>
+
+			<div class="form-group">
+				<div class="checkbox checkbox-standalone required">
+					<label for="agree1"><input id="agree1" name="agree1" type="checkbox" required> <span class="field-name">I agree</span> <strong class="required">(required)</strong></label>
+				</div>
+			</div>
+			<details class="mrgn-bttm-md">
+				<summary>View code</summary>
+				<pre><code>&lt;div class="wb-frmvld"&gt;
+	&lt;form action="#" method="get" id="validation-example"&gt;
+		...
+			&lt;div class="form-group"&gt;
+				&lt;div class="checkbox checkbox-standalone required"&gt;
+					&lt;label for="agree1"&gt;&lt;input id="agree1" name="agree1" type="checkbox" required&gt; &lt;span class="field-name"&gt;I agree&lt;/span&gt; &lt;strong class="required"&gt;(required)&lt;/strong&gt;&lt;/label&gt;
+				&lt;/div&gt;
+			&lt;/div&gt;</code></pre>
+			</details>
 
 			<fieldset class="chkbxrdio-grp">
 				<legend class="required"><span class="field-name">Favourite pets</span> <strong class="required">(required)</strong></legend>
--- a/src/plugins/formvalid/formvalid.js
+++ b/src/plugins/formvalid/formvalid.js
@@ -163,8 +163,8 @@ var componentName = "wb-frmvld",
 
 							//Std If we have a label and the input field is inside the label
 							// need to add a css-implicite-input
-							if ( $form.find( "label" ).find( "input[name='" + $element.attr( "name" ) + "']" ).length > 0 ) {
-								$error.insertBefore( $form.find( "input[name='" + $element.attr( "name" ) + "']" ) );
+							if ( $form.find( "label" ).find( ".wb-server-error + input.css-implicite-input[name='" + $element.attr( "name" ) + "']" ).length > 0 ) {
+								$error.insertBefore( $form.find( ".wb-server-error + input.css-implicite-input[name='" + $element.attr( "name" ) + "']" ) );
 								return;
 							}
 
--- a/site/pages/docs/ref/formvalid/formvalid-en.hbs
+++ b/site/pages/docs/ref/formvalid/formvalid-en.hbs
@@ -6,7 +6,7 @@
 	"categoryfile": "plugins",
 	"description": "Provides generic validation and error message handling for Web forms.",
 	"altLangPrefix": "formvalid",
-	"dateModified": "2020-12-23"
+	"dateModified": "2022-01-25"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
@@ -55,9 +55,12 @@
 		<li>Add <code>required="required"</code> to each mandatory form field
 			<pre><code>&lt;input id="fname1" name="fname1" type="text" autocomplete="given-name" required="required" /&gt;</code></pre>
 		</li>
-		<li><p><strong>Optional: </strong>For input fields, add one of these <a href="#SpecializedValidation">options</a> for specialized validation, or else you can use a <a href="#pattern">custom pattern validation</a> only if your pattern is very specific and is not part of the options for specialized validation.</p></li>
-		<li><strong>Optional: </strong>For ASP validators add the following attributes to enable the <a href="#MergeSCE">Merge Server-Client errors functionality</a>
-			<p><code>Display="Dynamic" EnableClientScript="false" CssClass="label label-danger wb-server-error"</code></p>
+		<li><strong>Optional:</strong>
+			<ul class="lst-spcd">
+				<li>To make a standalone checkbox field's label look consistent with text field labels (i.e. bolded), add the <code>checkbox checkbox-standalone</code> classes to the checkbox and place it in a <code>&lt;div class="form-group"&gt;</code> element.</li>
+				<li>For input fields, add one of these <a href="#SpecializedValidation">options</a> for specialized validation, or else you can use a <a href="#pattern">custom pattern validation</a> only if your pattern is very specific and is not part of the options for specialized validation.</li>
+				<li>For ASP validators add the following attributes to enable the <a href="#MergeSCE">Merge Server-Client errors functionality</a>:<br /><code>Display="Dynamic" EnableClientScript="false" CssClass="label label-danger wb-server-error"</code></li>
+			</ul>
 		</li>
 	</ol>
 </section>
--- a/site/pages/docs/ref/formvalid/formvalid-fr.hbs
+++ b/site/pages/docs/ref/formvalid/formvalid-fr.hbs
@@ -6,7 +6,7 @@
 	"categoryfile": "plugins",
 	"description": "Effectue la validation de formulaires Web selon un ensemble de règles de base avant qu'ils soient soumis.",
 	"altLangPrefix": "formvalid",
-	"dateModified": "2020-12-23"
+	"dateModified": "2022-01-25"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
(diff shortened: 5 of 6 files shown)


EricDunsworth:
FWIW I ended up deciding against doing horizontal offsetting. IMO it worked nicely for single line labels... but looked out of place in multi-line labels.

Btw ``$asterisk-width`` needs to be set to .668 to accurately represent the wildcard's width (to prevent super long labels from causing wildcards and labels to appear on separate lines). But that didn't work properly in IE11 [since it truncates em units to 2 decimals](https://cruft.io/posts/percentage-calculations-in-ie/). So I increased it a tiny bit more to .67em.

duboisp:
This will need to be discuss from a design perspective, do we want to allow bold/strong text in this context and what is the advantage or is it just a design personal preference.

EricDunsworth:
FWIW here's how WET currently deals with bolding in forms...
* Text field labels are bolded
* Fieldset legends **aren't** bolded (although they look giant - almost like headings)
* Fieldset legends are bolded for checkbox/radio button groups (i.e.``chkbxrdio-grp`` class)
* Checkbox/radio button labels **aren't** bolded (probably because they're often used in fieldsets)

IMO it seems like the mindset behind the status quo was to:
* Make standalone fields use bolded labels to make them stand out
* Do the opposite for sub-fields to de-emphasize them

Going by that logic, I feel like it makes sense to bold the labels of "standalone" checkboxes to stay in line with that way of doing things. But I didn't want to force my opinion onto anyone. So I made this PR add an optional bolding  class. That puts implementers in control of whether or not they want to bold their standalone checkboxes.

@duboisp I forgot to mention this during our discussion earlier, but this PR also changes the location of the error messages that appear near checkbox standalone fields. They now appear as label suffixes (same spot as legends and text fields).

duboisp (changes requested):

src/plugins/formvalid/formvalid.js
```diff
@@ -163,8 +163,8 @@ var componentName = "wb-frmvld",
 
 							//Std If we have a label and the input field is inside the label
 							// need to add a css-implicite-input
-							if ( $form.find( "label" ).find( "input[name='" + $element.attr( "name" ) + "']" ).length > 0 ) {
-								$error.insertBefore( $form.find( "input[name='" + $element.attr( "name" ) + "']" ) );
+							if ( $form.find( "label" ).find( "input.css-implicite-input[name='" + $element.attr( "name" ) + "']" ).length > 0 ) {
+								$error.insertBefore( $form.find( "input.css-implicite-input[name='" + $element.attr( "name" ) + "']" ) );
```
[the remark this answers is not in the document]
duboisp: We will be able to replace that CSS selector by something like : `.wb-server-error + input` which will require to update those line of code and rename the CSS selector here: https://github.com/wet-boew/wet-boew/blob/master/src/plugins/formvalid/_base.scss#L22

Note that additional explanation for the css class `css-implicite-input` as asked in PR #8397 was not submitted. /cc @StdGit

duboisp (changes requested):
Please see the requested changes.

I also find out the server-side validation extension don't work anymore. We would need to investigate it. For the working example, there is an undesirable inline onclick event on the submit button.

EricDunsworth:
@duboisp Regarding https://github.com/wet-boew/wet-boew/pull/9150#discussion_r740495448 (I'm unable to directly reply to it for some reason), would you like me to make that selector change as a part of this PR?

StdGit:
@EricDunsworth Can I help you with something?

duboisp:
> would you like me to make that selector change as a part of this PR?

Yes, unless if that will break the server side validation. @StdGit can you confirm or have objection?

@StdGit the server-side validation feature seem to be currently broken. Do you have a quick idea why? Is it only limited to our static working example?

EricDunsworth:
@StdGit Check out the [OP](https://github.com/wet-boew/wet-boew/pull/9150#issue-707492795), [Pierre's review comments](https://github.com/wet-boew/wet-boew/pull/9150#pullrequestreview-784833144) and [his latest reply](https://github.com/wet-boew/wet-boew/pull/9150#issuecomment-958141386) for extra context.

One of the updates that was made to the form validation plugin for the server side validation feature in #8397 changed the position of error messages in certain situations. It made it so that if an ``input`` element is situated **inside** a ``label``, error messages will appear **before** the input.

That was probably done for the "Beneficiary age" field in the [server-client demo page](https://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-server-en.html) to make the error message appear in-between the field name and the input. That example consists of a ``label`` element that contains a field name, required text and a text input. Makes sense to place the error message before the input in that scenario.

But there were multiple issues with how that was done:

1. Text fields aren't usually placed **inside** labels in WET's form validation plugin. None of the text fields in the [form validation demo page](https://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-en.html) are coded in that manner. Text fields are usually placed **outside** their labels, with both the label and the text input placed within ``<div class="form-group">``.
2. #8397's update didn't restrict its scope to server-client implementations of the form validation plugin. A line comment about the ``css-implicite-input`` class is near the problematic logic, but the logic itself wasn't tied to that class. Plus the class' purpose wasn't documented anywhere.
3. Checkbox and radio buttons are usually placed **inside** their respective labels, **before** field names. #8397's update caused the error message for this PR's standalone checkbox example to appear **before** the checkbox's field's name and required text. This can be seen in the [OP](https://github.com/wet-boew/wet-boew/pull/9150#issue-707492795)'s screenshots section.

To work around item 2, I changed the problematic logic in this PR by tying it to the ``css-implicite-input`` class. That fixed issues with standalone checkboxes. But @duboisp suggested adjusting it further.

Here's what I'd recommend for next steps:
* **Me:**
  * Revise the problematic logic to use something along the lines of the selector @duboisp suggested (``.wb-server-error + input``)
* @StdGit / @duboisp:
  * Look into why the server-side validation feature is currently broken.
* @StdGit:
  * Send in a separate PR to document the purpose of the ``css-implicite-input`` class
  * Consider moving the the text field outside of the label in the [server-client demo page](https://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-server-en.html)'s "Beneficiary age" example. That might necessitate accompanying changes to web apps that use the server-side validation feature.

EricDunsworth:
@duboisp Two questions:

1. Would you be cool if I replaced ``input.css-implicite-input`` with ``.wb-server-error + input.css-implicite-input``? Even though the ``css-implicite-input`` class isn't documented anywhere, the line comments above that jQuery logic would lead me to believe the class was intended for that scenario. So it'd make sense to specify the class somewhere in its selector.
2. You've stated that the server-side validation feature seems to currently be broken. Would it be possible to elaborate? Do you mean it's outright broken, or were you just referring to that jQuery logic we've been discussing? FWIW the [server-client demo page](https://wet-boew.github.io/v4.0-ci/unmin/demos/formvalid/formvalid-server-en.html) looks ok on my end at first glance.

StdGit:
image
this was working before

StdGit:
> Btw `$asterisk-width` needs to be set to .668 to accurately represent the required wildcard's width (to prevent super long labels from causing wildcards and labels to appear on separate lines). But that didn't work properly in IE11 [since it truncates em units to 2 decimals](https://cruft.io/posts/percentage-calculations-in-ie/). So I increased it a tiny bit more to .67em.

Note : I think the border and the asterisk are too close.

image

Also need to be test in a web form project.

EricDunsworth:
> Note : I think the border and the asterisk are too close.

I'd recommend spinning that off into a separate issue since it's unrelated to this PR.

That issue is related to the following factors:
* WET's required asterisks are inserted into padding space - which introduces a risk for that scenario
* WET's legends look giant by default (resulting in much larger asterisks and wider space characters)
* The [server-client demo page](https://wet-boew.github.io/v4.0-ci/unmin/demos/formvalid/formvalid-server-en.html) uses a panel - which causes the asterisk to stick against a dark border as opposed to the page's light grey background colour (resulting in a much starker difference). I don't blame the panel though. The panel's design just makes the issue look more obvious.

I did change the ``$asterisk-width`` variable's value in this PR to fix a layout issue with multi-line labels, but the difference is virtually imperceptible. I've provided before/after screenshots below for reference.

**Before:**
formvalid-server-contact-before

**After:**
formvalid-server-contact-after

> Also need to be test in a web form project.

You and/or @duboisp would probably need to look into that since I don't have a quick/easy way of setting up an ASP.NET environment to test it out myself. The only testing I'm equipped to do is on [WET's server-client demo page](https://wet-boew.github.io/v4.0-ci/unmin/demos/formvalid/formvalid-server-en.html). From what I can tell, this PR's jQuery logic change doesn't break anything on that page.

src/base/forms/_base.scss
```diff
@@ -31,6 +32,22 @@ input[type="checkbox"] {
...
+			color: $text-color;
+		}
+	}
+
+	.checkbox {
+		&.checkbox-bold {
+			label {
+				font-weight: 700;
+			}
+		}
```
duboisp, line 47: That is a new feature.

We will need to validate the use case and test locally.

And that new feature need to be described. It could be a simple paragraph explaining when to use and when not to use.

duboisp (changes requested):
This add a new feature "checkbox-bold" CSS for checkbox.

We need to define the use case and a tangible real life working example.

Todo:
* [x] Document the use case
* [x] Document the new feature
* [x] Test locally

src/base/forms/_base.scss
EricDunsworth: @duboisp What would your thoughts be about dropping the ``checkbox-bold`` class and making something along the lines of ``.wb-frmvld .form-group:not(ul) .checkbox label`` bold by default?

IMO it'd be more elegant to code and in line with the ``form-group`` class' intended use case (its purpose is to [space-out label/fields pairs](https://getbootstrap.com/docs/3.4/css/#forms-example)):
* Tying its selector to the form validation plugin's class (``.wb-frmvld``) would avoid conflicts with [WET](https://wet-boew.github.io/v4.0-ci/demos/tables/tables-fltrs-en.html)/[GCWeb](https://wet-boew.github.io/GCWeb/components/gc-chckbxrdio/gc-chckbxrdio-en.html)'s table filter examples.
* The ``:not(ul)`` selector would prevent it from clashing with the form validation plugin's [stacked checkbox examples](http://localhost/wet-boew/unmin/demos/formvalid/formvalid-en.html#fruits1a).

The downside is that I don't know if any pre-existing forms are already doing anything that might get picked up by my selector. Might need to something along the lines of https://github.com/wet-boew/wet-boew/pull/9147#issuecomment-956577561 to find out.

duboisp:
@EricDunsworth please provide a concrete example where a such checkbox bold is needed. It don't need to be public facing.

EricDunsworth:
@duboisp Here's one I managed to dig up just now:
* https://www.canada.ca/en/employment-social-development/services/foreign-workers/report-abuse/tool.html (check the very end of the form)

I believe a few other real-world examples are out there, but haven't tried digging them up as of yet. Not sure if any of the others have shorter labels.

duboisp:
Thanks @EricDunsworth 

We got a maybe related request under the reference number: MWS-2422

EricDunsworth:
@duboisp @GormFrank I just tried out MWS-2422's form in this PR's branch and everything was fine in Firefox + Edge (Chromium). Empty required checkboxes didn't shift positions upon trying to submit the form.

The cause of that layout issue was due to the form validation plugin placing the error messages at the beginning of the checkbox labels (**before** the checkbox inputs). This PR moves the error messages to the end of the labels (**after** the checkbox inputs). So the issue no longer occurs at all.

**PS:**
IMO that form wasn't coded very well. I gave @GormFrank an overview of various issues I noticed in it.

src/plugins/formvalid/formvalid-en.hbs
```diff
@@ -351,6 +351,23 @@
...
+					<label for="agree1"><input id="agree1" name="agree1" type="checkbox" required> <span class="field-name">I agree</span> <strong class="required">(required)</strong></label>
+				</div>
+			</div>
+			<details class="mrgn-bttm-md">
+				<summary>View code</summary>
+				<pre><code>&lt;div class="wb-frmvld"&gt;
+	&lt;form action="#" method="get" id="validation-example"&gt;
+		...
+			&lt;div class="form-group"&gt;
+				&lt;div class="checkbox checkbox-bold required"&gt;
```
GormFrank, line 365: @EricDunsworth Please try to come up with a name based on the function and not the effect, such as: `checkbox-standalone`

GormFrank (dismissed):

EricDunsworth:
FYI I was unable to get around to updating this PR this week :(. Won't be able to get back to it until mid/late-January.

@duboisp @GormFrank Feel free to push extra commits as you see fit or divert the JS logic change to a separate PR if push comes to shove for MWS-2422.

src/plugins/formvalid/formvalid-en.hbs
EricDunsworth: Just forced-pushed an update that renamed it from ``checkbox-bold`` to  ``checkbox-standalone``. Meant to do it back in December but ran out of time lol. Still need to get to documentation at some point.

src/base/forms/_base.scss
EricDunsworth: I've updated the documentation to try addressing the first two todos. Tried to keep it short and sweet.

duboisp (approved):
Approved, to be merged after successful local testing

duboisp:
I tested locally and it works as expected

PR #9159, opened 2021-08-30
ESLint: Allow IE11-friendly ES6

EricDunsworth:
* Set ESLint's environment to ES6 to prevent ``let``/``const``/``Map`` from triggering errors
* Use eslint-plugin-es5 to disallow most of ES6's other features
  * Relies on a fork (nkt/eslint-plugin-es5#39) to exempt jQuery's ``find()`` method
    * The exemption also covers ES6's native ``find()`` methods (which are unsupported by IE11...)
* Use eslint-plugin-ie11 to detect unsupported use cases in IE11
* Related to #9146

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,10 +1,16 @@
 {
+  "plugins": [
+    "es5",
+    "ie11"
+  ],
   "extends": [
     "eslint:recommended",
-    "jquery"
+    "jquery",
+    "plugin:es5/no-es2015"
   ],
   "env": {
     "browser": true,
+    "es6": true,
     "jquery": true
   },
   "globals": {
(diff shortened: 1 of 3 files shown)


EricDunsworth:
[Geomap's lib dependency](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/deps/geomap-lib.js) uses some ES6 features that are unsupported by IE11 (``values`` and ``includes``). So it's getting caught by this PR's ESLint settings (as it should be) and consequently causing the build to fail.

@GormFrank @jvanulde Thoughts?

EricDunsworth:
After discussing with @duboisp earlier, it looks like there isn't much of a point in moving forward with this kind of ESLint setup due to the following:
* IE11 is at death's door
* There isn't any appetite to ensure WET 4.0.x remains backwards compatible with IE11 going forward
* Revising this PR to produce IE11 warnings would be kind of pointless and bloaty

I'll send in another PR to supersede this one. It'll just change ESLint's environment to ES6 to allow #9146 to start passing the build.

**PS:**
I'll keep this PR as-is for posterity.

PR #9166, opened 2021-09-09
ESLint: Set environment to ES6

EricDunsworth:
* Prevents ``let``/``const``/``Map``/etc from triggering errors
* Supersedes and closes #9159
* Related to #9146

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,6 +5,7 @@
   ],
   "env": {
     "browser": true,
+    "es6": true,
     "jquery": true
   },
   "globals": {


EricDunsworth:
Btw [ESLint's environment](https://eslint.org/docs/user-guide/configuring/language-options#specifying-environments) can be set even higher than ES6. But I only need ES6 to be able to move forward with #9146.

I could've technically gone all the way up to ES2021, but that might've potentially allowed features that are unsupported in [modern-ish browsers that WET still supports](https://wet-boew.github.io/wet-boew-documentation/decision/2.html). Like Safari or enterprise versions of Edge/Firefox.

duboisp (approved):

PR #9169, opened 2021-09-13
add basic html notification section in a container class

polmih:
add basic html notification section in a container class adjust the user experience (confirmed with Francis).
current-display

with-container

### DIFF
--- a/src/base/accessibility/_base.scss
+++ b/src/base/accessibility/_base.scss
@@ -44,9 +44,6 @@
 }
 
 .wb-disable {
-	#wb-tphp {
-		background: #fff;
-	}
 
 	.wb-slc {
 		position: static;
@@ -67,6 +64,19 @@
 
 	#wb-dtmd {
 		float: none !important;
+	}
+
+	#wb-tphp + section h2 {
+		margin-left: 1.3em;
+
+		&::before {
+			color: $label-alert-warning-border-icon-color;
+			content: "\e107";
+			display: inline-block;
+			font-family: "Glyphicons Halflings";
+			margin-left: -1.3em;
+			position: absolute;
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


duboisp (changes requested):
Can you try to remove the alert. The alert pattern is designed to be used inside the page content area, not outside like before the page header. As an alternate solution, can you tryto add CSS to style it like a yellow background banner, like bg-warning, and keep the semantic markup.

duboisp (commented):
Try to improve the selector instead of using a new CSS class,

Like:
```
.wb-disable #wb-tphp .bg-warning h2::before
```

duboisp (changes requested):
Review the CSS selector

duboisp:
as discussed @GormFrank please check if we still avoid this failure: https://www.w3.org/WAI/WCAG22/Techniques/failures/F24

src/base/accessibility/_base.scss
```diff
@@ -68,6 +65,15 @@
...
 	}
+
+	#wb-tphp + section.bg-warning h2::before{
+		color: $label-alert-warning-border-icon-color;
+		content: "\e107";
+		display: inline-block;
+		font-family: "Glyphicons Halflings";
+		margin-left: -1.3em;
+		position: absolute;
+	}
```
GormFrank, line 76: As discussed:
```suggestion
	#wb-tphp + section.bg h2 {
		margin-left: 1.3em;
	
		&::before {
			color: $label-alert-warning-border-icon-color;
			content: "\e107";
			display: inline-block;
			font-family: "Glyphicons Halflings";
			margin-left: -1.3em;
			position: absolute;
		}
	}
```

GormFrank (changes requested):

src/base/accessibility/_base.scss
polmih: Done, thanks

GormFrank (approved):

duboisp (approved):

PR #9184, opened 2021-09-27
Data table date-picker - Fix result set when time not include

### LINKED ISSUE #9157: [BUG] Data table using date picker filter - incorrect result set when column does not contain a time
## Required information

**Describe the bug**
When a date picker input is used to filter results in a data table and the target column for the filter only contains a date in ISO 8601 format (YYYY-MM-DD), the result set is incorrect.

**To Reproduce**
Steps to reproduce the behaviour:

1. Rename date-filter-bug-data-table.txt to .html and open
2. Use the following two test scenarios

Note 1: The filter works in the first example. I believe this is due to the column containing a date AND time, which is being properly processed by the new min-max date range feature released here (line 295-327):
https://github.com/wet-boew/wet-boew/commit/fd309d95efb52187b7dafd8f28a1b95063db816f#diff-d71dbbb387c0890b97d1b1f67b2552fc39769583a602a94c08626b850f181035

Note 2: Following examples are taken from the first table example here: https://wet-boew.github.io/v4.0-ci/demos/tables/tables-fltrs-en.html. This example was modified/simplified to create the following 2 examples; the functionality (number of filters) and data set have been trimmed and data set is static (as opposed to ajaxSource).

Using example 1, steps that show working functionality (target column contains date and time):
1.	Filter using date 2021-07-24
PASS: no results

2.	Filter using date 2021-07-25
PASS: 1 result with correct date

3.	Filter using date 2021-07-26
PASS: 1 result with correct date

4.	Filter using date 2021-07-27
PASS: no results 

Using example 2, steps to reproduce the issue  (target column only contains date):
1.	Filter using date 2021-07-24
FAIL: 1 result with date 2021-07-25

2.	Filter using date 2021-07-25
FAIL: 1 result with date 2021-07-26

3.	Filter using date 2021-07-26
FAIL: no results

**Expected behaviour**
The date filter must return the correct results when the target column only contains a date in ISO 8601 format (YYYY-MM-DD). The target column must not require a time.

**Who shall do the work?**
I am asking for Principal publisher to please do the work

[date-filter-bug-data-table.txt](https://github.com/wet-boew/wet-boew/files/7021241/date-filter-bug-data-table.txt)

Ricokola:
When using table date-picker, results were incorrect when there was no time set in the data set.

Closes #9157
Related to: WET-193

### DIFF
--- a/src/plugins/tables/ajax/datasource.json
+++ b/src/plugins/tables/ajax/datasource.json
@@ -2,7 +2,7 @@
   "data": [{
     "TITLE": "<a href=\"http://news.gc.ca/web/index-en.do\">Biomass Program Continues to Support Transition to Renewable Energy in Manitoba</a>",
     "TEASER": "The governments of Canada and Manitoba are supporting a greener, more sustainable economy through the $1 million Biomass Energy Support Program, Federal Agriculture Minister Lawrence MacAulay and Manitoba Agriculture Minister Ralph Eichler announced today.",
-    "PUBDATE": "2016-07-29 12:07:00",
+    "PUBDATE": "2016-07-29",
     "TYPE": "News Releases",
     "DEPT": "Agriculture and Agri-Food Canada",
     "SUBJECT": "Agriculture",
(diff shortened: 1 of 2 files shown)


duboisp (changes requested):
Can you add a working example that don't contains time in the date?

Please remove the 2 unrelated commits.

duboisp (approved):

PR #9194, opened 2021-10-19
Details/Sum: Utility to limit width to max content

GormFrank:

Mostly used for Add to calendar at the moment. Details/summary with the .max-content class will have its width take up to the maximum of its own content and not stretch to the entirety of the container.


duboisp (changes requested):
Can you add a one liner near the working example of addCalendar?

duboisp (approved):

PR #9199, opened 2021-10-22
Adjust the structured data for breadcrumb

polmih:
Previous breadcrumb used typeof=”Breadcrumb”, which doesn’t offer a data structure as rich as the typeof=”BreadcrumbList”.
This PR will replace it with the typeof=”BreadcrumbList” approach, which includes things such as a typeof ListItem on each list items.

Here is a valid example: https://search.google.com/test/rich-results/result?id=zGU8HNwVFWXt4fBPF2cl8w

As per WET-148


duboisp (changes requested):
Once the requested change is completed, we only need to run a quick test locally to ensure it still work as expected.

duboisp (approved):

PR #9200, opened 2021-10-22
Data Ajax: Add the plugin only once

EricDunsworth:
* Fixes #9198
* Partially-reverts #6076


duboisp:
@EricDunsworth considering the recent changes, can you retest to see if you still have the issue?

EricDunsworth:
@duboisp Retried #9198's repro steps and got the same outcome as before :(.

PR #9204, opened 2021-11-03
Merge server-client errors: Remove noscript block

EricDunsworth:
Issues with what was removed:
* Was only present in the English demo page
* Duplicated the errors alert (which is already visible to all users in JS/basic HTML/noscript modes)
* Used escaped HTML tags (wasn't presented to users as rendered HTML)

CC @StdGit @duboisp

### DIFF
--- a/src/plugins/formvalid/formvalid-server-en.hbs
+++ b/src/plugins/formvalid/formvalid-server-en.hbs
@@ -7,7 +7,7 @@
 	"tag": "formvalid-server",
 	"parentdir": "formvalid",
 	"altLangPrefix": "formvalid-server",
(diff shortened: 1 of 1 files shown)


EricDunsworth:
@StdGit Are you ok with this change? Does the noscript block need to be kept for any reason?

If it's decided to keep the noscript block, I could revise this PR to fix the escaped HTML markup and port it into the French demo.

But I'd need proper French content for this part:
> This is user control wbDisplayErrors (It appears when JavaScript is disabled)

StdGit:
I wiil have a look tonight

duboisp (approved):

PR #9210, opened 2021-11-06
core: jQuery fixes

duboisp:
Apply fixes to jQuery and ensuring the HTML is sanitized

## Test case for the jQuery fixes bellow

Test sites with the jQuery fixes: http://universallabs.org/wet/labs/jquery-fixes/demos/index-en.html

### Test: `jQuery()`
Run in your browser console
```js
$( "<p>abc<iframe//src=jAva&Tab;script:alert(3)>def</p>" ).appendTo( "#wb-cont" );
```
**result:** No iframe inserted around the heading level 1

### Test:
`jQuery.html()`
`jQuery.append()`
`jQuery.prepend()`
`jQuery.before()`
`jQuery.after()`

* Use the data-ajax demo page: `/demos/data-ajax/data-ajax-en.html`
* Insert the following in the ajaxed in file: `<p>abc<iframe//src=jAva&Tab;script:alert(3)>def</p>`
* Check if the dirty HTML are cleaned for the example: "append", "prepend", "before", "after"
* The `data-ajax-replace` will also test the `jQuery.html()` function.

### Test: jQuery.replaceWith()
Run in your browser console
```js
$("#wb-cont").replaceWith( "<p>abc<iframe//src=jAva&Tab;script:alert(3)>def</p>" );
```
**result:** No iframe inserted around the heading level 1

### Test: jQuery.parseHTML()
Run in your browser console, pass unsafe HTML and it should return sanitized HTML
```js
jQuery.parseHTML( "<p>abc<iframe//src=jAva&Tab;script:alert(3)>def</p>" );
```
**result:** return an array with 1 element, `[ p ]`

### Test: jQuery.htmlPrefilter()
	- Run in your browser console
```js
jQuery.htmlPrefilter( "<tr/>" );
```
**result:** return `<tr/>`

### Test: jQuery.extend
	- Run in your browser console
```js
var dataFromAPI = JSON.parse( '{ "new": "property", "__proto__": { "polluted": "true" } }' );
var myObject = jQuery.extend( '{ "new": "property" }', dataFromAPI );
console.log( myObject.polluted );
```
**result:** `undefined`

## Note
* We removed the test for the feedback form because it was only for a demo purpose, was never integrated in wet-boew and its unit testing didn't worked because of the insertion of the HTML for the whole web page.

## Change required in your code if you it is depending on wet-boew

* Ensure there is no JSONP ajax call
* Ensure to not use self closing tag when a closing tag is usually needed.
ex: `<div />` need to change for `<div></div>`
This is applicable to all tags except the one in this list: area|br|col|embed|hr|img|input|link|meta|param
 * Review how `<tr>`, `<td>`, `<script>` and `<link>` are inserted with `$()`.
 Their insertion need to be completed differently like the by using the javascript DOM interface.
 * All content that related to jQuery DOM manipulation are sanitized with DOMPurify
 * You can't use jQuery for parsing XML document, you can use DOMParser() as an alternative
 * Ensure AJAX and fetch are sanitized before they response are used


GormFrank (changes requested):
[Line 649](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/calendar/test.js#L649) of **src/plugins/calendar/text.js** should be changed from:
```
inRange.wrap( "<a href='javascript:;' tabindex='-1'></a>" ).filter( ":eq(0)" ).parent().removeAttr( "tabindex" );
```
to
```
inRange.wrap( "<a tabindex='-1'></a>" ).filter( ":eq(0)" ).parent().removeAttr( "tabindex" );
inRange.parent().attr( "href", "javascript:;" );
```

duboisp:
Thanks @GormFrank

GormFrank (approved):
Love the approach with the jQuery fix!

PR #9222, opened 2021-11-12
Ajax-fetch - fetchData pointer option was empty

duboisp:
Fix a issue that prevented the megamenu to load when the data-ajax-replace is used with the mega menu plugin.

Note: The data-ajax-replace plugin work fine when it is used inside page content.


GormFrank (approved):

GormFrank:
This is a patch

PR #9231, opened 2021-11-24
Table filters - Fix date filter issue

Ricokola:
Related to: WET-214

Fixing the issue when filtering using a date from the date-picker for which there is no data available and still the search results is providing data that are dated between the 10 and 19 of every months.

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -229,6 +229,7 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 	var $prevCol = -1, $cachedVal = "";
 	$form.find( "[name]" ).each( function() {
 		var $elm = $( this ),
+			$val = $elm.val(),
 			$value = "",
 			$regex = "",
 			$column = parseInt( $elm.attr( "data-column" ), 10 ),
@@ -253,7 +254,7 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 		if ( $elm.is( "select" ) ) {
 			$value = $elm.find( "option:selected" ).val();
 		} else if ( $elm.is( "input[type='number']" ) ) {
-			var $val = $elm.val(), $minNum, $maxNum;
+			var $minNum, $maxNum;
 
 			// Retain minimum number (always the first number input)
 			if ( $cachedVal === "" ) {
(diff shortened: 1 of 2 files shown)


duboisp (changes requested):
as discussed, please apply the change for when the field value is retrieved in the other conditional, like number filter. And, as you mentioned, update the other default that seems to present the same kind of issue.

GormFrank:
To be tested locally and insure French working example also benefits from the extra entries in the table.

src/plugins/tables/tables.js
```diff
@@ -253,17 +253,17 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 		if ( $elm.is( "select" ) ) {
 			$value = $elm.find( "option:selected" ).val();
 		} else if ( $elm.is( "input[type='number']" ) ) {
-			var $val = $elm.val(), $minNum, $maxNum;
```
duboisp, line 256: Please keep the variable caching.

duboisp (changes requested):

duboisp (changes requested):
Change line 239, for number, similarly as you did for date.

src/plugins/tables/tables.js
```diff
@@ -253,12 +254,12 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
...
+			var $minNum, $maxNum;
 
 			// Retain minimum number (always the first number input)
 			if ( $cachedVal === "" ) {
 				$cachedVal = parseFloat( $val );
-				$cachedVal = ( $cachedVal ) ? $cachedVal : "-0";
+				$cachedVal = ( $cachedVal ? $cachedVal : $val );
```
duboisp, line 262: Humm, we will need to better understand the use of "-0" here and how it used later. May be we can look at the history to find out more information about the use case of using the value "-0"
 
If the value of in the input don't parse, then we might just fallback on a simple text search matching the input value.

thekodester: I originally coded a lot of these plugins. Quite a few of these changes are completely unrelated to the date filter and are related the the number filter.

Does the original poster have example data that can be used to show when the date filter breaks. If so I can see if I’m able to run some test.

thekodester: Yes for the date filter function yes I can see that issue. But the code you are changing on this line is unrelated to the date filter.

duboisp (approved):

GormFrank (approved):

PR #9255, opened 2022-01-19
helpers: add a function to identify most common PII in a string

polmih:

helper function that will required a string as first parameter and will find potential PII.
The second parameter is optional and if  set to `true`, this function will remove the identified PII from that string and will return the modified string.
If the second parameter is set to `false`, this function will return `true `if any PII found, false otherwise.

### DIFF
--- a/src/core/helpers.js
+++ b/src/core/helpers.js
@@ -1250,6 +1250,41 @@ wb.escapeAttribute = function( str ) {
 	return str.replace( /'/g, "&#39;" ).replace( /"/g, "&#34;" );
 };
 
+/*
+* Find most common Personal Identifiable Information (PII) in a string and return either the cleaned string either true/false
+* @param {string} str
+* @param {boolean} toClean
+* @return {string | true | false}
+* @example
+* wb.findPotentialPII( "email:test@test.com, phone:123 123 1234", true )
+* returns "email:, phone:",
+* wb.findPotentialPII( "email:test@test.com, phone:123 123 1234", false )
+* returns true
+*/
+wb.findPotentialPII = function( str, toClean ) {
+
+	if ( typeof str  !== "string" ) {
+		return false;
+	}
+	var regEx = [
+			/\b(?:\w*[\s\\.-]*?\d[\s\\.-]*?){5,}\b/ig, //5digits or more pattern
+			/\b[A-Za-z]\d[A-Za-z][ -]?\d[A-Za-z]\d\b/ig, //postal code pattern
+			/\b([a-zA-Z0-9_\-\\.]+)@([a-zA-Z0-9_\-\\.]+)\.([a-zA-Z]{2,5})\b/ig //email pattern
+		],
+		isFound = false;
+
+	for ( var key in regEx ) {
+		if ( str.match( regEx[ key ] ) ) {
+			isFound = true;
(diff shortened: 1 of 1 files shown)


duboisp (changes requested):
Approved after the requested change is completed

src/core/helpers.js
```diff
@@ -1250,6 +1250,41 @@ wb.escapeAttribute = function( str ) {
...
+* @param {string} str
+* @param {boolean} toClean
+* @return {string | true | false}
+* @example
+* wb.catchKnownPII( "email:test@test.com, phone:123 123 1234", true )
+* returns "email:, phone:",
+* wb.catchKnownPII( "email:test@test.com, phone:123 123 1234", false )
+* returns true
+*/
+wb.catchKnownPII = function( str, toClean ) {
```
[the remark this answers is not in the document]
polmih: Done. Thanks

duboisp (approved):

src/core/helpers.js
```diff
@@ -1250,6 +1250,41 @@ wb.escapeAttribute = function( str ) {
...
+* wb.findCommonPII( "email:test@test.com, phone:123 123 1234", false )
+* returns true
+*/
+wb.findCommonPII = function( str, toClean ) {
+
+	if ( typeof str  !== "string" ) {
+		return false;
+	}
+	var regEx = [
+			/\b(?:\w{0,2}[\s\\.-]*?\d[\s\\.-]*?){5,}\b/ig, //5digits or more pattern
```
delisma, line 1270: This regex may return false positive.
We should have specific regex for specific number patterns 
The other thing if letters and numbers are submitted together the numbers won't be detected (i.e. tel5142825555)

CC @HamzaAburaneh

delisma (changes requested):

src/core/helpers.js
duboisp: @polmih can you adapt the regex to support the test case of delisma where it don't need start with a space.

src/core/helpers.js
```diff
@@ -1250,6 +1250,41 @@ wb.escapeAttribute = function( str ) {
...
+* @param {string} str
+* @param {boolean} toClean
+* @return {string | true | false}
+* @example
+* wb.findCommonPII( "email:test@test.com, phone:123 123 1234", true )
+* returns "email:, phone:",
+* wb.findCommonPII( "email:test@test.com, phone:123 123 1234", false )
+* returns true
+*/
+wb.findCommonPII = function( str, toClean ) {
```
duboisp, line 1264: findProbableCommonPII ?
findPotentialPII ?

duboisp (changes requested):

src/core/helpers.js
duboisp: ex: `\b(?:\w{0,}[\s\\.-]*?\d[\s\\.-]*?){5,}\b`

polmih: changed. It will take in consideration any word characters 0 or multiple times at the beginning of the 5 digits

src/core/helpers.js
polmih: adjusted. Thanks

duboisp (approved):

PR #9272, opened 2022-02-07
Core: Set WET's lang to first 2 chars of page lang

### LINKED ISSUE #8842: Language interpolation fails scripts loading
When loading a page with the `lang="en-CA"` set, CDTS trys to load scripts that don't exist.

I had set the default language attribute to `en-Ca` on my page in the `html` element like:
```html
<html lang="{{ site.lang | default: 'en-Ca' }}" xmlns="http://www.w3.org/1999/xhtml" class="no-js" dir="ltr">
```

Which ended up returning this issue:
> SEC7112: Script from https://www.canada.ca/etc/designs/canada/cdts/gcweb/v4_0_32/js/i18n/en-Ca.min.js was blocked due to mime type mismatch
> 2 HTTP404: NOT FOUND - The server has not found anything matching the requested URI (Uniform Resource Identifier).
GET - https://www.canada.ca/etc/designs/canada/cdts/gcweb/v4_0_32/js/i18n/en-Ca.min.js

Which causes features in WET to stop working.

I think it would make sense to have some interpolation on the `lang` element to know that if `en-CA` is entered the `en.min.js` script should be loaded; not the `en-CA.min.js`.

EricDunsworth:
Certain web app platforms (e.g. Microsoft PowerApps) can only be configured to use country language codes. So they end up generating HTML elements that use lang attributes along the lines of "en-CA" or "fr-CA". By comparison, most of WET's built-in i18n languages use two-letter language codes such as "en" or "fr".

That's a recipe for disaster. WET sets its language based on the current page's HTML lang attribute, then tries loading a namesake i18n file. If WET fails to load its i18n file, plugins don't get initialized. So users end up with broken pages.

This change reconciles the differences by deriving WET's language from the HTML lang attribute's first two characters, unless:
* The page's language is built-into WET and uses a longer language code (i.e. "pt-BR" and "zh-Hans")
* WET's ``script`` element (e.g. wet-boew.js or wet-boew.min.js) contains both of the following attributes:
  * ``data-wb-core`` (doesn't need a value)
  * ``data-lang-long="[longer language codes, space-separated]"`` (in case a site wants to use custom i18n files for longer language codes)

A core documentation page has also been added to document how WET will handle language codes going forward.

Fixes #8842.

Co-authored-by:
* @fsnoddy
* @duboisp

### DIFF
--- a/site/pages/docs/ref/core/core-en.hbs
+++ b/site/pages/docs/ref/core/core-en.hbs
@@ -1,0 +1,68 @@
+---
+{
+	"title": "WET core",
+	"language": "en",
+	"category": "Other",
+	"categoryfile": "other",
+	"description": "Core WET object.",
+	"altLangPrefix": "core",
+	"dateModified": "2022-04-20"
+}
+---
+<div class="wb-prettify all-pre hide"></div>
+
+<div class="alert alert-warning mrgn-tp-lg">
+	<h2>Work in progress</h2>
+	<p>This page is a work in progress.</p>
+</div>
+
+<section>
+	<h2>Purpose</h2>
+	<p>Core WET object.</p>
+</section>
+
+<section>
+	<h2>Language (<code>wb.lang</code>)</h2>
+	<p>WET's language is derived from the <code>html</code> element's <code>lang</code> attribute. A formatted version of the page's language code is stored in the <code>wb.lang</code> variable.</p>
+	<p>By default, the variable's value will be derived from the first two characters of the page's language code. If a longer code is specified, it'll be re-used as the variable's value so long as it corresponds to one of WET's built-in languages (i.e. <code>pt-BR</code> for Portuguese [Brazilian] or <code>zh-Hans</code> for Chinese [Simplified]).</p>
+	<p>Implementations that want to provide custom i18n files for longer language codes (e.g. <code>en-CA</code>, <code>en-US</code>, etc) can do so by adding <code>data-wb-core</code> and <code>lang-lang-long=""</code> attributes to WET's <code>script</code> element.</p>
+	<p>WET will also attempt to load an i18n file named after the variable. Plugins won't initialize if the i18n file fails to load.</p>
+</section>
+
+<section>
+	<h2>Configuration</h2>
+	<div class="table-responsive">
+		<table class="table">
+			<thead>
+				<tr>
+					<th>Type</th>
+					<th>Option</th>
+					<th>Description</th>
+					<th>How to configure</th>
+					<th>Values</th>
+				</tr>
+			</thead>
+			<tbody>
+				<tr>
+					<td>HTML attribue</td>
+					<td><code>data-wb-core</code></td>
+					<td>Customizes WET's core variables. Use with other <code>data-*</code> attributes.</td>
+					<td>Add a <code>data-wb-core</code> attribute to WET's <code>script</code> element</td>
+					<td>None</td>
+				</tr>
+				<tr>
+					<td>HTML attribue</td>
+					<td><code>data-lang-long</code></td>
+					<td>Extends <code>wb.lang</code>'s support for long language codes (&gt; 2 characters). Use with the <code>data-wb-core</code> attribute.</td>
+					<td>Add a <code>data-wb-core</code> attribute to WET's <code>script</code> element</td>
+					<td>Space-separated <a href="https://www.rfc-editor.org/info/bcp47" rel="external">BCP 47</a> language codes</td>
+				</tr>
+			</tbody>
+		</table>
+	</div>
+</section>
+
+<section>
+	<h2>Source code</h2>
+	<p><a href="https://github.com/wet-boew/wet-boew/tree/master/src/core">WET core source code on GitHub</a></p>
+</section>
--- a/src/core/wb.js
+++ b/src/core/wb.js
@@ -57,9 +57,33 @@ var getUrlParts = function( url ) {
 
 	/**
 	 * @variable i18n
-	 * @return {string} of HTML document language
+	 * @return {string} of WET document language
 	 */
-	lang = document.documentElement.lang,
+	lang = ( function( ele ) {
+		let lang = document.documentElement.lang;
+		const shortLangLength = 2;
+
+		// Perform extra checks if the page uses a long language code
+		if ( lang.length > shortLangLength ) {
+			let longLangs = [ "pt-BR", "zh-Hans" ]; // Built-in long language codes
+
+			// Check if any custom long language codes have been specified
+			// Specify by adding data-wb-core and data-lang-long="en-CA en-US etc" attributes to WET's script element (e.g. wet-boew.js or wet-boew.min.js)
+			if ( ele[ 0 ].hasAttribute( "data-wb-core" ) &&  ele[ 0 ].hasAttribute( "data-lang-long" ) ) {
+				const longLangsCustom = ele.attr( "data-lang-long" ).split( " " );
+
+				// Add extra language codes to the beginning of the longLangs array to match them more quickly
+				longLangs = longLangsCustom.concat( longLangs );
+			}
+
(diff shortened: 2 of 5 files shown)


EricDunsworth:
@duboisp @GormFrank Any quick thoughts on this?

If you're cool with the premise, I'll update the PR to document the ``data-wb-lang`` attribute.

duboisp (changes requested):
It will be preferable to not have that parameter be configurable. I prefer to extract from the HTML element instead.

As discussed, see this suggestion
```js
var lang = document.root.lang;

var whitelistLang = [ "pt-BA", "zh-HANZ" ];

whitelistLang = &.extend( whitelistLang, window[ "wb-core-whitelist" ], $src]

if ( lang.indexOf( "-" ) && !whitelistLang.indexOf( lang ) ) {
	// Clean the language first section.
      // LANG official syntax see ref. https://www.rfc-editor.org/rfc/rfc5646.html#section-2.1
	
}

// Continue...

```

EricDunsworth:
@duboisp FYI I've reworked this PR using your suggestion as a starting point and added you as a co-committer :).

As far as configuration goes, I mimicked what ``wb.paths`` does with its``data-*`` attributes. So if an implementer wants to add custom i18n files that use longer language codes, they'll need to add ``data-wb-core`` and ``data-lang-long="en-CA fr-CA etc"`` attributes to any ``script`` element within their site's pages.

The only other pitfalls I can think of are as follows:
* None of this functionality is documented outside of this PR and wb.js' source code
* ``pt-BR`` and ``zh-Hans`` are hardcoded in wb.js:
  * That could potentially be revised to be generated via a grunt task and inserted into the built version of wb.js similar to what used to be done with "WET_BOEW_VERSION_MATHJAX" getting replaced with a real version number. But IMO that'd be overkill since WET's i18n languages haven't changed in ages.

GormFrank:
@EricDunsworth Can you provide an HTML blurb using this data attribute (as some sot of a working example/how to implement kind of thing)?

EricDunsworth:
I've drafted some preliminary documentation below :).

If you're cool with it, give me a heads up and I'll provide an HTML version of it. I might also edit this comment if I decide to make any revisions.

---

## Language (``wb.lang``)

WET's language is derived from the ``html`` element's ``lang`` attribute. A formatted version of the page's language code is stored in the ``wb.lang`` variable.

By default, the variable's value will be derived from the first two characters of the page's language code. If a longer code is specified, it'll be re-used as the variable's value so long as it corresponds to one of WET's built-in languages (i.e. ``pt-BR`` for Portuguese [Brazilian] or ``zh-Hans`` for Chinese [Simplified]).

Implementations that want to provide custom i18n files for longer language codes (e.g. ``en-CA``, ``en-US``, etc) can do so by adding ``data-wb-core`` and ``lang-lang-long=""`` attributes to WET's ``script`` element.

WET will also attempt to load an i18n file named after the variable. Plugins won't initialize if the i18n file fails to load.

## Configuration

| Type | Option | Description | How to configure | Values |
| -- | -- | -- | -- | -- |
| HTML attribue | ``data-wb-core`` | Customizes WET's core variables. Use with other ``data-*`` attributes. | Add a ``data-wb-core`` attribute to WET's ``script`` element | None |
| HTML attribue | ``data-lang-long`` | Extends ``wb.lang``'s support for long language codes (containing more than 2 characters). Use alongside the ``data-wb-core`` attribute. | Add a ``data-wb-core`` attribute to WET's ``script`` element | ``""``:<br />Space-separated language codes |

EricDunsworth:
Just force-pushed a few more adjustments:
* Update JSDoc's @return annotation to reflect that it returns WET's document language (as opposed to that of the HTML document)
* Fixed a JS error that occurred if ``data-wb-core`` was used without ``data-lang-long``

src/core/wb.js
```diff
@@ -57,9 +57,33 @@ var getUrlParts = function( url ) {
 
 	/**
 	 * @variable i18n
-	 * @return {string} of HTML document language
+	 * @return {string} of WET document language
```
EricDunsworth, line 60: Would this count as a major change?

If so, should this PR be reworked to use something along the lines of a ``wb.wbLang`` variable and make WET use it instead of ``wb.lang``?

src/core/wb.js
```diff
@@ -57,9 +57,33 @@ var getUrlParts = function( url ) {
...
 	 */
-	lang = document.documentElement.lang,
+	lang = ( function( ele ) {
+		let lang = document.documentElement.lang;
+		const shortLangLength = 2;
+
+		// Perform extra checks if the page uses a long language code
+		if ( lang.length > shortLangLength ) {
+			let longLangs = [ "pt-BR", "zh-Hans" ]; // Built-in long language codes
```
EricDunsworth, line 68: Regarding the request in last week's technical review meeting to move this variable's declaration near here: https://github.com/wet-boew/wet-boew/blob/12d829e8a12940be2c0c89e80898e47e151efb1f/src/core/wb.js#L145

I don't think that'd be feasible. The ``lang`` function runs before ``wb`` gets declared. So if I were to move ``longLangs`` down there, it would be unavailable at the moment it's needed the most. Or maybe I'm missing something?

duboisp: How the default setting is set in the core is not consistent. So it is fine to have it here.

src/core/wb.js
duboisp: As long we don't need to modify existing code that call `wb.lang` don't need to pass a parameter.

duboisp (commented):
Next step, is to do the testing:
* [x] Ensure that a plugin that use `wb.lang` continue to work as is
* [x] Test is a experimental page that have the default language set to "fr-CA" and "en-CA" and "en-US"
* [x] Ensure we can add a new long lang. For this test, we would need to manually create a such language by modifying an existing "i18n/lang.js" file
* [x] Ensure we are providing the instruction on how to extend for additional language. At least have a PR that will publish those instruction if not included in this PR.

Once that testing is completed, we will be good to approve and then merge.

EricDunsworth:
Just force-pushed the documentation updates:
* Added a no-frills core documentation page that regurgitates what I posted in https://github.com/wet-boew/wet-boew/pull/9272#issuecomment-1079474468
* Added a link to it in the [other components page](https://wet-boew.github.io/v4.0-ci/unmin/docs/ref/other-en.html) (might not be the best place for it - but I wouldn't consider it to be a plugin either)

duboisp (dismissed):

site/pages/docs/ref/core/core-en.hbs
```diff
@@ -0,0 +1,66 @@
...
+					<td>Customizes WET's core variables. Use with other <code>data-*</code> attributes.</td>
+					<td>Add a <code>data-wb-core</code> attribute to WET's <code>script</code> element</td>
+					<td>None</td>
+				</tr>
+				<tr>
+					<td>HTML attribue</td>
+					<td><code>data-lang-long</code></td>
+					<td>Extends <code>wb.lang</code>'s support for long language codes (&gt; 2 characters). Use with the <code>data-wb-core</code> attribute.</td>
+					<td>Add a <code>data-wb-core</code> attribute to WET's <code>script</code> element</td>
+					<td><code>""</code>:<br />Space-separated language codes</td>
```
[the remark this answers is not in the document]
EricDunsworth: Ended up changing this to:
> Space-separated [BCP 47](https://www.rfc-editor.org/info/bcp47) language codes

EricDunsworth:
Just force-pushed a few more minor docs updates :). I don't believe there's anything further that needs to be done from my end at this point.

Btw it doesn't look like the demos index page is listing any documentation pages from what I can tell. So the core page I added will still be somewhat hidden since the only page that links to it is pretty obscure.

duboisp:
The code look good.

**TODO**: finalize a local testing. 

\cc @polmih

GormFrank (approved):
Works as intended!

PR #9287, opened 2022-02-18
Footnote - Relink footer reference when multiple and not first

duboisp:
Relink footer reference when multiple and not first.

Rational: A similar JS code was added into all pages in Canada.ca and we are removing those. 

Purpose: Fix an editing issue when a footnote is added later during a review step and that new footnote is before the one that already exist when the footnote ID match the wet-boew footnote example. This ensure to provide a consistent user experience.

---
For Principal Publisher: This request is related to MWS-2053

### DIFF
--- a/src/plugins/footnotes/footnotes-en.hbs
+++ b/src/plugins/footnotes/footnotes-en.hbs
@@ -22,14 +22,14 @@
 		<li>Progressive enhancement approach</li>
 		<li>Support for Firefox, Opera, Safari, Chrome, and IE 7+<sup id="fn2-2-rf"><a class="fn-lnk" href="#fn2"><span class="wb-inv">Footnote </span>2</a></sup></li>
 		<li>Support for English and French</li>
-		<li>Configurable layout and design<sup id="fn2-3-rf"><a class="fn-lnk" href="#fn2"><span class="wb-inv">Footnote </span>2</a></sup>&#160;<sup id="fn3-rf"><a class="fn-lnk" href="#fn3"><span class="wb-inv">Footnote </span>3</a></sup></li>
+		<li>Configurable layout and design<sup id="fn2-3-rf"><a class="fn-lnk" href="#fn2"><span class="wb-inv">Footnote </span>2</a></sup>&#160;<sup id="fn3-rf"><a class="fn-lnk" href="#fn3"><span class="wb-inv">Footnote </span>3</a></sup>&#160;<sup id="fn4-0-rf"><a class="fn-lnk" href="#fn4"><span class="wb-inv">Footnote </span>4</a></sup></li>
 	</ul>
 </section>
 
 <section>
 	<h2>Recommended usage</h2>
 	<ul>
-		<li>Implementing footnotes in Web pages<sup id="fn*-rf"><a class="fn-lnk" href="#fn*"><span class="wb-inv">Footnote </span>*</a></sup></li>
+		<li>Implementing footnotes in Web pages<sup id="fn*-rf"><a class="fn-lnk" href="#fn*"><span class="wb-inv">Footnote </span>*</a></sup>&#160;<sup id="fn4-1-rf"><a class="fn-lnk" href="#fn4"><span class="wb-inv">Footnote </span>4</a></sup></li>
 	</ul>
 </section>
 
@@ -53,6 +53,11 @@
 			<p>Example of a footnote containing multiple paragraphs (third paragraph).</p>
 			<p class="fn-rtn"><a href="#fn3-rf"><span class="wb-inv">Return to footnote </span>3<span class="wb-inv"> referrer</span></a></p>
 		</dd>
+		<dt>Footnote 4</dt>
+		<dd id="fn4">
+			<p>Example of a footnote where first time clicked the user will get redirected to it's first instance of the footnote even if the coded link say otherwise. Prevent undesirable behaviour when footnote are added during a content review process.</p>
+			<p class="fn-rtn"><a href="#fn4-1-rf"><span class="wb-inv">Return to <span>first</span> footnote </span>4<span class="wb-inv"> referrer</span></a></p>
+		</dd>
 		<dt>Footnote *</dt>
 		<dd id="fn*">
 			<p>Example of a standard footnote, denoted by a symbol.</p>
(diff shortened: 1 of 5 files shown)


Ricokola (approved):
Seems all good!

duboisp (commented):
Todo: check if it don't conflict with the other code that resolve duplicate id issue.

EricDunsworth:
Something similar to this has been built-into AEM for ages to work around limitations in its WYSIWYG components. That hack is is present in pretty much every page on Canada.ca. So I guess this functionality is probably taken for granted by AEM users.

But is it really appropriate to be porting a CMS hack into the plugin itself? AFAIK there's no precedence for WET's plugins using JavaScript to correct mistakes in hardcoded HTML markup. IMO it'll encourage bad coding practices and make it more difficult to spot coding mistakes when pages undergo QA.

Furthermore, this PR won't affect noscript/basic HTML mode. So users will encounter a link named "Return to **first** footnote 4 referrer" that doesn't actually go to the first referrer in those scenarios. Things will only work properly when JS is fully enabled.

src/plugins/footnotes/footnotes-en.hbs
```diff
@@ -22,14 +22,14 @@
 		<li>Progressive enhancement approach</li>
 		<li>Support for Firefox, Opera, Safari, Chrome, and IE 7+<sup id="fn2-2-rf"><a class="fn-lnk" href="#fn2"><span class="wb-inv">Footnote </span>2</a></sup></li>
 		<li>Support for English and French</li>
-		<li>Configurable layout and design<sup id="fn2-3-rf"><a class="fn-lnk" href="#fn2"><span class="wb-inv">Footnote </span>2</a></sup>&#160;<sup id="fn3-rf"><a class="fn-lnk" href="#fn3"><span class="wb-inv">Footnote </span>3</a></sup></li>
+		<li>Configurable layout and design<sup id="fn2-3-rf"><a class="fn-lnk" href="#fn2"><span class="wb-inv">Footnote </span>2</a></sup>&#160;<sup id="fn3-rf"><a class="fn-lnk" href="#fn3"><span class="wb-inv">Footnote </span>3</a></sup>&#160;<sup id="fn4-0-rf"><a class="fn-lnk" href="#fn4"><span class="wb-inv">Footnote </span>4</a></sup></li>
```
EricDunsworth, line 25: Code example at the bottom of the page will need to be updated to cover footnote 4.

src/plugins/footnotes/footnotes-fr.hbs
```diff
@@ -53,6 +53,12 @@
 			<p>Exemple de note de bas de page qui compte plusieurs paragraphes (troisième paragraphe).</p>
 			<p class="fn-rtn"><a href="#fn3-rf"><span class="wb-inv">Retour à la référence de la note de bas de page </span>3</a></p>
 		</dd>
+		<dt>Note de bas de page 4</dt>
+		<dd id="fn4">
+			<p>Example of a footnote where first time clicked the user will get redirected to it's first instance of the footnote. That even if the coded link say otherwise. It prevent undesirable behaviour when footnote are added during a content review process.</p>
```
EricDunsworth, line 58:
```suggestion
```

src/plugins/footnotes/footnotes-fr.hbs
```diff
@@ -53,6 +53,12 @@
...
+			<p>Exemple de note de bas de page lorsqu'il est initialement cliqué et que l'utilisateur est redirigé vers la première instance de la note de base de page. Ceci même si cela a été codifié autrement. Cela prévient un comportement indésirable lorsqu'une note de bas de page est ajouté durant une revue du contenu.</p>
+			<p class="fn-rtn"><a href="#fn4-1-rf"><span class="wb-inv">Retour à la <span>quatrième</span> référence de la note de bas de page </span>4</a></p>
```
EricDunsworth, line 60:
```suggestion
			<p class="fn-rtn"><a href="#fn4-1-rf"><span class="wb-inv">Retour à la <span>première</span> référence de la note de bas de page </span>4</a></p>
```

EricDunsworth (changes requested):
Even though I disagree with the premise of this PR, I decided to review it anyway :).

src/plugins/footnotes/footnotes-en.hbs
duboisp: I fixed in both language

duboisp:
As discussed at the technical review, I added some documentation and a log for when the relink workaround is activated. That way quality assessor would be able to detect if the link in the footer need to be adjusted for an more optimal content.

src/plugins/footnotes/footnotes.js
```diff
@@ -69,21 +71,38 @@ var componentName = "wb-fnote",
...
 					// Focus on associated referrer link (if the return link points to an ID)
 					if ( ref.charAt( 0 ) === "#" ) {
-						refId = "#" + wb.jqEscape( ref.substring( 1 ) );
+						refId = wb.jqEscape( ref.substring( 1 ) );
+
+						// When first clicked, ensure we send the user on the first instance when the id follow the recommend pattern
+						refIdDashIdx = refId.indexOf( "-" );
+						if ( refIdDashIdx !== -1 && !$elmTarget.attr( modFlag ) ) {
+							searchRefId = refId.substring( 0, refIdDashIdx + 1 );
+							refIdSrc = $( "sup[id^='" + searchRefId + "']:first()" ).attr( "id" );
```
EricDunsworth, line 95:
```suggestion
							refIdSrc = wb.jqEscape( $( "sup[id^='" + searchRefId + "']:first()" ).attr( "id" ) );
```

Escape ``refIdSrc`` before it gets compared to ``refId`` later on.

Prevents the plugin from treating footnote \*'s return link as incorrect.

src/plugins/footnotes/footnotes.js
```diff
@@ -69,21 +71,38 @@ var componentName = "wb-fnote",
...
+							if ( !refIdSrc || refId !== refIdSrc ) {
+								console.warn( componentName + " - Relink first reference of " + ref + " for #" + refIdSrc );
+								$elmTarget
```
EricDunsworth, line 98:
```suggestion
								refId = refIdSrc;
								$elmTarget
```

Change ``refId``'s value to match ``refIdSrc``.

Without it, clicking on footnote 4's return link will navigate to the second reference link instead of the first one. Why? Because the plugin focuses onto ``refId`` a few lines later.

The updated ``href`` attribute won't be taken into consideration unless the user tries repeatedly clicking footnote 4's return link.

src/plugins/footnotes/footnotes.js
```diff
@@ -69,21 +71,38 @@ var componentName = "wb-fnote",
...
+									.attr( "href", "#" + refIdSrc )
```
EricDunsworth, line 99:
```suggestion
									.attr( "href", "#" + refId )
```

Given my preceding change suggestion, maybe it'd be worthwhile to change this spot to ``refId`` too (since its value is correct by this point in the logic).

EricDunsworth (changes requested):
I spotted a few bugs while re-reviewing :(.

So I played around with the JS logic and wanted to suggest some bug fixes :).

EricDunsworth (approved):

PR #9289, opened 2022-02-24
Add plugin - Data fusion query

duboisp:
Map a query parameter value into an input value.

Also, improve the URL query parsing in the wb core.

Note: This plugin existed in Canada.ca for a while and we are moving/integrating it into wet-boew project and make it available to all wet-boew users.

Principal Publisher reference: MWS-2053

### DIFF
--- a/src/core/wb.js
+++ b/src/core/wb.js
@@ -31,18 +31,13 @@ var getUrlParts = function( url ) {
 
 			// A collection of the parameters of the query string part of the URL.
 			params: ( function() {
-				var results = {},
-					queryString = encodeURI( decodeURI( a.search.replace( /^\?/, "" ) ) ).replace( /'/g, "%27" ).split( "&" ),
-					len = queryString.length,
-					key, strings, i;
-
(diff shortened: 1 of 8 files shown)


src/core/wb.js
```diff
@@ -31,18 +31,13 @@ var getUrlParts = function( url ) {
...
-					if ( ( key = queryString[ i ] ) !== null ) {
-						strings = key.split( "=" );
-						results[ strings[ 0 ] ] = strings[ 1 ];
-					}
+				var queryString = a.search.replace( /(^\?)/, "" );
+				if ( !queryString ) {
+					return {};
 				}
-				return results;
+				return a.search.replace( /(^\?)/, "" ).split( "&" ).map( function( n ) {
```
GormFrank, line 38: @duboisp Can you use the variable you created?
```suggestion
				return queryString.split( "&" ).map( function( n ) {
```

GormFrank (changes requested):
Will approve after small change.

GormFrank (approved):

PR #9296, opened 2022-03-02
add new plugin: Data Scrub

polmih:
A clear and concise description of what the code of this PR accomplishes. /// Une description claire et concise de ce que le code de cette PR accompli.

This plugin will delete the potential Personal Identifiable Information (PII) before a form submit.

A popup message will inform the user if any fields contain personal information and will show the output of the scrubbed data. The user can choose to go back to the form to delete the data or to proceed with scrubbing the personal information and the form submission;

This plugin will work with any standard form.
This plugin will not support Form validation plugin or Postback plugin to be executed on the same form.

In order to activate this plugin on a form:

the form fields that needs to be scrubbed should have the following data attribute : `data-rule-wbscrub` ;


src/plugins/wb-data-scrub/wb-data-scrub.js
```diff
@@ -0,0 +1,145 @@
...
+								scrubbedVal = wb.findPotentialPII( elmInputVal, true );
+							$elmInput.removeAttr( "data-first-to-scrub" );
+
+							// if the helper find any PII values clean the PII and assign new value to the form element with the output string
+							if ( scrubbedVal !== false ) {
+								countPiifield++;
+								if ( countPiifield === 1 ) {
+									$elmInput.focus();
+								}
+								msgBody2 = msgBody2 + "<li>" + $elmInput.prev( "label[for=" + $elmInput.attr( "id" ) + "]" ).text() + ": " + ( scrubbedVal !== "" ? scrubbedVal : "" ) + "</li>";
```
duboisp, line 76: $elmInput.prev is not accurate, search for the "for" attribute associate with the input instead.
We need to use the fieldname, not the full label
We need to use some markup to differenciate the field label and the value. And use markup to clearly identify what is the value. Like it is not clear when the value is fully emptied.


May be we can provide the possible of the author to update that patter if they define a template element in the webform?

src/plugins/wb-data-scrub/wb-data-scrub.js
```diff
@@ -0,0 +1,145 @@
...
+										"<details class='mrgn-tp-md mrgn-bttm-md'>" +
+										"<summary>" + viewMore + "</summary>" +
+											msgBody3 +
+											msgBody4 +
+										"</details>" +
+										"</div>" +
+										"<div class='modal-footer'>" +
+										"<ul class='list-inline text-center'>" +
+											"<li><button class='btn btn-default popup-modal-dismiss pull-left' data-wbscrub-submit>" + yesBtn + "</button></li>" +
+											"<li><button class='btn btn-primary popup-modal-dismiss pull-left'>" + cancelBtn + "</button></li>" +
```
duboisp, line 104: Do we need 2 pull-left here?

src/plugins/wb-data-scrub/wb-data-scrub.js
```diff
@@ -0,0 +1,145 @@
...
+				checkFormValues();
+				tpl.innerHTML = "<section id='" + moDalId + "' " + "class='modal-dialog modal-content overlay-def'>" +
+										"<header class='modal-header'><h2 class='modal-title'>" + msgboxHeader + "</h2></header>" +
+										"<div class='modal-body'>" +
+										"<p>" + msgBody1 + "</p>" +
+										"<p>" + msgBody2 + "</p>" +
+										"<details class='mrgn-tp-md mrgn-bttm-md'>" +
+										"<summary>" + viewMore + "</summary>" +
+											msgBody3 +
+											msgBody4 +
```
duboisp, line 98: 2 msgBody ?

src/plugins/wb-data-scrub/wb-data-scrub.js
```diff
@@ -0,0 +1,145 @@
...
+					};
+				if ( document.getElementById( moDalId ) ) {
+					document.getElementById( moDalId ).remove();
+				}
+				checkFormValues();
+				tpl.innerHTML = "<section id='" + moDalId + "' " + "class='modal-dialog modal-content overlay-def'>" +
+										"<header class='modal-header'><h2 class='modal-title'>" + msgboxHeader + "</h2></header>" +
+										"<div class='modal-body'>" +
+										"<p>" + msgBody1 + "</p>" +
+										"<p>" + msgBody2 + "</p>" +
```
duboisp, line 94: 2 paragraph? Can we stick to 1 if possible. When following the example the second paragraph become empty

src/plugins/wb-data-scrub/wb-data-scrub.js
```diff
@@ -0,0 +1,145 @@
...
+										"<details class='mrgn-tp-md mrgn-bttm-md'>" +
+										"<summary>" + viewMore + "</summary>" +
+											msgBody3 +
+											msgBody4 +
+										"</details>" +
+										"</div>" +
+										"<div class='modal-footer'>" +
+										"<ul class='list-inline text-center'>" +
```
duboisp, line 102: (Best practice) We should leave the button in their default location. Having in center might create issue for user that using zoom software to navigate.

duboisp (changes requested):
Default action should be first and in primary color. Do we want the person to "agree" the removal of the information or do we want the user to let's scrub and move forward?

duboisp (commented):
I suggest to have the "test" case after the working example.

src/plugins/wb-data-scrub/wb-data-scrub.js
polmih: Took note for the template element. Will be implemented in the next step, as discussed.
For the rest of the points, all done.

src/plugins/wb-data-scrub/wb-data-scrub.js
polmih: the 2 paragraph are merged and an eslint exception added, as discussed

src/plugins/wb-data-scrub/wb-data-scrub.js
polmih: fixed.

src/plugins/wb-data-scrub/wb-data-scrub.js
polmih: fixed

polmih:
> I suggest to have the "test" case after the working example.

Done. Thanks

src/plugins/wb-data-scrub/wb-data-scrub.js
```diff
@@ -0,0 +1,144 @@
...
+						},
+						fr: {
+							"msgboxHeader": "Supprimer les informations personnelles",
+							"msgBodyIntro": "Certaines informations ajoutées dans les champs suivants sont identifiées comme des informations personnelles et seront supprimées et soumises comme suit:",
+							// eslint-disable-next-line max-len
+							"msgBodyViewMore": "<p>Veuillez noter que toute information correspondant aux cas de la liste ci-dessous est considérée comme une information personnelle et est supprimé du formulaire avant sa soumission:</p> <ul><li>8 chiffres ou plus séparés ou non par les caractères suivants : ' ', '/', '-', '.'</li><li>Numéro de série du passeport</li><li>E-mail</li><li>Code Postal</li><li>Des paramètres considérés comme non sécurisés dans une url</li></ul>",
+							"yesBtn": "J'accepte, supprime les informations personnelles et soumettre",
+							"cancelBtn": "Retourner au formulaire",
+							"viewMoreLink": "En savoir plus"
+						}
```
duboisp, line 47: Those i18n would need to be added into our i18n csv file

duboisp (changes requested):

src/plugins/wb-data-scrub/wb-data-scrub-en.hbs
```diff
@@ -0,0 +1,85 @@
...
+
+<h2>Test case</h2>
+<p>Fill the following fields with the following values:</p>
+<ul>
+	<li>Phone: <code>514 111 1111</code></li>
+	<li>Email: <code>1@example.com</code></li>
+	<li>Search: <code>Passport Nr: AB123456</code></li>
+	<li>Comments: <code>visa: 4916 0739.3667-6891, master:5428735149026050, SIN:123 123-123, ZIP:K2C 3N2, Driving Licence Nr:P12345678912345, Bank account example: 003-1234567</code></li>
+	<li>No scrub field example: <code>John Doe</code></li>
+</ul>
```
GormFrank, line 83: @duboisp We have some test cases strings here. Would you like @polmih to move them somewhere else instead?

src/plugins/wb-data-scrub/wb-data-scrub-en.hbs
```diff
@@ -0,0 +1,85 @@
...
+<p>The form will prompt a message informing the user that any Personal Information will not be submitted.</p>
+<p>After the form submit, the output of url parameters will be : <code>?preferredNumber=&email=&search=Passport+Nr%3A+&comments=visa%3A+%2C+master%3A%2C+SIN%3A%2C+ZIP%3A%2C+Driving+Licence+Nr%3AP%2C+Bank+account+example%3A+&noScrubField=John+Doe</code></p>
```
GormFrank, line 85: @polmih Please escape the & character, like so: `&amp;`

(Same in French.)

GormFrank (changes requested):

src/plugins/wb-data-scrub/wb-data-scrub-en.hbs
duboisp: Yes, I will provide some instruction in the other PR.

src/plugins/wb-data-scrub/wb-data-scrub.js
polmih: Done. Thanks

src/plugins/wb-data-scrub/wb-data-scrub-en.hbs
polmih: Done. Thanks

src/plugins/wb-data-scrub/wb-data-scrub.js
```diff
@@ -0,0 +1,144 @@
...
+
+			var	form = elm.form;
+
+			$( elm.form ).data( "scrub-inited", "" );
+
+			form.addEventListener( "submit", function( event ) {
+				event.preventDefault();
+				var isPii = false,
+					moDalId = componentName + "-modal",
+					i18nDict = {
```
duboisp, line 29: No need to define the i18nDict, instead retrieve it from wb.

Something like: https://github.com/wet-boew/wet-boew/blob/master/src/plugins/multimedia/multimedia.js#L108-L120

duboisp (changes requested):

src/plugins/wb-data-scrub/wb-data-scrub.js
```diff
@@ -0,0 +1,144 @@
...
+								}
+							}
+						} );
+						msgBodyListFields += "</dl>";
+					};
+				if ( document.getElementById( moDalId ) ) {
+					document.getElementById( moDalId ).remove();
+				}
+				checkFormValues();
+				tpl.innerHTML = "<section id='" + moDalId + "' " + "class='modal-dialog modal-content overlay-def'>" +
```
duboisp, line 91: 🙄

duboisp (changes requested):

Garneauma:
Closed by #9823

PR #9298, opened 2022-03-07
helpers: adjust passport pattern in findPotentialPII

polmih:
In helpers: 

- adjust passport pattern in findPotentialPII;
- create test cases for findPotentialPII;

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -376,6 +376,7 @@ module.exports = (grunt) ->
 					"src/core/dep/modernizr-custom.js"
 					"src/core/wb.js"
 					"src/core/helpers.js"
+					"src/shims/**/*.js"
 					"src/plugins/**/*.js"
 					"!src/plugins/**/test.js"
 					"!src/plugins/**/assets/*.js"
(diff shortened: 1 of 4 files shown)


duboisp:
@polmih can you provide all the strings that have been used for testing to find potential PII? Thanks

duboisp:
You can create the test into a "test.js" in the core folder beside the helper file.

You can structure you file like the "src/plugin/feeds/test.js". You will need to describe the test. There is no need to using the other helper there. You can call directly `wb.findPotentialPII()` and assert by using the function

```js
it( "should match canadian nr passport", function() {
  expect( wb.findPotentialPII( "PATTERN" ) ).to.equal( "EXPECTED TRIMMED STRING" );
});
```

polmih:
> You can create the test into a "test.js" in the core folder beside the helper file.
> 
> You can structure you file like the "src/plugin/feeds/test.js". You will need to describe the test. There is no need to using the other helper there. You can call directly `wb.findPotentialPII()` and assert by using the function
> 
> ```js
> it( "should match canadian nr passport", function() {
>   expect( wb.findPotentialPII( "PATTERN" ) ).to.equal( "EXPECTED TRIMMED STRING" );
> });
> ```

Test cases added. Thanks

duboisp (approved):
Tested locally and work as expected.

Thanks @polmih for adding some unit testing

PR #9300, opened 2022-03-08
Theme: Add colcount-xxs-1 proximity class

EricDunsworth:
Meant to be used alongside any ``colcount-xs-*`` class. Makes it possible to revert a multi-column layout intended for extra-small view into a single-column layout in extra-extra-small view.

### DIFF
--- a/src/base/proximity/_base.scss
+++ b/src/base/proximity/_base.scss
@@ -255,6 +255,10 @@
 	}
 }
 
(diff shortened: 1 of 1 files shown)


EricDunsworth:
Closing this since it's irrelevant.

I was mistaken about what ``colcount-xs-*``'s minimum view width was (thought it was setup as 0... turns out it's actually 480px).

PR #9302, opened 2022-03-08
Equal height CSS - make col fullwidth on xs viewport

GormFrank:
**Que fait cette demande « pull » (PR)?**

Facilitates a glitch on xs viewport where a div with `col-*` class would take as much space as it's max content takes up creating a non-uniform look.

**Captures d'écrans**

Before:
Screen Shot 2022-03-08 at 4 51 37 PM

After:
Screen Shot 2022-03-08 at 4 52 05 PM

### DIFF
--- a/src/base/_wet-boew.scss
+++ b/src/base/_wet-boew.scss
@@ -138,6 +138,7 @@
 /* Extra-small view and under */
 @media screen and (max-width: $screen-xs-max) {
 	@import "views/screen-xs-max";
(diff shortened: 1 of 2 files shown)


duboisp (changes requested):
Can you update the working example that will show the isssue.

Like making shorter the paragraph in the "short container" column

duboisp:
Note: Once this PR is merged, we will need to update the import list for GCWeb theme

duboisp (approved):
We leave the working example as is.

PR #9307, opened 2022-03-11
Navcurrent: Add aria-current="page" attribute

EricDunsworth:
The navcurrent plugin is a perfect fit for the `aria-current` attribute. The former adds a class ("wb-navcurr") to visually-denote links to the current page, whereas the latter represents it in semantics.

**Scenarios this covers:**
* Any links to the current page the plugin tries adding the "wb-navcurr" class to

**Scenarios this won't cover:**
* Top-level mega menu links that point to dropdown IDs (`aria-current` is inappropriate in those scenarios)

### DIFF
--- a/src/plugins/navcurrent/navcurrent.js
+++ b/src/plugins/navcurrent/navcurrent.js
@@ -121,6 +121,7 @@ var componentName = "wb-navcurr",
 
 				if ( match ) {
 					link.className += " " + className;
+					link.setAttribute( "aria-current", "page" );
 					if ( menu.className.indexOf( "wb-menu" ) !== -1 && link.className.indexOf( "item" ) === -1 ) {
 						$( link ).closest( ".sm" ).parent().children( "a" ).addClass( className );
 					}


duboisp (changes requested):
The `[aria-current=page]` is only when the link is **visually styled** to represent the currently-displayed page. This is not the case with your proposed change.

I tested your propose fix with the `/index-en.html` page and looked at the link in the footer. There is link where the attribute is added and they are not styled appropriately. For example the link: About the Web Experience Toolkit

ref. https://www.w3.org/TR/2021/CRD-wai-aria-1.2-20211208/#aria-current

EricDunsworth:
> I tested your propose fix with the ``/index-en.html`` page and looked at the link in the footer. There is link where the attribute is added and they are not styled appropriately. For example the link: About the Web Experience Toolkit

@duboisp IMO that's a CSS bug in WET. #9308 will fix it.

Garneauma (approved):

Garneauma:
API changes:
- Patch - Navcurrent: add aria-current="page" attribute to link

EricDunsworth:
Going to mark this as a draft for the time being.

Given that ``aria-current="page"`` would be inappropriate in the context of anchor ID links... I think this PR will need to be reworked to implement https://github.com/wet-boew/wet-boew/pull/9308#issuecomment-4239943786 's suggestions.

PR #9308, opened 2022-03-14
Navcurrent: Make current page's footer link bold

EricDunsworth:
The "wb-navcurr" class changes colour schemes in the mega and secondary menus to visually-emphasize the current page.

This adds a similar effect to footer links via bolding. It wouldn't of been practical to change text/background colours in that context since WET's footer links don't resemble buttons.

### DIFF
--- a/theme/footer/_base.scss
+++ b/theme/footer/_base.scss
@@ -5,4 +5,8 @@
 	li {
 		margin-bottom: .75em;
 	}
+
+	.wb-navcurr {
+		font-weight: 700;
+	}
 }
--- a/theme/footer/_base.scss
+++ b/theme/footer/_base.scss
@@ -6,7 +6,7 @@
 		margin-bottom: .75em;
 	}
 
-	.wb-navcurr {
+	a.wb-navcurr {
 		font-weight: 700;
 	}
 }


duboisp (changes requested):
The change in this PR should be merged with your other related navcurrent PR #9307 

Prior to apply this change, the style should only be applied when the link exactly match the current page, not an anchor. Like the footer link "About the Web Experience Toolkit" when navigating in the `index.en.html` page
Also, we will need to identify all the location where we are using that plugin. We need to evaluate the impact of this propose change.

TODO:
* [ ] Merge change with PR #9307 
* [ ] Identify all the instance that use this plugin. In GCWeb and wet-boew
* [ ] Ensure the style is only applied for exact match including the page anchor, not how the wb-navcurr plugin is currently working.

theme/footer/_base.scss
```diff
@@ -5,4 +5,8 @@
 	li {
 		margin-bottom: .75em;
 	}
+
+	.wb-navcurr {
```
Garneauma:
```suggestion
	a.wb-navcurr {
```

Garneauma:
@duboisp After some research, it seems like the wb-navcurr plugin is only used in the following instances:
- WET-BOEW footer
- Old version of the GCWeb menu (now deprecated)
- Secondary menu (which I believe is not recommended for use)
- Theme template (which currently adds "aria-current=page")

Therefore, I believe the scope of this PR should be focused on the WET-BOEW footer. I don't think we should enforce a style for the link with "aria-current" attribute globally.

EricDunsworth:
@Garneauma Just merged-in your suggestion :).

As for these...

**1)**
> Merge change with PR https://github.com/wet-boew/wet-boew/pull/9307

@Garneauma Did you want me to proceed with this? It'd facilitate testing, but I wouldn't consider this PR and #9307 to go hand-in-hand. I suppose they complement each other, but both serve distinct purposes (hence why I sent them separately).

This PR only cares about adding a visual indicator in the WET theme's footer, whereas #9307's goal is to add ``aria-current`` anywhere the plugin runs.

**2)**
> Ensure the style is only applied for exact match including the page anchor, not how the wb-navcurr plugin is currently working.

I agree that this should be done and isn't the plugin's current behaviour.

It looks like the plugin is currently designed to:
* Look for _pages_ (it's oblivious to anchor IDs)
* Match the final link to the current _page_

Example:
* Check the [WET index page's about section](https://wet-boew.github.io/v4.0-ci/index-en.html#about):
  * Its header link ("[About the Web Experience Toolkit](https://wet-boew.github.io/v4.0-ci/index-en.html#about)") isn't highlighted... but the "[Benefits](https://wet-boew.github.io/v4.0-ci/index-en.html#benefits)' link is.

Note:
* That behaviour exists in the header's mega menu, mobile menu overlay and footer.

IMO the best way of dealing with that flaw would be to rework #9307 to:
* Take anchor ID links into consideration
* Be pickier about ``aria-current`` values (such as using ``aria-current="page"`` for page links and ``aria-current="true"`` for anchor ID links... more info can be found in [ARIA 1.3's ``aria-current`` state spec](https://w3c.github.io/aria/#aria-current))

@Garneauma Thoughts?

Garneauma:
@EricDunsworth We might be better off simply adding to the documentation that this plugin is to be used for multi-page navigation, not single-page navigation using anchors since I believe the `aria-current="true"` would no longer be valid once the user scrolls past the selected item's section and we don't have any way to update the dynamically. What do you think?

EricDunsworth:
> @EricDunsworth We might be better off simply adding to the documentation that this plugin is to be used for multi-page navigation, not single-page navigation using anchors...

@Garneauma I don't think the navcurrent plugin is currently covered anywhere in WET's documentation.

My take on this is that there's already an expectation that the plugin should support anchors :P (given that WET has used anchors in its header/footer for ages). So I think the best course of action would still be to fix the plugin's anchor support.

> ... since I believe the aria-current="true" would no longer be valid once the user scrolls past the selected item's section...

That's one way of looking at it, but in this case, I'd expect the "current" anchor to match the [``:target`` pseudo-class in CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:target) (i.e. current URL hash). Not whether the anchor ID is currently on-screen.

As far as the ``aria-current`` attribute's value goes, based on the [ARIA spec's ``aria-current`` definition](https://w3c.github.io/aria/#aria-current) (whose examples imply that "sets" of items should be of the same type)... I think it'd make sense to:
* Set the value to "page" if the set only contains pages
* OR set the value to "true" if anchors exist anywhere within the set (even if the current item is a link to a page)

> ... and we don't have any way to update the dynamically. What do you think?

IMO it'll probably be necessary to add dynamic updating support. It might not be _too_ bad though.

The plugin could probably be easily adjusted to cache all the links it loops through. A [``hashchange`` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/hashchange_event) could also be thrown-in to make it update the current item based on what it cached earlier. The ``aria-current`` type could also be cached.

Garneauma:
@EricDunsworth Up to you if you want to spend time to add those features. For now, all we'll expect is a `aria-current="page"` on the anchor that has the current page's URL.

EricDunsworth:
@Garneauma I know navcurrent isn't as relevant in 2026 as I believe it was when I sent these PRs in 2022... but since I hate unfinished business, I'll try to put in the time to tackle anchor support to my satisfaction :P.

PR #9316, opened 2022-03-29
Nested control: NEW plugin to nest field within checkbox/radio

GormFrank:
Provide ability to have a nested field within a checkbox or radio form control.

## Additional information

- [ ] Translate documentation to French
- [ ]  Validate changes against WCAG for accessibility and test for screen readers
- [ ] Test against other checkbox and radio button list patterns (may collide with GC-Chckbxrdio in GCweb)
- [ ] Define limitations

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -1004,6 +1004,8 @@ module.exports = (grunt) ->
 						"W010" # Using `.pull-left` or `.pull-right` as part of the media object component is deprecated as of Bootstrap v3.3.0. Use `.media-left` or `.media-right` instead.
 						"E013" # Only columns (`.col-*-*`) may be children of `.row`s
 						"E014" # Columns (`.col-*-*`) can only be children of `.row`s or `.form-group`s
+						"E017" # WET wants to support explicit labels when using checkboxes and radio inputs under the .wb-nested-control class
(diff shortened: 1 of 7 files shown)


GormFrank:
Up here for testing purposes: https://universallabs.org/wet/labs/pr-9316/demos/nested-control/nested-control-en.html

duboisp:
In the working example, the first bullet items are right aligned instead of left aligned.

EricDunsworth:
FWIW it looks like it's only happening in Firefox. I checked the [Universal labs demo](https://universallabs.org/wet/labs/pr-9316/demos/nested-control/nested-control-en.html) in Edge Chromium earlier and it looked ok in that browser.

Seems similar to #9147 but with a different root cause.

EricDunsworth:
If this PR ultimately leads to anything viable (i.e. a stable feature - not a provisional "use at your own risk" class or plugin :P), it might be worth modifying the OP or commit message to make it close #8836.

IMO this PR's code is much simpler and more elegant than what @fsnoddy and I proposed in #8836. But it lacks ``aria-*`` attributes (some of which would've been brought in by the toggle plugin if #8836 had been finalized).

More research and testing will probably be needed to figure out how best to leverage ``aria-*`` attributes in the scenario this PR addresses. For instance, could you get away with just ``aria-live``? Would ``aria-controls`` be a better bet, or would screen readers not care about in practice?

PR #9321, opened 2022-04-19
WCAG SC 2.4.13 - Testing Script for page break

Ricokola:
Based on the change log of WCAG 2.2: [https://www.w3.org/TR/WCAG22/#changelog](https://www.w3.org/TR/WCAG22/#changelog), This success criterion has been removed on July 15 2022

Creation of new JS file (wcag.js) for testing Web pages against certain identified success criteria.

### DIFF
--- a/src/core/wcag.js
+++ b/src/core/wcag.js
@@ -1,0 +1,29 @@
+/*
+ * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)
+ * @title WCAG Script for testing purposes
+ * @overview Test certain WCAG Success Criteria
+ * @license wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
+ * @author @ricokola @duboisp
+ */
+
+function( $, wb ) {
+
+	// Execution of any action after WET is ready
+	wb.doc.one( "wb-ready.wb", setTimeout( function( ) {
+		var $pgbrk = $( ".pg-brk-aft" ),
+			msg = "WCAG SC 2.4.13 - The page break (.pg-brk-aft) is not ";
+
+		if ( $pgbrk.length > 0 ) {
+			$pgbrk.each( function() {
+				var id = $( this ).next().attr( "id" );
+				if ( id ) {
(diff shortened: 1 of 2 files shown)


duboisp (changes requested):
* [x] Please wrap the code to ensure there is no interference with any other JS code.
* [x] Modify the build script to include this file.
* [x] We need to run this code only after WET has been ready and all other WET component has been initialized and marked ready.

src/core/wcag.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+var $pgbrk = $( ".pg-brk-aft" ),
+	msg = "WCAG SC 2.4.13 - The page break (.pg-brk-aft) is not ";
+
+if ( $pgbrk.length > 0 ) {
+	$pgbrk.each( function() {
+		var id = $( this ).next().attr( "id" );
+		if ( id ) {
+			if ( !$( "a[href*=" + id + "]" ).length ) {
```
duboisp, line 16: We will need to change the CSS attribute operator

duboisp (changes requested):

GormFrank:
@Ricokola Please rebase to upstream/master and @duboisp Have your requested changes been fixed / is it good to go?

GormFrank:
Concerned about it running every page load to look for page breaks when 99% of all pages using WET don't use page breaks.

Maybe a different approach could be done to programmatically load the script instead of having the check for class. Something  "softer" in terms of performance (even though we are not talking about a big hit anyway).

CC:  @duboisp @Ricokola

duboisp (commented):
- [ ] Can you add a link, in comments, of a valid and invalid page with page-break. The invalid page should enumerate all the use case.
- [x] Add un delay de 250 milisecondes (setTimeout) pour ne pas interférer avec les autres plugins. 
- [ ] Add a blurb dans notre page de documentation dans les utilities.

Ricokola:
Based on the change log of WCAG 2.2: [https://www.w3.org/TR/WCAG22/#changelog](https://www.w3.org/TR/WCAG22/#changelog), This success criterion has been removed on July 15 2022

PR #9328, opened 2022-05-04
helpers: findPotentialPii changes

polmih:
Update: 2022-06-14:

WET wb.findPotentialPII() function is updated to:

- Switch 8-digit wildcard to a 9-digit wildcard;
- New parameters added : 

   1.  `scope`  - can be:
        - a Boolean - if true, it will scrub the content that match the patterns;
        - an object (optional) that will the define the scope of replacing the content. Can have the following properties:
           - {string} any key name of the default patterns e.g. email, digits, etc. with the value 1. The function will only scrub the content that match the regex of the default patterns passed in this object;
           - {regex} customCase - this param is a regex. It will search and replace the values corresponding that pattern

   2. `opts` (optional) - an object that can have the following properties:
       - {boolean} isCustomExclusive - if true, it will scrub only the custom regex if the regex is the only property of the "scope" object
       - {bolean} useFullBlock - if true, it will replace the scrubbed characters with the "█" symbol;
       - {string} useString - this string will replace the scrubbed content

### DIFF
--- a/site/pages/docs/ref/core/core-en.hbs
+++ b/site/pages/docs/ref/core/core-en.hbs
@@ -63,6 +63,65 @@
 </section>
 
 <section>
+	<h2>JS API</h2>
+	<h3>wb.findPotentialPII(str, scope, opts) - Find most common Personal Identifiable Information (PII) in a string</h3>
+	<p>Find most common Personal Identifiable Information (PII) in a string and return either the cleaned string either true/false</p>
+	<p>This function can have the following parameters</p>
+	<ul>
+		<li>
+			<code>str</code> (required) - string representing the content that needs to be verified
+		</li>
+		<li>
+			<code>scope</code> this parameter can be:
+			<ul>
+				<li>either a boolean (required) - If passed as true will scrub the content</li>
+				<li>either an object (optional) - having the following properties (optional):
+					<ul>
+						<li>{string} any key name of the default patterns e.g. <code>email</code>, <code>digits</code>, etc. with the value 1. The function will only scrub the content that match the regex of the default patterns passed in this object
+						<p>The list of the regular expressions that will be applied by default:</p>
+						<ul>
+							<li>digits</li>
+							<li>passport</li>
+							<li>email</li>
+							<li>postalCode</li>
+							<li>username</li><li>password</li>
+						</ul>
+						</li>
+						<li>{regex} <code>customCase</code> - this param is a regular expression. It will search and replace the values corresponding that pattern</li>
+					</ul>
+				</li>
+			</ul>
+		</li>
+		<li>
+			<code>opts</code> (optional) - the 3rd param of the function is an object that can contain the following properties (optional):
+			<ul>
+				<li>{boolean} <code>isCustomExclusive</code> - if the value is 1(true), it will scrubb only the custom regex if the regex is the only property of the <code>scope</code> object
+</li>
+				<li>{string} <code>replaceWith</code> - this string will replace the scrubbed content</li>
+				<li>{bolean} <code>useFullBlock</code> - if the is value 1(true), it will replace the scrubbed characters with the "█" symbol;</li>
+			</ul>
+		</li>
+	</ul>
+	<p>Working example</p>
+<pre><code>
+wb.findPotentialPII( "email:test@test.com, phone:123 123 1234", true )
+</code></pre>
+<p>It will return: "email:, phone:" </p>
+<pre><code>
+	wb.findPotentialPII( "email:test@test.com, phone:123 123 1234", false )
+</code></pre>
+<p>It will return: <code>true</code> </p>
+<pre><code>
+wb.findPotentialPII( "email:test@test.com, phone:123 123 1234", { email:1 }{ replaceWith: [REDACTED/CAVIARDÉ] } )
+</code></pre>
+<p>It will return:"email:[REDACTED/CAVIARDÉ], phone:123 123 1234" </p>
+<pre><code>
+wb.findPotentialPII( "email:test@test.com, phone:123 123 1234, numéro de cas 12345678", { "customCase":/\b(?:case[\s-]?number[\s\-\\.]?(?:\d{5,10}))|(?:numéro[\s-]?de[\s-]?cas[\s\-\\.]?(?:\d{5,10}))/ig }, { useFullBlock:1})
+</code></pre>
+<p>It will return: "phone:████████████, email:█████████████, postalCode:██████, ██████████████████████"</p>
+</section>
+
+<section>
 	<h2>Source code</h2>
 	<p><a href="https://github.com/wet-boew/wet-boew/tree/master/src/core">WET core source code on GitHub</a></p>
 </section>
--- a/src/core/helpers.js
+++ b/src/core/helpers.js
@@ -1252,35 +1252,91 @@ wb.escapeAttribute = function( str ) {
 
 /*
 * Find most common Personal Identifiable Information (PII) in a string and return either the cleaned string either true/false
-* @param {string} str
-* @param {boolean} toClean
+* @param {string} str (required) - the content that needs to be verified
+*
+* @param {boolean} scope - if true will scrub the content
+* @param {object} (optional) the 2nd param (scope) can also be an object having the following properties (optional):
+* 	{string} any key name of the default patterns e.g. email, digits, etc. with the value 1. The function will only scrub the content that match the regex of the default patterns passed in this object
+* 	{regex} customCase - this param is a regex. It will search and replace the values corresponding that pattern
+*
+* @param {object} opts (optional) - the 3rd param of the function that can contain the following properties (optional):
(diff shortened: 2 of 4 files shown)


GormFrank (dismissed):
Just go simple, no need to allow for different redacted marker based on the different regex.

duboisp:
See the technique for redacted text here: https://wet-boew.github.io/GCWeb/components/redacted-en.html

polmih:
> Just go simple, no need to allow for different redacted marker based on the different regex.

Switched back to simple version. Thanks!

duboisp:
Note: We will need to re-think and discuss how we can change the PII function to support the various requirement change/needs. Like the action to be taken: removal, redact, replace

duboisp:
As discussed, do we need to review the unit testing or/and add some new testing item?
Other than that, this are ready for review

\cc @GormFrank

polmih:
> 

Unit tests are updated and all tests passed.

src/core/helpers.js
```diff
@@ -1252,40 +1252,59 @@ wb.escapeAttribute = function( str ) {
...
+* returns "email:[REDACTED/CAVIARDÉ], phone:123 123 1234"
+*
+* wb.findPotentialPII( { "str":"email:test@test.com, phone:123 123 1234, numéro de cas 12345678", "toClean":true, "bMarker":true, "customCase":/\b(?:case[\s-]?number[\s\-\\.]?(?:\d{5,10}))|(?:numéro[\s-]?de[\s-]?cas[\s\-\\.]?(?:\d{5,10}))/ig })
+* returns "email:[REDACTED/CAVIARDÉ], phone:[REDACTED/CAVIARDÉ], [REDACTED/CAVIARDÉ]"
 */
-wb.findPotentialPII = function( str, toClean ) {
-
-	if ( typeof str  !== "string" ) {
+wb.findPotentialPII = function( args ) {
+	if ( args.str && typeof args.str  !== "string" ) {
```
duboisp, line 1277: We will need to avoid the breaking change. You will need a way to support the old way to sending the parameter.

duboisp (changes requested):

src/core/helpers.js
polmih: Fixed! Now the function accepts 2 params:

1. str  - the content to be verified;
2.  opts that can be passed: 
- as Boolean:  if set as true, it will simply scrub the data (this will avoid the major change);
- as object: here can be passed the other scrubbing options (bMarker, caseOnly, customCase)

src/core/helpers.js
```diff
@@ -1252,39 +1252,61 @@ wb.escapeAttribute = function( str ) {
...
+	var regEx = {
+			"digits": /\d(?:[\s\-\\.\\/]?\d){8,}(?!\d)/ig, //9digits or more pattern
+			"passport": /\b[A-Za-z]{2}[\s\\.-]*?\d{6}\b/ig, //canadian nr passport pattern
+			"email": /\b(?:[a-zA-Z0-9_\-\\.]+)(?:@|%40)(?:[a-zA-Z0-9_\-\\.]+)\.(?:[a-zA-Z]{2,5})\b/ig, //email pattern
+			"postal-code": /\b[A-Za-z]\d[A-Za-z][ -]?\d[A-Za-z]\d\b/ig, //postal code pattern
+			"username": /\b(?:(username|user)[:=][a-zA-Z0-9_\-\\.]+)\b/ig,
+			"password": /\b(?:(password|pass)[:=][^\s#&]+)\b/ig
+		},
+		isFound = false,
+		txMarker =  "[REDACTED/CAVIARDÉ]",
```
duboisp, line 1289: We will need to localize it. We would need to add an entry in the i18n spreadsheet

Check example that use "i18nText".

Or we can just fall back on the FullBlock character by default when no maker is provided.

The current practice is to replace every character by the marker, but your proposal seem is to replace the whole redacted text with a single phrase

src/core/helpers.js
```diff
@@ -1252,39 +1252,61 @@ wb.escapeAttribute = function( str ) {
...
+	}
 
 	for ( var key in regEx ) {
-		if ( str.match( regEx[ key ] ) ) {
+		var arrIndex = ( opts && opts.caseOnly && Object.prototype.hasOwnProperty.call( regEx, opts.caseOnly ) ? opts.caseOnly : key );
+		if ( str.match( regEx[ arrIndex ] ) ) {
 			isFound = true;
 			if ( toClean ) {
-				str = str.replaceAll( regEx[ key ], "" );
+				str = str.replaceAll( regEx[ arrIndex ], ( opts.bMarker && opts.bMarker === true ? txMarker : "" ) );
```
duboisp, line 1303:
```
var myString = "phone:514-514-5144, email:1@example.com";
var myReturnVal = wb.findPotentialPII( myString );
```

Est-ce que myString contient le même texte que myReturnVal? Car je crois qu'il est passé par référence.

Au pire utlise une autre variable pour la valeur retourné.

src/core/helpers.js
```diff
@@ -1252,39 +1252,61 @@ wb.escapeAttribute = function( str ) {
...
+* wb.findPotentialPII( "email:test@test.com, phone:123 123 1234", { "bMarker":true, "caseOnly":"email" } )
+* returns "email:[REDACTED/CAVIARDÉ], phone:123 123 1234"
+*
+* wb.findPotentialPII( "email:test@test.com, phone:123 123 1234, numéro de cas 12345678", { "bMarker":true, "customCase":/\b(?:case[\s-]?number[\s\-\\.]?(?:\d{5,10}))|(?:numéro[\s-]?de[\s-]?cas[\s\-\\.]?(?:\d{5,10}))/ig })
+* returns "email:[REDACTED/CAVIARDÉ], phone:[REDACTED/CAVIARDÉ], [REDACTED/CAVIARDÉ]"
 */
-wb.findPotentialPII = function( str, toClean ) {
-
-	if ( typeof str  !== "string" ) {
+wb.findPotentialPII = function( str, opts ) {
```
duboisp, line 1276: Example d'idée

option (2e paramètre) par défaut lorsque boolean ou undefined.

```js
{
  digits: 1,
  passport: 1,
  email: 1,
  postalCode: 1,
  username: 1,
  password: 1
}
```

Custom parameter like:
```js
{
   email: /kekchose/gi,
   digits: 1
   custom1: /kekchosedautre/gi
}
```

Iterate to build all the regex to execute with a for...in statement

```js
{
   email: /kekchose/gi,
   digits: 1
   custom1: /kekchosedautre/gi,
   replaceWithMarker: 1,
   toClean: 1 // if true replace if false or undefine don't replace
}
```

duboisp, line 1276:
```suggestion
wb.findPotentialPII = function( str, scope, opts ) {
```
where:
* str: string to find Potential PII
* scope: Object or list to run the find to detect PII (if False => we use all current regex)
* opts: Option of hte PII transformation: toClean replace with marker, replace with one phrase.....

3e param

```
{
   markerText = FullBlock character,
   repeatMarker = true | false,
   allRuleDefaultIncluded = true | false,
   allRuleExcluded = true | false // default true,
}

var settings = $.extend( {}, defaultOpts, opts );
// object.assign

```

duboisp (changes requested):

src/core/helpers.js
polmih: Testé en local, ne remplace pas le string initial

src/core/helpers.js
polmih: As discussed, the function will have 3 params:

1. `str ` (required) - the string that needs to ne verified;
2. `scope` if passed as Boolean is required as true if the content needs to be scrubbed; if passed as object is optional. 
3. `opts` (optional) - object that will customize the way the content is replaced;

duboisp (changes requested):
Code look good and work as expected.

Can you just document this findPII function here: site/pages/docs/ref/core/core-en.hbs with the same info that you added in the comments before the function declaration.

duboisp (changes requested):

site/pages/docs/ref/core/core-en.hbs
```diff
@@ -62,6 +62,56 @@
...
+<section>
+	<h2>JS API</h2>
+	<h3>Find most common Personal Identifiable Information (PII) in a string </h3>
+	<p>Find most common Personal Identifiable Information (PII) in a string and return either the cleaned string either true/false</p>
+	<p>This function can have the following parameters</p>
+	<ul>
+		<li>
+			<code>str</code> (required) - string representing the content that needs to be verified
+		</li>
+		<li>
```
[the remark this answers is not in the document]
polmih: Done. Thanks

site/pages/docs/ref/core/core-en.hbs
```diff
@@ -62,6 +62,56 @@
 	</div>
 </section>
 
+<section>
+	<h2>JS API</h2>
+	<h3>Find most common Personal Identifiable Information (PII) in a string </h3>
```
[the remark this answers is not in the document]
polmih: done.thanks

duboisp (approved):
Tested locally and documentation was fixed as discussed

PR #9332, opened 2022-05-06
Build: Bump grunt-html to ^15.2.2 and related changes

EricDunsworth:
* Bump ``grunt-html`` dependency to ^15.2.2
  * Many pre-existing ``htmllint`` errors weren't being reported due to a bug in ^14.3.0 (validator/grunt-html#152)... ``grunt-html`` 15.0.0 fixed it
* Drop node 10.x from GitHub Actions (``grunt-html`` 15.0.0 and up require node 12.x at minimum)
* Fix various ``htmllint`` errors:
  * License:
    * Remove "See " / "Voir " from an ``href`` attribute's value
    * Remove ``data-proofer-ignore`` attributes (no longer needed, botched ``href`` attributes were causing ``html-proofer``'s errors)
  * Demos:
    * Exit script:
      * Add ``>`` (greater than sign) to the end of a closing ``</details>`` tag
    * Lightbox:
      * Add an extra ``htmllint`` task that ignores lightbox's ``longdesc`` attributes... removing them might necessitate other changes to the plugin
    * Postback:
      * Add number suffixes to field IDs to make them unique
      * Add a space character before ``/`` (forward slash) in ``<input>`` tags
    * Postback (French version):
      * Move ``aria-hidden="true"`` alert after ``summary`` element
    * Utilities (French version):
      * Add ``>`` (greater than sign) to the end of a closing </h3> tag
  * Docs:
    * Background image:
      * Remove "Proprety name/type" column header... no content existed for it in data rows and its presence was causing a mismatch in the number of thead/tbody columns
      * Remove empty ``td`` element from the end of the ``data-bgimg`` attribute's row
    * Exit script:
      * Remove stray closing ``</tr>`` tag
      * Adjust indentation of an opening ``<tr>`` tag
    * Plugins (French version):
      * Add closing ``</a>`` tag after "Image d'arrière plan"
    * WET core (French version):
      * Add closing ``</div>`` tag for untranslated English content

Spinoff of #9146.

CC @GormFrank

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -386,7 +386,7 @@
     "absolute": {
       "version": "0.0.1",
       "resolved": "https://registry.npmjs.org/absolute/-/absolute-0.0.1.tgz",
-      "integrity": "sha512-Zs3Sz3lmEIH76fvnc4jto4CpuI5bhU2TD+jtBYVIfXnjFm+49y1opKxNy4gtNF+xSLBL7hXRpLwypc+o2MzBIQ==",
+      "integrity": "sha1-wigi+H4ck59XmIdQTZwQnEFzgp0=",
       "dev": true
     },
     "accepts": {
@@ -498,7 +498,7 @@
     "align-text": {
       "version": "0.1.4",
       "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
-      "integrity": "sha512-GrTZLRpmp6wIC2ztrWW9MjjTgSKccffgFagbNDOX95/dcjEcYZibYTeaOntySQLcdw1ztBoFkviiUvTMbb9MYg==",
+      "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
       "dev": true,
       "optional": true,
       "requires": {
@@ -528,7 +528,7 @@
     "ansi-red": {
       "version": "0.1.1",
       "resolved": "https://registry.npmjs.org/ansi-red/-/ansi-red-0.1.1.tgz",
-      "integrity": "sha512-ewaIr5y+9CUTGFwZfpECUbFlGcC0GCw1oqR9RI6h1gQCd9Aj2GxSckCnPsVJnmfMZbwFYE+leZGASgkWl06Jow==",
+      "integrity": "sha1-jGOPnRCAgAo1PJwoyKgcpHBdlGw=",
       "dev": true,
       "requires": {
         "ansi-wrap": "0.1.0"
@@ -549,7 +549,7 @@
     "ansi-wrap": {
       "version": "0.1.0",
       "resolved": "https://registry.npmjs.org/ansi-wrap/-/ansi-wrap-0.1.0.tgz",
-      "integrity": "sha512-ZyznvL8k/FZeQHr2T6LzcJ/+vBApDnMNZvfVFy3At0knswWd6rJ3/0Hhmpu8oqa6C92npmozs890sX9Dl6q+Qw==",
+      "integrity": "sha1-qCJQ3bABXponyoLoLqYDu/pF768=",
       "dev": true
     },
     "anymatch": {
@@ -617,7 +617,7 @@
     "array-differ": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/array-differ/-/array-differ-1.0.0.tgz",
-      "integrity": "sha512-LeZY+DZDRnvP7eMuQ6LHfCzUGxAAIViUBliK24P3hWXL6y4SortgR6Nim6xrkfSLlmH0+k+9NYNwVC2s53ZrYQ==",
+      "integrity": "sha1-7/UuN1gknTO+QCuLuOVkuytdQDE=",
       "dev": true
     },
     "array-each": {
@@ -716,7 +716,7 @@
     "assert": {
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/assert/-/assert-1.3.0.tgz",
-      "integrity": "sha512-5aKcpD+XnHpZ7EGxsuo6uoILNh0rvm0Ypa17GlkrF2CNSPhvdgi3ft9XsL2ajdVOI2I3xuGZnHvlXAeqTZYvXg==",
+      "integrity": "sha1-A5OaYiWCqBLMICMgoLmlbJuBWEk=",
       "dev": true,
       "requires": {
         "util": "0.10.3"
@@ -931,7 +931,7 @@
     "bluebird": {
       "version": "3.4.7",
       "resolved": "https://registry.npmjs.org/bluebird/-/bluebird-3.4.7.tgz",
-      "integrity": "sha512-iD3898SR7sWVRHbiQv+sHUtHnMvC1o3nW5rAcqnq3uOn07DSAppZYUkIGslDz6gXC7HfunPe7YVBgoEJASPcHA==",
+      "integrity": "sha1-9y12C+Cbf3bQjtj66Ysomo0F+rM=",
       "dev": true
     },
     "bn.js": {
@@ -1122,7 +1122,7 @@
     "brorand": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/brorand/-/brorand-1.1.0.tgz",
-      "integrity": "sha512-cKV8tMCEpQs4hK/ik71d6LrPOnpkpGBR0wzxqr68g2m/LB2GxVYQroAjMJZRVM1Y4BCjCKc3vAamxSzOY2RP+w==",
+      "integrity": "sha1-EsJe/kCkXjwyPrhnWgoM5XsiNx8=",
       "dev": true
     },
     "browser-pack": {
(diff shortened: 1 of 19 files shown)


EricDunsworth:
Switching this back to a draft for the time being. I installed ``grunt-html`` v15.2.2 just now and ``grunt htmllint`` reported more errors than this PR fixed. I'll try to get around to fixing up the other errors and promoting this out of draft later today. That'll help set the stage for #9333 and a ``grunt-html`` bump PR.

EricDunsworth:
Finally out of draft now :).

Here's a quick overview of all that's changed since last time:
* Rebased
* Bumped ``grunt-html`` dependency to ^15.2.2
* Dropped node 10.x from GitHub Actions (grunt-html 15.0.0 and up require node 12.x at minimum)
* Fixed way more ``htmllint`` errors. Many new ones appeared after updating to ``grunt-html`` 15.2.2 - probably mostly due to errors reliably appearing now (via validator/grunt-html#152 having been fixed).
* Didn't restore the license files' ``data-proofer-ignore`` attributes. The errors that were causing ``html-proofer`` to complain about the license links' ``href`` attributes have been fixed in this PR. I force-pushed a test commit earlier to see how GitHub Actions reacted (couldn't test it locally since ``html-proofer`` is a Ruby gem).
* Rewrote the commit message and OP... refer to them for an outline of specific ``htmllint`` fixes.

src/plugins/lightbox/lightbox-en.hbs
```diff
@@ -210,7 +210,7 @@
 					</li>
 					<li>
 						<a href="demo/2_b.jpg" title="Image 2">
-							<img src="demo/2_s.jpg" alt="Image 2" aria-describedby="image2-desc" longdesc="#image2-desc" />
+							<img src="demo/2_s.jpg" alt="Image 2" aria-describedby="image2-desc" />
```
duboisp, line 213: Can you make this an exception because `longdesc` is defined by https://www.w3.org/TR/html-longdesc/

src/plugins/wb-postback/wb-postback-fr.hbs
```diff
@@ -15,113 +15,113 @@
 
 <p>Pour voir cet example en action, vérifier l'onglet réseaux de la console du fureteur afin de voir la requête Ajax envoyé.</p>
 
-<form action="wb-postback-fr.html" class="wb-postback" data-wb-postback='{"success":".success-message","failure":".failure-message"}'>
+<form action="wb-postback-en.html" class="wb-postback" data-wb-postback='{"success":".success-message","failure":".failure-message","content":".form-content"}'>
 	<div class="form-content">
 		<div class="form-group">
-			<label for="firstname"><span class="field-name">Prénom</span></label>
-			<input class="form-control" id="firstname" name="firstname" type="text" data-rule-minlength="2" />
+			<label for="firstname1"><span class="field-name">First Name</span></label>
```
duboisp, line 21: Keep French content

duboisp (changes requested):

src/plugins/lightbox/lightbox-en.hbs
EricDunsworth: Are you sure you want to go in that direction?

I did a bit more research this afternoon and most of the reputable sources I found downplayed the ``longdesc`` attribute.

**Pros with keeping it:**
* Certain screen readers can detect ``longdesc`` attributes, announce their presence and access them:
  * [WebAIM's ``longdest`` test cases page](https://webaim.org/techniques/alttext/longdesctestcases.htm) shows that NVDA and JAWS supported the attribute in 2015.
  * My slightly-outdated version of NVDA (2020.4) is able to detect it and optionally open a new tab that navigates to ``#image2-desc``.

**Cons with keeping it:**
* The [HTML5 Image Description Extension (longdesc) spec](https://www.w3.org/TR/html-longdesc/) became a W3C recommendation in early 2015. But [the W3C subsequently began to recognize the WHATWG HTML spec as authoritative](https://en.wikipedia.org/wiki/HTML5#W3C_and_WHATWG_conflict) 4 years later in May 2019. Wouldn't that effectively make the ``longdesc`` spec obsolete?
* Both ``htmllint`` and the [W3C's variant of the NU validator](https://validator.w3.org/nu/) treat the ``longdesc`` attribute as obsolete. Here's the specific error they produce:
  > The “longdesc” attribute on the “img” element is obsolete. Use a regular “a” element to link to the description.
* All of the images that use the ``longdesc`` attribute in the lightbox demo also use ``aria-describedby``. In fact, in my NVDA test, I got a ``longdesc`` hint upon navigating to image links... followed by the long description being automatically read out thanks to ``aria-describedby`` (both attributes refer to ``#image2-desc``).
* No one's ever suggested bringing back the ``longdesc`` attribute in [whatwg/html](https://github.com/whatwg/html).
* [PowerMapper removed its``longdesc`` tests in 2019](https://www.powermapper.com/products/sortsite/rules/w3chtml5obsoleteattributelongdesc/) after it became obsolete.
* [Deque University's long descriptions page](https://dequeuniversity.com/tips/long-descriptions) states that the ``longdesc`` attribute isn't usually recommended due to limited screen reader support. Not sure if that predates NVDA and JAWS' support, but I strongly doubt they'll change their stance about it in the future.

**Cons with removing it:**
* Screen readers that still support ``longdesc`` but that don't support ``aria-describedby`` on non-interactive elements won't be able to read the lightbox demo's long descriptions when the lightbox overlays are active. They'll only read out ``aria-describedby``'s long description on the image thumbnail links. Might not be much of a problem in practice though. Users will encounter the thumbnails links first and most screen readers probably wouldn't bother trying to view lightbox image galleries in the first place.

src/plugins/wb-postback/wb-postback-fr.hbs
EricDunsworth: Fixed :).

EricDunsworth:
Just force-pushed the following updates and adjusted my commit message + OP:
* Postback:
  * Add a space character before ``/`` (forward slash) in ``<input>`` tags
* Postback (French version):
  * Move ``aria-hidden="true"`` alert after ``summary`` element

src/plugins/lightbox/lightbox-en.hbs
duboisp: Or can you update the working example that will provide an alternative for the long description.

src/plugins/lightbox/lightbox-en.hbs
```diff
@@ -403,7 +403,7 @@
 					</li>
 					<li>
 						<a href="demo/2_b.jpg" title="Image 2" data-title="image_2_b">
-							<img src="demo/2_s.jpg" alt="Image 2" aria-describedby="image2-desc" longdesc="#image2-desc" />
+							<img src="demo/2_s.jpg" alt="Image 2" aria-describedby="image2-desc" />
 						</a>
 					</li>
```
duboisp, line 408: I am looking for something like this or similar. The goal is to keep the link between the image and it's long description
```suggestion
							<img src="demo/2_s.jpg" alt="Image 2" aria-describedby="image2-desc" />
						</a>
						<p><a href="#image2-desc">View long description for image 2</a></p>
					</li>
```


```suggestion
							<img src="demo/2_s.jpg" alt="Image 2" aria-describedby="image2-desc" />
						</a>
						<details>
						   <summary>View long description for image 2</summary>
						   <p>Description here</p>
						</details>
					</li>
```

src/plugins/lightbox/lightbox-en.hbs
EricDunsworth: Upon further reflection, I think I'll fall back on your initial suggestion and keep ``longdesc`` in place for the time being with an exclusion rule.

What I have in mind for how to resolve this properly will require JS revisions. IMO that goes way beyond this PR's scope and would be better off as a spinoff PR.

src/plugins/lightbox/lightbox-en.hbs
EricDunsworth: I realized yesterday afternoon that the lightbox plugin re-uses image 2's ``img`` element (from the thumbnail links) and dynamically copies it into lightbox popups... So whatever we end up doing will require changes to the plugin itself. IMO that's goes way beyond the scope of this PR.

I'm going to mark this conversation as resolved since I've just updated the PR to exempt lightbox's ``longdesc`` attributes from the ``htmllint`` task.

package-lock.json
```diff
@@ -386,7 +386,7 @@
     "absolute": {
       "version": "0.0.1",
       "resolved": "https://registry.npmjs.org/absolute/-/absolute-0.0.1.tgz",
-      "integrity": "sha512-Zs3Sz3lmEIH76fvnc4jto4CpuI5bhU2TD+jtBYVIfXnjFm+49y1opKxNy4gtNF+xSLBL7hXRpLwypc+o2MzBIQ==",
+      "integrity": "sha1-wigi+H4ck59XmIdQTZwQnEFzgp0=",
```
EricDunsworth, line 389: Looks like node.js v12.11.1 downgraded the integrity hashes of certain pre-existing dependencies from SHA512 to SHA1... might be a show-stopper.

If we push this back to next week I could try updating to a newer version and seeing if this fixes itself.

src/plugins/lightbox/lightbox-en.hbs
duboisp: Now lightbox was put as an exception

duboisp (approved):

duboisp:
CI build has passed, that means no error with npm install and grunt dist

PR #9333, opened 2022-05-06
Build: Add htmllint/bootlint tasks to GH Actions

EricDunsworth:
These tasks are called by grunt dist, but were missing from GitHub Actions. End result was that contributors could technically send in pull requests that introduced HTML and/or Bootlint validation errors that could slip under the radar.

* Spinoff of #9146
* Depends on #9332

C @GormFrank

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,5 +37,7 @@ jobs:
       # - run: grunt dist
       - if: ${{ !env.ACT }}
         run: grunt test
+      - run: grunt htmllint
+      - run: grunt bootlint
       - run: npm test
       - run: rake


EricDunsworth:
I'm glad this PR is passing the CI checks "as-is"... but I was expecting it to fail without #9332.

#9332 fixes ``htmllint`` errors I came across when I ran ``grunt`` and ``grunt htmllint`` in my fork's ``master`` branch on Friday. Today, I tried running ``grunt``, ``grunt docs``, ``grunt htmllint`` and only got a subset of those errors. When I tried running ``grunt dist`` (similar to what GitHub Actions does), I got no errors at all on the ``htmllint`` task. My  ``master`` branch was "clean" and fully up-to-date.

What's happening is that WET uses ``grunt-html`` ^14.3.0, which is afflicted by validator/grunt-html#152 (a bug that causes ``htmllint`` to overwrite its results before reporting them when many files are in play). The bug is fixed in [v15.0.0 and up](https://github.com/validator/grunt-html/releases/tag/v15.0.0) - but those versions are incompatible with node v10.x.

@duboisp @GormFrank If I send in yet another a PR to update grunt-html... would either of you have any concerns with dropping WET's node v10.x support?

duboisp (changes requested):
Great,

Please remove the condition "ACT" if that is not needed. The ACT is a software that emulate Github action locally

EricDunsworth:
Just removed the ``if`` statements :).

Btw I decided against moving ``htmllint`` and ``bootlint`` before ``grunt test`` since I wanted to keep a similar task order as ``grunt dist`` (it calls ``htmllint`` and ``bootlint`` after ``test`` and theme/minification tasks).

**PS:**
I'd recommend merging #9332 before this PR. But all in all, the merging order won't matter in practice since the ``grunt-html`` bug #9332 fixes is preventing GitHub Actions from failing on this PR lol.

duboisp (approved):
Approved but depend on PR #9332 to be merged first

PR #9334, opened 2022-05-10
Overlay - Prevent background content to scroll

Christopher-O:
_(Pick the language of your choice to fill out the following information. / Choisissez la langue de votre choix pour remplir l'information ci-dessous.)_

Adds style to body for overlay [mid-screen, full-screen, full-screen hidden-header, mid-screen with background] in order to remove scrollbar of content behind the overlay when active.  This will avoid the user from scrolling the background content while navigating the overlay content.

### DIFF
--- a/src/plugins/overlay/_base.scss
+++ b/src/plugins/overlay/_base.scss
@@ -90,6 +90,8 @@
 }
 
 .wb-overlay-dlg {
+	overflow: hidden;
+	
 	.overlay-bg {
 		box-shadow: 0 0 1000px 1000px black;
 	}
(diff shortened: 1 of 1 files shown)


duboisp (approved):

Ricokola (changes requested):
Tested Middle Screen Overly locally and noticed that the background still scroll when the overlay is active so this functionality failed.

Only have one scrolling bar when reducing my viewport so this functionality pass.

Ricokola (approved):
Re-tested the changes locally and work as expected so good to merged.

PR #9337, opened 2022-05-12
Release v4.0.50.1

Ricokola:
Release v4.0.50.1


duboisp (changes requested):
package.json need to be updated too

duboisp (approved):

PR #9371, opened 2022-05-24
Forms: Use global variable for asterisk width

EricDunsworth:
* Rename ``$asterisk-width`` to ``$wb-forms-asterisk-width``
* Declare ``$wb-forms-asterisk-width`` as a global SCSS variable to allow other WET themes to override it
* **Note:** The asterisk's width can vary based on font face. The default width isn't suitable for themes that use wider fonts.

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -110,6 +110,12 @@ $wb-so-text-align: right !default;
 $wb-so-padding: 1em 0 0 !default;
 
 
+//== Forms
+//
+// WET Custom
+$wb-forms-asterisk-width: .67em !default; // Asterisk offset for required field labels.
+
+
 //== Tables
 //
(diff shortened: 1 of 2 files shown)


EricDunsworth:
FYI I'm planning to send in a GCWeb PR that'll depend on this. It'll increase the asterisk variable's width a bit to account for GCWeb using a wider font face.

My goal was to send it in prior to the technical review meeting and hopefully get it merged today. But IE11 is acting up (``:before`` pseudo-element's negative margin is affecting its parent element). So it'll probably have to wait until next week.

duboisp (approved):

duboisp (approved):
Approved, I tested locally and the asterisk for the forms still work as expected

PR #9388, opened 2022-06-28
Repo: Drop pre-IE11 support

EricDunsworth:
* Remove:
  * SCSS styles, JS logic, comments and build tasks/configs/dependencies that tied into pre-IE11 versions of Internet Explorer
  * Content dealing with pre-IE11 support or troubleshooting
* Replace:
  * ``dist`` folder's IE8 CSS/JS files with deprecation comments
  * General mentions of IE or pre-IE11 with Edge in content about supported browsers
* Retain:
  * Support for IE version detection in WET core
  * Polyfills (except IE8 SCSS)
  * External dependencies containing some pre-IE11 logic
  * JS comments copied from other sources (i.e. ``jQuery.extend`` in jQuery fixes)
  * Casual mentions of IE (e.g. changelogs, example content, etc...)
  * Anything related to IE11 (or ambiguous IE logic/references that might apply to it)
* Other:
  * Force IE10 and below into basic HTML mode
  * Remove logic dependent on ``wb.ie`` since it's falsy in IE11 (likely a bug... but means none of that logic ever ran in IE11 in the first place)
  * Remove server configuration page... the only content in it that didn't deal with pre-IE11 support was no longer relevant (i.e. configuring SVG MIME types in obsolete IIS versions)
  * Speeds-up SCSS/CSS/JS linters


src/plugins/cal-events/cal-events.js
```diff
@@ -188,16 +188,6 @@ var componentName = "wb-calevt",
...
 				linkId = event.id || wb.getId();
 				event.id = linkId;
-
-				/*
-				 * Fixes IE tabbing error:
-				 * http://www.earthchronicle.com/ECv1point8/Accessibility01IEAnchoredKeyboardNavigation.aspx
-				 */
-
-				// TODO: Which versions of IE should this fix be limited to?
-				if ( wb.ie ) {
```
EricDunsworth, line 198: Removed these kinds of conditions since ``wb.ie`` isn't truthy in IE11 (technically a bug... but as a result IE11 never tried running them).

EricDunsworth:
@duboisp I had a few questions for you regarding this PR. I'll refine it depending on your answers.

1. Should we get rid of irrelevant polyfills instead of neutering them to break their pre-IE11 support? Most of those polyfills were designed over a decade ago for browsers that are completely obsolete in 2023.
2. Should WET drop support for IE11 and pre-Chromium Edge (EdgeHTML)?
3. Should unsupported versions of IE/Edge be forced into basic HTML mode? Looks like [it was already being done for pre-IE7](https://github.com/wet-boew/wet-boew/blob/7e90698884b17de703ae5e7e0d84fd76bc5b04b4/src/plugins/wb-disable/disable.js#L51)...

I'll update the PR depending on what comes out of this.

duboisp:
> Should we get rid of irrelevant polyfills inst

No let's keep it

> Should WET drop support for IE11 and pre-Chromium Edge (EdgeHTML)?

Don't know, we will need to gather a fresh the web statistic first for old Edge version and IE

> Should unsupported versions of IE/Edge be forced into basic HTML mode? 

Yes please

EricDunsworth:
FYI I just re-tested ``wb.ie`` in IE11... it returned ``false``. Checked in [my WET fork on GitHub pages (v4.0.51 - summer 2022)](https://ericdunsworth.github.io/wet-boew/unmin/index-en.html).

[Latest upstream](https://wet-boew.github.io/v4.0-ci/unmin/index-en.html) is currently broken in IE11 due to an [unsupported ``for... of`` condition in the tag filter plugin](https://github.com/wet-boew/wet-boew/blob/f288097b1f807c3a6ca35a9f5a3be82735d1e940/src/plugins/tagfilter/tagfilter.js#L143).

This illustrates why I was asking about killing-off IE11 support. If Microsoft no longer supports it and WET keeps breaking in it due to no one testing in it anymore, IMO there isn't really any point in trying to support it at all. Ditto for most polyfills.

EricDunsworth:
Just forced-pushed a few updates:
* Keep polyfills mostly "as-is" (i.e. restored their pre-IE11 logic... but not IE8 CSS)
* Divert dynamic datalist sync changes into a separate PR (IMO they're out of scope now if polyfills are being kept as-is)
* Force IE10 and below into basic HTML mode
* OP changes:
  * Revise polyfils retention comment
  * Mention that IE10 and below are now forced into basic HTML mode
  * Note that logic tied to ``wb.ie`` was removed since it's falsy in IE11

src/base/details/_ie8.scss
EricDunsworth, line 1: I'll keep this file deleted even though it belongs to a polyfill.

The main driver for this PR is to stop compiling IE8-specific CSS/JS files. So it doesn't make sense to keep this IE8 CSS file if it'll no longer be compiled into anything.

duboisp:
I will do a review

duboisp:
For the record and I didn't check: Should we include a message when we do detect an unsupported browser, like the IE message proposed by pr#8800

EricDunsworth:
Just rebased to fix the merge conflict. Also updated changed modified dates to today.

site/pages/docs/server-en.hbs
```diff
@@ -4,7 +4,7 @@
 	"language": "en",
 	"description": "Instructions for configuring the server.",
 	"altLangPrefix": "server",
-	"dateModified": "2014-10-15"
+	"dateModified": "2023-06-05"
 }
 ---
```
duboisp, line 10: I think we could remove this page completely as it seems to only discuss about old IE issue.

duboisp (dismissed):
Third way through the review.

Todo:
* [x] Review and confirm the JS code change
* [x] Test every impacted plugin to ensure they still work as expected.

site/pages/docs/server-en.hbs
EricDunsworth: That page's scope goes slightly beyond IE, but I agree that it's irrelevant at this point:

* [IIS 6.0-7.x are no longer supported.](https://learn.microsoft.com/en-us/lifecycle/products/internet-information-services-iis)... wouldn't surprise me if newer versions have SVG as a built-in MIME type.
* WET/GCWeb's page templates stopped using IE conditional comments long ago
* IE11 never supported the ``X-UA-Compatible`` flag
* IE compatibility mode is no longer a thing in Edge Chromium
* Edge Chromium's "IE mode" (IE11 Trident mode) can't be used for debugging... Edge's devtools are disconnected in that mode, making it impossible to view IE11's console if anything breaks

EricDunsworth: Just scrapped this page and links that went towards it.

Garneauma (approved):
Reviewed all updated pages and plugins. Everything is working as it was.

EricDunsworth:
I wouldn't consider these to be show-stoppers, but just wanted to re-iterate them here:
* Putting older versions of IE into basic HTML mode sounds good in theory (and is already being done for IE7)... but probably doesn't work in practice. Why? Because wet-boew.js contains most of WET's JS logic... and IIRC IE is prone to run into parsing errors on newer JS syntax. End result is that those browsers might fail immediately, long before they encounter basic HTML mode's logic.
* IE11 should probably be dropped too, but that can wait until a future PR.

duboisp:
This PR has been pre-approved during the wet-boew technical review.

PR #9394, opened 2022-07-18
Country content: Fix plugin... again

EricDunsworth:
* Replace all references to [freegeoip.app](https://freegeoip.app/) (which stopped working in late April) with [country.is](https://country.is/) and [api.country.is](https://api.country.is/)
* Rename ``country_code`` JSON key to ``country``
* Overview of [country.is](https://country.is/):
  * Has existed since 2014-2015
  * Site claims not to collect request logs
  * Doesn't have any usage limits according to hakanensari/country#17
  * Discovered it in [public-apis/public-apis#geocoding](https://github.com/public-apis/public-apis#geocoding)
* Successor to #8928

PR #9395, opened 2022-07-20
Rakefile: Update HTMLProofer options for v4

EricDunsworth:
* Rename ``url_ignore`` to ``ignore_urls``
* Add ``allow_missing_href`` and set it to ``true`` (suppresses "'a' tag is missing a reference" errors on JSON manager and data JSON templates)
* Add ``enforce_https`` and set it to ``false`` (suppresses "* is not an HTTPS link" errors on HTTP links... might be possible to update some to HTTPS, but many of them are literally HTTP-only)
* Remove ``empty_alt_ignore`` (renamed to ``ignore_missing_alt`` in v4, but WET no longer needs it)
* Remove ``ext`` (renamed to ``extensions`` in v4, isn't needed in practice since the default extension is already ".html")

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,7 +6,7 @@ task :default => [:test]
 
 task :test do
 	HTMLProofer.check_directory("dist/", {
-		:url_ignore => [
+		:ignore_urls => [
 			"#",
 			# The additional anchor link is picked up from the Geomap JSON, but shouldn't be flagged
 			"\\\"#\\\"",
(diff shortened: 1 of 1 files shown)


EricDunsworth:
See https://github.com/wet-boew/wet-boew/pull/9394#issuecomment-1190478191 for extra context.

Rakefile
```diff
@@ -20,8 +20,8 @@ task :test do
 			"texthighlight/texthighlight-fr.html?txthl=influenza%20aviaire+monde+suffis+symptômes%20semblables%20à%20ceux%20de%20l'influenza+À%20titre%20de%20rappel...+À%20de%20rares%20occasions,%20des%20humains%20ont%20été%20infectés%20par%20ce%20virus.",
 			"../../../demos/texthighlight/texthighlight-en.html?txthl=influenza%20aviaire+monde+suffis+sympt%C3%B4mes%20semblables%20%C3%A0%20ceux%20de%20l'influenza+%C3%80%20titre%20de%20rappel...+%C3%80%20de%20rares%20occasions,%20des%20humains%20ont%20%C3%A9t%C3%A9%20infect%C3%A9s%20par%20ce%20virus.#exemple"
 		],
+		:allow_missing_href => true,
 		:disable_external => true,
-		:empty_alt_ignore => true,
-		:ext => ".html"
+		:enforce_https => false
```
GormFrank, line 25: @EricDunsworth Would be great to eventually convert our http links to https / replace them by others instead of ignoring them. I'll create an issue so we track it but please @EricDunsworth feel free to open it if you feel like it :) Pretty sure you'll be quicker than me.

GormFrank (approved):
This is great, thanks :)

Rakefile
duboisp: @GormFrank : Eric already did that change a long time ago. Here, I do think it is because this test are running on a local http server.

PR #9397, opened 2022-07-22
Site: Update more HTTP links to HTTPS

EricDunsworth:
* Update links to HTTP sites that now support HTTPS and/or redirect to new locations on HTTPS sites
* Change links to HTTP sites that no longer exist to point towards Wayback Machine mirrors
* Spinoff of #9395's review comment
* Successor to #8674

CC @GormFrank @duboisp

### DIFF
--- a/site/pages/docs/opt-en.hbs
+++ b/site/pages/docs/opt-en.hbs
@@ -111,7 +111,7 @@ console.log( ( new Date() ).getTime() - test );</code></pre>
 		<h3>Additional techniques</h3>
 		<ul>
 			<li><a rel="external" href="https://www.queness.com/post/588/15-ways-to-optimize-css-and-reduce-css-file-size">15 Ways to Optimize CSS and Reduce CSS File Size</a></li>
-			<li><a rel="external" href="http://www.dailycoding.com/Posts/top_7_tip_for_optimizing_css.aspx">Top 7 Tip for Optimizing CSS</a></li>
+			<li><a rel="external" href="https://web.archive.org/web/20190119144722/http://www.dailycoding.com/Posts/top_7_tip_for_optimizing_css.aspx">Top 7 Tip for Optimizing CSS</a></li>
 			<li><a rel="external" href="https://perishablepress.com/how-to-micro-optimize-your-css/">How to Micro-Optimize Your CSS]</a></li>
 		</ul>
 	</section>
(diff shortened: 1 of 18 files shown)


site/layouts/core.hbs
```diff
@@ -8,7 +8,7 @@
 		{{>metadata}}
 		{{>headresources}}
 	</head>
-	<body {{#if pageclass}}class="{{pageclass}}" {{/if}}vocab="http://schema.org/" resource="#wb-webpage" typeof="WebPage">
+	<body {{#if pageclass}}class="{{pageclass}}" {{/if}}vocab="https://schema.org/" resource="#wb-webpage" typeof="WebPage">
```
EricDunsworth, line 11: I was previously asked to revert to http://schema.org/ in #8674's review comments due to the [W3C's RDFa initial context spec](https://www.w3.org/2011/rdfa-context/rdfa-1.1).

The W3C spec's schema.org URI protocol hasn't changed since then... but schema.org itself started using HTTPS in ``vocab="https://schema.org/"`` in [all of its documentation's examples](https://schema.org/docs/gs.html) as of [version 12.0](https://schema.org/docs/releases.html). The W3C's spec was last updated on 2020-05-09, whereas schema.org v12 was released almost a year later on 2021-03-08.

I can revert it back to HTTP again if needed, but don't see any value in continuing to hold off from using HTTPS just on account of the W3C spec having fallen behind.

GormFrank: @duboisp What's your take on this? I don't see any issues switching to https but it will definitely have domino effect all the way up to Canada.ca (in terms of work intake).

duboisp: Use the HTTP protocol for schema.org when referring it as a vocabulary. 

FYI- https://www.w3.org/TR/cooluris/#semweb

EricDunsworth: I'll revert it, but still personally disagree with going backwards to HTTP given that schema.org itself agrees with my stance.

IMO outdated W3C specs shouldn't take precedence over schema.org itself (although [schema.org is technically ok with both HTTP and HTTPS variants of its URL](https://schema.org/docs/faq.html#19)).

Regarding https://www.w3.org/TR/cooluris/#semweb, it's more than a decade old (last updated in 2008) and doesn't specifically advise using HTTP over HTTPS. HTTPS isn't even mentioned anywhere in it apart from the URLs of a couple of links in the reference section (probably because HTTPS wasn't very common back then).

The reality of the web in 2022 is that HTTPS is used by default on the (vast?) majority of modern-day sites. That couldn't of been predicted back in 2008. In fact, some of the vocabulary prefixes listed in the [W3C's RDFa core initial context](https://www.w3.org/2011/rdfa-context/rdfa-1.1) page use HTTPS. So even the W3C itself is moving forward with the times.

EricDunsworth:
Just force-pushed a PR that reverted ``vocab="https://schema.org/"`` to ``vocab="http://schema.org/"`` in all the files I applied that change to.

GormFrank (approved):
Maybe eventually it should be worth looking into adding a marker besides hyperlinks to the wayback machine, to indicate that those destination pages are archived / removed.

Issue #9123, opened 2021-06-15
[BUG] Multimedia player (YouTube) - conflict with Doaction plugin


Christopher-O:
## Required information

**Describe the bug**
When running a doaction plugin with the multimedia plugin importing a YouTube video, the doaction plugin does not work.  It seems the script for the YouTube video is conflicting or not at a ready state.

**To Reproduce**
I have provided an example to show that it does not work with the YouTube video render
http://test.canada.ca/cra/bugs/video-yt-render.html 
(When pressing the button the doaction should addClass "bg-info" to the text).

When adding a 1 to the class wb-mltmd to the element that has the youtube video in the following example (thus making the youtube video not render), the doaction works as expected
http://test.canada.ca/cra/bugs/video-yt-norender.html

**Expected behavior**
The YouTube video and doaction should work independently on the page.  I have provided an example where I have added a wb.ready element `wb.ready($(eventTarget), componentName)` to the inpage multimedia player script, changed the component class name so there is no conflicts and now both the YouTube video player and the doaction function as normal
http://test.canada.ca/cra/bugs/video-yt-render-fix.html  

**Who shall do the work?**
I would like Principle Publisher or someone to review if my fix is the correct course of action, or if there is another way to resolve the conflicts

EricDunsworth:
Good catch :)! I also ran across this bug a few weeks ago and was trying to troubleshoot it recently to no avail.

@dubbs @GormFrank Thoughts?

Christopher-O:
I just tested my "fix" and realized that it's not the correct solution, as it makes wb.ready "undefined" rather than true or false, which is probably why it just renders both scripts.  So it will need a more stable fix.

Christopher-O:
A somewhat related (depending on the issue with whatever script is causing the issue), but when using multiple source files in a video element, one from youtube and one hosted locally, in case the user is blocked from YouTube (e.g. a government worker), the alternative file does not play, so a local file cannot be used as a fallback if YouTube is blocked.  I don't know if the mp4 files were not intended to be a fallback to YouTube or not...but it would be a nice way to support people who have blocked access, but still utilizing the more optimal streaming of YouTube's servers.

EricDunsworth:
@Christopher-O Might be a better bet to create a separate issue for that since it's unrelated to this issue's bug.

But I'll chime in any case :).

My take on setting a video's sources to a mix of YouTube and real video formats is that it'd make sense to support that scenario since YT was implemented as a "fake" source type.

[...]

PR #9400, opened 2022-07-28
multimedia: renderUIEvent - load Modernizr without return

polmih:

Multimedia player with youtube video and Do action plugin conflict together on the same page.

This issue is present only loading the youtube video, multimedia will never call `wb.ready( $this, componentName ) `which is needed for the do-action plugin to trigger its action.

A solution for that is to call wb.ready() once the onReady event is fired, meaning that the player has finished loading and is ready to begin receiving API calls (play the video, etc.);

For the scenario when youtube is blocked in a corporate network, the solution would be to check if iframe api ressource is loaded then check if youtubeEvent has fired, if not call wb.ready() after a second timeout, although this solution will not reduce the delay time for Modernizr to load the api. @duboisp , @EricDunsworth , any thoughts on how to tackle this?  :)

Related to WET-218


duboisp (changes requested):

EricDunsworth:
FWIW this might fix #9123.

src/plugins/multimedia/multimedia.js
```diff
@@ -679,7 +679,7 @@ $document.on( initializedEvent, selector, function( event ) {
 			}
 
 			// finally lets load safely
-			return Modernizr.load( {
+			Modernizr.load( {
```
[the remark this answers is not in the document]
EricDunsworth: There might also be value in implementing a failure/timeout check of sorts in case the YouTube API request fails. Like if YouTube is blocked in a corporate network. That would prevent other plugins (like the do action plugin - which GCWeb's page feedback widget depends on) from breaking on pages that contain YouTube videos in those kinds of networks.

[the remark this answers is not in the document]
polmih: wb.ready() is now called on the onReady event. Works fine on unblocked networks.
on blocked networks, this solution works too, but there is a big delay in loading the YT api with Modernizr

EricDunsworth:
Regarding this:
> For the scenario when youtube is blocked in a corporate network, the solution would be to check if iframe api ressource is loaded then check if youtubeEvent has fired, if not call wb.ready() after a second timeout, **although this solution will not reduce the delay time for Modernizr to load the api.** @duboisp , @EricDunsworth , any thoughts on how to tackle this? :)

I tried prepping a similar PR to this one around 1-2 months ago ago but was unable to get it to work to my liking (and my setup made it a pain to test). So I never got around to sending it.

IIRC when the YouTube API fails to load, Modernizr (or yepnope) waits for 5-10 seconds before reaching its ``complete``/``callback`` events. I didn't try validating this, but it's probably because when the firewall returns a warning page instead of the API, it falsely returns an HTTP 200 status code... tricking Modernizr into thinking it retrieved what it was looking for and waiting for something to happen until eventually timing out. But its timeout limit can be overridden.

Here's some old documentation I previously dug up that covers the iteration of Modernizr WET 4.0 uses:
* https://web.archive.org/web/20140625052645/http://modernizr.com/docs/
* https://web.archive.org/web/20140612053652/http://yepnopejs.com/ (older versions of Modernizr used to leverage yepnope)
* http://html5doctor.com/using-modernizr-to-detect-html5-features-and-provide-fallbacks/

IMO it'd be cool to find a way of running a test condition immediately after Modernizr loads *something* in (whether it be the YT API itself or the firewall's warning page) instead of testing users' patience with timeouts that need to be long-enough to accommodate slow connections. But WET's version of Modernizr isn't that sophisticated and I was unable to come up with a workaround in my attempts :(.

duboisp:
This API might help to find out if Youtube work or not: performance.getEntries()

We can also make a fetch "HEAD" http call to see if the domain is reachable.

src/plugins/multimedia/multimedia.js
```diff
@@ -680,19 +680,29 @@ $document.on( initializedEvent, selector, function( event ) {
...
+				load: "https://www.youtube.com/iframe_api",
+
+				//possible solution for multimedia and doaction conflict in corporate network
+				complete: function() {
+					setTimeout( function() {
+						if ( $this.hasClass( "youtube" ) === false ) {
+							console.warn( "There might be an issue loading the Youtube API" );
+							wb.ready( $this, componentName );
+						}
+					}, 1500 );
```
duboisp, line 692: Try to see if there is an element "iframe" initiator from the `window.performance.getEntries()`

duboisp (changes requested):

src/plugins/multimedia/multimedia.js
```diff
@@ -680,19 +681,47 @@ $document.on( initializedEvent, selector, function( event ) {
...
+
+						} );
+
+
+						/* if none found, most probably wb is loaded in restricted network so wb.ready() is triggered for not preventing other wb components to load*/
+						if ( arrIframesYt.length < 1 ) {
+							if ( !wb.isReady ) {
+
+								// show the video notification error
+								$this.prepend( "<div id='" + id + "-notif-text" + "' aria-live='polite' class='pstn-lft-xs bg-dark text-white'><p class='mrgn-tp-md mrgn-rght-md mrgn-bttm-md mrgn-lft-md'>" + i18n( "msgYoutubeNotLoad" ) + "</p></div>" );
```
duboisp, line 705:
```suggestion

								$this.data().notifText = $this.prepend( "<div  aria-live='polite' class='pstn-lft-xs bg-dark text-white'><p class='mrgn-tp-md mrgn-rght-md mrgn-bttm-md mrgn-lft-md'>" + i18n( "msgYoutubeNotLoad" ) + "</p></div>" );
```

src/plugins/multimedia/multimedia.js
```diff
@@ -719,15 +748,22 @@ $document.on( youtubeEvent, selector, function( event, data ) {
 			},
 			events: {
 				onReady: function( event ) {
+					$( "#" + $this.context.id + "-notif-text" ).hide();
```
duboisp, line 751:
```suggestion
					$this.data().notifText.hide();
```

duboisp (changes requested):

src/plugins/multimedia/multimedia.js
```diff
@@ -680,19 +681,47 @@ $document.on( initializedEvent, selector, function( event ) {
 
 			// finally lets load safely
 			return Modernizr.load( {
-				load: "https://www.youtube.com/iframe_api"
+				load: "https://www.youtube.com/iframe_api",
+
+				//possible solution for multimedia and doaction conflict in corporate network
+				complete: function() {
+					setTimeout( function() {
+						var resources = window.performance.getEntriesByType( "resource" );
```
duboisp, line 689: Declare variable at the beginning of the function.

```suggestion
						var resources, arrIframesYt;

						resources = window.performance.getEntriesByType( "resource" );
```

duboisp (changes requested):

src/plugins/multimedia/multimedia.js
```diff
@@ -680,19 +681,47 @@ $document.on( initializedEvent, selector, function( event ) {
 
 			// finally lets load safely
 			return Modernizr.load( {
-				load: "https://www.youtube.com/iframe_api"
+				load: "https://www.youtube.com/iframe_api",
+
+				//possible solution for multimedia and doaction conflict in corporate network
```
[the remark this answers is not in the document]
polmih: Thanks for the suggestion. We will tackle this down later

src/plugins/multimedia/multimedia.js
polmih: Shortly different approach applied.

src/plugins/multimedia/multimedia.js
polmih: Applied. Thanks

src/plugins/multimedia/multimedia.js
```diff
@@ -680,19 +681,50 @@ $document.on( initializedEvent, selector, function( event ) {
...
+				complete: function() {
+					setTimeout( function() {
+						var resources, arrIframesYt, notifText;
```
duboisp, line 689: We prefix our variable that contain jQuery object with "$" 
```suggestion
						var resources, arrIframesYt, $notifText;
```

duboisp (changes requested):
Approved at the technical review once the editorial code change are applied.

src/plugins/multimedia/multimedia.js
polmih: Done. Thanks

src/plugins/multimedia/multimedia.js
```diff
@@ -680,19 +681,50 @@ $document.on( initializedEvent, selector, function( event ) {
...
+
+						resources = window.performance.getEntriesByType( "resource" );
+
+						/* get all the iframe initiators that have the same YT url id */
+						arrIframesYt  = resources.filter( function( obj ) {
+
+							//return obj.initiatorType === "iframe" && obj.name.includes( data.youTubeId );
```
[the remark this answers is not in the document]
polmih: Done. Thanks

src/plugins/multimedia/multimedia.js
```diff
@@ -680,19 +681,50 @@ $document.on( initializedEvent, selector, function( event ) {
 
 			// finally lets load safely
 			return Modernizr.load( {
-				load: "https://www.youtube.com/iframe_api"
+				load: "https://www.youtube.com/iframe_api",
+
+				//possible solution for multimedia and doaction conflict in corporate network
+				complete: function() {
+					setTimeout( function() {
```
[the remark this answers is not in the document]
polmih: Done. Thanks

duboisp (approved):
Requested changes was completed.
I also did a local test by blocking the URL youtube.com in my dev console and the message was successfully displaying.

PR #9401, opened 2022-07-29
Add new plugin: Tag filter

Garneauma:
Adding new plugin called Tag filter. Similar to Filter, but using checkboxes instead of a text box, with plans to expand to radio buttons and select tags.

Based off of @duboisp's initial idea and hard work: https://github.com/wet-boew/wet-boew/pull/8464.

### DIFF
--- a/src/plugins/tagfilter/_base.scss
+++ b/src/plugins/tagfilter/_base.scss
@@ -1,0 +1,3 @@
+.wb-tgfltr-out {
+	display: none !important;
+}
--- a/src/plugins/tagfilter/tagfilter-en.hbs
+++ b/src/plugins/tagfilter/tagfilter-en.hbs
@@ -1,0 +1,1195 @@
+---
+{
+"title": "Tag filter",
+"language": "en",
+"category": "Plugins",
+"description": "Filters through content and shows/hides content that match a certain tags.",
+"tag": "filter",
+"parentdir": "tagfilter",
+"altLangPrefix": "tagfilter",
+"dateModified": "2022-09-10"
+}
+---
+
+<div class="alert alert-warning">
+	<h2>Unstable feature</h2>
+	<p>To be used at <strong>your own risk</strong>. This feature described below can be removed in any subsequent
+		minor/major release</p>
+	<p><small><a href="https://wet-boew.github.io/wet-boew-documentation/decision/7.html">Learn more about the design
+				decision around provisional features.</a></small></p>
+	<p><small><a href="../../docs/ref/provisional-en.html">Check other provisional features available.</a></small>
+	</p>
+</div>
+
+<span class="wb-prettify all-pre"></span>
+
+<h2>On this page:</h2>
+<ul>
+	<li><a href="#filteredEventsList">Example 1 - Events list</a></li>
+	<li><a href="#filteredAppsList">Example 2 - Applications list</a></li>
+	<li><a href="#filteredFAQ">Example 3 - FAQ leveraging Filter plugin</a></li>
+	<li><a href="#filteredTable">Example 4 - Table</a></li>
+</ul>
+
+<section id="filteredEventsList" class="wb-tagfilter provisional">
+	<h2>Example 1 - Events list</h2>
+	<div class="row">
+		<div class="col-xs-12 col-sm-4 col-lg-3">
+			<details open>
+				<summary>
+					<h3>Filter options</h3>
+				</summary>
+				<div class="form-group mrgn-bttm-0">
+					<fieldset>
+						<legend class="h5 mrgn-tp-0"><span class="field-name">Language</span></legend>
+						<ul class="list-unstyled">
+							<li class="radio">
+								<label>
+									<input type="radio" name="language" class="wb-tagfilter-ctrl" value="" checked> All
+								</label>
+							</li>
+							<li class="radio">
+								<label>
+									<input type="radio" name="language" class="wb-tagfilter-ctrl" value="language__french"> French
+								</label>
+							</li>
+							<li class="radio">
+								<label>
+									<input type="radio" name="language" class="wb-tagfilter-ctrl" value="language__english"> English
+								</label>
+							</li>
+						</ul>
+					</fieldset>
+				</div>
+				<div class="form-group mrgn-bttm-0">
+					<fieldset>
+						<legend class="h5 mrgn-tp-0"><span class="field-name">Event location</span></legend>
+						<ul class="list-unstyled">
+							<li class="checkbox">
+								<label>
+									<input type="checkbox" name="location" class="wb-tagfilter-ctrl" value="location__montreal"> Montreal
+								</label>
+							</li>
+							<li class="checkbox">
+								<label>
+									<input type="checkbox" name="location" class="wb-tagfilter-ctrl" value="location__online"> Online
+								</label>
+							</li>
+							<li class="checkbox">
+								<label>
+									<input type="checkbox" name="location" class="wb-tagfilter-ctrl" value="location__ottawa"> Ottawa
+								</label>
+							</li>
+							<li class="checkbox">
+								<label>
+									<input type="checkbox" name="location" class="wb-tagfilter-ctrl" value="location__toronto"> Toronto
+								</label>
+							</li>
+							<li class="checkbox">
+								<label>
+									<input type="checkbox" name="location" class="wb-tagfilter-ctrl" value="location__vancouver"> Vancouver
+								</label>
+							</li>
+						</ul>
+					</fieldset>
+				</div>
+				<div class="form-group mrgn-bttm-0">
+					<label for="dayList">Day of the week</label>
+					<select id="dayList" name="dayOTW" class="form-control full-width wb-tagfilter-ctrl">
+						<option value="">All</option>
+						<option value="dayOTW_monday">Monday</option>
+						<option value="dayOTW_tuesday">Tuesday</option>
+						<option value="dayOTW_wednesday">Wednedsday</option>
+						<option value="dayOTW_thursday">Thursday</option>
+						<option value="dayOTW_friday">Friday</option>
+						<option value="dayOTW_saturday">Saturday</option>
+						<option value="dayOTW_sunday">Sunday</option>
+					</select>
+				</div>
+			</details>
+		</div>
+		<div class="col-xs-12 col-sm-8 col-lg-9">
+			<ul class="list-unstyled events-list wb-tagfilter-items">
+				<li data-wb-tags="language__english language__french location__online dayOTW_friday">
+					<h3>Ask PAC Anything!</h3>
+					<p>Friday August 5th, 2022 | 12:00 PM to 1:00 PM</p>
+					<p><span><strong>Location:</strong> Online</span> <strong>Language</strong>: English, French</p>
+				</li>
+				<li data-wb-tags="language__english location__vancouver dayOTW_wednesday">
+					<h3>Bidding on Opportunities</h3>
+					<p>Wednesday August 10th, 2022 | 2:00 PM to 3:30 PM</p>
+					<p><span><strong>Location:</strong> Vancouver</span> <strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__english location__ottawa dayOTW_thursday">
+					<h3>Doing business with the Government of Canada (Webinar)</h3>
+					<p>Thursday August 11th, 2022 | 12:00 PM to 2:00 PM</p>
+					<p><span><strong>Location:</strong> Ottawa</span> <strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__english location__online dayOTW_friday">
+					<h3>Ask PAC Anything!</h3>
+					<p>Friday August 12th, 2022 | 12:00 PM to 1:00 PM</p>
+					<p><span><strong>Location:</strong> Online</span> <strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__english location__ottawa dayOTW_thursday">
+					<h3>Bidding on opportunities (Webinar)</h3>
+					<p>Thursday August 18th, 2022 | 12:00 PM to 2:00 PM</p>
+					<p><span><strong>Location:</strong> Ottawa</span> <strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__english location__online dayOTW_friday">
+					<h3>Ask PAC Anything!</h3>
+					<p>Friday August 19th, 2022 | 12:00 PM to 1:00 PM</p>
+					<p><span><strong>Location:</strong> Online</span> <strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__french location__ottawa dayOTW_thursday">
+					<h3>Workshop to register in the Supplier Registration Information (SRI) system and in the Indigenous ness Directory (IBD) for Indigenous-led businesses</h3>
+					<p>Thursday August 25th, 2022 | 9:30 AM to 11:00 AM</p>
+					<p><span><strong>Location:</strong> Ottawa</span> <strong>Language</strong>: French</p>
+				</li>
+				<li data-wb-tags="language__english location__ottawa location__montreal dayOTW_thursday">
+					<h3>Workshop to register in the Supplier Registration Information (SRI) system and in the Indigenous ness Directory (IBD) for Indigenous-led businesses</h3>
+					<p>Thursday August 25th, 2022 | 1:00 PM to 2:30 PM</p>
+					<p><span><strong>Location:</strong> Ottawa and Montreal</span> <strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__french location__montreal dayOTW_thursday">
+					<h3>Doing Business with the Government of Canada</h3>
+					<p>Thursday August 25th, 2022 | 1:30 PM to 3:00 PM</p>
+					<p><span><strong>Location:</strong> Montreal</span> <strong>Language</strong>: French</p>
+				</li>
+				<li data-wb-tags="language__english location__vancouver dayOTW_thursday">
+					<h3>Doing Business with the Government of Canada</h3>
+					<p>Thursday August 25th, 2022 | 2:00 PM to 3:30 PM</p>
+					<p><span><strong>Location:</strong> Vancouver</span> <strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__english location__online dayOTW_friday">
+					<h3>Ask PAC Anything!</h3>
+					<p>Friday August 26th, 2022 | 12:00 PM to 1:30 PM</p>
+					<p><span><strong>Location:</strong> Online</span> <strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__english location__toronto dayOTW_tuesday">
+					<h3>Mythbusting Federal Government Procurement and Overview of Innovation for Defence, Excellence and rity (IDEaS) </h3>
+					<p>Tuesday August 30th, 2022 | 10:00 AM to 11:00 AM</p>
+					<p><span><strong>Location:</strong> Toronto</span> <strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__french location__montreal dayOTW_tuesday">
+					<h3>Register as supplier and find opportunities on BuyAndSell.gc.ca</h3>
+					<p>Tuesday August 30th, 2022 | 1:30 PM to 3:00 PM</p>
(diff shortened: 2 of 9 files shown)


GormFrank:
@Garneauma Please co-author @duboisp :)

src/plugins/tagfilter/tagfilter.js
```diff
@@ -0,0 +1,122 @@
...
+		relatedExclusiveRem = [],
+		i, i_len, j, j_len,
+		currentElm, tagList, tag, lastIndex;
+
+	// Filter down the ones that matched the initial DOM search
+	i_len = relatedPotential.length;
+	for ( i = 0; i < i_len; i = i + 1 ) {
+		currentElm = relatedPotential[ i ];
+
+		tagList = currentElm.dataset.wb5Tags.split( " " );
```
GormFrank, line 60: @Garneauma Please call it `wbTags`:
```suggestion
		tagList = currentElm.dataset.wbTags.split( " " );
```

GormFrank (changes requested):
@Garneauma This was done in the beginning of WET5 talks, so wb5 was used, but we don't want it necessarily nowadays.

Remarque : je me demande rapidement ce que le `wb-tgfltr-persistent` fait.

Ricokola (changes requested):
I noticed some strange behavior during the testing exercise.

Test case 1
1.   Uncheck the "Odd Numbers" and "Even numbers" so every numbers disappear
2.  Check and them uncheck the "Hide multiples of 3".
Noticed that the multiples of 3 suddenly appears even if no numbers where showing.

Test case 2
1.  Uncheck the "Odd Numbers" and "Even numbers" so every numbers disappear
2.  Check and them uncheck the "Hide multiples of 3"
3.  Check the "Odd Numbers"
Noticed that the odd numbers showing but also the even numbers that are multiples of 3

I question the fact that some filter options are checked by default on page load. Would it be more accurate to have no filter checked by default and having all the options showing on page load and then the user select filters to narrow down to the desired results?

Now is the problem is from the initial state or from the conditional process in the script?

GormFrank:
> > @Garneauma Please co-author @duboisp :)
> 
> Already done. It is done at the commit level.

There is probably something wrong with having the "@" in front of his username, because the commit should be showing two people just like it does here: https://github.com/wet-boew/GCWeb/pull/1837/commits

Garneauma:
> I noticed some strange behavior during the testing exercise.
> 
> Test case 1
> 
> 1. Uncheck the "Odd Numbers" and "Even numbers" so every numbers disappear
> 2. Check and them uncheck the "Hide multiples of 3".
>    Noticed that the multiples of 3 suddenly appears even if no numbers where showing.
> 
> Test case 2
> 
> 1. Uncheck the "Odd Numbers" and "Even numbers" so every numbers disappear
> 2. Check and them uncheck the "Hide multiples of 3"
> 3. Check the "Odd Numbers"
>    Noticed that the odd numbers showing but also the even numbers that are multiples of 3
> 
> I question the fact that some filter options are checked by default on page load. Would it be more accurate to have no filter checked by default and having all the options showing on page load and then the user select filters to narrow down to the desired results?
> 
> Now is the problem is from the initial state or from the conditional process in the script?

Interesting. I will take a look at it.

As for the default state, it might be worth discussing as a team. It was designed like this by @duboisp, but we can discuss how to better it.

CC @GormFrank

src/plugins/tagfilter/_base.scss
```diff
@@ -0,0 +1,25 @@
+/**
+ * @title Tag filter CSS
+ * @license wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
+ * @author @duboisp
+ *
+ * Note: Depend also on the CSS for the filter WET plugin
+ */
+
+.wb-tagfilter.provisional {
+	.wb-tgfltr-out {
```
duboisp, line 10: Side note:
Originally the intend was to also extend the filter plugin to display persistent faded text when text was hidden. We can look at that later on.

duboisp (changes requested):
Question:

Should exclusive tag selection be controlled by the tagged elements or by the controller? 
I do think it might make more sense if that are defined at the controller level considering it will also contains instruction saying that it toggle exclusive selection. It's similar to the "only" button in the "How to meet WCAG - Quickref"

For the reverse tag (with the exclamation mark prefix) it makes senses to have it tagged at the element level. Open question: Should the reverse tag be used instead of the persistent one? How their behaviour differ from each other. Can one be faded and the other one not?

duboisp:
@Garneauma 

> > > > @Garneauma Please co-author @duboisp :)
> > > 
> > > 
> > > Already done. It is done at the commit level.
> > 
> > There is probably something wrong with having the "@" in front of his username, because the commit should be showing two people just like it does here: https://github.com/wet-boew/GCWeb/pull/1837/commits
> 
> The commit message has the email address of the co-author, so Github can find it. I'm not sure what's @duboisp's email address used on Github, so I couldn't do that...
> 
> Edit: Found a hack to find it

For the co-authored, please use my corresponding noreply Github email like: duboisp@users.noreply.github.com. Or use my email explicitly attached to the original commit which you can see beside "Author" when you run `git log` command.

Ricokola (dismissed):
Seems all good to me! Pre-approved upon clarification of the French link.

Ricokola (dismissed):
All good!

src/plugins/tagfilter/tagfilter-en.hbs
```diff
@@ -0,0 +1,1035 @@
...
+	<p>To be used at <strong>your own risk</strong>. This feature described below can be removed in any subsequent
+		minor/major release</p>
+	<p><small><a href="https://wet-boew.github.io/wet-boew-documentation/decision/7.html">Learn more about the design
+				decision around provisional features.</a></small></p>
+	<p><small><a href="../../../docs/ref/provisional-en.html">Check other provisional features available.</a></small>
+	</p>
+</div>
+
+<span class="wb-prettify all-pre"></span>
+
```
GormFrank, line 25: Can you add a "On this page:" H2 here? (et en français : "Sur cette page :")

GormFrank (changes requested):
@Garneauma Love the number of comments and `console.warn`, this makes for great quality product!

Minor changes required + Just wondering if there should be any aria-controls and/or aria-live attributes somewhere for accessibility purposes?

Garneauma:
> Minor changes required + Just wondering if there should be any aria-controls and/or aria-live attributes somewhere for accessibility purposes?

Added aria controls on the form controls and aria-live on the tagged items container.

GormFrank (dismissed):
Small changes

src/plugins/tagfilter/tagfilter-en.hbs
```diff
@@ -0,0 +1,1278 @@
...
+					<legend class="h5 mrgn-tp-0"><span class="field-name">Platform</span></legend>
+					<ul class="list-unstyled list-inline">
+						<li class="checkbox">
+							<label for="plat__ios">
+								<input type="checkbox" name="platform" id="plat__ios" class="wb-tagfilter-ctrl" value="plat__ios"> iOS
+							</label>
+						</li>
+						<li class="checkbox">
+							<label for="plat__android">
+								<input type="checkbox" name="platform" id="plat__android" class="wb-tagfilter-ctrl" value="plat__android"> Android
```
duboisp, line 396: What is the risk of having this filter control to be mixed with an external form control? Like have 2 control sibling with different purpose?

If all controls is about tag filter, then we could only identify that section from a higher common parent.

duboisp (changes requested):
I did a quick code review and we should try to reduce the use of CSS to modify orignal component styling and to try to simplify a little bit more how that plugin can be use. See inline comments for more details.

src/plugins/tagfilter/tagfilter.js
```diff
@@ -0,0 +1,281 @@
...
+			elm.activeFilters = [];
+
+			// Get all form inputs (radio buttons, checkboxes and select) within filters form
+			const filterControls = document.querySelectorAll( "#" + elm.id + " .wb-tagfilter-ctrl" ),
+				taggedItems = document.querySelectorAll( "#" + elm.id + " [data-wb-tags]" ),
+				taggedItemsWrapper = document.querySelector( "#" + elm.id + " .wb-tagfilter-items" );
+
+			if ( taggedItemsWrapper ) {
+				taggedItemsWrapper.id = taggedItemsWrapper.id || wb.getId(); // Ensure the element has an ID
+				taggedItemsWrapper.setAttribute( "aria-live", "polite" );
```
duboisp, line 32: Note: Have we tested the different with and without it? Should we only have one global or can we add it on individual element that have tag on it?

src/plugins/tagfilter/tagfilter.js
```diff
@@ -0,0 +1,281 @@
...
+			wb.ready( $( elm ), componentName );
+		}
+	},
+
+	// Add every tagged item to an array of objects with their DOM ID, list of associated tags, and default isMatched attribute
+	buildTaggedItemsArr = function( taggedItems ) {
+		let taggedItemsArr = [];
+
+		taggedItems.forEach( function( taggedItem ) {
+			let tagsList = taggedItem.dataset.wbTags.split( " " );
```
duboisp, line 67: I don't know if there is a quick way to use the DOM Token List interface to parse the content? https://dom.spec.whatwg.org/#domtokenlist

src/plugins/tagfilter/tagfilter.js
```diff
@@ -0,0 +1,281 @@
...
+
+	// Utility method to update stored active filters, update stored items and update visibility of tagged items
+	update = function( instance ) {
+		refineFilters( instance );
+		matchItemsToFilters( instance );
+		updateDOMItems( instance );
+	};
+
+// When a filter is updated
+$document.on( "change", selectorCtrl, function( event )  {
```
duboisp, line 237: Didn't check if we already have that but We will need to provide the instruction where the filter form interface need to be before the content being tag filtered.

src/plugins/tagfilter/tagfilter-en.hbs
```diff
@@ -0,0 +1,1278 @@
...
+					<h3 class="h4"><a href="#" rel="external">Bidding on Opportunities</a></h3>
+					<p class="mrgn-bttm-sm">Wednesday August 10th, 2022 | 2:00 PM to 3:30 PM</p>
+					<p><span class="mrgn-rght-lg"><strong>Location:</strong> Vancouver</span><strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__english location__ottawa dayOTW_thursday">
+					<h3 class="h4"><a href="#" rel="external">Doing business with the Government of Canada (Webinar)</a></h3>
+					<p class="mrgn-bttm-sm">Thursday August 11th, 2022 | 12:00 PM to 2:00 PM</p>
+					<p><span class="mrgn-rght-lg"><strong>Location:</strong> Ottawa</span><strong>Language</strong>: English</p>
+				</li>
+				<li data-wb-tags="language__english location__online dayOTW_friday">
```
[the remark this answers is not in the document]
Garneauma: Simply tried to reduce the tags name where it made sense. Ultimately, those are defined by the author so they can name them whatever they want.

src/plugins/tagfilter/tagfilter-en.hbs
```diff
@@ -0,0 +1,1278 @@
...
+						<option value="inst__SC">Statistics Canada</option>
+						<option value="inst__VAC">Veterans Affairs Canada</option>
+					</select>
+				</fieldset>
+			</div>
+		</div>
+		<div class="col-sm-8 col-md-7 col-lg-6">
+			<div class="form-group mrgn-bttm-0">
+				<fieldset>
+					<legend class="h5 mrgn-tp-0"><span class="field-name">Platform</span></legend>
```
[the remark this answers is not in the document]
Garneauma: It is so that it looks better. But I can remove it.

src/plugins/tagfilter/tagfilter-en.hbs
```diff
@@ -0,0 +1,1278 @@
...
+							<label for="plat__windows">
+								<input type="checkbox" name="platform" id="plat__windows" class="wb-tagfilter-ctrl" value="plat__windows"> Windows
+							</label>
+						</li>
+					</ul>
+				</fieldset>
+			</div>
+		</div>
+	</div>
+	<div class="row wb-eqht-grd wb-tagfilter-items">
```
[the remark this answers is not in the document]
duboisp: Will be studied at the stabilization step

src/plugins/tagfilter/tagfilter.js
duboisp: To be discussed during the stabilization

src/plugins/tagfilter/tagfilter.js
duboisp: To be explored during plugin stabilization if we can use that interface here or a variant with CSS selector.

src/plugins/tagfilter/tagfilter.js
duboisp: This is for "Accessibility guidance"

duboisp (commented):
To test and validate before review for stabilization:
* [ ] The impact of the use of aria-controls from an user perspective
* [ ] The use of aria-live, having it one global vs multiple individual instance (like 1 per tagged element)
* [ ] Discuss best strategy to identify tag filtering control, by individual (like now), by group (parent element) or mixed. Or when input have the tagfilter class control, all input that share its `[name]` become in scope.
* [ ] For managing the list of tag applied on the elements, check if there is a quick way to use the DOM Token List interface to parse the content? https://dom.spec.whatwg.org/#domtokenlist or leverage CSS selector like `[data-wb-tags~=myTag]`

Accessibility guidance to add
* [x] Mention the tag filter control must be before the filtered items

Garneauma:
Discussion for stabilization: https://github.com/wet-boew/wet-boew/discussions/9551

duboisp:
Here a code snipped for an example to filter columns

```html
<section class="wb-tagfilter provisional">
	<h2>Filter table columns</h2>

	<fieldset>
		<legend class="h5 mrgn-tp-0"><span class="field-name">Filtre les colonnes du tableau</span></legend>
		<ul class="list-unstyled">
			<li class="checkbox">
				<label>
					<input type="checkbox" name="column-2" class="wb-tagfilter-ctrl" value="col-2" checked> Column 2
				</label>
			</li>
			<li class="checkbox">
				<label>
					<input type="checkbox" name="column-3" class="wb-tagfilter-ctrl" value="col-3" checked> Column 3
				</label>
			</li>
		</ul>
	</fieldset>
	<table class="wb-tagfilter-items table table-bordered">
		<thead>
			<tr>
				<th>Col heading 1
				<th data-wb-tags="col-2">Col heading 2
				<th data-wb-tags="col-3">Col heading 3
		</thead>
		<tbody>
			<tr>
				<th>Row heading 1
				<td data-wb-tags="col-2">Data cell 2a
				<td data-wb-tags="col-3">Data cell 3a
			<tr>
				<th>Row heading 2
				<td data-wb-tags="col-2">Data cell 2b
				<td data-wb-tags="col-3">Data cell 3b
			<tr>
				<th>Row heading 3
				<td data-wb-tags="col-2">Data cell 2c
				<td data-wb-tags="col-3">Data cell 3c
			<tr>
				<th>Row heading 4
				<td data-wb-tags="col-2">Data cell 2d
				<td data-wb-tags="col-3">Data cell 3d
		</tbody>
	</table>
</section>
```

Let's check how we can support it with the tag filter plugin

Garneauma:
Here's a working example:
```
<section class="wb-tagfilter provisional">
	<h2>Table example</h2>
	<fieldset>
		<legend class="h5 mrgn-tp-0"><span class="field-name">Columns</span></legend>
		<ul class="list-unstyled">
			<li class="checkbox">
				<label>
					<input type="checkbox" name="columns" class="wb-tagfilter-ctrl" value="col_1" checked> Column 1
				</label>
			</li>
			<li class="checkbox">
				<label>
					<input type="checkbox" name="columns" class="wb-tagfilter-ctrl" value="col_2" checked> Column 2
				</label>
			</li>
			<li class="checkbox">
				<label>
					<input type="checkbox" name="columns" class="wb-tagfilter-ctrl" value="col_3" checked> Column 3
				</label>
			</li>
		</ul>
	</fieldset>
	<table class="wb-tagfilter-items table">
		<thead>
			<tr>
				<th data-wb-tags="col_1">Column 1</th>
				<th data-wb-tags="col_2">Column 2</th>
				<th data-wb-tags="col_3">Column 3</th>
			</tr>
		</thead>
		<tbody>
			<tr>
				<td data-wb-tags="col_1">Column 1</td>
				<td data-wb-tags="col_2">Column 2</td>
				<td data-wb-tags="col_3">Column 3</td>
			</tr>
			<tr>
				<td data-wb-tags="col_1">Column 1</td>
				<td data-wb-tags="col_2">Column 2</td>
				<td data-wb-tags="col_3">Column 3</td>
			</tr>
			<tr>
				<td data-wb-tags="col_1">Column 1</td>
				<td data-wb-tags="col_2">Column 2</td>
				<td data-wb-tags="col_3">Column 3</td>
			</tr>
			<tr>
				<td data-wb-tags="col_1">Column 1</td>
				<td data-wb-tags="col_2">Column 2</td>
				<td data-wb-tags="col_3">Column 3</td>
			</tr>
			<tr>
				<td data-wb-tags="col_1">Column 1</td>
				<td data-wb-tags="col_2">Column 2</td>
				<td data-wb-tags="col_3">Column 3</td>
			</tr>
		</tbody>
	</table>
</section>
```

duboisp:
Pre-approve upon subsequent review and local testing.

duboisp (approved):
I reviewed and locally tested, and it's all good

* [ ] Not forget that we will need to add the tagfilter base CSS into GCWeb too. `@import "../plugins/tagfilter/base";`

I added a comment about a few items in the discussion.

src/plugins/tagfilter/tagfilter-en.hbs
```diff
@@ -0,0 +1,1205 @@
...
+
+<section id="filteredFAQ" class="wb-tagfilter provisional wb-filter" data-wb-filter='{
+		"source": "#searchFilter",
+		"selector": "article",
+		"section": ".wb-tagfilter-items",
+		"uiTemplate": "#searchuiTemplate"
+	}'>
+	<div id="searchuiTemplate" class="hidden">
+		<div class="col-sm-12">
+			<h3 class="wb-fltr-info mrgn-tp-sm">_NBITEM_ results</h3>
```
duboisp, line 846: L'exemple est broken car il n'utilise pas la notation pour personnaliser l'interface du le "filtre" plugin.
Les 2 span sont requis sinon on obtient cet erreur: `Uncaught TypeError: can't access property "textContent", document.querySelector(...) is null` lancé par: https://github.com/wet-boew/wet-boew/blob/master/src/plugins/filter/filter.js#L126-L127

Aussi, ce titre est ambigu et devrais être complémenté avec du texte additionnelle indiquant sa nature et répondant à la question "x résultats de quoi?".

Le plugin tag-filter throw une erreur car il ne reconnaît pas l'input utilisé pour le filter plugin interface: `wb-tagfilter: Filter controls require an attribute 'name'.`

Suggestion à titre informatif seulement a des fins d'essaie
```suggestion
<h3 class="wb-fltr-info mrgn-tp-sm"><span data-nbitem></span> results <span data-total hidden></span></h3>
```

Idem en français

duboisp (changes requested):
Après une seconde revue considérant que j'ai fusionné les changements préalable lié au filter plugin. Le tag filter n'est plus entièrement opérationnelle. C'est seulement l’exemple qui combine les 2 plugins ne fonctionne pas tel que prévue. Les autres exemples pratiques fonctionnent bien et sont approuvés.

Voici une liste des problèmes à corriger:

* [x] Il y a une erreur lancé indiquant qu'un control de filtre n'a pas de `[name]`, peut-être lancé une console.info à la place avec un peut plus de détails sur le contrôle "input" problématique afin de facilité le débogage et le repérage.
* [x] L'interface de recherche personnalisé du filtre ne fonctionne pas
* [x] Le pattern HTML pour l'affichage du résultat de recherche ainsi que l'interface de filtre personnalisé doit être revu conformément au WCAG.

src/plugins/tagfilter/tagfilter-en.hbs
Garneauma: Effectivement, j'attendais que la PR du filter UI soit fusionnée avant d'effectuer les changements nécessaires.

duboisp (changes requested):
La version française ne correspond à la version anglaise pour ce qui est de la combinaison du tag filter avec le filter plugin.

Outre, tous est ok.

duboisp (approved):
Reviewed and tested,

Good to be merged

PR #9426, opened 2022-09-07
Bootstrap 4: Backport alignment/flexbox sm classes

EricDunsworth:
* Backport the following small view classes:
  * Alignment: 
    * ``text-sm-left``
    * ``text-sm-right``
  * Flexbox:
    * ``d-sm-flex``
    * ``flex-sm-wrap``
    * ``align-items-sm-center``
* Cover the backported classes in the utilities page

Spinoff of wet-boew/GCWeb#1916.

PR #9428, opened 2022-09-08
Utility: Add position options to ease vertical centering

GormFrank:

**From Bootstrap 4:** This helps position elements that are side-by-side to be vertically centered to each other

### General checklist

- [x] Create/update documentation
- [ ] Build locally and test

### DIFF
--- a/src/base/_bootstrap-4.scss
+++ b/src/base/_bootstrap-4.scss
@@ -43,9 +43,18 @@
 	}
 }
 
-/* Position relative */
+/* Positioning */
 .position-relative {
 	position: relative !important;
+}
+
+/* Flexbox */
+.d-flex {
+	display: flex !important;
+}
+
+.align-self-center {
+	align-self: center !important;
 }
 
 /* Vertical alignment */
--- a/src/other/utility/utility-en.hbs
+++ b/src/other/utility/utility-en.hbs
@@ -1,6 +1,6 @@
 ---
 {
-	"title": "Uility",
+	"title": "Utilities",
 	"language": "en",
 	"category": "Other",
 	"description": "Utilities page including working examples.",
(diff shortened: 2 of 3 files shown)


src/other/utility/utility-en.hbs
```diff
@@ -116,76 +116,96 @@
...
-<h4>working example:</h4>
-<p class="position-relative">This is is a pragraph relatively positioned.</p>
-<h4>Code sample:</h4>
-<pre><code>&lt;p class="<strong>position-relative</strong>"&gt;This is is a pragraph relatively positioned.&lt;/p&gt;</code></pre>
+<h4>Working example</h4>
+<p class="position-relative">This is a paragraph relatively positioned.</p>
+<h4>Code sample</h4>
+<pre><code>&lt;p class="<strong>position-relative</strong>"&gt;This is a paragraph relatively positioned.&lt;/p&gt;</code></pre>
+
+<h3 id="d-flex"><code>d-flex</code></h3>
```
EricDunsworth, line 184: How should these descriptions/examples be handled when both a generic class and a breakpoint-specific version have been backported?

Just for context, I sent in #9426 a day before this PR. Among other things, it proposes adding ``d-sm-flex`` - with a different description and example than what this PR is pitching for the ``d-flex`` class.

src/base/_bootstrap-4.scss
```diff
@@ -43,7 +43,15 @@
 	}
 }
 
-/* Position relative */
+/* Positioning */
+.d-flex {
+	display: flex !important;
+}
```
duboisp, line 49: We will need to remove/update this because it was introduced by PR #9426

duboisp (changes requested):
Please rebase

duboisp:
I validated and those new propose CSS exist in Bootstrap 4

src/base/_bootstrap-4.scss
EricDunsworth: Just wanted to clarify that #9426 added ``d-sm-flex``, which only activates flexbox layouts in small view and up.

This PR's ``d-flex`` class activates flexbox in all views.

So both classes target different scenarios and therefore might need to co-exist together.

duboisp (approved):

src/base/_bootstrap-4.scss
duboisp: Agreed, but since all those CSS are in a single page, let wait until that page began to not be usable for web author and then let's reorganize it.

duboisp:
Before to merge, quick test locally first. Approved at the technical review.

duboisp (approved):
Tested locally and work as expected.

PR #9436, opened 2022-09-16
Calevt: Adding redraw feature and small fixes

Garneauma:
Following changes:
- Handling of `wb-redraw` event that deletes the current calendar and generates a new one with updated HTML.
- Changed solution to get list of events.
    - Previous solution would get every `li` directly under an `ol` or `ul` inside the specified element, which could trigger unwanted behaviour (i.e. if an add to calendar was initiated inside a `li`). 
    - New solution finds the first of either `ol` or `ul` and get its immediate `li` children.
- Hiding events in calendar on focus out.
    - Events would remain visible when moving focus to another element or when selecting an event.
- Adding option to exclude items using `data-calevt-exclude`.
- Updated lists of events to use `ul` instead of `ol`. `ul` is now the recommended approach, however `ol` is still supported.

### DIFF
--- a/site/pages/docs/ref/cal-events/cal-events-en.hbs
+++ b/site/pages/docs/ref/cal-events/cal-events-en.hbs
@@ -33,21 +33,25 @@
 	<h2>How to implement</h2>
 	<ol>
 		<li>Add a div element to the Web page with a <code>class</code> attribute containing "wb-calevt" and a unique name in a <code>data-calevt-src</code> attribute (e.g.: <code>&lt;div class="wb-calevt" data-calevt-src="unique-name"&gt;&lt;/div&gt;</code>).</li>
-		<li>Include in the div element a ordered list of events (ol element) with the following structure for each event:
+		<li>Include in the div element an unordered list of events (ul element) with the following structure for each event:
 			<ol>
 				<li>The list element must start with a heading element (<code>h2</code> to <code>h6</code>, depending on the context of the page).</li>
 				<li>The heading element must include a link to the event description with the event title for the link text.</li>
 				<li>The event date must be provided in an HTML5 <code>time</code> element contained within a paragraph element (following the heading element). The date to be displayed is contained in the <code>time</code> element and an ISO-8601 date (e.g., 2010-03-11) is provided in the <code>datetime</code> attribute.</li>
+				<li>Note: to exclude an item from being shown on the calendar and in the list, simply add the CSS class <code>wb-fltr-out</code> to the list element (<code>&lt;li&gt;</code>).</li>
 			</ol>
 		</li>
 		<li>Add a second <code>div</code> element to the Web page with an <code>id</code> attribute containing your <code>data-calevt-src</code> attribute's unique name from above (e.g., <code>&lt;div id="unique-name"&gt;&lt;/div&gt;</code>). This will hold the calendar interface.</li>
 	</ol>
+	<div class="alert alert-info">
+		<p><strong>Note</strong>: the plugin still supports <code>&lt;ul&gt;</code> and <code>&lt;ol&gt;</code>, but, as of October 4th 2022, we recommend using a <code>&lt;ul&gt;</code> since items numbers in a <code>&lt;ol&gt;</code> will always start at 1, even when some items are hidden.</p>
+	</div>
 
 	<h3>Example Code</h3>
 	<pre><code>&lt;div id="calendar1"&gt;&lt;/div&gt;
 
 &lt;div class="wb-calevt" data-calevt-src="calendar1"&gt;
-	&lt;ol&gt;
+	&lt;ul&gt;
 		&lt;li&gt;
 			&lt;section&gt;
 				&lt;h4&gt;&lt;a href="https://www.canada.gc.ca"&gt;Single-Day Event&lt;/a&gt;&lt;/h4&gt;
@@ -62,7 +66,7 @@
 				&lt;p&gt;Event Description&lt;/p&gt;
 			&lt;/section&gt;
 		&lt;/li&gt;
-	&lt;/ol&gt;
+	&lt;/ul&gt;
 &lt;/div&gt;</code></pre>
 
 	<section>
@@ -71,7 +75,7 @@
 		<pre><code>&lt;div id="calendar2"&gt;&lt;/div&gt;
 
 &lt;div class="wb-calevt evt-anchor" data-calevt-src="calendar2"&gt;
-	&lt;ol&gt;
+	&lt;ul&gt;
 		&lt;li&gt;
 			&lt;section&gt;
 				&lt;h4&gt;Single-Day Event&lt;/h4&gt;
@@ -96,7 +100,7 @@
 				&lt;/ul&gt;
 			&lt;/section&gt;
 		&lt;/li&gt;
-	&lt;/ol&gt;
+	&lt;/ul&gt;
 &lt;/div&gt;</code></pre>
 	</section>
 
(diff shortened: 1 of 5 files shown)


src/plugins/cal-events/cal-events.js
```diff
@@ -168,15 +168,16 @@ var componentName = "wb-calevt",
 					}
 				]
 			},
-			objEventsList = obj.find( "ol > li, ul > li" ),
-			iLen = objEventsList.length,
+			objEventsList = obj.find( "ol, ul" ).first(),
```
duboisp, line 171:
```suggestion
			objEventsList = obj.find( "ul, ol" ).first(),
```

And apply the same fix on our working example. + Add a note about this.

src/plugins/cal-events/cal-events.js
```diff
@@ -168,15 +168,16 @@ var componentName = "wb-calevt",
...
+			objEventsListItems = objEventsList.find( "> li:not(.wb-fltr-out)" ),
```
duboisp, line 172: We will need a working example that illustrate this new functionality of filtering + documentation

src/plugins/cal-events/cal-events.js
```diff
@@ -372,7 +373,27 @@ var componentName = "wb-calevt",
...
+		$elm = $( "#" + event.target.id ),
+		calendarId = event.currentTarget.dataset.calevtSrc;
+
+	switch ( eventType ) {
+	case "timerpoke":
+	case "wb-init":
+		init( event );
+		break;
+
+	case "redraw":
```
duboisp, line 391: What is the use case behind this new event? + need to be documented in the list of events.

What is the difference with `wb-updated.wb-calevt`

src/plugins/cal-events/cal-events.js
```diff
@@ -436,6 +457,10 @@ $document.on( "keydown", ".wb-calevt-cal .cal-days td > ul li", function( event
 	}
 } );
 
+$document.on( "focusout", ".wb-calevt-cal .cal-days td > ul > li > a", function( event ) {
```
duboisp, line 460: Can we use a different method to detect a focus out, like with our overlay plugin.

However, the CSS selector for this type of event is too much precise.

duboisp (changes requested):

site/pages/docs/ref/cal-events/cal-events-en.hbs
```diff
@@ -154,8 +157,44 @@
...
 				&lt;/section&gt;
 			&lt;/li&gt;
-		&lt;/ol&gt;
+		&lt;/ul&gt;
 	&lt;/section&gt;
+&lt;/div&gt;</code></pre>
+	</section>
+
+	<section>
+		<h3>Exclude items (<code>data-calevt-exclude</code>)</h3>
```
duboisp, line 166: Just document the CSS class "filter-out" instead of making it customizable.

duboisp (changes requested):

src/plugins/cal-events/cal-events.js
```diff
@@ -436,6 +458,10 @@ $document.on( "keydown", ".wb-calevt-cal .cal-days td > ul li", function( event
 	}
 } );
 
+$document.on( "focusout", ".wb-calevt-cal td > ul", function( event ) {
```
duboisp, line 461: Let's take the opportunity to optimize the code and reduce number of duplicata

```suggestion
$document.on( "focusout", selectorEvent + ". td > ul", function( event ) {
```

At the top of the plugin, declare that variable near line 18 https://github.com/wet-boew/wet-boew/blob/master/src/plugins/cal-events/cal-events.js#L18
```
var componentName = "wb-calevt",
	selector = "." + componentName,
	componentEventName = componentName + "-cal",
	selectorEvent = "." + componentEventName,
```

And replace line 91 https://github.com/wet-boew/wet-boew/blob/master/src/plugins/cal-events/cal-events.js#L91

```
		$container = $( "#" + containerId ).addClass( componentEventName );
```

And replace the other event that rely on an hard coded selector like: `.wb-calevt-cal`

src/plugins/cal-events/cal-events.js
```diff
@@ -372,7 +374,27 @@ var componentName = "wb-calevt",
...
+		calendarId = event.currentTarget.dataset.calevtSrc;
+
+	switch ( eventType ) {
+	case "timerpoke":
+	case "wb-init":
+		init( event );
+		break;
+
+	case "wb-redraw":
+		$( "#" + calendarId + " .wb-clndr" ).remove();
```
duboisp, line 392: Can you find it relatively from the $elm?

Like
```suggestion
		$( ".wb-clndr", $elm ).remove();
```

or

```suggestion
		$elm.find( ".wb-clndr" ).remove();
```

src/plugins/cal-events/cal-events-fr.hbs
```diff
@@ -365,7 +365,126 @@
...
+		</section>
+	</section>
+
+	<section>
+		<h3>Exclure des éléments</h3>
+		<p>Dans cet exemple, certains éléments sont exclus de l'affichage sur le widget Calendrier en utilisant l'attribut <code>data-calevt-exclude</code>.</p>
+
+		<div id="cal-exclude"></div>
+
+		<div class="wb-calevt evt-anchor" data-calevt-src="cal-exclude" data-calevt-exclude=".exclude">
```
duboisp, line 381: as discussed, let use a static selector for now considering it is not currently supported by an current use case.

The data attribute `data-calevt-src` don't seem to be documented. What is it use case? I am not sure if that is needed when I look how it is used in the timepoke event. It seem to contain an ID, if that is the case may be named it like "src-id" | "bind-id" or let the author to provide a CSS selector as the source.

duboisp (changes requested):

src/plugins/cal-events/cal-events.js
Garneauma: Unfortunately, none of those options would work as the calendar widget is added outside of the calendar of events element. So, I cannot start from the $elm reference.

src/plugins/cal-events/cal-events-fr.hbs
Garneauma: The `data-calevt-src` is documented at the beginning of the documentation page:

"How to implement
1. Add a div element to the Web page with a class attribute containing "wb-calevt" and a unique name in a data-calevt-src attribute (e.g.: <div class="wb-calevt" data-calevt-src="unique-name"></div>).
...
3. Add a second div element to the Web page with an id attribute containing your data-calevt-src attribute's unique name from above (e.g., <div id="unique-name"></div>). This will hold the calendar interface."

src/plugins/cal-events/cal-events.js
```diff
@@ -372,9 +375,29 @@ var componentName = "wb-calevt",
 	};
 
 // Bind the init event of the plugin
-$document.on( "timerpoke.wb " + initEvent, selector, init );
+//$document.on( "timerpoke.wb " + initEvent, selector, init );
```
duboisp, line 378: Remove it, you already replaced it

```suggestion
```

src/plugins/cal-events/cal-events.js
duboisp: What about using the `$elm.parentUntil( ".wb-clndr" )` jquery function? That way the user/web author would not need to provide that technical information to the component which it don't provide any personalization advantage.

duboisp: well, ignore my previous comment. 
There is several example where the web author identify where to insert the calendar.

duboisp (changes requested):
Small changes.

* [x] Replace line 91 to leverage the new private variable `componentEventName` instead of computing it.
* [x] Remove unused code
* [x] **To not action, but has been check**, the attribute is already been used in other working example. ---Remove the attribute ` data-calevt-src="cal-exclude"` in the working example (E&F) and re-work the detection of `wb-clndr`. Regarding line 92, 93---

Garneauma:
> Small changes.
> 
> * [x]  Replace line 91 to leverage the new private variable `componentEventName` instead of computing it.
> * [x]  Remove unused code
> * [x]  **To not action, but has been check**, the attribute is already been used in other working example. ---Remove the attribute ` data-calevt-src="cal-exclude"` in the working example (E&F) and re-work the detection of `wb-clndr`. Regarding line 92, 93---

Done!

duboisp (approved):
I reviewed the code and tested locally.

Note:
* The focus out work as expected, the listbox popup don't stay open.
* The filtering out work

PR #9446, opened 2022-10-05
Calevt: Fixing accessibility issues (WCAG and SR compatibility)

Garneauma:
Fixing a few accessibility issues. Some related to WCAG failures, but most related to SR compatibility.

1. Left and right arrows ("previous month" and "next month" buttons) were being communicated by SR.
2. Visually hidden events related to other months were still being offered to SR. Changed class to prevent this.
3. `aria-label` on the calendar was not associated with `role="application"`. Moved `aria-label` to the correct `div`.
4. `select` dropdowns used the `title` attribute instead of a label. Added label which needed a dynamically generated `for` attribute that was the same as the dropdown's `id`.
5. Removed unnecessary `role` attributes in calendar table.
6. Clicking on the "previous" or "next" month buttons did not provide feedback that the month had changed. Added element with `aria-live="polite"` to the calendar nav bar that announces the new month when clicking those buttons.

Changes related to WET-270

### DIFF
--- a/src/plugins/cal-events/cal-events.js
+++ b/src/plugins/cal-events/cal-events.js
@@ -20,7 +20,7 @@ var componentName = "wb-calevt",
 	setFocusEvent = "focus",
 	dataAttr = componentName,
 	$document = wb.doc,
-	i18n, i18nText,
+	hiddenClass = "hidden",
 
 	/**
 	 * @method init
@@ -36,14 +36,6 @@ var componentName = "wb-calevt",
 
 		if ( elm ) {
 			$elm = $( elm );
-
-			// Only initialize the i18nText once
-			if ( !i18nText ) {
-				i18n = wb.i18n;
-				i18nText = {
-					calendar: i18n( "cal" )
-				};
-			}
 
 			// Load ajax content
 			$.when.apply( $, $.map( $elm.find( "[data-calevt]" ), getAjax ) )
@@ -123,8 +115,6 @@ var componentName = "wb-calevt",
 			events: events.list,
 			$events: $elm
 		} );
-
-		$container.attr( "aria-label", i18nText.calendar );
 	},
 
 	daysBetween = function( dateLow, dateHigh ) {
@@ -345,10 +335,10 @@ var componentName = "wb-calevt",
 
 	filterEvents = function( year, month ) {
 		this.find( "li.cal-disp-onshow" )
-			.addClass( "wb-inv" )
+			.addClass( hiddenClass )
 			.has( ":header[class*=filter-" + year + "-" +
 				wb.string.pad( parseInt( month, 10 ) + 1, 2 ) + "]" )
-			.removeClass( "wb-inv" );
+			.removeClass( hiddenClass );
 	},
 
 	showEvents = function() {
--- a/src/plugins/calendar/calendar.js
+++ b/src/plugins/calendar/calendar.js
@@ -28,25 +28,27 @@ var i18nText,
 			goToMonth: i18n( "cal-goToMnth" ),
 			dayNames: i18n( "days" ),
 			currDay: i18n( "currDay" ),
-			format: i18n( "cal-format" )
+			format: i18n( "cal-format" ),
+			calendar: i18n( "cal" )
 		};
 
 		textWeekDayNames = i18nText.dayNames;
 		textMonthNames = i18nText.monthNames;
 
-		$calBase = $( "<div class='wb-clndr' role='application'>" +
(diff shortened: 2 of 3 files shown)


src/plugins/cal-events/cal-events.js
```diff
@@ -37,14 +37,6 @@ var componentName = "wb-calevt",
 		if ( elm ) {
 			$elm = $( elm );
 
-			// Only initialize the i18nText once
-			if ( !i18nText ) {
-				i18n = wb.i18n;
-				i18nText = {
-					calendar: i18n( "cal" )
```
duboisp, line 44: Check if "cal" is used elsewhere, if not we will need to remove it from our CSV.

Garneauma: It was moved to calendar.js line 38. No need to remove it from the CSV.

src/plugins/calendar/calendar.js
```diff
@@ -28,25 +28,27 @@ var i18nText,
...
+			calendar: i18n( "cal" )
 		};
 
 		textWeekDayNames = i18nText.dayNames;
 		textMonthNames = i18nText.monthNames;
 
-		$calBase = $( "<div class='wb-clndr' role='application'>" +
+		$calBase = $( "<div class='wb-clndr' role='application' aria-label='" + i18nText.calendar + "'>" +
 			"<div class='cal-nav'>" +
+				"<span class='wb-inv current-month' aria-live='polite'></span>" + // Added for screen-readers
```
duboisp, line 40: Should we change the aria label on `.wb-clndr` instead or at least change it also and keep both synchronized after the first interaction?

If possible, @Garneauma can you test with a SR: Ideally we will have an heading, figcaption, caption... something to identify the calendar title. Here I know it's the aria-label but I don't know how will be experience if `aria-live="polite"` is set on the `.wb-clndr` element when we change the aria-label and all the inner controls. I do think we don't necessary want the inner control changes to be announced but to announce the whole `role=application` change.

duboisp (changes requested):
Other than the requested change, it looks good.

* [x] Review how we sync and keep updated the whole component labeling (aria-polite)
* [ ] Use implicit label instead of the explicit one.

src/plugins/calendar/calendar.js
Garneauma: Adding aria-live to the .wb-clndr element makes the SR announce too much when clicking on the previous or next arrows. This results in the SR announcing something like: "March Month: March Monday, March 1, 2013 Tuesday, March 2, 2013, ..." and is very annoying. By defining a specific element to keep track, it is much more understandable for a SR user to understand what their action did. However, I realized that only announcing the Month might not be enough. I have just added the year too. So it now reads something like: "March 2022".

Garneauma:
> Other than the requested change, it looks good.
> 
> * [x]  Review how we sync and keep updated the whole component labeling (aria-polite)
> * [x]  Use implicit label instead of the explicit one.

Done!

src/plugins/calendar/calendar.js
```diff
@@ -324,6 +326,9 @@ $document.on( "click vclick touchstart", ".cal-month-prev, .cal-month-next", fun
 		modifier = className.indexOf( "cal-month-prev" ) !== -1 ? -1 : 1,
 		date = new Date( calendar.lib.year, calendar.lib.month + modifier, 1 );
 
+	// Added declaration of current month in aria-live="polite" for screen-readers
+	$calendar.find( ".current-month" ).html( i18nText.monthNames[ date.getMonth() ] + " " + calendar.lib.year );
```
duboisp, line 330: The "current-month" span are not updated when the user change the month or the year directly via the select elements.
Idem when we navigate with the days with the "left" and "right" arrow and the current month change

You might need to hook other event and apply the same/similar update to the current-month.

duboisp (changes requested):
Tested and reviewed the code, everything work except the following

* [ ] Aria live current month are not updated when: year or month dropdown are updated and when the user navigate with the arrow on a event that span multiple months. The expectation is the value for the current month match the current state of the calendar.

Garneauma:
> Tested and reviewed the code, everything work except the following
> 
> * [x]  Aria live current month are not updated when: year or month dropdown are updated and when the user navigate with the arrow on a event that span multiple months. The expectation is the value for the current month match the current state of the calendar.

@duboisp Moved the update of the live element to the navigate event. Added a condition to prevent its update on initialization.

GormFrank:
Pre-approved upon successful testing from @duboisp

duboisp (approved):
I reviewed the code and tested locally, all changes requested are completed and work as expected.

PR #9454, opened 2022-10-17
wcag upgrade fix F101 - remove touchstart

### LINKED ISSUE #9453: SC 2.5.2 - Remove "touchstart" - Fail F101
SC 2.5.2 - Pointer Gestures

**Goal:** Don't initiate an action on touch start, mouse down, key down event. 

**At the same time**: Check and identify any key event that might apply to SC 2.1.4 Character key shortcuts

**Todo:**
* Review all component and remove "mousedown", "touchstart", "pointerdown" and "vclick" to leverage only "click", all that by respecting the exception as described in the understanding.

**Technical note:**
* **Slider polyfill:** Check at lower priority unless if we need to support vertical slider because it seems to be unsupported by FF (vertical slider). Horizontal slider seems to be supported in all major browser.

shiva-sc:
relate to WET-283

close #9453

### DIFF
--- a/src/plugins/deps/geomap-lib.js
+++ b/src/plugins/deps/geomap-lib.js
@@ -1645,7 +1645,7 @@ var componentName = "wb-geomap",
 
 						// bind events to the options
 						// TODO: do this in the wb-update event
-						$( ".al-opt a" ).on( "keydown click vclick", function( event ) {
+						$( ".al-opt a" ).on( "keydown click", function( event ) {
 							var link = event.target,
 								eventType = event.type,
 								which = event.which;
@@ -1657,8 +1657,6 @@ var componentName = "wb-geomap",
 								}
 								break;
 							case "click":
-							case "vclick":
-							case "touchstart":
 
 								// Ignore middle/right mouse buttons
 								if ( !which || which === 1 ) {
--- a/src/polyfills/events/mobile.js
+++ b/src/polyfills/events/mobile.js
@@ -1,4 +1,5 @@
 /*
+* This file is to enable touch on browser
 * jQuery Mobile v1.3.2
 * https://jquerymobile.com
 *
--- a/src/plugins/calendar/calendar.js
+++ b/src/plugins/calendar/calendar.js
@@ -317,7 +317,7 @@ $document.on( "change", selector, function( event ) {
 	} );
 } );
 
-$document.on( "click vclick touchstart", ".cal-month-prev, .cal-month-next", function( event ) {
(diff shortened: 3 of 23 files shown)


src/plugins/deps/geomap-lib.js
```diff
@@ -1657,8 +1657,6 @@ var componentName = "wb-geomap",
 								}
 								break;
 							case "click":
-							case "vclick":
```
duboisp, line 1660: Remove also the "vclick" event handler at line 1648.

duboisp (changes requested):
Changes look good except for geomap. Also let's remove all instances of `vclick` in wet-boew which are an artifact of jQuery Mobile used by WET 3.x

Todo:
* [x] Remove the vclick event handler in geomap
* [x] Remove all "vclick" event handler instance? I quickly searched and I found ~21 files that use it which is more than the number of modified file in this PR currently (7 files). No need to test it as long they are accompanied with the standard "click" event handler.

GormFrank:

src/polyfills/events/mobile.js
```diff
@@ -70,13 +71,13 @@
...
 // The current version exposes the following virtual events to jQuery bind methods:
-// "vmouseover vmousedown vmousemove vmouseup vclick vmouseout vmousecancel"
+// "vmouseover vmousedown vmousemove vmouseup vmouseout vmousecancel"
 
 (function( $, window, document, undefined ) {
 
 var dataPropertyName = "virtualMouseBindings",
 	touchTargetPropertyName = "virtualTouchID",
-	virtualEventNames = "vmouseover vmousedown vmousemove vmouseup vclick vmouseout vmousecancel".split( " " ),
+	virtualEventNames = "vmouseover vmousedown vmousemove vmouseup vmouseout vmousecancel".split( " " ),
```
duboisp, line 80: Please undo those change, let keep this polyfill as is. It should not get loaded based on our list of supported browser.

Thanks

duboisp (changes requested):
One small thing

* [x] Undo the code change made to the mobile.js files

I did locally tested and reviewed the code and it all looks good and work as expected. Once the todo is completed, we will be able to merge this PR.

src/polyfills/events/mobile.js
duboisp: @shiva-sc can you undo this change in this file, just re-add `vclick` thanks.

duboisp:

duboisp (approved):
Tested locally and reviewed the latest changes. It works as expected.

Thanks @shiva-sc

PR #9459, opened 2022-10-24
Steps form: Add type to next previous buttons

andrewdp:
Fix: The default WET action on pressing Enter in an input field is to trigger a click event on the first submit type button located on the form. Since the first button on a step form is often the generated "next" button on the first step and no type is specified for this button the form will open to the second step. Specifying the button type should fix this issue.

### DIFF
--- a/src/plugins/stepsform/stepsform.js
+++ b/src/plugins/stepsform/stepsform.js
@@ -143,6 +143,7 @@ var componentName = "wb-steps",
 
 		// set default attributes
 		control.className = ( type === "prev" ? "btn btn-md btn-default" : "btn btn-md btn-primary" ) + " " + style;
+		control.setAttribute("type", "button");
 		control.innerHTML = text;
 
 		return control;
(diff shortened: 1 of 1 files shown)


GormFrank:
**To do:**
- [ ] Build and test locally (@duboisp)

Pre-approved upon successful completion of "To do" above.

duboisp:
I tested locally and work as expected. Once the eslint error is fixed this PR can be merged.

@andrewdp: There is two small eslint issue which is to add spacing in the function call.

ref. https://contribute.jquery.org/style-guide/js/#spacing

andrewdp:
Ok The spacing on the brackets has been added.

duboisp:
Pre-approved upon successful completion of successful build and testing.

It's ok from a code review perspective

duboisp (approved):

duboisp:
I builded and tested locally. 
It works as expected and make the produced HTML code compliant to specification because the default type for a button is submit which is not the case here. Reference: https://html.spec.whatwg.org/multipage/form-elements.html#attr-button-type

Thanks @andrewdp for fixing that issue.

PR #9467, opened 2022-11-03
Bootstrap 4: Backport pr-sm-3 class

EricDunsworth:
**Notes:**
* Unit measurement differences:
  * Bootstrap 4-5 use ``.5rem`` for this class... which is technically ``16px``
  * Went with ``15px`` to stay consistent with pre-existing backported classes
* Using ``!important`` for the following reasons:
  * It's needed for my use case
  * Bootstrap 4-5 themselves use it for their versions of this class
* This class name no longer exists in Bootstrap 5:
  * It was renamed to ``pe-sm-3`` (``r`` for "right" changed to ``e`` for "end")
  * Went with the outdated name since pre-existing backported classes are already using ``r`` for "right"

Spinoff of wet-boew/GCWeb#1916.

### DIFF
--- a/src/base/views/_bootstrap-4-screen-sm-min.scss
+++ b/src/base/views/_bootstrap-4-screen-sm-min.scss
@@ -38,3 +38,7 @@
 	padding-left: 15px;
 	padding-right: 15px;
 }
+
+.pr-sm-3 {
+	padding-right: 15px !important;
+}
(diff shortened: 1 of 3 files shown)


duboisp:
Pre-approved upon successful completion of successful build and testing.

duboisp (approved):
Tested locally and reviewed the code and it works as expected.

Thanks @EricDunsworth

PR #9469, opened 2022-11-09
JSON Manager | Core: Add support for Base64 decode and HTML escape

GormFrank:

Adds helpers in the core as well as operations in JSON Manager to:

- Decode Base64 content; and
- Escape HTML characters from an HTML string.

This comes in handy when generating code samples for documentation purposes, by allowing to grab content directly from GitHub (see working example in PR). Note that the GitHub API has a rate limit however (see note in PR).

Related to WET-294


duboisp (commented):
Approved upon full code review and successful testing.

This is a minor change.

duboisp (approved):
Tested locally and reviewed the code. Look good and work as expected.

PR #9478, opened 2022-11-17
Form validation: fix french alphanumeric validation and french error message

polmih:
Form validation: fix french alphanumeric validation and french error message ( the error message in french was saying it is accepting the spaces too and its not true).

Related to WET-259

### DIFF
--- a/src/plugins/formvalid/formvalid.js
+++ b/src/plugins/formvalid/formvalid.js
@@ -93,6 +93,19 @@ var componentName = "wb-frmvld",
 						summaryHeading = settings.hdLvl,
 						i, len, validator;
 
+					if ( wb.lang === "fr" ) {
+
+						// alphanumeric regex is changed to allow french characters;
+						$.validator.addMethod( "alphanumeric", function( value, element ) {
+							return this.optional( element ) || /^[a-zàâçéèêëîïôûùüÿæœ0-9_]+$/i.test( value );
+						}, "Letters, numbers, and underscores only please." );
+
+						// error french text is adjusted to remove the word "spaces"
+						$.extend( $.validator.messages, {
+							alphanumeric: "Veuillez fournir seulement des lettres, nombres et soulignages."
+						} );
+					}
+
 					// Append the aria-live region (for provide message updates to screen readers)
 					$elm.append( "<div class='arialive wb-inv' aria-live='polite' aria-relevant='all'></div>" );
 


RobJohnston:
This matches the changes to the jQuery Validation plugin done here 10 years ago:
https://github.com/jquery-validation/jquery-validation/pull/375/files

However, their French translation was never updated:
https://github.com/jquery-validation/jquery-validation/blob/master/src/localization/messages_fr.js

polmih:
> This matches the changes to the jQuery Validation plugin done here 10 years ago: https://github.com/jquery-validation/jquery-validation/pull/375/files
> 
> However, their French translation was never updated: https://github.com/jquery-validation/jquery-validation/blob/master/src/localization/messages_fr.js

Although their regex will not allow french characters :)

duboisp:
Pre-approved upon successful code review and testing.

@polmih can you submit the same fix upstream in the jQuery validation plugin. and add reference in your code here. Thank you

polmih:
> 

PR sent : https://github.com/jquery-validation/jquery-validation/pull/2456

RobJohnston:
Changing the regex to include French characters surely isn't going to fly with jQuery Validation.  When I earlier said that they made the same change, I was just referencing the translation of the error message.

Even WET-BOEW supports more than 2 languages.  Not sure that the regex should be changed. 🤔

duboisp:
Pre-approved upon successful code review and testing.

Thanks @polmih for the PR

duboisp (approved):
Tested locally and work as expected

duboisp:
@polmih after considering @RobJohnston comment, let's localize that regular expression override for only the French pages.

@RobJohnston I do agree that might don't fly with jQuery Validation group, but it can help other.

src/plugins/formvalid/formvalid.js
```diff
@@ -93,6 +93,19 @@ var componentName = "wb-frmvld",
...
 						i, len, validator;
 
+
+					// alphanumeric regex is changed to allow french characters;
+					$.validator.addMethod( "alphanumeric", function( value, element ) {
+						return this.optional( element ) || /^[a-zàâçéèêëîïôûùüÿæœ0-9_]+$/i.test( value );
+					}, "Letters, numbers, and underscores only please." );
+
+					// // error french text is adjusted to remove the word "spaces"
+					if ( wb.lang === "fr" ) {
```
duboisp, line 103:
```suggestion

					if ( wb.lang === "fr" ) {
						// alphanumeric regex is changed to allow french characters;
						$.validator.addMethod( "alphanumeric", function( value, element ) {
							return this.optional( element ) || /^[a-zàâçéèêëîïôûùüÿæœ0-9_]+$/i.test( value );
						}, "Letters, numbers, and underscores only please." );

					// error french text is adjusted to remove the word "spaces"
```

duboisp (changes requested):

src/plugins/formvalid/formvalid.js
polmih: Done. Thanks

duboisp:
Pre-approved upon successful local testing. Code review is ok.

duboisp (approved):
Tested locally and reviewed the code. All good

PR #9493, opened 2022-12-06
bower.json - remove unnecessary properties

polmih:
bower.json - remove unnecessary properties .
The wet version is loaded from package.json


duboisp (approved):
I reviewed the code and I those meta properties are not used by bower.

Approved at the technical review

PR #9482, opened 2022-11-20
Performance: use regex.test or string.includes instead of string.match

radmorecameron:
_(Pick the language of your choice to fill out the following information. / Choisissez la langue de votre choix pour remplir l'information ci-dessous.)_

## What does this pull request (PR) do? / Que fait cette demande « pull » (PR)?
This PR replaces occurrences of string.match with regex.test or string.includes when possible. Which should slightly increase the performance.

## Additional information (optional) / Information additionnelle (optionel)
Benchmark of string.includes, string.match and regex.test: https://www.measurethat.net/Benchmarks/Show/13675/0/regextest-vs-stringincludes-vs-stringmatch#latest_results_block

**General checklist / Liste de contrôle générale**
Make your own list for the purpose of your Pull request. /// Faites votre propre liste en fonction des besoins de votre demande « pull ».

- [x] Build and test PR's code /// Rouler le script de compilation et tester le code de la PR

**Screenshots / Captures d'écrans**
benchmark graph

PR #9505, opened 2023-01-13
Shim to resolve conflict with IE11 due to usage of .includes()

Ricokola:
Related to: #9482 

The usages of `.includes()` instead of `.match()` has created conflicts with IE11 and therefore, some plugins have stop working. A shim has been created to fix the issues encounter in IE11.


src/shims/includes.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * String.prototype.includes() polyfill
+ * @author Ricokola
+ * @license MIT
+ */
+if ( !String.prototype.includes ) {
+	String.prototype.includes = function( string ) {
+		console.log( string );
```
duboisp, line 8: Please remove the console.log

duboisp (changes requested):

EricDunsworth:
@duboisp Regarding this:
> Note: The IE11 detection is made through the user agent detection and not on the availability of the "includes" function on String object.

It's true that IE11 detection is performed by checking its user agent string, but that check currently uses ``includes()``. So this shim is needed to avoid a JS console error when the check runs.

**PS:**
The check used to use ``match()`` in the past:
https://github.com/wet-boew/wet-boew/blob/d93b46e8b56dd50009b8192705d004867298f5ff/src/core/wb.js#L258

But #9482 recently changed it to use ``includes()``:
https://github.com/wet-boew/wet-boew/blob/999b5e021718a8fe945a8b4054333c9b62de3722/src/core/wb.js#L258

duboisp:
fyi - Test page: http://universallabs.org/wet/labs/pr9505/

duboisp (approved):
Tested locally in my old IE11 and work as expected. Now what was working continu to work as expected. ex. expand/collapse.

PR #9512, opened 2023-02-01
Postback: Fix demo/code sample inconsistencies

EricDunsworth:
Some of these prevented the demos from working properly when copying/pasting their code samples (like missing class name dots).

**Specific changes:**
* Working examples:
  * Remove ``data-wb-postback`` attribute's ``content`` variable (same as code samples)
* Code samples:
  * Change quote format to single quotes for attribute values and double quotes for JSON strings (same as working examples)
  * Add dots in front of the ``data-wb-postback`` attribute's success/failure class names (same as working examples)
  * Add prettify to French demo page
  * Second demo: Change "First name" to title case in English (same as working example)
* Both:
  * Normalize spacing/indentation

CC @therealolivergross @GormFrank

### DIFF
--- a/site/pages/docs/ref/wb-postback/wb-postback-en.hbs
+++ b/site/pages/docs/ref/wb-postback/wb-postback-en.hbs
@@ -6,7 +6,7 @@
 	"categoryfile": "plugins",
 	"description": "Submit web form through AJAX call.",
 	"altLangPrefix": "wb-postback",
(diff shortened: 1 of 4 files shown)


duboisp (approved):
I reviewed and tested locally, work as expected.

This change is only site content change.

Thanks for that documentation fix.

PR #9522, opened 2023-02-09
JSON Manager: Fix JSON pointer not always loaded

Garneauma:
JSON pointer was not loaded in some cases which broke JSON manager.

### DIFF
--- a/src/plugins/wb-jsonmanager/jsonmanager.js
+++ b/src/plugins/wb-jsonmanager/jsonmanager.js
@@ -243,7 +243,10 @@ var componentName = "wb-jsonmanager",
 			Modernizr.load( {
 
 				// For loading multiple dependencies
-				load: "site!deps/json-patch" + wb.getMode() + ".js",
+				load: [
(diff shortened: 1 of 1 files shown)


duboisp (changes requested):
When I did the local test, keeping only the fist data extractor working example, it is failing but it is when the provided CSS selector return an empty array.

So we will to apply an additional fix in this PR to ensure the data extraction still continue to work with returned empty array value.

The empty array are returned by the `querySelectorAll` function around line 403

```js
targetTag = document.querySelectorAll( selectedTag.selector || "" );
```

It generate a console error at the line 443 because `targetTag` is an undefined object and the properties `textContent` don't exist

``` js
targetTag [ 0 ].textContent;
```

Todo:
* [ ] Fix the use case where the CSS query selector return an empty array and the JSON manager will continue to work as expected.

duboisp:
Pre-approved after the requested change are completed and locally tested.

duboisp (approved):
Reviewed and tested locally, work as expected without error.

PR #9529, opened 2023-02-15
Twitter embedded timeline: Improve loading phase

### LINKED ISSUE #9503: [BUG] Twitter's spinning icon still shows
The spinning icon which should show getting feeds from Twitter is doubled and only one is removed after the feeds are received. 
It is present even in the demo
https://wet-boew.github.io/wet-boew/demos/twitter/twitter-en.html
Thank you,
Ionut

EricDunsworth:
The plugin's loading icon was previously tied to the "twitter-timeline" class... which is used by both the Twitter link and timeline container.

That resulted in the following behaviour:
1. The link's loading icon immediately appeared
2. A second loading icon appeared when Twitter's widget script injected an empty timeline container into the page
3. The link's loading icon got removed when the timeline finished loading... leaving the timeline's loading icon in place forever

The following issues also occurred:
* The empty space to the right of the link was a linked area
* The link's loading icon was a linked area
* The link's loading icon appeared "on top" of the link in mobile viewports
* The loading icon was center-aligned in a full-width overlay... causing it to look like it was in the middle of nowhere in larger viewports
* The link's loading icon appeared forever in IE11

This fixes it by reworking the loading phase as follows:
* The link's linked area now only covers the link text
* The plugin now injects a loading icon ``div`` below the link
* The loading icon now...
  * Appears below the link instead of "on top" of it
  * Is either left/right-aligned or center-aligned relative to the link if a browser supports the ``:has()`` pseudo-class
  * Disappears when Twitter's timeline widget removes the link (CSS immediately hides it, then the plugin removes it behind-the-scenes)
  * Never appears in IE11

Other:
* Only shows one loading icon per plugin container
* Plugin containers with extra Twitter links or non-standard structures won't show loading icons
* Disables the plugin in IE11 (Twitter's widget script no longer supports that browser)

Related:
* Fixes #9503
* Depends on #9539

**Screenshots...**

**Before:**
* **Phase 1 (loading icon on top of the link... centered relative to viewport)**
twitter-before-1
* **Phase 2 (two loading icons on top of the link)**
twitter-before-2
* **Phase 3 (centered loading icon on top of the timeline... forever)**
twitter-before-3

**After:**
* **Phase 1 (link before plugin init)**
twitter-after-1
* **Phase 2 (loading icon below the link... left-aligned by default, centered relative to link if ``:has()`` is supported)**
twitter-after-2-without-has OR twitter-after-2-with-has
* **Phase 3 (timeline)**
twitter-after-3

### DIFF
--- a/src/plugins/twitter/_base.scss
+++ b/src/plugins/twitter/_base.scss
@@ -3,33 +3,78 @@
  * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
  */
 
-.wb-disable {
-	.wb-twitter {
-		.twitter-timeline {
-			text-indent: 0 !important;
-		}
-	}
-}
-
 .wb-twitter {
-	.twitter-timeline {
-		display: block;
+	/* Align the loading icon relative to the center of the Twitter link in browsers that support :has() */
+	&:has( a.twitter-timeline + .twitter-timeline-loading ) {
+		display: inline-block;
+
+		a {
+			&.twitter-timeline {
+				+ {
+					.twitter-timeline-loading {
+						&::before {
+							left: 0;
+							right: 0;
+						}
+
+						&::after {
+							margin: auto;
+						}
+					}
+				}
+			}
+		}
 	}
 
-	iframe {
-		&[style] {
-			width: 100% !important;
+	/* Display the loading icon as normal content */
+	a {
+		&.twitter-timeline {
+			+ {
+				.twitter-timeline-loading {
+					@extend %global-loading;
+					margin-top: 5px;
+					min: {
+						height: 100px;
+						width: 100px;
+					}
+					position: relative;
+
+					/* Align the loading icon leftwards by default */
+					&::before {
+						right: auto;
+					}
+
+					&::after {
+						margin-left: calc((100px - 1em) / 2);
+					}
+				}
+			}
 		}
 	}
 }
 
-.wb-enable {
+[dir=rtl] {
 	.wb-twitter {
-		.twitter-timeline {
-			@extend %global-loading;
-			min-height: 100px;
-			min-width: 100px;
-			position: relative;
+		a {
+			&.twitter-timeline {
+				+ {
+					.twitter-timeline-loading {
+
+						/* Align the loading icon rightwards by default */
+						&::before {
+							left: auto;
+							right: 0;
+						}
+
+						&::after {
+							margin: {
+								left: auto;
+								right: calc((100px - 1em) / 2);
+							}
+						}
+					}
+				}
+			}
 		}
 	}
 }
--- a/src/plugins/twitter/twitter.js
+++ b/src/plugins/twitter/twitter.js
@@ -31,6 +31,49 @@ var componentName = "wb-twitter",
 			protocol = wb.pageUrlParts.protocol;
 
 		if ( eventTarget ) {
+			const twitterLink = eventTarget.firstElementChild;
+
+			// Ignore IE11
+			// Note: Twitter's widget no longer supports it...
+			if ( wb.ie11 ) {
+				wb.ready( $( eventTarget ), componentName );
+				return;
+			}
+
+			// If the plugin container's first child element is a Twitter link...
+			if ( twitterLink && twitterLink.matches( "a.twitter-timeline" ) ) {
+				const loadingDiv = document.createElement( "div" );
+				let observer;
+
+				// Add a loading icon below the link
+				loadingDiv.className = "twitter-timeline-loading";
+				twitterLink.after( loadingDiv );
+
+				// Remove the loading icon after the timeline widget appears
+				// Note: Twitter's widget script removes "a.twitter-timeline" upon filling-in the timeline's content... at which point the loading icon is no longer useful
+				observer = new MutationObserver( function( mutations ) {
+
+					// Check for DOM mutations
+					mutations.forEach( function( mutation ) {
+
+						// Deal only with removed HTML nodes
+						mutation.removedNodes.forEach( function( removedNode ) {
+
+							// If the removed node was a Twitter link, remove its adjacent loading icon and stop observing
+							if ( removedNode === twitterLink && mutation.nextSibling === loadingDiv ) {
+								loadingDiv.remove();
+								observer.disconnect();
+							}
+						} );
+					} );
+				} );
+
+				// Observe changes to the plugin container's direct child elements
+				observer.observe( eventTarget, {
+					childList: true
+				} );
+			}
+
 			Modernizr.load( {
 				load: ( protocol.indexOf( "http" ) === -1 ? "http:" : protocol ) + "//platform.twitter.com/widgets.js",
 				complete: function() {


EricDunsworth:
The build failures are being caused by CSSLint lacking support for ``:has()`` and nested ``calc()`` expressions.

Since [CSSLint](https://github.com/CSSLint/csslint) has gone stale (latest release was in fall 2016), it'll probably be necessary to migrate WET to a more modern CSS linter. I'm gonna try looking into [stylelint](https://github.com/stylelint/stylelint).

On a random side note, the reason ``:has()`` didn't trigger any build failures in wet-boew/GCWeb#1916 is because GCWeb doesn't currently use any CSS linters at all :S. CSSLint was probably removed from it when it adopted Jekyll.

src/plugins/twitter/_base.scss
```diff
@@ -3,33 +3,78 @@
  * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
  */
 
-.wb-disable {
-	.wb-twitter {
-		.twitter-timeline {
-			text-indent: 0 !important;
+.wb-twitter {
+	/* Align the loading icon relative to the center of the Twitter link in browsers that support :has() */
+	&:has( a.twitter-timeline + .twitter-timeline-loading ) {
```
duboisp, line 8: Add an exception for the build sasslint error.

duboisp (changes requested):

duboisp:
Pre-approved upon successful testing and after the sasslint exception is added for both selector or the file if not possible.

EricDunsworth:
Just tried adding inline CSSLint ignore comments (``/* csslint ignore:start */`` and ``/* csslint ignore:stop */``) without any luck :(.

From what I could surmise from issue discussions in [CSSLint/csslint](https://github.com/CSSLint/csslint), it looks like its ignore syntax isn't unconditional. Therefore, the things it dislikes about my ``:has()`` and ``calc()`` syntax are probably impossible to ignore.

Btw its ignore syntax isn't formally documented anywhere :S. Closest I found was a [wiki link to a non-existent page](https://github.com/CSSLint/csslint/wiki#ignoring-parts-of-css-during-linting).

As far as ignoring ``src/plugins/twitter/_base.scss`` in ``Gruntfile.coffee`` goes, I don't think that'll be possible either since CSSLint is used to lint *compiled* CSS files. So my only quick-fix solution would to exclude ``dist/theme-wet-boew/css/theme.css`` and ``dist/theme-wet-boew/css/ie8-theme.css`` from  CSSLint... which IMO is a non-starter since it'd effectively exclude almost all of WET's CSS.

> Pre-approved upon successful testing and after the sasslint exception is added for both selector or the file if not possible.

@duboisp Just to clarify - SASSLint lints SCSS source files and has no trouble with this PR. CSSLint lints compiled CSS and is causing my hurdles. I think migrating to stylelint will be my only viable option to get out of this mess. So this PR will probably need to be held back until next week.

src/plugins/twitter/_base.scss
EricDunsworth: Unfortunately this doesn't seem to be fixable (see https://github.com/wet-boew/wet-boew/pull/9529#issuecomment-1437329855), but I'll hold off marking this as a resolved until I get around to opening a stylelint PR.

duboisp:
@EricDunsworth humm, I think it's `/* csslint ignore:end */` not "stop", see: https://github.com/CSSLint/csslint/blob/master/src/core/CSSLint.js#L211-L225

It need both in order to work.

EricDunsworth:
@duboisp Just tried ``/* csslint ignore:start */`` and ``/* csslint ignore:end */`` without any luck :(.

Btw good catch on the ``end`` syntax. I stumbled upon ``stop`` via CSSLint/csslint#564 while searching around the repo. Guess it got renamed to ``end`` later on.

src/plugins/twitter/_base.scss
```diff
@@ -3,33 +3,78 @@
...
+					/* Align the loading icon leftwards by default */
+					&::before {
+						right: auto;
+					}
 
-	iframe {
-		&[style] {
-			width: 100% !important;
+					&::after {
+						margin-left: calc( ( 100px - 1em ) / 2 );
```
duboisp, line 48: Stylelint error: Unexpected whitespace after "(" in a single-line function

```suggestion
						margin-left: calc(( 100px - 1em ) / 2 );
```

src/plugins/twitter/_base.scss
```diff
@@ -3,33 +3,78 @@
...
+						/* Align the loading icon rightwards by default */
+						&::before {
+							left: auto;
+							right: 0;
+						}
+
+						&::after {
+							margin: {
+								left: auto;
+								right: calc( ( 100px - 1em ) / 2 );
```
duboisp, line 72: Stylelint error: Unexpected whitespace after "(" in a single-line function

```suggestion
								right: calc(( 100px - 1em ) / 2 );
```

duboisp (changes requested):
See the build error, meanwhile I am continuing my local testing.

src/plugins/twitter/_base.scss
EricDunsworth: I came super close to doing that locally, but noticed that the trailing white space after ``2`` wasn't triggering an error (even though [``function-parentheses-space-inside``'s docs](https://stylelint.io/user-guide/rules/function-parentheses-space-inside/) suggest that it's supposed to).

I'm gonna change it up to this:
```suggestion
						margin-left: calc((100px - 1em ) / 2);
```

That's in line with how all of WET and most of GCWeb's pre-existing ``calc()`` pseudo-classes are spaced-out.

src/plugins/twitter/_base.scss
EricDunsworth: Gonna revise this to be in line with https://github.com/wet-boew/wet-boew/pull/9529#discussion_r1128536180.

src/plugins/twitter/_base.scss
EricDunsworth: Marking this as resolved since #9539 fixed the build error.

src/plugins/twitter/twitter.js
```diff
@@ -31,6 +31,61 @@ var componentName = "wb-twitter",
...
+			if ( !wb.ie11 ) {
+				const linkSelector = "a.twitter-timeline",
+					twitterLinks = eventTarget.querySelectorAll( linkSelector ),
+					loadingClass = "twitter-timeline-loading";
+				let loadingRemovals = 0,
+					observer;
+
+				// Insert loading icons below Twitter links
+				twitterLinks.forEach( function( twitterLink ) {
+					const loadingDiv = document.createElement( "div" );
```
duboisp, line 46:
```suggestion
					let loadingDiv = document.createElement( "div" );
```

`const` in this context are not sealing the inner object.

src/plugins/twitter/twitter.js
```diff
@@ -31,6 +31,61 @@ var componentName = "wb-twitter",
...
+					loadingDiv.className = loadingClass;
+
+					twitterLink.after( loadingDiv );
+				} );
+
+				// Remove loading icon after the timeline widget appears
+				// Note: Twitter's widget script removes "a.twitter-timeline" upon filling-in the timeline's content... at which point the loading icon is no longer useful
+				observer = new MutationObserver( function( mutations ) {
```
duboisp, line 54: Comment:

What about to use the CSS animation + JS event. Like binded on this CSS selector "`.twitter-timeline-rendered`"

For an example, see wet-boew-experimental here: https://github.com/wet-boew/wet-boew-experimental/blob/main/src/wet-boew.js#L97-L101

src/plugins/twitter/twitter.js
```diff
@@ -31,6 +31,61 @@ var componentName = "wb-twitter",
...
+						// Deal only with removed HTML nodes
+						mutation.removedNodes.forEach( function( node ) {
+							const twitterLink = node.matches( linkSelector );
+
+							// Continue if the removed node was a Twitter link
+							if ( twitterLink ) {
+								const linkSibling = mutation.nextSibling;
+
+								// Remove the link's adjacent loading icon
+								if ( linkSibling instanceof HTMLElement && linkSibling.matches( "." + loadingClass ) ) {
```
duboisp, line 68: Propose change:

```suggestion
								if ( linkSibling.classList && linkSibling.classList.contains( loadingClass ) ) {
```

src/plugins/twitter/twitter.js
```diff
@@ -31,6 +31,61 @@ var componentName = "wb-twitter",
...
+									linkSibling.remove();
+									loadingRemovals++;
+
+									// Stop observing after every link's loading icon has been removed
+									if ( loadingRemovals === twitterLinks.length ) {
+										observer.disconnect();
+									}
```
duboisp, line 75: Ça m'indique qu'il y a probablement un autre problème, comme si l'init est appeler 2 fois ou si l'observer est au niveau du document et pas de l'élement.

Je sais que quelque de nos plugin utilise ceci pour éviter ce genre de problème: `event.currentTarget === event.target`

duboisp (changes requested):

src/plugins/twitter/twitter.js
```diff
@@ -31,6 +31,61 @@ var componentName = "wb-twitter",
 			protocol = wb.pageUrlParts.protocol;
```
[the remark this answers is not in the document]
EricDunsworth: I noticed your suggestion was to place this *before* the ``if ( eventTarget ) { ... }`` condition.

Was that intentional? I figure it'd better to put it inside ``if`` to guarantee it's truthy before proceeding. But maybe I'm overlooking something?

src/plugins/twitter/twitter.js
EricDunsworth: Good catch :)!

I dug around a bit and it looks like this suggestion would be simpler + more performant + more compatible (with older browsers) vs what I was doing.

src/plugins/twitter/twitter.js
EricDunsworth: I think that'd be pretty cool and am considering going in that direction. It might make for cleaner/shorter code than a mutation observer. No idea which would have better performance though.

As far as the  ``.twitter-timeline-rendered`` selector goes, it looks like that gets injected by Twitter's third party script when it inserts the timeline's ``iframe`` into the page. But it still takes several seconds for the timeline to appear afterwards.

I'll probably do something along the lines of checking for the ``animationcancel`` event on ``".twitter-timeline-loading``. When Twitter's script removes the link, the loading ``div`` stops being adjacent to it - which should cause the loading icon to instantly disappear and trigger an ``animationcancel`` event.

src/plugins/twitter/twitter.js
EricDunsworth: It isn't supposed to :P.

I felt like ``const`` was a better bet than ``let`` for this since the ``div`` element that's getting created isn't intended to turn into a different element later on. Only its properties are subject to change (like getting a class name).

As we chatted about yesterday, I could change this to ``let`` if you want since that's more in line with what most programming languages usually expect a constant to represent (i.e. set-in-stone value). But I feel like that would betray how JavaScript ``const`` variables are meant to work :D lol.

I know that ``const`` is kind of quirky in JavaScript since it doesn't require DOM node properties to be totally set-in-stone. But at the same time, IMO it would be inappropriate to forego using ``const`` in those quirky situations just to follow a common convention that only applies to *other* programming languages.

src/plugins/twitter/twitter.js
EricDunsworth: I experimented a bit with the``animationcancel`` event listener yesterday and it looks like it can work in this scenario :).

But it just hit me that making the plugin dependent on CSS animation events might not work as well as a mutation observer in practice :(.

Here's why:
* If a user were to use client-side CSS overrides to disable animations, it'd probably disable animation events. In that scenario, the loading icon would still visually-disappear when the Twitter link gets removed, but its ``div`` element would never get removed from the DOM.
* The ``a11y/media-prefers-reduced-motion`` rule from stylelint's ``stylelint-a11y`` plugin checks for ``@media (prefers-reduced-motion)`` overrides. If we activate that rule in the future and update WET's CSS to follow it, we'll probably want to disable the loading icon's spinning animation... which'll probably mess up the the animation event listener.

src/plugins/twitter/twitter.js
EricDunsworth: Even though I disagreed with what was being asked here, I ultimately ended changing ``const`` to ``let`` for ``loadingDiv`` due to other unrelated changes.

``forEach()`` is no longer being used at this spot... Each plugin instance is now expected to have only one loading icon at most.

``loadingDiv`` is now declared at the start of the plugin since it's now being used for both adding and removing the loading icon.

I could've technically kept using ``const``, but a loading icon only gets added if a suitable Twitter link exists... so there's no point in creating a useless HTML element for ``loadingDiv`` unless a link is actually found. So I went with ``let`` since it's more flexible.

src/plugins/twitter/twitter.js
EricDunsworth: J'ai décidé d'arrêter de supporter plus qu'un lien de Twitter à la fois. Alors cette logique a été éliminé :).

EricDunsworth:
Just force-pushed an update :)!

**Here's what's changed:**
* Revise IE11 ignore logic to disable the entire plugin in that browser
* Rework logic to treat each plugin instance as having only one Twitter link situated at the beginning (i.e. strictly adhere to WET's demo/documentation structure)
  * Plugin containers with extra links or non-standard structures will still load all their timelines, but won't show loading icons
* Only use the mutation observer if a loading icon gets inserted

**Heads up:**
@duboisp We're probably going to need to discuss variable declarations and ``const`` some more.

At one point earlier today, I had them setup like what the requested changes wanted and marked most as resolved. But then I significantly reworked more logic and didn't feel like that way of declaring variables made sense anymore. So I spread-out the declarations and went back to ``const`` for ``loadingDiv``.

I think we're gonna need to "pick our poison" 😛:
1. Status quo (scoping out variables)
2. Readability (grouping all variables at the start of the plugin
3. Hybrid (move scoped variable declarations around so they come right after the comments for the logic they tie into):
   * Arguably the best of both worlds
   * Would improve scoping since it'd allow the mutation observer to be declared as ``const`` (it's currently using ``let`` since I wanted to keep scoped variable declarations at the beginning of their code blocks)
   * Readability would be improved since scope variable declarations would always be "in context" (i.e. directly after the comments for the logic they're needed for)
   * Downside is that this approach would betray common programming conventions (i.e. everyone usually expects to see variable declarations at the beginning of functions or code blocks)

Thoughts?

src/plugins/twitter/twitter.js
```diff
@@ -31,6 +31,53 @@ var componentName = "wb-twitter",
...
+					mutations.forEach( function( mutation ) {
+
+						// Deal only with removed HTML nodes
+						mutation.removedNodes.forEach( function( removedNode ) {
+
+							// Continue if the removed node was a Twitter link
+							if ( removedNode === twitterLink ) {
+
+								// Remove the link's adjacent loading icon and stop observing
+								if ( mutation.nextSibling === loadingDiv ) {
```
EricDunsworth, line 66: Maybe it'd be combining these two ``if``s together. I overlooked it last week.

GormFrank:
Pre-approved upon successful review & testing

src/plugins/twitter/twitter.js
EricDunsworth: Just combined them :).

duboisp (approved):
Reviewed and tested locally (in FF and Edge) and work as expected.

Thanks @EricDunsworth

PR #9531, opened 2023-02-16
Build: Bump grunt-contrib-uglify to 5.2.1 to support new JS features

Garneauma:
Needed for the upcoming dropdown plugin which leverages ES6 "class" feature.

PR #9532, opened 2023-02-17
Dropdown: implementing dropdown plugin

Garneauma:
Dependent on PR #9531.

- [x] Once PR #9531 is merged, remove `package.json` and `package-lock.json` from this commit.

Changes related to WET-274 - Stabilize add to calendar plugin.

### DIFF
--- a/site/pages/docs/ref/dropdown/dropdown-en.hbs
+++ b/site/pages/docs/ref/dropdown/dropdown-en.hbs
@@ -1,0 +1,94 @@
+---
+{
+	"title": "Dropdown button",
+	"language": "en",
+	"category": "Plugins",
+	"categoryfile": "plugins",
+	"description": "Dropdown button to display a list of actions or links on click",
+	"altLangPrefix": "dropdown",
+	"dateModified": "2023-02-15"
+}
+---
+<div class="alert alert-warning">
+	<h2>Unstable feature</h2>
+	<p>To be used at <strong>your own risk</strong>. This feature described below can be removed in any subsequent minor/major release</p>
+	<p><small><a href="https://wet-boew.github.io/wet-boew-documentation/decision/7.html">Learn more about the design decision around provisional features.</a></small></p>
+	<p><small><a href="../provisional-en.html">Check other provisional features available.</a></small></p>
+</div>
+
+<span class="wb-prettify all-pre"></span>
+
+<h2>Purpose</h2>
+<p>Dropdown button to display a list of actions or links on click</p>
+
+<h2>Working example</h2>
+<ul>
+	<li><a href="../../../demos/dropdown/dropdown-en.html">English example</a></li>
+	<li><a href="../../../demos/dropdown/dropdown-fr.html">French example</a></li>
+</ul>
+
+<h2>Evaluation and report</h2>
+<p>There is no evaluation and report available for this component.</p>
+
+<h2>API (Version 1.0)</h2>
+<dl class="dl-horizontal">
+	<dt>Function</dt>
+	<dd>Version 1.0</dd>
+	<dt>Configuration</dt>
+	<dd>Not applicable</dd>
+	<dt>User interface (Template)</dt>
+	<dd>Version 1.0</dd>
+	<dt>Data source</dt>
+	<dd>Version 1.0</dd>
+	<dt>View and style</dt>
+	<dd>Version 1.0</dd>
+	<dt>i18n string</dt>
+	<dd>Version 1.0</dd>
+</dl>
+
+<h3>Function</h3>
+<p>(Version 1.0)</p>
+<table class="table">
+	<tr>
+		<th>Function type</th>
+		<th>Name</th>
+		<th>How to implement</th>
+		<th>What it does</th>
+	</tr>
+	<tr>
+		<td>jQuery Event</td>
+		<td><code>wb-init.dropdown</code></td>
+		<td>Triggered manually (e.g., <code>$( ".dropdown" ).trigger( "wb-init.dropdown" );</code>).</td>
+		<td>Initializes the <code>.dropdown</code> plugin. This plugin will be initialized automatically unless the <code>.dropdown</code> element is added after the page load and wet-boew was initialized.</td>
+	</tr>
+	<tr>
+		<td>jQuery Event</td>
+		<td><code>wb-ready.dropdown</code></td>
+		<td>Triggered automatically after the plugin initializes.</td>
+		<td>Used to identify when and where the plugin initializes.
+			<pre><code>$( document ).on( "wb-ready.dropdown", ".dropdown", function( event ) {
+});</code></pre>
+			<pre><code>$elm.on( "wb-ready.dropdown", function( event ) {
+});</code></pre>
+		</td>
+	</tr>
+</table>
+
+<h3>Configuration</h3>
+<p>Not applicable for this plugin.</p>
+
+<h3>User interface (Template)</h3>
+<p>Version 1.0</p>
+
+<pre><code>&lt;div class="dropdown"&gt;
+	&lt;button class="btn btn-default dropdown-toggle" type="button"&gt;Dropdown&lt;/button&gt;
+	&lt;ul class="dropdown-menu"&gt;
+		&lt;li&gt;&lt;a href="#"&gt;Link 1&lt;/a&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;a href="#"&gt;Link 2&lt;/a&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;a href="#"&gt;Link 3&lt;/a&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;a href="#"&gt;Link 4&lt;/a&gt;&lt;/li&gt;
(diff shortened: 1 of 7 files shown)


polmih (changes requested):
Tested on local environment, working as expected on Chrome, Firefox, Edge, for all 3 types of devices (cell, tablet, desktop). 
Minor change requests related to translation and code examples

polmih (dismissed):

src/plugins/dropdown/dropdown-fr.hbs
```diff
@@ -0,0 +1,210 @@
...
+	&lt;button class="btn btn-default btn-xs dropdown-toggle" type="button"&gt;Menu déroulant &lt;span class="caret"&gt;&lt;/span&gt;&lt;/button&gt;
+	&lt;ul class="dropdown-menu"&gt;
+		&lt;li&gt;&lt;a href="#"&gt;Un lien&lt;/a&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;a href="#"&gt;Un autre lien&lt;/a&gt;&lt;/li&gt;
+		&lt;li&gt;&lt;a href="#"&gt;Autre chose&lt;/a&gt;&lt;/li&gt;
+	&lt;/ul&gt;
+&lt;/div&gt;</code></pre>
+</details>
+
+<h2>Couleurs</h2>
```
duboisp, line 111:
```suggestion
<h2>Menu déroulant basé sur un état</h2>
```

We can't rely on colour. We need to rely on the meaning provided by that use of some specific colour.

duboisp (changes requested):
Here a partial review, I only reviewed the design pattern for now.
The interaction pattern, once enhanced, seem to works as expected. 

* [ ] Regroup the bootstrap CSS with this plugin CSS

* [ ] Is the dropdown pattern is limited at the following?

```
<div class="wb-dropdown">
  <button></button>
  <ul>
      <li></li>
   </ul>
</div>
```

* Where: `.wb-dropdown` is the component scope
* The component only have 2 allowed children in this sequence:
   1. `<button>`
   2. `<ul>`
* The menu item is restricted (by the design pattern) to be either anchor or button

* [x] Can we have dropdown inside dropdown or is that pattern not supported yet? If so, let add a note or an example about that.

* [ ] Can you adjust the focus visible style for a conformity to [Focus Appearance (Level AA)](https://www.w3.org/TR/WCAG22/#focus-appearance)

* [x] Enhance the pattern with JS instead of relying on the author. See the inline notes.

* [x] The various "color" is attached to convey a state, our working example and section title need to be adjusted to convey the same intention. The working example would need to be updated and same thing for the button name to avoid a failure to [SC 1.4.1 Use of Color](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html)

* [ ] Perform a second review on the design pattern, review the CSS and review the JS

Garneauma:
Fixed all issues mentioned in comments.

# Can you adjust the focus visible style for a conformity to [Focus Appearance (Level AA)](https://www.w3.org/TR/WCAG22/#focus-appearance)

I am not sure what you mean. There already is a border around the item on keyboard focus. I don't know what else should be added.

src/plugins/dropdown/dropdown.js
```diff
@@ -0,0 +1,369 @@
...
+		this.buttonNode.classList.add( "dropdown-toggle" );
+		this.buttonNode.addEventListener( "keydown", this.onButtonKeydown.bind( this ) );
+		this.buttonNode.addEventListener( "click", this.onButtonClick.bind( this ) );
+
+		// Set ID's
+		this.buttonNode.id = this.buttonNode.id || wb.getId();
+		this.menuNode.id = this.menuNode.id || wb.getId();
+
+		// Set attributes for accessibility
+		this.menuNode.setAttribute( "role", "menu" );
```
duboisp, line 62:
```suggestion
```

Is it always a menu?

src/plugins/dropdown/dropdown.js
```diff
@@ -0,0 +1,369 @@
...
+	}
+
+	onButtonKeydown( event ) {
+		var key = event.key,
+			flag = false;
+
+		switch ( key ) {
+		case " ":
+		case "Enter":
+		case "ArrowDown":
```
duboisp, line 226:
```suggestion
		case "ArrowDown":
```

Not sure if we should open up the dropdown when the user use the arrow key. Some user do read the content by using the arrow key. I think relying on an action key like the "enter" or the "space bar" only to open the dropdown may provide a better user experience for screen reader. This need to be validated with pattern described in the WAI-ARIA practice.

duboisp (changes requested):
Great work, 

Here another partial review, see my observation bellow. I am going to complete my review+testing this afternoon.

Observations:

* [ ] The dropdown button is still displayed in basic HTML mode and it is not operable. Suggest to hide it in basic mode.
* [x] Add a little JS action for the demo that use buttons
* [x] Mouse hover contrast is barely visible on the inner interactive element of the dropdown. Especially when compared an anchor or a button outside the dropdown pattern
* [x] The link and the button rendering are exactly the same in the drop down. Can we render them differently? A link bring you to somewhere and a button do an action on the current page.
* [x] When clicking the button, there is no style when clicked like a normal button.
* [x] Is the items in a drop down is always a menu? Overriding the role on anchor and button can be problematic here. Let's discuss.
* [x] Missing an aria-hidden on the caret icon.

src/plugins/dropdown/dropdown.js
duboisp: This is a non-issue, that behaviour for "up" and "down" are identified in the Aria pattern here: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#keyboardinteraction

site/pages/docs/ref/dropdown/dropdown-en.hbs
```diff
@@ -0,0 +1,94 @@
...
+});</code></pre>
+		</td>
+	</tr>
+</table>
+
+<h3>Configuration</h3>
+<p>Not applicable for this plugin.</p>
+
+<h3>User interface (Template)</h3>
+<p>(Provisional)</p>
```
duboisp, line 81:
```suggestion
```

site/pages/docs/ref/dropdown/dropdown-fr.hbs
```diff
@@ -0,0 +1,95 @@
...
+	<dd>N'est pas applicable</dd>
+	<dt>Interface utilisateur (Modèle)</dt>
+	<dd>Provisoire</dd>
+	<dt>Source de données</dt>
+	<dd>Provisoire</dd>
+	<dt>Affichage et style</dt>
+	<dd>Provisoire</dd>
+	<dt>Chaîne de caractères i18n</dt>
+	<dd>Provisoire</dd>
+</dl>
```
duboisp, line 48:
```suggestion
<dl class="dl-horizontal">
	<dt>Fonction</dt>
	<dd>Version 1.0</dd>
	<dt>Configuration</dt>
	<dd>N'est pas applicable</dd>
	<dt>Interface utilisateur (Modèle)</dt>
	<dd>Version 1.0</dd>
	<dt>Source de données</dt>
	<dd>Version 1.0</dd>
	<dt>Affichage et style</dt>
	<dd>Version 1.0</dd>
	<dt>Chaîne de caractères i18n</dt>
	<dd>Version 1.0</dd>
</dl>
```

site/pages/docs/ref/dropdown/dropdown-fr.hbs
```diff
@@ -0,0 +1,95 @@
...
+});</code></pre>
+		</td>
+	</tr>
+</table>
+
+<h3>Configuration</h3>
+<p>N'est pas applicable pour cette composante.</p>
+
+<h3>Interface utilisateur (Modèle)</h3>
+<p>(Provisoire)</p>
```
duboisp, line 82:
```suggestion
```

src/plugins/dropdown/dropdown-en.hbs
```diff
@@ -0,0 +1,212 @@
...
+&lt;/div&gt;</code></pre>
+</details>
+
+<h2>Dropdown with action buttons</h2>
+<div class="wb-dropdown">
+	<button class="btn btn-default">Dropdown</button>
+	<ul>
+		<li><button>Action</button></li>
+		<li><button>Another action</button></li>
+		<li><button>Something else</button></li>
```
duboisp, line 45: Let use a different label for the button vs the link.

```suggestion
		<li><button>Action button</button></li>
		<li><button>Another action button</button></li>
		<li><button>Something else button</button></li>
```

duboisp (changes requested):
Just a few small change and a note to not forget to apply the same content change to the French version.

This complete my review and testing until all the requested change is addressed.

duboisp (changes requested):
Tested and the following todo:
* [ ] Transform the dropdown button into paragraph in basic HTML mode
* [ ] Apply the 2 little content propose change

Other than that, it should be good to go upon a subsequent local testing and code review of the additional changes.

Garneauma:
Closing as we transition to GCDS, we do not want to include more components here. This can be implemented in GCDS directly.

PR #9534, opened 2023-02-17
Postback: Fix missing submit button variable

EricDunsworth:
The postback plugin was previously excluding submit button variables in the following scenarios:
1. When the submit button was a type-less ``button`` element (the ``button`` element's default type is ``submit`` according to the HTML spec)
2. When the form validation plugin was used

This fixes it by:
1. Adjusting button selectors as follows:
   * Expand the ``$buttons`` selector to cover type-less button elements
   * Remove ``[type=submit]`` from the ``$btn`` selector (what remains should be able to pick up any engaged submit button associated to a variable)
2. Removing the ``data-wb-engaged`` attribute from a clicked submit button if form validation fails (opposite of what was previously being done)

Also updates the demo page's examples to better represent the fixed scenarios:
* Add submit button variable examples (``name``/``value`` attributes):
  * Standard form with just a submit button
  * Form validation plugin with one required field and a submit button
* Remove all submit buttons' ``type="submit"`` attributes
* Gruntfile: Add exclusion for bootlint's ``W007`` rule to suppress errors on type-less ``button`` elements
* Place all examples into their own sections

### DIFF
--- a/src/plugins/wb-postback/wb-postback.js
+++ b/src/plugins/wb-postback/wb-postback.js
@@ -27,7 +27,7 @@ var $document = wb.doc,
 					wb.getData( $elm, componentName )
 				),
 				attrEngaged = "data-wb-engaged",
-				$buttons = $( "[type=submit]", $elm ),
+				$buttons = $( "[type=submit], button:not([type])", $elm ),
 				multiple = typeof $elm.data( componentName + "-multiple" ) !== "undefined",
 				classToggle = settings.toggle || "hide",
 				selectorSuccess = settings.success,
@@ -48,15 +48,15 @@ var $document = wb.doc,
 				if ( elm.parentElement.classList.contains( "wb-frmvld" ) ) {
 					if ( !$elm.valid() ) {
 						$( this ).attr( attrEngaged, true );
-					} else {
 						$buttons.removeAttr( attrEngaged );
+					} else {
 						$( this ).attr( attrEngaged, "" );
 					}
 				}
 
 				if ( !$( this ).attr( attrEngaged ) ) {
 					var data = $elm.serializeArray(),
-						$btn = $( "[type=submit][name][" + attrEngaged + "]", $elm ),
+						$btn = $( "[name][" + attrEngaged + "]", $elm ),
 						$selectorSuccess = $( selectorSuccess ),
 						$selectorFailure = $( selectorFailure );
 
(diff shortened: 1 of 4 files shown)


src/plugins/wb-postback/wb-postback.js
```diff
@@ -27,7 +27,7 @@ var $document = wb.doc,
 					wb.getData( $elm, componentName )
 				),
 				attrEngaged = "data-wb-engaged",
-				$buttons = $( "[type=submit]", $elm ),
+				$buttons = $( "[type=submit], button:not([type=button]):not([type=reset])", $elm ),
```
duboisp, line 30: Here simplified version

```suggestion
				$buttons = $( "[type=submit], button:not([type]])", $elm ),
```

This match the HTML5 specification: https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element:missing-value-default

duboisp (changes requested):
See the CSS selector improvement,

We will need at least 1 example where it don't use `[type=submit]` for testing + another one for the other change if we can't reproduce with the existing example.

src/plugins/wb-postback/wb-postback.js
EricDunsworth: Good idea :)! I like this more than what I started out with since it's simpler and more future-proof.

EricDunsworth:
Just force-pushed a few updates :)!

Here's an overview of net-new changes:
* Simplified the script's first selector as per https://github.com/wet-boew/wet-boew/pull/9534#discussion_r1112087204
* Updated the demo page's examples as follows:
  * Added submit button variables (``name``/``value`` attributes)
  * Removed ``type="submit"`` from the submit buttons

As for why I modified pre-existing examples rather than tack in new ones... I personally like how the postback plugin's demo page is already arranged (two takes on the same example) and didn't want to add any bloat to it. IMO tinkering with the pre-existing examples was good-enough to show off this PR's use cases without compromising what was already being demoed.

EricDunsworth:
Looks like bootlint dislikes type-less ``button`` elements :S...

duboisp:
> Looks like bootlint dislikes type-less button elements :S...

We can add an exception in our Gruntfile.coffee for that bootlint rule specifically or change it for a warning only. The HTML5 spec allow it but I understand being explicit more explicit make it clear about the author intend on how that button should behave.

duboisp (changes requested):
I was thinking about 1 modified working example (or a new one) that showcase your need, not to modifying all of them.

We must keep one standard and simple example and having "name" + "value" on button are not so common on simple web form.

EricDunsworth:
Just force-pushed an update for this :).

EricDunsworth:
Ugh sorry - had to do another force-push to get rid of a couple of commented-out task calls and a random empty line in my Gruntfile. Should be all sorted out now.

duboisp (approved):
I reviewed the code and tested locally, it works as expected.

duboisp:
Considering a gray zone we do have currently in our versioning API, we will consider this change as a Minor because it represent a new coded variant of the postback plugin.

PR #9536, opened 2023-02-20
Core - wb: add separate initialization for components when in basic HTML mode

polmih:

- add separate initialization for components when in basic HTML mode;
- create basicInit function for data-ajax component to allow to be executed in basic HTML mode. For that a new data attribute is added: `data-wb-basic-ajax=true`


EricDunsworth:
Was this superseded by #9628?

PR #9539, opened 2023-02-24
Build: Replace csslint with stylelint

EricDunsworth:
Csslint has become a burden on WET's build system:
* It's stale (last updated in 2016)
* Triggers false errors on modern CSS features (e.g. nested ``calc()`` expressions and ``:has()``)
* Inline ignore syntax is unable to unconditionally suppress false errors
* WET's activated rules mostly-focused on supporting obsolete versions of IE

Stylelint seems best-suited to replace it for the following reasons:
* Still being maintained to this day
* Can replace sass-lint (which has also gone stale)
* Many former csslint users have jumped ship to it
* Has a prior history with csslint
* Supports plugins

Implementation details:
* Dependencies:
  * New dependencies target stylelint 14.x (15.0.0 no longer supports Node.js 12.x)
  * Adds:
    * ``grunt-stylelint``
    * ``stylelint``
    * ``stylelint-config-standard``
    * ``@double-great/stylelint-a11y``
  * Removes:
    * ``grunt-contrib-csslint``
* Gruntfile:
  * Renames the ``csslint`` task to ``stylelint``
  * No longer outputs errors to a file (doing it would cause them to stop getting listed in the CLI)
* Lint configuration:
  * Removes .csslintrc
  * Adds .stlylelintrc.yml:
    * Extends the ``stylelint-config-standard`` ruleset:
      * Overrides negate pre-existing errors
    * Uses the ``stylelint-a11y`` plugin:
      * Activates rules that already pass
      * Disables rules that trigger errors (didn't need to declare them, but wanted to set the stage to activate more later on)
      * Thanks @radmorecameron for suggesting it :)

Notes:
* Stylelint is slower than csslint:
  * #9388 should provide a substantial speed boost (it'll practically halve the amount of compiled CSS that needs to be linted)
* Stylelint's rulesets differ substantially from csslint's:
  * Stylelint mainly focuses on modern-day CSS
  * Csslint focused a lot on supporting obsolete browsers
* Csslint rules that were previously activated in WET's build system:
  * Rules with stylelint successors:
    * ``errors`` (``CssSyntaxError``)
    * ``vendor-prefix`` (``property-no-vendor-prefix``... sort of)
  * No stylelint equivalents:
    * ``bulletproof-font-face``
    * ``display-property-grouping``
    * ``font-faces``
    * ``import``
    * ``import-ie-limit``
    * ``performant-transitions``
    * ``rules-count`` (possibly unused... didn't have a default value)
    * ``star-property-hack``
    * ``underscore-property-hack``


duboisp:
If needed, we can upgrade the nodejs version we can use to build wet-boew. Like with the following NodeJS version: LTS 16, LST 18, 19.

radmorecameron:
Might be a good idea to include this stylelint plugin for accessibility @double-great/stylelint-a11y

https://www.npmjs.com/package/@double-great/stylelint-a11y
https://github.com/double-great/stylelint-a11y

EricDunsworth:
@duboisp I don't think it'll be necessary at this time. Node.js 12.x support was dropped in ``stylelint`` 15.0.0. But that was super recently (2-3 weeks ago IIRC). Plus ``grunt-stylelint`` still currently depends on ``stylelint`` ^14.5.0.

I think the show-stopper failures on my first draft's GitHub Actions builds were due to my direct ``stylelint`` dependency being the latest version (15.2.0)... I was on Node.js 12.x and my local build failed. I wanted to see how GH Actions would fare with its parallel 12.x and 14.x builds. IIRC both failed. So ``grunt-stylelint`` probably didn't take kindly to ``stylelint`` 15.x.

My recent force-push downgraded the direct ``stylelint`` dependency to ^14.6.1 and seems to have fixed everything. I also used ruleset versions that predate ``stylelint`` 15.x.

It might be a bad sign of things to come in the future though if stylelint functionality starts getting affected by dependency hell. Particularly since ``grunt-stylelint`` is an unofficial repo (maintained by the Wikimedia foundation). But it's still a better bet than CSSLint or no CSS linter at all.

---

@radmorecameron Thanks for the heads up :) !

I checked it out and am currently experimenting with it. Not totally sure if I'll include it for real yet. But it might make a cameo my next force push.

I was unable to extend its recommended rules in my ``.eslintrc.yml`` file for some reason :S. But that only activated like 2-3 rules anyway.

I tried setting all of its rules to ``true`` in my config file as a test and got bombarded with errors :D:
* The only errors that didn't trigger any failures out-of-the-box were ``a11y/no-obsolete-attribute`` and ``a11y/no-spread-text``.
* I'd of expected ``a11y/no-obsolete-element`` to be ok too, but it mistreated the ``hgroup`` and ``menu`` elements as obsolete.

What kind of spooks me about that plugin is that I'm not sure how much WET 4.0 could benefit from it. I feel like many of its rules are kind of opinionated or redundant. For example:
* ``no-display-none`` treats ``display: none;`` as a show-stopper... even though from an accessibility standpoint, there's nothing inherently wrong with it if the goal is to deliberately hide something from all users (including AT users).
* ``no-obsolete-attribute`` and ``no-obsolete-element`` would probably already get caught by ``htmllint``. Although I suppose they could potentially catch outdated CSS selectors that don't exist anywhere in the repo's HTML code.
* I am on board with things like ``media-prefers-reduced-motion``.

In any case, I'm not really planning on introducing ``stylelint`` in a manner that'll trigger a ton of new errors in what already exists. I'm thinking the best way of rolling it in will be to tune it to pass "as-is" for now (to open the door to merging #9529), then update the repo to follow stricter rulesets later on.

Btw it looks like [YozhikM/stylelint-a11y (upstream)](https://github.com/YozhikM/stylelint-a11y) went stale a few years ago. Last release was in late 2019 and its latest commit is from spring 2020. What you suggested ([double-great/stylelint-a11y](https://github.com/double-great/stylelint-a11y)) is one of [multiple forks of it that are still being kept up-to-date](https://github.com/YozhikM/stylelint-a11y/issues/69).

I haven't tried comparing the forks in-depth, but at first glance they all look comparable. None have added any rules beyond what was already available in upstream. YozhikM/stylelint-a11y#69 is trying to encourage the fork maintainers to consolidate their efforts to create a true successor, but hasn't made any progress as of yet.

EricDunsworth:
I'm gonna hold off replacing ``sass-lint`` as a part of this PR to give it a more-targeted focus and get it merged more quickly (#9529 can't wait forever :P).

It should be feasible to replace ``sass-lint`` with ``stylelint`` in a future PR. But I expect that it'll be more involved than meets the eye. It's probably going to need a mix of these:
* ``stylelint-config-sass-guidelines`` (extras for SCSS-focused linting)
* ``prettier`` (to handle [opinionated rules stylelint 15.0.0 is deprecating](https://stylelint.io/user-guide/rules#deprecated)).

Later on, we'll probably want to consider doing more to ensure the same rules aren't needlessly re-evaluated on both source SCSS/CSS files and "compiled" CSS.

For example:
* ``indentation`` makes sense for source files, but not really for compiled/minified CSS.
* Ditto for things like ``media-prefers-reduced-motion`` (can't really evaluate media query usage until everything is combined together).

It should be possible to enforce stricter rules on source files since they're not at the mercy of any questionable things originating from external dependencies (like Bootstrap 3.x's ``text-justify`` class failing ``a11y/no-text-align-justify``).

EricDunsworth:
This should be ready to rock and roll now :).

Check out the updated OP for more info.

package-lock.json
```diff
@@ -458,7 +524,7 @@
     "absolute": {
       "version": "0.0.1",
       "resolved": "https://registry.npmjs.org/absolute/-/absolute-0.0.1.tgz",
-      "integrity": "sha512-Zs3Sz3lmEIH76fvnc4jto4CpuI5bhU2TD+jtBYVIfXnjFm+49y1opKxNy4gtNF+xSLBL7hXRpLwypc+o2MzBIQ==",
+      "integrity": "sha1-wigi+H4ck59XmIdQTZwQnEFzgp0=",
```
EricDunsworth, line 527: Looks like some SHA-512 hashes somehow got downgraded to SHA-1 :S.

I don't think that'll be show-stopper, but wanted to point it out just in case.

duboisp: that is fine

duboisp:
pre-approve upon successful testing

.stylelintrc.yml
```diff
@@ -0,0 +1,65 @@
+extends:
+  - stylelint-config-standard
+
+plugins:
+  - "@double-great/stylelint-a11y"
+
+rules:
+  # Suppress stylelint-config-recommended errors
```
duboisp, line 8: I don't think we should have all those exception but instead we should work toward to remove all of them and fix our code base to match as close as possible to apply all the standard recommendation set of rules.

I do know the immediate need with this feature. I am proposing to merge this one after this is completed
* Add an explicit comment in our config regarding the task of removing the stylelint rule exception
* Create a new Github issue to perform the clean up exercise to remove the stylelint rule exception in our config or to add a justification beside of each rule explaining why we do have such exception on.

duboisp (changes requested):
I tested and work as expected, but I do have an administrative/maintenance concern which once the following action/change are done we will be good to merge.

* [x] Add an explicit comment in our config regarding the task of removing the stylelint rule exception + a reference to the match github issue number
* [x] Create a new Github issue to perform the clean up exercise required to remove the stylelint rule exception in our config.

.stylelintrc.yml
EricDunsworth: The immediate need for this PR is to open a path forward for #9529. It fixes a 3 month old issue reported in #9503 and impacts all Twitter timeline widgets.

Csslint fails on it and there's nothing I can do to get around it without either scrapping or replacing it. Other options like using ignore comments aren't feasible because csslint's ignore syntax is inherently flawed (doesn't unconditionally ignore - it still fails in my scenario). Excluding WET's main compiled CSS file from the ``csslint`` task isn't practical either since not linting that file would negate the task's purpose.

---

Regarding exemptions, I agree that there should be plenty of room to update WET's SCSS/CSS to better align with ``stylelint-config-standard``'s ruleset and enable more of ``stylelint-a11y``'s rules. But I don't think enough pieces of the puzzle are in place yet to start tackling those.

As mentioned in https://github.com/wet-boew/wet-boew/pull/9539#issuecomment-1454119035, I expect that this'll need to be tackled in phases across a few PRs. IMO doing it all at once would be very messy and time-consuming.

IMO it also probably won't be realistic to fully follow ``stylelint-config-standard``'s ruleset nor activate all of ``stylelint-a11y``'s rules. Following some of the former's rules would guarantee breaking IE11 (which I'm personally on board with - but I don't think PP is yet) and shift indentation styles to spaces instead of tabs. Some of the latter's rules are opinionated and IMO non-starters for WET.

duboisp: I am not saying to hold on, but to open a new issue to ensure we don't forget about it with a comments in that new config file to add emphasis that we are working on it and a reference (the new issue number) where we can track progress and discussion.

@EricDunsworth Do you want me to create that new issue or you will do?

.stylelintrc.yml
```diff
@@ -0,0 +1,65 @@
+extends:
+  - stylelint-config-standard
+
+plugins:
+  - "@double-great/stylelint-a11y"
+
+rules:
```
duboisp, line 7:
```suggestion
# See #9545 for the cleanup and the re-activation tasks related to the following rules
rules:
```

duboisp (changes requested):

.stylelintrc.yml
duboisp: See my another request of change.

EricDunsworth:
Just rebased to fix the merge conflict, combined into one commit (with a co-author credit added) and force-pushed :).

duboisp (approved):
Tested locally and the changes requested was completed (new request + comment)

PR #9544, opened 2023-03-06
Overlay-Accessibility Report

HeansiH:


src/plugins/overlay/reports/a11y-1.json
```diff
@@ -0,0 +1,902 @@
+{
+    "@context": "https://wet-boew.github.io/vocab/context/2022/acr-new-wcag-sc.json-ld",
+
+    "@id": "https://wet-boew.github.io/wet-boew/demos/add-cal/test/a11y-1.json",
```
duboisp, line 4: I think this file (`a11y-1.json`) was added by mistake

src/plugins/overlay/reports/a11y-2.json
```diff
@@ -0,0 +1,764 @@
...
+            "earl:outcome": "earl:fail",
+            "notes": "Title attribute, text within <a> element. Role attribue is missing within <a> element.",
+            "techniques": [
+                "WCAG21Tech:HTML/H91",
+                "WCAG21Tech:general/G94",
+                "WCAG21Tech:general/G95",
+                "WCAG21Tech:general/G82",
+                "WCAG21Tech:failures/F3",
+                "WCAG21Tech:failures/F13",
+                "WCAG21Tech:failures/F20"
```
duboisp, line 40: Note: We have recently changed/clarified the guidance regarding the use of the "techniques" property

Please only list the technique that we observed it was fully applied and the techniques being discussed in the notes, like applicable failure and sufficient technique that should be applied.

duboisp (changes requested):
Todo:

* [x] Remove or modify the file `a11y-1.json` (if removed, please rename a11y-2 for a11y-1)
* [x] Add the EN and FR HTML page template to visually render the a11y-2.json file. For example : 
   * https://github.com/wet-boew/wet-boew/blob/master/src/plugins/add-cal/reports/a11y-1-en.hbs
* [x] Add link to the accessibility report in the overlay documentation (EN and FR)
   *  Create a new section for that purpose here: https://github.com/wet-boew/wet-boew/blob/master/site/pages/docs/ref/overlay/overlay-en.hbs#L46

duboisp (dismissed):
You will be able to create the HTML file which will be used to display the report.

See the following file as an example (minimum the in page alert) where you only need to update the `reportURL `
* https://github.com/wet-boew/wet-boew/blob/master/src/other/wamethod/demo-acr/a11y-1-en.hbs
* https://github.com/wet-boew/wet-boew/blob/master/src/other/wamethod/demo-acr/a11y-1-fr.hbs

src/plugins/overlay/reports/a11y-1.json
duboisp: Fixed

GormFrank (approved):
@duboisp Overall looks good. 

The only odd looking section is the "Assessment audits" when there is none to show (just like for this ACR):

Screen Shot 2023-06-05 at 2 21 56 PM

duboisp:
The audit of the assessment still need to be performed. I only added the living ACR as a placeholder.

I am working on showing the attachment that could be contained in those JSON report, once the data-json update required for that change is included in wet-boew, we will have a way to manage that kind of situation and display a clear message to the users.

PR #9546, opened 2023-03-07
Filter: uiTemplate option, mutationObserver

Garneauma:
- Adding "uiTemplate" option
  - Define the HTML generated by the plugin
  - Must contain an `<input type="search">`
  - Info string can be modified by adding an element with the class `.wb-fltr-info`
- Adding mutation observer on info element to be able to work with the Tag-filter plugin
- Changing `aria-live="polite"` to `role="status"`


duboisp (changes requested):
Is the interface in the template element are going to be re-used during the plugin instance lifetime? If not, I suggest to look for an alternative where static binding on element could occurs instead like with a flag on the element to be update.

Note that the [`<output>` element](https://html.spec.whatwg.org/multipage/form-elements.html#the-output-element) could be interesting to provide the status update. See [its aria equivalent](https://w3c.github.io/html-aria/#el-output) role is `role=status`.

src/plugins/filter/filter.js
```diff
@@ -98,11 +123,20 @@ var componentName = "wb-filter",
 				$elm.before( filterUI );
 			}
 
+			itemsObserver = new MutationObserver( function() {
+				let itemsLength = $elm.find( secSelector + settings.selector ).length,
+					itemsVisible = $elm.find( secSelector + notFilterClassSel + settings.selector + visibleSelector ).length;
+
+				$( "#" + $elm.get( 0 ).id + "-info" ).html( infoFormater( itemsVisible, itemsLength, elm.infoText ) );
```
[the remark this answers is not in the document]
Garneauma: From my testing locally, the mutation event is only called once, not once per item.

src/plugins/filter/filter-en.hbs
```diff
@@ -823,3 +824,52 @@
...
+<ul class="wb-filter" data-wb-filter='{
+	"uiTemplate": "#filterTemplate"
+}'>
+	<template id="filterTemplate">
+		<p class="wb-fltr-info">_NBITEM_ / _TOTAL_ provinces</p>
+		<div class="form-group">
+			<label for="customSearchUI">Search</label>
+			<input type="search" class="form-control" placeholder="Type a province name..." id="customSearchUI">
+		</div>
+	</template>
```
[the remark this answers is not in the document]
Garneauma: Changed it to use a simple `div`, or whatever element the author decide to use. The option will point to a CSS selector.

Garneauma:
> Note that the [`<output>` element](https://html.spec.whatwg.org/multipage/form-elements.html#the-output-element) could be interesting to provide the status update. See [its aria equivalent](https://w3c.github.io/html-aria/#el-output) role is `role=status`.

After discussion, it has been decided to not use the `<output>` element as it would not be appropriate for this use as per MDN's documentation: "The <output> element is a container element into which a site or app can inject the results of a calculation or the outcome of a user action."

src/plugins/filter/filter-fr.hbs
```diff
@@ -794,3 +795,52 @@
...
+<p>L'élément doit, au minimum, contenir un <code>&lt;input type="search"></code>.</p>
+
+<p>Pour afficher le nombre d'items visibles et le nombre total d'items, ajoutez un élément avec la classe <code>.wb-fltr-info</code>. Le plugiciel remplacera automatiquement "_NBITEM_" par le nombre d'éléments visibles et "_TOTAL_" par le total des éléments.</p>
+
+<div id="filterTemplate">
+	<div class="form-group">
+		<label for="customSearchUI">Rechercher</label>
+		<input type="search" class="form-control" placeholder="Tapez le nom d'une province..." id="customSearchUI">
+	</div>
+	<p class="wb-fltr-info">_NBITEM_ / _TOTAL_ provinces</p>
```
duboisp, line 812: The baseline requirement need to be document for when a custom search interface is used. Otherwise the author will only know by reverse engineering our working example and javascript.

src/plugins/filter/filter.js
```diff
@@ -82,13 +82,40 @@ var componentName = "wb-filter",
...
-				"<p aria-live=\"polite\" id=\"" + elm.id + "-info\">" + infoFormater( totalEntries, totalEntries ) + "</p>" );
+					if ( uiInfo ) {
+						elm.infoText = uiInfo.textContent;
+						uiInfo.id = uiInfo.id || elm.id + "-info";
+						uiInfo.setAttribute( "role", "status" );
+						uiInfo.textContent = infoFormater( totalEntries, totalEntries, elm.infoText );
+					}
+
+					filterUI = uiTemplate.innerHTML;
+					uiTemplate.remove();
```
duboisp, line 107: Can we simple leave the custom filter UI where it is? That way it will allow to have the filter in a side panel, like along with some tag filter controls.

src/plugins/filter/filter.js
```diff
@@ -98,11 +125,24 @@ var componentName = "wb-filter",
 				$elm.before( filterUI );
 			}
 
+			itemsObserver = new MutationObserver( function() {
+				let itemsLength = $elm.find( secSelector + settings.selector ).length,
+					itemsVisible = $elm.find( secSelector + notFilterClassSel + settings.selector + visibleSelector ).length,
+					infoElm = $( "#" + $elm.get( 0 ).id + "-info" );
+
+				if ( infoElm ) {
+					infoElm.html( infoFormater( itemsVisible, itemsLength, elm.infoText ) );
```
duboisp, line 134: May be it will be better and safer to use `textContent` instead of the `html` function

duboisp (changes requested):
The following need to be documented
* [x] A section about the use of custom search interface
* [x] A code sample
* [x] Mention the input must be a `[type=search]`
* [x] Mention about the `wb-fltr-info` flag along with it's two required contained variable.
* [x] Mention about the possible error being throw

Todo:
* [x] Don't move the filter ui when provided by the author. Keep it in place instead.
* [x] Use a safer method to update the filter status info.

---
A functional test and a second code review must be performed

src/plugins/filter/filter.js
duboisp: May be we can adapt to allow the use of an `<span>` element inside to scope where the text content update are going to happen.

May be it will be easier to use a data attribute or a CSS flag to identify the field to be updated instead of replacing a placeholder in the text content.

Like:

```html
<p class="wb-fltr-info"><span data-nbitem>_NBITEM_</span> / <span data-total>_TOTAL_</span> provinces</p>
```

duboisp:
Pre-approve upon second review and local testing.

Note: This PR need to be merged before TagFilter provisional plugin.

duboisp (approved):
I reviewed and locally tested. It pass and the following can be applied later this week.

Can be done in a subsequent PR considering the following change are about content change and not about our versioning API for this plugin.
* [ ] in the documentation - Change the data type for the config `uiTemplate`. it's not a jQuery but a CSS selector instead.
* [ ] Remove the use of "placeholder" attribute in the working example as it don't pass the color contrast.
* [ ] Apply the change to the csv into our master google sheet.

PR #9566, opened 2023-04-11
Datalist polyfill: Future-proof dynamic demo's JS code sample

EricDunsworth:
The dynamic demo pages use a custom JavaScript plugin. A code sample of the plugin also exists in them. But it was hardcoded and hadn't been kept in sync with updates to the plugin itself.

This updates the code sample and future-proofs it by:
* Setting up the custom JS plugin as an include (i.e. handlebars partial)
* Adding two custom handlebars helpers:
  * ``escape``: Encodes special characters (like HTML tags) for code samples
  * ``stripbanner``: Removes the banner (i.e. first multi-line comment) from CSS/JavaScript code
* Demo page: Automating the JS code sample via the include and new helpers

Spinoff of #9388.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -479,7 +479,10 @@ module.exports = (grunt) ->
 				production: false
 				data: "site/data/**/*.{yml,json}"
 				helpers: "site/helpers/helper{,s}-*.js"
-				partials: "site/includes/**/*.hbs"
+				partials: [
+					"site/includes/**/*.hbs"
+					"src/polyfills/datalist/demo/datalist_dynamic.js"
+				]
 				layoutdir: "site/layouts"
 				layout: "default.hbs"
 				environment:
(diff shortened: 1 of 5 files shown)


EricDunsworth:
Btw I initially tried modifying the build system to simply copy the JS file's contents into the demo page's code sample... but was unable to get it working properly :(.

IIRC at one point I had managed to get the code sample "filled-into" the demo pages, but couldn't get handlebars/assemble to turn escape anything nor turn special characters (like HTML tags) into entities. Plus the JS file's WET banner added a lot of bloat.

EricDunsworth:
I managed to do a U-turn... it's now fully automated :). Like what I wanted to do in the first place.

duboisp (approved):
Reviewed and tested locally,

Work as expected and I confirm this is only a content change

Issue #9271, opened 2022-02-07
Report and evaluation tool - Conformity with WCAG 2.1/2.2 Level AA


duboisp:
Goal: Report on accessibility by using a machine readable format.

JSON-LD sample (only for WCAG 2.1/2.2 Level AA Success Criterion) :
```json
{
	"@context":{
		
		"dcterms":"http://purl.org/dc/terms/",
		"rdfs":"http://www.w3.org/2000/01/rdf-schema#",
		"foaf":"http://xmlns.com/foaf/spec/#",
		"gh": "http://github.com/",
		"earl":"http://www.w3.org/ns/earl#",
		"WCAG21":"http://www.w3.org/TR/WCAG21/#",
		"WCAG22":"http://www.w3.org/TR/WCAG22/#",
		
		"WCAG21:orientation": { "@type": "earl:TestRequirement" },
		"WCAG21:identify-input-purpose": { "@type": "earl:TestRequirement" },
		"WCAG21:reflow": { "@type": "earl:TestRequirement" },
		"WCAG21:non-text-contrast": { "@type": "earl:TestRequirement" },
		"WCAG21:text-spacing": { "@type": "earl:TestRequirement" },
		"WCAG21:content-on-hover-or-focus": { "@type": "earl:TestRequirement" },
		"WCAG21:character-key-shortcuts": { "@type": "earl:TestRequirement" },
		"WCAG21:pointer-gestures": { "@type": "earl:TestRequirement" },
		"WCAG21:pointer-cancellation": { "@type": "earl:TestRequirement" },
		"WCAG21:label-in-name": { "@type": "earl:TestRequirement" },
		"WCAG21:motion-actuation": { "@type": "earl:TestRequirement" },
		"WCAG21:status-messages": { "@type": "earl:TestRequirement" },
		"WCAG22:accessible-authentication": { "@type": "earl:TestRequirement" },
		"WCAG22:dragging-movements": { "@type": "earl:TestRequirement" },
		"WCAG22:consistent-help": { "@type": "earl:TestRequirement" },
		"WCAG22:page-break-navigation": { "@type": "earl:TestRequirement" },
		"WCAG22:focus-appearance-minimum": { "@type": "earl:TestRequirement" },
		"WCAG22:visible-controls": { "@type": "earl:TestRequirement" },
		"WCAG22:target-size-minimum": { "@type": "earl:TestRequirement" },
		"WCAG22:redundant-entry": { "@type": "earl:TestRequirement" },

[...]

PR #9579, opened 2023-05-09
Data-JSON - Template - Keep default template text + Core - Log element on error

duboisp:
Data-JSON - Template - Keep default template text when JSON node don't exist
Core - Logging the elm to ease debugging of plugin config

This PR is required for the revised accessibility conformance report methodology as discussed in #9271


Garneauma (approved):
Works well! Nothing to report. Approved.

PR #9582, opened 2023-05-09
dl horizontal - Fix characters hidding when a word overflow

duboisp:
Fix an accessibility issue which use non text content to hide the overflow without providing a text alternative. It's only happen when the word in the description title `<dt>` are larger than the available width.

cc @nfrenette

### DIFF
--- a/src/base/bootstrap-overrides/_core.scss
+++ b/src/base/bootstrap-overrides/_core.scss
@@ -48,9 +48,11 @@ a {
 
 		dt {
 			border-top: 1px solid $clrMedium;
+			hyphens: auto;
 			padding: 10px 10px 10px 0;
 			text-align: left;
 			white-space: normal;
+			word-break: break-word;
 		}
 
 		dd {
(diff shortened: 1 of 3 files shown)


duboisp:
@polmih for your review this week

src/base/bootstrap-overrides/_core.scss
```diff
@@ -51,6 +51,7 @@ a {
 			padding: 10px 10px 10px 0;
 			text-align: left;
 			white-space: normal;
+			word-break: break-word;
```
polmih, line 54: word-break: break-word;
hyphens: auto;

polmih (changes requested):

duboisp:
pre-approved after the proposed update is applied and reviewed.

src/base/bootstrap-overrides/_core.scss
duboisp: fixed

polmih (approved):

PR #9586, opened 2023-05-11
Plugin Charts: If reverseTblParsing, use column header not row header.

### LINKED ISSUE #9583: Pie chart -
When using attribute "reversettblparsing": true and the pie Chart, there is a bug where the "subtitle" is the first category instead of the data column header. In the image below, you can see that the subtitle above the chart is "General", and it should be "Task". 
https://wet-boew.github.io/v4.0-ci/demos/charts/testing-en.html

image

wewhite:
Incorrect row header shows in figure caption, if using reverse table parsing, use the column header for the figure caption. Fixes #9583.

There is an error with the pie chart figure caption text when reverseTblParsing is true.
The figure caption text should be taken from the column, not the row.
This fixes this issue as demonstrated with the Charts and graphs - Specific test cases example page.

Screenshots of problem and fix located in #9583 .

### DIFF
--- a/src/plugins/charts/charts.js
+++ b/src/plugins/charts/charts.js
@@ -894,7 +894,7 @@ var componentName = "wb-charts",
 
 				} else {
 
-					header = currentRowGroup.row[ rIndex ].header;
+					header = !reverseTblParsing ? dataCell.row.header : dataCell.col.header;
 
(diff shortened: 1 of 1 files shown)


polmih (approved):
Tested on local instance, no issues found

polmih (approved):

PR #9587, opened 2023-05-11
Plugin Tables: Handle sorting/ordering only on thead.

### LINKED ISSUE #9578: Row header in datatables getting sorting classes in v4.0.45+
There seems to be an issue in v4.0.45 and up of WET.  I have a datatable with a scope="row" in one of the cells and WET is adding the column header formatting with the button and sorting classes.
```
<table id="employees" class="wb-tables table" data-wb-tables='{ "ordering" : true, "paging": true, "searching": true, "info": false, "aaSorting": [[1, "asc"]]}'>
    <thead>
        <tr>
            <th class="hide" scope="col">Id</th>
            <th scope="col">@Localization.EmployeeFirstName</th>
            <th scope="col">@Localization.EmployeeLastName</th>
            <th scope="col">@Localization.EmployeeJobDescription</th>
             </tr>
    </thead>
    <tbody>
        @foreach (var employee in Model.Employees)
        {
            <tr>
                <td class="hide">@employee.Id</td>
                <th scope="row">@employee.FirstName</th>
                <td>@employee.LastName</td>
                <td>@employee.JobDescription</td>
               </tr>
    }
    </tbody>
</table>
```
Using CDTS template and v4_0_44 it renders like this:
image

If I change to 45 or up, it renders like this:
image

and the html looks like this:
image

wewhite:
If using th elements in tr, sorting/ording is applied, and should only be applied to the thead th. Fixes #9578

This prevents  tr th data cells from being rendered as sorting/ordering buttons.
This allows only table header thead th columns to be rendered as sorting/ordering buttons.

See #9578 for screen shots.

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -223,7 +223,7 @@ $document.on( "init.dt", function( event ) {
 	// Handle sorting/ordering
 	var ordering = ( settings && settings.ordering === false ) ? false : true;
 	if ( ordering ) {
-		$elm.find( "th" ).each( function() {
+		$elm.find( "thead th" ).each( function() {
 			var $th = $( this ),
 				label = ( $th.attr( "aria-sort" ) === "ascending" ) ? i18nText.aria.sortDescending : i18nText.aria.sortAscending;
 			if ( $th.attr( "data-orderable" ) !== "false" ) {


wewhite:
Fixes https://github.com/wet-boew/wet-boew/issues/9268

Garneauma (approved):

PR #9476, opened 2022-11-16
Table: Don't create column sort button when sort is disabled

radmorecameron:
## What does this pull request (PR) do?
DataTables has support for adding `data-orderable="false"` to `th` elements in order to disable sorting on a specific column. When using `data-orderable` with wet, it creates the sort button but sorting is disabled. This PR stops that button from being created.

## Additional information (optional)
This is my first PR for this project, let me know if I should create documentation for this

**General checklist**
Make your own list for the purpose of your Pull request.

- [x] Build and test PR's code
- [x] Validate changes against WCAG for accessibility 

**Related issues / Requêtes associées**
None.

**Screenshots / Captures d'écrans**
Tested with `data-orderable='false'` on the `Engine Version` column.

DataTable where all columns excluding the Engine Version column are sortable

PR #9590, opened 2023-05-11
Plugin Tables: Do not create sort/order button when column sort is di…

### LINKED ISSUE #9588: [BUG] Tables - removing sorting from a column generates aria-controls="undefined"
## Required information

**Describe the bug**
We use a table with 1 column that will not be orderable.  Upon rendering, the header contains a button
button type="button" aria-controls="undefined" title="activate for ascending sort"

The aria-controls generate a W3C validation error:
Error: The aria-controls attribute must point to an element in the same document.

Here is the table definition
```
<table  id="ResultsTable" class="cards wb-tables table draftLayout"
                   data-wb-tables='{"paging":"true", "ordering":"true", "info":"true", "columnDefs": [ {"orderable": false, "targets": [5]}], "aaSorting": [[2, "asc"]], "lengthMenu": [[9, 18, 36, 54, -1], [9, 18, 36, 54, "@Common.AllFilter"]]}'>
```

**Expected behaviour**
No button at all as the sorting is not available

**Who shall do the work?**
I am asking for Principal publisher to please do the work

**Version of WET-BOEW/GCWeb you are using**
v4.0.47

wewhite:
…sabled.

There are multiple ways to disable column sorting. This patch does not allow button to be created if th column class sorting_disabled is present.

Related to #9476
Fixes #9588

Column sorting can be enabled/disabled different ways:
- using th data-orderable attribute
- using columns in data-wb-table settings
- using columnDefs in data-wb-table settings

When using data-wb-table settings:
```
"columns": [
    { "orderable": false },
    null,
    null,
    null,
    null
  ]  }'
```
or 

```
"columnDefs": [ {
"targets": 0,
"orderable": false
} ]
```

The class name `sorting_disabled` is added to the table header, and WET tables still adds the button code to it.
Adding another check in the if statement, does not add the button code if `sorting_disabled` class exists.

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -226,7 +226,7 @@ $document.on( "init.dt", function( event ) {
 		$elm.find( "th" ).each( function() {
 			var $th = $( this ),
 				label = ( $th.attr( "aria-sort" ) === "ascending" ) ? i18nText.aria.sortDescending : i18nText.aria.sortAscending;
-			if ( $th.attr( "data-orderable" ) !== "false" ) {
+			if ( ( $th.attr( "data-orderable" ) !== "false" ) && !( $th.hasClass( "sorting_disabled" ) ) ) {
 				$th.html( "<button type='button' aria-controls='" + $th.attr( "aria-controls" ) +  "' title='" + $th.text().replace( /'/g, "&#39;" ) + label + "'>" + $th.html() + "<span class='sorting-cnt'><span class='sorting-icons' aria-hidden='true'></span></span></button>" );
 				$th.removeAttr( "aria-label tabindex aria-controls" );
 			}


src/plugins/tables/tables.js
```diff
@@ -226,7 +226,7 @@ $document.on( "init.dt", function( event ) {
 		$elm.find( "th" ).each( function() {
 			var $th = $( this ),
 				label = ( $th.attr( "aria-sort" ) === "ascending" ) ? i18nText.aria.sortDescending : i18nText.aria.sortAscending;
-			if ( $th.attr( "data-orderable" ) !== "false" ) {
+			if ( ( $th.attr( "data-orderable" ) !== "false" ) && !( $th.hasClass( "sorting_disabled" ) ) ) {
```
polmih, line 229: Question: what is the purpose of manually adding the class "`sorting_disable`" in the table head? Is it to disable the column sorting? Why we should add it since sorting functionality and the visual button is handled well by the attribute `data-orderable`?
Thank you

wewhite: DataTables adds the class `sorting_disable` automatically, it is not manually added.

Disable column sorting using DataTables API example:
`data-wb-tables='{ "ordering" : false }'>`

There are multiple ways to disable sorting/ordering with DataTables.

If only use the DataTables API config, and not using `data-orderable` in the `<th...`, the buttons get added, with `aria-controls=undefined` and shows the buttons.

polmih (approved):
locally tested. No issues found

PR #9593, opened 2023-05-17
wb-paginate: Adding new provisional plugin

Garneauma:
Adding new pagination plugin and support for this new plugin in Filter and Tagfilter plugins.

### DIFF
--- a/src/plugins/paginate/_base.scss
+++ b/src/plugins/paginate/_base.scss
@@ -1,0 +1,31 @@
+.wb-paginate-pager {
+	text-align: center;
+
+	.paginate-prev::before,
+	.paginate-next::after {
+		font-family: "Glyphicons Halflings";
+		font-weight: 400;
+		line-height: 1em;
+		position: relative;
+		top: .1em;
+	}
+
+	.paginate-prev::before {
+		content: "\e091";
+		margin-right: .5em;
+	}
+
+	.paginate-next::after {
+		content: "\e092";
+		margin-left: .5em;
+	}
+
+	li.active:nth-last-child(2) a {
+		border-bottom-right-radius: 4px;
+		border-top-right-radius: 4px;
+	}
+}
+
+html:not(.wb-disable) .wb-pgfltr-out {
+	display: none !important;
+}
--- a/src/plugins/paginate/paginate-en.hbs
+++ b/src/plugins/paginate/paginate-en.hbs
@@ -1,0 +1,1191 @@
+---
+{
+	"title": "Paginate",
+	"language": "en",
+	"category": "Plugins",
+	"description": "Adds pagination at the bottom of a list of items",
+	"tag": "paginate",
+	"parentdir": "paginate",
+	"altLangPrefix": "paginate",
+	"dateModified": "2023-05-15"
+}
+---
+
+<div class="alert alert-warning">
+	<h2>Unstable feature</h2>
+	<p>To be used at <strong>your own risk</strong>. This feature described below can be removed in any subsequent
+		minor/major release</p>
+	<p><small><a href="https://wet-boew.github.io/wet-boew-documentation/decision/7.html">Learn more about the design
+				decision around provisional features.</a></small></p>
+	<p><small><a href="../../docs/ref/provisional-en.html">Check other provisional features available.</a></small>
+	</p>
+</div>
+
+<span class="wb-prettify all-pre"></span>
+
+<h2>On this page:</h2>
+<ul>
+	<li><a href="#paginatedList">Example 1 - List</a></li>
+	<li><a href="#paginatedTable">Example 2 - Table</a></li>
+	<li><a href="#paginatedArticles">Example 3 - Articles</a></li>
+	<li><a href="#filteredEventsList">Example 4 - Events list</a></li>
+</ul>
+
+<section id="paginatedList">
+	<h2>Example 1 - List</h2>
+	<ul class="wb-paginate provisional">
+		<li>Lorem ipsum, dolor sit amet consectetur adipisicing elit. Illum, at!</li>
+		<li>Soluta maiores deserunt fugit dolorum ratione cum necessitatibus dolore ipsum?</li>
+		<li>Libero debitis dicta porro modi nobis odit labore dolor eveniet.</li>
+		<li>Corrupti est quos ad vero. Hic repellendus numquam officia beatae.</li>
+		<li>Amet ullam, ipsam adipisci ad nisi facere veritatis fuga pariatur?</li>
+		<li>Eum ipsum doloribus nisi natus dolorum maxime omnis, mollitia hic!</li>
+		<li>Aperiam odit, sapiente cum modi amet labore iste dolorum minima.</li>
+		<li>Sunt praesentium voluptatibus aliquid cumque quae eaque aspernatur incidunt sint?</li>
+		<li>Quae illo veritatis deserunt optio corporis repellendus! Qui, deleniti suscipit?</li>
+		<li>Expedita quae temporibus ex labore ipsa excepturi inventore deleniti magnam!</li>
+		<li>Facere reprehenderit natus voluptatibus perspiciatis. Ipsam id quam facilis maxime?</li>
+		<li>Provident assumenda fugiat, nulla temporibus accusantium incidunt laudantium ea consequuntur!</li>
+		<li>Nulla, non aspernatur ipsam provident libero ratione corrupti necessitatibus dicta.</li>
+		<li>Possimus minus autem voluptate suscipit porro non culpa officiis doloribus?</li>
+		<li>Alias modi eius repudiandae natus molestiae ad, unde aliquid praesentium.</li>
+		<li>Ut incidunt corporis quos vel accusantium non assumenda dolorum vitae?</li>
+		<li>Velit magnam laudantium expedita recusandae similique voluptates est, doloribus odit!</li>
+		<li>Officia quasi non, ex dolore fuga reiciendis unde alias blanditiis!</li>
+		<li>Repellendus, libero sint at nam tempore blanditiis recusandae ipsa cupiditate?</li>
+		<li>Maiores exercitationem ipsa fuga accusamus impedit dolorum cum voluptatum alias.</li>
+		<li>Nulla blanditiis ut ipsa nesciunt officiis error quisquam, reprehenderit quam!</li>
+		<li>Libero sint optio sunt beatae ex, quibusdam autem dignissimos et?</li>
+		<li>Mollitia cum tenetur sed blanditiis unde sint tempore quaerat ratione?</li>
+		<li>Optio dicta harum placeat tempore dolorem? Temporibus enim libero vel.</li>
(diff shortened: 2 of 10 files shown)


src/plugins/paginate/paginate-en.hbs
```diff
@@ -0,0 +1,839 @@
...
+	<li><a href="#paginatedList">Example 1 - List</a></li>
+	<li><a href="#paginatedTable">Example 2 - Table</a></li>
+	<li><a href="#paginatedArticles">Example 3 - Articles</a></li>
+	<li><a href="#filteredEventsList">Example 4 - Events list</a></li>
+</ul>
+
+<section id="paginatedList">
+	<h2>Example 1 - List</h2>
+	<ul class="wb-paginate provisional">
+		<li class="wb-pgfltr-out">Lorem ipsum, dolor sit amet consectetur adipisicing elit. Illum, at!</li>
```
duboisp, line 37: Is the initial state is hidden? I suggest to do the reverse that way if there is a JS error in the plugin, the content will still be displayed, as per the progressive enhancement design approach.

src/plugins/paginate/paginate-en.hbs
```diff
@@ -0,0 +1,839 @@
...
+		<li class="wb-pgfltr-out">Accusamus inventore ad explicabo earum quidem qui soluta iusto officia.</li>
+		<li class="wb-pgfltr-out">At rerum doloremque nostrum perspiciatis repellat velit nihil aperiam laborum!</li>
+	</ul>
+</section>
+
+<section id="paginatedTable">
+	<h2>Example 2 - Table</h2>
+	<p>This example leverages the following option: <code>startPage</code></p>
+	<table class="table wb-paginate provisional" data-wb-paginate='{
+		"startPage": 3
```
duboisp, line 126: I will need more info about this use case of when we need to configure statically the start page.

I do have the impression is for when, on page load, an element in deeper page are receiving the focus. In that situation we can probably use the `document.activeElement` to figure out what page to auto-navigate toward.

I can also see it configurable with do-action plugin or the url-mapping plugin for example

src/plugins/paginate/paginate-en.hbs
```diff
@@ -0,0 +1,839 @@
...
+		<article class="col-sm-6 col-md-4 col-lg-3">
+			<div class="well">
+				<h3 class="mrgn-tp-0">Assumenda dignissimos beatae doloremque dolores.</h3>
+				<p>Aliquid temporibus sunt quis illum, voluptatem atque. Blanditiis voluptatum iure cupiditate recusandae hic sint aperiam.</p>
+			</div>
+		</article>
+	</div>
+</section>
+
+<section id="filteredEventsList">
```
duboisp, line 569: In this example the pagination are outside the scope of the section element. Would be preferable to add the pagination plugin via an inner div inside the section?

src/plugins/paginate/paginate-en.hbs
```diff
@@ -0,0 +1,839 @@
...
+		"uiTemplate": "#filteredEventsList",
+		"section": ".events-list",
+		"selector": "article"
+	}'
+	data-wb-paginate='{
+		"itemsPerPage": 5,
+		"section": ".events-list",
+		"selector": "article"
+	}'>
+		<h3 class="wb-fltr-info h3"><span data-nbitem></span> event(s)</h3>
```
duboisp, line 584: Adding the role status overwrite the default role heading level 3 of the HTML element `<h3>`.

Here the heading element seem to be used for styling purpose only and not for semantic.

src/plugins/paginate/_base.scss
```diff
@@ -0,0 +1,31 @@
+.wb-paginate-pager {
```
duboisp, line 1: Is there some pagination style that exist in the data table CSS which will be preferable to be moved here?

src/plugins/paginate/paginate.js
```diff
@@ -0,0 +1,262 @@
...
+			if ( !i18nText ) {
+				i18n = wb.i18n;
+				i18nText = {
+					prv: i18n( "prv" ),
+					nxt: i18n( "nxt" )
+				};
+			}
+
+			// Setup configurations
+			if ( [ "UL" ].indexOf( elmTagName ) >= 0 ) {
```
duboisp, line 55: If that list `[ "UL" ]` limited to one element or the intend is to add a few more?

src/plugins/paginate/paginate.js
```diff
@@ -0,0 +1,262 @@
...
+					},
+					grp: {
+						selector: "> *",
+						appendUI: true
+					},
+					tbl: {
+						selector: "tr",
+						section: ":scope > tbody"
+					}
+				};
```
duboisp, line 44: The default should be declared at the top, near line 12 to 21.

duboisp (changes requested):
I didn't completed my review.

Overall, that look to be an amazing plugin. I love it.

Just noted a few items for discussion.

src/plugins/paginate/paginate-en.hbs
Garneauma: Removed "startPage" option. The feature to navigate to active element's page on init will be added at a later date.

src/plugins/paginate/paginate-en.hbs
```diff
@@ -0,0 +1,1060 @@
...
+		"section": ".events-list",
+		"selector": "article",
+		"uiTarget": "#paginationCntnr"
+	}'>
+		<p class="h3 wb-fltr-info"><span data-nbitem></span> event(s)</p>
+		<div class="row">
+			<div class="col-sm-6 col-md-5 col-lg-4">
+				<details open>
+					<summary>
+						<h4>Filter events</h4>
```
duboisp, line 669: I suggest to remove the heading inside the summary because it is not well supported by UA and AT. The summary element do have an implicit button role and that invalidate the heading level semantic in the accessibility tree.

ref. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary

src/plugins/paginate/paginate-en.hbs
```diff
@@ -0,0 +1,1060 @@
...
+		"section": ".events-list",
+		"selector": "article"
+	}'
+	data-wb-paginate='{
+		"itemsPerPage": 5,
+		"section": ".events-list",
+		"selector": "article",
+		"uiTarget": "#paginationCntnr"
+	}'>
+		<p class="h3 wb-fltr-info"><span data-nbitem></span> event(s)</p>
```
duboisp, line 664: This style of the status can create some confusion on the info & relationship. Can we add something that will make it different from the perceptible missing heading level 3 looks?

src/plugins/paginate/paginate.js
```diff
@@ -0,0 +1,267 @@
...
+				break;
+			default:
+				setDefault = defaults.grp;
+				break;
+			}
+
+			elm.id = elm.id || wb.getId();
+			elm.pgSettings = { ...setDefault, ...userConf };
+			elm.pgSettings.currPage = 1;
+			elm.pgSettings.itemsPerPage = elm.pgSettings.itemsPerPage || 25;
```
duboisp, line 69: I suggest to move the default "25" at the top of the plugin js code. 

Comment: Would 7, 10 or 12 be a better choice compared to 25? That way it will avoid a content scrolling when a page is selected. The data table do have a default of 10 items. I know it is a minimal scrolling in wet-boew but it's going to be more prominent in GCWeb because the font are different.

src/plugins/paginate/paginate.js
```diff
@@ -0,0 +1,267 @@
...
+			case "TABLE":
+				setDefault = defaults.tbl;
+				break;
+			default:
+				setDefault = defaults.grp;
+				break;
+			}
+
+			elm.id = elm.id || wb.getId();
+			elm.pgSettings = { ...setDefault, ...userConf };
```
duboisp, line 67: todo: Check if the ``...` instruction would cause a parsing error in old browser like IE11?

src/plugins/paginate/_base.scss
duboisp: If both are render the same, then leave it as is for now.

src/plugins/paginate/paginate.js
```diff
@@ -0,0 +1,267 @@
...
+
+			wb.ready( $( elm ), componentName );
+		}
+	},
+
+	// Set or reset pagination UI
+	generateUI = function( elm ) {
+		var paginationUI = "",
+			currPage = elm.pgSettings.currPage,
+			pagesCount = elm.pgSettings.pagesCount,
```
duboisp, line 103: Comment

There is no demo for the configuration:
* currPage
* pagesCount

But I figured out it is a dynamic data managed by the plugin. I like the idea and it allow other plugin to gather the current state of the pagination plugin.

src/plugins/paginate/paginate.js
```diff
@@ -0,0 +1,267 @@
...
+			selector: "tr",
+			section: ":scope > tbody"
+		}
+	},
+
+	init = function( event ) {
+		const elm = wb.init( event, componentName, selector );
+
+		if ( elm ) {
+			var userConf = JSON.parse( elm.getAttribute( "data-wb-paginate" ) ),
```
duboisp, line 39: Can you use the wet-boew function to get the plugin setting? 


```suggestion
			var userConf = wb.getData( $( elm ), componentName ),
```

duboisp (changes requested):
I tested and reviewed and it seems functional and the plugin look like to be compliant to accessibility from my quick review. The plugin don't seem to introduce or create any accessibility issue. Although I added a note about the use of some style and a missing header in the working example which is an accessibility issue.

Eventually a formal WCAG assessment will be required and I do suggest to run a test with an AT software to ensure if the ARIA usage are producing the expected end-user experience.

Once all comments are addressed and after a third quick round of testing, I think this plugin will be ready to be merged.

src/plugins/paginate/paginate.js
Garneauma: At this point, it is very subjective, but I'll set the default as 10 and add it to the defaults object.

src/plugins/paginate/paginate.js
```diff
@@ -0,0 +1,267 @@
...
+				setDefault = defaults.tbl;
+				break;
+			default:
+				setDefault = defaults.grp;
+				break;
+			}
+
+			elm.id = elm.id || wb.getId();
+			elm.pgSettings = { ...setDefault, ...userConf };
+			elm.pgSettings.currPage = 1;
```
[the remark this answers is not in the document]
Garneauma: Yes, the "startPage" feature is planned to be implemented in the future.

src/plugins/paginate/paginate.js
Garneauma: Removed.

Garneauma:
Ready to be re-reviewed @duboisp

duboisp:
Pre-approved upon successful testing + review

duboisp (approved):
Reviewed and tested locally, work great and its ready to be merged

PR #9595, opened 2023-05-23
Tagfilter: Adding no result message feature

Garneauma:
Adding the possibility to define a "No result" message element that will show when there are no items visible.

### DIFF
--- a/src/plugins/tagfilter/tagfilter-fr.hbs
+++ b/src/plugins/tagfilter/tagfilter-fr.hbs
@@ -125,6 +125,9 @@
 					</li>
 				</template>
 			</ul>
+			<div class="wb-tagfilter-noresult">
+				<p>Aucun item ne correspond à cette combinaison de filtres.</p>
+			</div>
 		</div>
 	</div>
 </section>
@@ -228,6 +231,9 @@
 					&lt;/li>
 				&lt;/template>
 			&lt;/ul>
+			&lt;div class="wb-tagfilter-noresult">
+				&lt;p>Aucun item ne correspond à cette combinaison de filtres.&lt;/p>
+			&lt;/div>
 		&lt;/div>
 	&lt;/div>
 &lt;/section></code></pre>
@@ -572,6 +578,9 @@
 			</div>
 		</div>
 	</div>
+	<div class="wb-tagfilter-noresult">
+		<p>Aucune application.</p>
+	</div>
 </section>
 
 <details>
@@ -741,6 +750,9 @@
 		&lt;/div&gt;
 		...
 	&lt;/div&gt;
+	&lt;div class="wb-tagfilter-noresult">
+		&lt;p>Aucune application.&lt;/p>
+	&lt;/div>
 &lt;/section&gt;</code></pre>
 </details>
 
@@ -785,6 +797,7 @@
 			</div>
 		</div>
 	</div>
+
 	<div class="wb-tagfilter-items">
 		<article data-wb-tags="subject_mwa subject_hwm">
 			<h3>Les employés du MDN dont le régime de travail ne sera pas modifié par la mise en œuvre du nouveau modèle de travail hybride devront-ils tout de même soumettre une entente à l'aide de l'application mesRégimesDeTravail (mRT)?</h3>
@@ -858,6 +871,9 @@
 			<p>D'autres directives seront fournies au cours des prochaines semaines afin que tout le monde puisse travailler efficacement dans ce milieu de travail hybride officiel.</p>
 		</article>
 	</div>
+	<div class="wb-tagfilter-noresult">
+		<p>Aucun résultat.</p>
+	</div>
 </section>
 
 <details>
@@ -903,6 +919,7 @@
 			&lt;/div>
 		&lt;/div>
 	&lt;/div>
+
 	&lt;div class="wb-tagfilter-items">
 		&lt;article data-wb-tags="subject_mwa subject_hwm">
 			&lt;h3>Les employés du MDN dont le régime de travail ne sera pas modifié par la mise en œuvre du nouveau modèle de travail hybride devront-ils tout de même soumettre une entente à l'aide de l'application mesRégimesDeTravail (mRT)?&lt;/h3>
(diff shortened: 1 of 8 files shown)


duboisp (changes requested):
(quick partial review)

* [x] Rename `sr-only` by `wb-inv`

src/plugins/tagfilter/tagfilter.js
```diff
@@ -52,6 +54,11 @@ const componentName = "wb-tagfilter",
 				console.warn( componentName + ": You have no tagged items. Please add tags using the 'data-wb-tags' attribute." );
 			}
 
+			// Add accessibility to no result element
+			if ( noResultWrapper ) {
+				noResultWrapper.setAttribute( "aria-live", "polite" );
```
duboisp, line 59: Not sure if the only use of the `[aria-live=polite]` is enough to satisfy at [SC 4.1.3 Status messages](https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html).

I do think the "Situation A" is applicable here because the status message advises on the results of an action.

According to the sufficient technique, it might require to use the role "status".

Also, to be tested with 1 AT, is the show/hide of the content block do get announced to AT user when:
1 - There is a few result
2 - There is no result (the block is displayed for the first time)
3 - There is a few result
4 - There is no result ( the block is re-displayed for a second time)

What about then the component do have 2 status content block? like the "Example 3". I do think the content block representing the status "no results" should be marked with the role "status" also.

duboisp (changes requested):

src/plugins/tagfilter/tagfilter.js
Garneauma: When I tested using the Mac screen-reader, I correctly got the message for no results every time (hiding, then showing, then hiding, etc.). However, I can add the role status to make it more obvious for the screen reader, and test.

duboisp: ok it is compatible with AT on mac. Have you compared against the SC 4.1.3 statement? Is it satisfied and applicable technique applied?

Garneauma: Switched `aria-live="polite"` to `role="status"` which satisfies success criterion 4.1.3. However, while fixing this, I have discovered an issue with the current implementation which does not helpfully declare de changes in the DOM when a filter is selected. This issue is going to be fixed in an upcoming PR.

duboisp:
Pre-approved upon second review and testing.

duboisp (changes requested):
Le "no result" message n'est pas fonctionnel. Je pense qu'il manque ce bloc de code avec le mutation observer:

https://github.com/wet-boew/wet-boew/commit/e061bdcce03adbb4616657943bce94301a918969#diff-40069d6c91d0286504cff47f4416f809a249c4f1b151a99974e78fc682c34880R61-R82

src/plugins/tagfilter/tagfilter.js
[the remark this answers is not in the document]
Garneauma: @duboisp From our discussion, the feature does not work on Firefox due to the use of the CSS selector `:has()`. A fallback solution will need to be implemented for Firefox.

duboisp:
Pre-approved upon successful second review and local testing.

src/plugins/tagfilter/tagfilter-fr.hbs
duboisp, line 1: The no result don't appear in the French example. But it do work for the English example.

src/plugins/tagfilter/tagfilter.js
```diff
@@ -214,6 +222,21 @@ const componentName = "wb-tagfilter",
 		refineFilters( instance );
 		matchItemsToFilters( instance );
 		updateDOMItems( instance );
+
+		// Show no result message if on Firefox -- Remove once Firefox supports ":has()"
+		if ( navigator.userAgent.toLowerCase().indexOf( "firefox" ) > -1 ) {
```
duboisp, line 227: It will be preferable to do a [feature detection](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection) instead of being browser specific. Being being browser specific could let to unexpected behaviour when the browser would start to support the feature, especially when we know that some site take a lot of time before to update their WET-BOEW implementation to the latest version.

Can you check if it would be possible to use a polyfill?
I didn't tested but like: https://github.com/jplhomer/polyfill-css-has/blob/master/src/polyfill-css-has.ts
Can we leverage somehow the [@supports CSS at-rule](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports) for this? Like if supported we style a little something that we can detect in JS to use an alternate implementation?

duboisp (changes requested):
I did a review and local testing, here my observation:

* [x] French example don't work
* [x] Example 3 in English don't work in Firefox
* [x] Use feature detection instead of being browser specific

duboisp:
Pre-approved once the change is pushed + review and local testing completed.

src/plugins/tagfilter/tagfilter.js
```diff
@@ -261,6 +271,25 @@ $document.on( "change", selectorCtrl, function( event )  {
 	$elm.trigger( "wb-contentupdated" );
 } );
 
+// Show no result message if on Firefox -- Remove once Firefox supports ":has()"
+$document.on( "wb-contentupdated", selector, function()  {
+	let supportsHas = getComputedStyle( document.documentElement ).getPropertyValue( "--supports-has" );
```
duboisp, line 276: Considering the closure scope currently defined for this plugin, this call should be made from the `window` variable instead of calling the global directly.

```suggestion
	let supportsHas = window.getComputedStyle( document.documentElement ).getPropertyValue( "--supports-has" );
```

duboisp (changes requested):
Awesome way to detect the `:has` CSS selector, I love it. 

I just have a little typographic change which are normal caught, I thought,  by the "use strict" instruction or our eslint task but I don't know why it didn't.

Todo:
* [x] Call `getComputedStyle` via the `window` scoped variable
* [x] Fix the merge conflict

Note: Once this PR is merged, you will need to replicate in GCWeb and adapt the change made at the build script. That sync is required in order to release GCWeb along with this changes.

src/plugins/tagfilter/tagfilter-fr.hbs
```diff
@@ -785,6 +797,10 @@
 			</div>
 		</div>
 	</div>
+
+	<div class="wb-tagfilter-noresult">
+		<p>Aucun résultat.</p>
+	</div>
```
duboisp, line 803: I think this was added by mistake because it is already there at the end of the list of articles.

src/plugins/tagfilter/tagfilter.js
```diff
@@ -253,12 +262,10 @@ $document.on( "change", selectorCtrl, function( event )  {
 
 	// Update list of visible items
 	update( elm );
-
-	$elm.trigger( "wb-contentupdated" );
 } );
 
 // Reinitialize tagfilter if content on the page has been updated
-$document.on( "wb-contentupdated", selector, function() {
+$document.on( "wb-contentupdated", selector, function()  {
```
duboisp, line 268: There is an extra space

```suggestion
$document.on( "wb-contentupdated", selector, function() {
```

duboisp (changes requested):

duboisp (approved):
Reviewed and tested locally, work as expected in Chrome and in FF.

PR #9601, opened 2023-05-26
Data-ajax: init ajaxed WET elements if WET is done init

Garneauma:
Adding trigger on content updated to re-run WET for the newly ajaxed-in elements if wet is done initializing beforehand.


duboisp (approved):
Reviewed and tested to see if that was fixing the issue with the server-less GCWeb feedback.

Work as expected.

I suggest to consider this as a patch because for light pages that do use data-ajax will trigger wet for the ajaxed in content.

Next (outside this request):
* [ ] Remove the "trigger" wet functionality + corresponding docs used by the wb-actionmanager and fieldflow

PR #9612, opened 2023-06-16
Filter: fixing visible items bug

Garneauma:
Fixing bug where the count of visible items would return an incorrect number when the "selector" property was not set to a tag, but rather a CSS selector. For example: "> li" instead of "li".

### DIFF
--- a/src/plugins/filter/filter.js
+++ b/src/plugins/filter/filter.js
@@ -132,7 +132,7 @@ var componentName = "wb-filter",
 				uiNbItems.textContent = totalEntries;
 
 				itemsObserver = new MutationObserver( function() {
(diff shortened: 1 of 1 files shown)


duboisp:
The issue is visible on the GCWeb home page.

To reproduce:
* Do a filter until there is no result,
* Remove the filter text
* Expected: The status is:  total of item / total of item (ex: 20 items on 20 items )
* Error seen: The status is: Greater number of item / total of item (ex: 54 items on 20 items)

Pre-approved upon successful review and local testing

duboisp (dismissed):
I tested the fix and work as expected.

duboisp (approved):

PR #9613, opened 2023-06-16
Stretched-link: removing underline on hover for accessibility

Garneauma:


duboisp:
This seem to go beyond the stretched-link style as it act exactly like any other link.

The issue seem to be limited to the ":visited" link style and may be a different fix should be applied.

Further investigation is required and identify the SC that are not satisfied.

Resource that might be helpful:
* WCAG 2.2 - https://www.w3.org/TR/WCAG22/#focus-appearance

GormFrank:
Agreed with @duboisp that the scope of this ticket should be changed to focus on the :visited accessibility flaw overall and not in the context of stretched-link only.

Garneauma:
Closing this pull request as the current implementation does not fail any WCAG guideline.

PR #9615, opened 2023-06-19
Demos: Replace CSS code samples with SCSS includes

EricDunsworth:
The equal heights demo's hardcoded CSS sample was outdated compared to its "real" SCSS source file.

Rather than updating the affected CSS sample, this replaces all demo CSS samples with future-proof SCSS ones. The SCSS samples are setup as includes (handlebars partials) that are sourced from the demos' "real" SCSS files. #9566's custom handlebars helpers are also used to make the SCSS samples more presentable.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -491,6 +491,10 @@ module.exports = (grunt) ->
 				helpers: "site/helpers/helper{,s}-*.js"
 				partials: [
 					"site/includes/**/*.hbs"
+					"src/other/archived/demo/archived.scss"
+					"src/plugins/eqht-css/demo/eqht-css.scss"
+					"src/plugins/equalheight/demo/equalheight.scss"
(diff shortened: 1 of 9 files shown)


EricDunsworth:
FWIW I initially wanted to stick with CSS samples and derive them from the dist folder's unminified CSS files... but ran into snags while testing. Build failures could potentially occur if the demos were compiled before the demo CSS files (like running ``grunt demos`` without a pre-existing dist folder).

duboisp:
This seem to be a content change only. Approved upon review and local testing.

Garneauma (changes requested):
@EricDunsworth Please fix merge conflicts.

EricDunsworth:
@Garneauma Just rebased and updated modified dates 👍.

Garneauma (approved):
Tested locally and everything works as expected. Thank you @EricDunsworth for your contribution!

PR #9616, opened 2023-06-20
Data-ajax: fixing plugins not initializing

Garneauma:
Fixing plugins not being initialized after being loaded with data-ajax in some cases.


duboisp (approved):
tested locally and work as expected.

PR #9618, opened 2023-06-22
Tagilfter: adding reinitialization on content updated

Garneauma:
Adding feature to reinitialize Tagfilter plugin when content on the page has been added through data-json.

Also removing warning messages as it might happen that the items or filters are added through data-json.


duboisp:
We will need a working example to test it.

duboisp:
Pre-approved upon successful review and local testing

duboisp (approved):
Reviewed and tested locally, work as expected.

Issue #8045, opened 2017-06-06
Session Timeout timer being throttled in the background of certain browsers.


CalvinRodo:
One of the implementers of the GoC Web Templates has reported an issue with the session timeout plugin that seems to be related to this: https://blog.chromium.org/2017/03/reducing-power-consumption-for.html

They are finding that the JS Timer is either stopping or slowing down when the tab is in the background. 

I ran a quick test on my cellphone and the session timeout plugin page and it looks like there are two issues here, number one is the modal dialog popup doesn't trigger until the user returns to the page. 

Then the time on that modal dialogue runs at about half the speed (in my test).
 
This could result in a user leaving a page for significant amount of time and then someone getting on the device at a later date  and accessing the first users data. 

Based on the blog post linked above this will be an issue with any Chrome Browser, this could be an issue with GoC Services as the kiosks in Service Canada Centre's use Chromium. 

If possible we might want to base this on the users local time as opposed to a timer to determine the time. 

The issue on GC/Code can be found here: 
http://gitlab.ssc.etg.gc.ca/GOCWebTemplates/DotNetTemplates/issues/76 (Available only on GCIntranet)

PR #9620, opened 2023-07-04
Session-Timeout: Use current time to check if session is expired

radmorecameron:
_(Pick the language of your choice to fill out the following information. / Choisissez la langue de votre choix pour remplir l'information ci-dessous.)_

This PR uses the current date and time to check if a session is expired as timeout and intervals can be throttled when in the background.

- [x] Build and test PR's code /// Rouler le script de compilation et tester le code de la PR

#8045

### DIFF
--- a/src/plugins/session-timeout/session-timeout.js
+++ b/src/plugins/session-timeout/session-timeout.js
@@ -109,19 +109,15 @@ var $modal, $modalLink, countdownInterval, i18n, i18nText,
 	 * @param {Object} settings Key-value object
 	 */
 	initEventTimeout = function( $elm, eventName, time, settings ) {
-		var startTime = getCurrentTime(),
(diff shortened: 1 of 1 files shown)


duboisp:
Are going to be review by PP this week. We may propose some change request because the CI script is failing. 

Note: Some additional change/adjustment might be required for the corresponding test file here: https://github.com/wet-boew/wet-boew/blob/master/src/plugins/session-timeout/test.js

duboisp:
@polmih poke

polmih (changes requested):
These proposed changes should make the plugin's timing mechanisms compatible with the test expectations, and the test suite should pass without errors and still using the current time in order 
 to check if the session is expired

polmih (approved):
code review done, tested locally

PR #9623, opened 2023-07-06
Content - Overlay - Fix typo in a11y-1 JSON report

duboisp:
Fix a typo in the report that prevent the attached image to be displayed now that we do have the conditional template feature available.


GormFrank (approved):
Approved assuming build passes.

PR #9624, opened 2023-07-07
Tagfilter: Fixing issues with paginate and no result with data-json

Garneauma:
Issue: when having tagfilter and pagination on the same element, the pagination is not updated when items are filtered.
Solution: changing the event name to the standard name `"wb-contentupdated"` and adding data to the tagfilter event trigger only, to be able to differentiate it from other plugins.

Issue: when loading tagfilter items through `data-json`, the no result message is briefly visible on page load.
Solution: added CSS to hide the no result message when no tagged items are defined in the `.wb-tagfilter-items` element


duboisp:
Pre-approved upon successful local testing and after a test example is added. 

I suggest to add that edge case example in a separate web page.

duboisp (approved):
Reviewed and tested locally, everything work as expected.

Thanks @Garneauma for adding the edge case working example.

PR #9628, opened 2023-07-11
wb-disable: Adding option to allow some plugins on basic mode

Garneauma:
Adding option to allow a few plugins to run on basic mode:
- Data-ajax
- Data-JSON
- JSON manager
- Postback

Also adding wb-disable documentation.

### DIFF
--- a/src/plugins/wb-disable/wb-disable-fr.hbs
+++ b/src/plugins/wb-disable/wb-disable-fr.hbs
@@ -1,0 +1,24 @@
+---
+{
+"title": "HTML simplifié",
+"language": "fr",
+"category": "Plugiciels",
+"description": "Passer à la version HTML simplifiée",
+"tag": "disable",
+"parentdir": "wb-disable",
+"altLangPrefix": "wb-disable",
+"dateModified": "2023-07-11"
+}
+---
+
+<div class="wb-prettify all-pre"></div>
+
+<h2 id="basicMode">Mode HTML simplifié</h2>
+<p>Cliquez sur le lien ci-dessous pour activer le mode HTML simplifié</p>
+<p><a href="?wbdisable=true">Basculer vers le mode HTML simplifié</a></p>
+<p><a href="?wbdisable=false">Basculer vers la page régulière</a></p>
+
+<h3>Code</h3>
+<pre><code>&lt;p>Cliquez sur le lien ci-dessous pour activer le mode HTML simplifié&lt;/p>
+&lt;p>&lt;a href="?wbdisable=true">Basculer vers le mode HTML simplifié&lt;/a>&lt;/p>
+&lt;p>&lt;a href="?wbdisable=false">Basculer vers la page régulière&lt;/a>&lt;/p></code></pre>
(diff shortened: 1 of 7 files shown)


src/plugins/wb-disable/wb-disable-fr.hbs
```diff
@@ -0,0 +1,24 @@
+---
+{
+"title": "HTML simplifié",
+"language": "fr",
+"category": "Plugins",
```
duboisp, line 5: Voici le correctif pour l'erreur relevé par le link checker de notre CI script

```suggestion
"category": "Plugiciels",
```

duboisp (dismissed):
I locally and manually tested every wb-disable-allowed init and it worked as expected.

It's great to see the introduction of our GCWeb index.json-ld here which will allow us to use it in the future.

There is an CI error that need to be fixed, see my inline comment.

Great work on this.

src/plugins/wb-disable/wb-disable-fr.hbs
Garneauma: Unfortunate that `grunt dist` or `grunt test` don't notice that error. It is fixed now!

duboisp (dismissed):

GormFrank (dismissed):
In this PR, we must take into consideration the fact that a domino effect is possible here, since both `data-ajax` as well as `wb-data-json` can trigger init events after dynamic data is loaded, assuming they have WET plugins within that data; resulting in a potential Basic HTML mode that isn't so basic after all.

Garneauma:
> In this PR, we must take into consideration the fact that a domino effect is possible here, since both `data-ajax` as well as `wb-data-json` can trigger init events after dynamic data is loaded, assuming they have WET plugins within that data; resulting in a potential Basic HTML mode that isn't so basic after all.

Data loaded via data-ajax are now taken into consideration. Users must add the class `.wb-disable-allow` to elements loaded in order for them to be initialized in basic mode.

duboisp (changes requested):
I tried locally,  and I do confirm:
* Data ajax is initiated in basic mode
   * In my previous review I tested all init method and it worked. Here I only tested the init with `[data-ajax-replace]`
* json-manager component in the ajaxed contetn is initiated if activated do run in basic mode
* data-json component in the ajaxed content is initiated if activated do run in basic mode
  * I only tested the init with : `data-wb-json`

But the wb-postback plugin don't start.

TODO:
* [x] Ensure the wb-postback plugin can start in basic mode

Garneauma:
> I tried locally, and I do confirm:
> 
> * Data ajax is initiated in basic mode
>   
>   * In my previous review I tested all init method and it worked. Here I only tested the init with `[data-ajax-replace]`
> * json-manager component in the ajaxed contetn is initiated if activated do run in basic mode
> * data-json component in the ajaxed content is initiated if activated do run in basic mode
>   
>   * I only tested the init with : `data-wb-json`
> 
> But the wb-postback plugin don't start.
> 
> TODO:
> 
> * [x]  Ensure the wb-postback plugin can start in basic mode

Fixed wb-postback.

Apparently, using `.wb-postback` as the selector for the init event does not work. It only works when using `[data-wb-postback]`. I can't figure out why by looking at the code.

duboisp (approved):
Reviewed and tested locally,

This is a minor change - Add a new basic init feature for data-ajax, data-json, json-manager and postback.

PR #9629, opened 2023-07-12
Data-json - Fix templating with null value

duboisp:
I find out today the current ACR report don't work because it do contains a null value and null value are interpreted as object.

This is patch to Data-JSON plugin and the [ACR English theoretical report](https://wet-boew.github.io/wet-boew/unmin/demos/acr/demo-wcag21/acr-wcag21-en.html) can be use as the test case


Ricokola (approved):
Tested locally and pass successfully!

PR #9630, opened 2023-07-12
Data-ajax - Add flag to HTML encode the ajaxed content

duboisp:
Add flag to HTML encode the ajaxed content

This feature is needed to support and allow the full implementation of the GCWeb Global header documentation prototype as discussed with DTO.


duboisp:

polmih (approved):
review completed and tested on localhost

PR #9632, opened 2023-07-12
JSON manager - Improve wb-swap to use @id prop instead of only the object key

duboisp:


duboisp:

site/pages/docs/ref/wb-jsonmanager/wb-jsonmanager-en.hbs
```diff
@@ -201,7 +201,7 @@
 		<tr>
 			<td>Swap data with reference</td>
 			<td><code>{ "op": "wb-swap", "path": &lt;path to current value&gt;, "ref": &lt;path to reference value&gt; }</code></td>
-			<td>Swaps data from "path" with data from "ref" using the value as a key in the reference. The value from "path" can either be a string or an array.</td>
+			<td>Swaps data from "path" with data from "ref" using the value as a key in the reference or as the value set in the <code>@id</code> property of an array of object in the reference. The value from "path" can either be a string or an array.</td>
```
Garneauma, line 204:
```suggestion
			<td>Swaps data from "path" with data from "ref", using either the value as a key in the reference or the value set matching the <code>@id</code> property in an array of objects in the reference. The value from "path" can either be a string or an array.</td>
```

Garneauma (changes requested):
Works well. I have made an attempt to improve the description of the "wb-swap" option. Let me know what you think.

site/pages/docs/ref/wb-jsonmanager/wb-jsonmanager-en.hbs
duboisp: J'ai appliqué le changement de texte.

Garneauma (approved):
Last change request applied. Approved.

PR #9633, opened 2023-07-13
Adding an avoid note on the Carousel working examples

ouafaaetta:
This is to add a note on the Carousel working example pages on the WET website, as recommended


duboisp (changes requested):

GormFrank:

duboisp:
This is only a content change

duboisp (approved):
Tested locally, work as expected

This is a content change only

PR #9640, opened 2023-08-09
Tagfilter: fixing no result element blink on load with Data-JSON

Garneauma:
Fixing rare issue where the "No result" element would briefly appear on page load when loading items from Data-JSON before the items would show up. 

Only happens on Firefox since it does not support the `:has()` CSS selector.

Changes related to WET-405.


duboisp (approved):
Test and I noticed the "no result message" is not quickly displayed and then get replaced by the content. Now the new behaviour is it will show all the result first prior to apply the filter.

PR #9641, opened 2023-08-10
Lightbox: WGAG 2.1 AA Pre-Assessment

BrahimMahadi:
### This pull request includes the WCAG 2.1 Accessibility pre-assessments done for the lightbox plugin

General checklist
- [x] Updated lightbox documentation to include the accessibility pre-assessment reports
- [x] Lightbox was pre-assessed against WCAG for accessibility
- [x] Documentation is bilingual

### DIFF
--- a/src/plugins/lightbox/index.json-ld
+++ b/src/plugins/lightbox/index.json-ld
@@ -1,0 +1,58 @@
+{
+	"@context": {
+		"@version": 1.1,
+		"dct": "http://purl.org/dc/terms/",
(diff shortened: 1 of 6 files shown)


src/plugins/lightbox/index.json-ld
duboisp, line 1: This file is not needed for plugin in wet-boew, but let's keep it as eventually we are going to add it.

duboisp (approved):
Reviewed and tested locally

This is a content change

PR #9642, opened 2023-08-11
Data-JSON, JSON manager extrator - Support null mapping

duboisp:
Minor change for Data-JSON - Support null mapping to display template without further data processing
Patch change for JSON manager - Return null when trying to extract content from an inexistant HTML node


duboisp:
Pre-approved upon successful tested and review

GormFrank (dismissed):
I would've enjoyed a small rationale or use case of this new feature, but from a technical perspective it's all looking good. No impact on existing features either. 

Juste des petites erreurs grammaticales, mais good to go quand même!

duboisp:
@GormFrank the rational is we found that an empty object was returned instead of null when the CSS selector found nothing. Having an empty object increased the difficulties to test that situation with our conditional template. That situation was found by @Garneauma when he was working to merge the 2 ajax template for the page feedback PR with our conditional template.

I made the typo fix.

GormFrank (approved):

PR #9643, opened 2023-08-11
Data-JSON: do not display null values on JSON render

polmih:

In Data-JSON, null values are actually being displayed as “null” on the page. This solution will detect and ignore null values once the json values are rendered on the page
_WET-402_


duboisp:

duboisp (approved):
Reviewed and tested locally, work as expected

PR #9651, opened 2023-08-21
data-json: fixing issue with null value

Garneauma:
Fixing issue where Data-JSON would throw an error when the value of a property is equal to `null`.

Related to WET-408


GormFrank (dismissed):

duboisp (changes requested):
Approved but I am going to add a test cases.

duboisp (approved):
Tested the fix and now there is a test example.

Before to merge, I will just wait for either @Garneauma or @GormFrank confirmation about the test cases that was added.

This change is a Patch

duboisp:
Confirmed with @Garneauma the test-case addition I made is ok.

PR #9656, opened 2023-08-23
Feed - Fix loading issue for the ajaxed JSON feed

duboisp:
Fix loading issue for the ajaxed JSON feed

Fix the console error "Uncaught TypeError: e.fetch.response.get is not a function" that is happening in the Feeds working example page.


Garneauma (approved):
Tested locally and works as intended.

PR #9658, opened 2023-08-23
minor fixes for dl-horizontal

Christopher-O:
This fixes a margin left issue with the dd margin left design for dl-horizontal with the new implementation of the 20ch width for the dt. Only visible when the dd content wraps to a new line.

Current
image
With fix
image


duboisp:
Pre-approve

duboisp (approved):
Reviewed, tested locally, reproduced the issue and I confirm this fix the alignment issue when there is a long description definition `<dd>`.

I added a commit which add a longer `<dd>` content for testing and to ensure we do have a test condition that specific edge case.

PR #9662, opened 2023-09-01
Bootstrap 4-5: Backport more classes

EricDunsworth:
**WIP**

Specifically:
* Borders:
  * border-top-0
  * border-bottom-0
* Color scheme:
  * text-black (5.1)
* Flexbox:
  * d-md-flex
  * flex-sm-column
  * flex-sm-grow-1 (4.1)
  * flex-md-column
  * flex-md-grow-1 (4.1)
  * flex-md-wrap
  * flex-wrap
* Shadows:
  * shadow-none (4.1)

Notes:
* FR typo fix
* col-xs-12 class name typo fixes in flex-sm-wrap examples
* Some backported classes are from newer versions of Bootstrap than 4.0
* Used !important in classes tied to common CSS properties
  * Omitted it from classes tied to uncommon properties (e.g. property names containing "flex") or where precents existed (e.g. d-sm-flex didn't use !important in its display property... so d-md-flex won't either)
* The backported text-black class is inconsistent with WET's implementation of text-white:
  * text-black uses !important... text-white doesn't
  * text-white contains link and button colour overrides... text-black doesn't (Bootstrap 4 has no solution for this, 5 uses link-* classes)
  * text-black is more in line with Bootstrap's version than WET's version of text-white

### DIFF
--- a/src/base/_bootstrap-4.scss
+++ b/src/base/_bootstrap-4.scss
@@ -3,6 +3,15 @@
   @title: Bootstrap 4 suplemental style for WET-BOEW
  */
 
+/* Borders */
+.border-top-0 {
+	border-top: 0 !important;
+}
+
+.border-bottom-0 {
+	border-bottom: 0 !important;
+}
+
 /* Spacing */
 .p-0 {
 	padding: 0 !important;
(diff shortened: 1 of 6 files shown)


duboisp (commented):
@EricDunsworth may be during one of our weekly office hours, can you show us your department use case which would require leverage those new CSS backport.

PR #9668, opened 2023-09-13
Data-Ajax: fix for when ajaxing templating logic

Garneauma:


duboisp (changes requested):
Can we add a simple example that show the issue we were having?

Garneauma:
> Can we add a simple example that show the issue we were having?

Working example added. I also changed my solution. This one fixes the issue with very little change to the code.

duboisp (changes requested):

src/plugins/ajax-fetch/ajax-fetch.js
```diff
@@ -81,7 +81,15 @@ $document.on( "ajax-fetch.wb", function( event ) {
 				fetchData.pointer = $( "<div id='" + wb.getId() + "' data-type='" + responseType + "'></div>" )
 					.append( responseType === "string" ? response : "" );
 
-				response = !xhr.responseJSON ? $( response ) : xhr.responseText;
+				if ( !xhr.responseJSON ) {
+					try {
+						response = $( response );
+					} catch ( e ) {
+						response = DOMPurify.sanitize( xhr.responseText );
```
duboisp, line 88: DOMPurify need to be mapped in the plugin closure mapping.

duboisp (changes requested):
Tested and reviewed the code. Everything work as expected.

Just one little change, see inline comment.

duboisp (approved):
Change requested was completed

PR #9669, opened 2023-09-14
Randomize: Adding attribute values randomization option

Garneauma:
"isAttr": true/false -> set to true if the the selector is an attribute of the randomize element.
"values": list of possible values for the attribute.

### DIFF
--- a/src/plugins/wb-randomize/wb-randomize-en.hbs
+++ b/src/plugins/wb-randomize/wb-randomize-en.hbs
@@ -186,3 +186,87 @@
 	&lt;/tbody&gt;
 &lt;/table&gt;</code></pre>
 </details>
+
+<h3>Randomize attribute value</h3>
+<p>Select a random option within a list of options and append it as the value to the defined attribute</p>
+
+<h4>Example 1</h4>
+<div class="bg-cover well"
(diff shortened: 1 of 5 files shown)


src/plugins/wb-randomize/wb-randomize-en.hbs
```diff
@@ -186,3 +186,93 @@
...
 &lt;/table&gt;</code></pre>
 </details>
+
+<h3>Randomize attribute value</h3>
+<p>Select a random option within a list of options and append it as the value to the defined attribute</p>
+
+<h4>Example 1</h4>
+<div class="bg-cover well"
+	data-wb-randomize='{
+		"selector": "data-bgimg-srcset",
```
duboisp, line 196:
```suggestion
		"attribute": "data-bgimg-srcset",
```

duboisp (changes requested):
+ remove the setting "isAttr"

duboisp:
Pre-approved once the change requested are completed + review + local testing

duboisp (commented):
reviewed and tested locally,

Work as expected

This is minor change

duboisp (approved):

PR #9678, opened 2023-10-04
Geomap: adding zoom to bounding box on load option

Garneauma:
Adding `mapExtent` configuration option to allow users to zoom to a specific area on page load.

### DIFF
--- a/src/plugins/deps/geomap-lib.js
+++ b/src/plugins/deps/geomap-lib.js
@@ -192,8 +192,10 @@ var componentName = "wb-geomap",
 		// Add map layers
 		this.addMapLayers();
 
-		// If an extent was configured, fit the map to it
-		if ( viewOptions.extent ) {
+		// If an extent was configured, fit the map to it (use mapExtent property, then basemap.extent)
+		if ( this.settings.mapExtent ) {
+			this.zoomAOI( this.settings.mapExtent );
+		} else if ( viewOptions.extent ) {
 			this.map.getView().fit( viewOptions.extent, this.map.getSize() );
 		}
 
--- a/src/plugins/geomap/geomap-en.hbs
+++ b/src/plugins/geomap/geomap-en.hbs
@@ -30,6 +30,7 @@
 			<li><a href="#osm">OSM basemap example</a></li>
 			<li><a href="#tile-source">Tile Source (XYZ) basemap example</a></li>
 			<li><a href="#filtering">Filtering example</a></li>
+			<li><a href="#aoiZoomOnLoad">Zoom to AOI on page load (extent)</a></li>
 		</ul>
 	</div>
 </div>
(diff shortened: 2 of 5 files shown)


src/plugins/deps/geomap-lib.js
```diff
@@ -192,8 +192,10 @@ var componentName = "wb-geomap",
 		// Add map layers
 		this.addMapLayers();
 
-		// If an extent was configured, fit the map to it
-		if ( viewOptions.extent ) {
+		// If an extent was configured, fit the map to it (use mapExtent property, then basemap.extent)
+		if ( this.settings.mapExtent ) {
```
GormFrank, line 196: Could we do simply `extent` instead of `mapExtent` since we are already in the context of a map? Or is it because the word is already reserved for something else?

src/plugins/geomap/geomap-en.hbs
```diff
@@ -1303,3 +1303,53 @@ var wet_boew_geomap = {
 		</details>
 	</section>
 </section>
+
+<section id="aoiZoomOnLoad">
+	<h2>Zoom to AOI on page load</h2>
```
GormFrank, line 1308: Could you add the word extent in parathensis here since you are interchangeably using zoom AOI and extent for this feature?
```suggestion
	<h2>Zoom to AOI on page load (extent)</h2>
```

src/plugins/geomap/geomap-en.hbs
```diff
@@ -1303,3 +1303,53 @@ var wet_boew_geomap = {
...
+
+	<form class="wb-geomap-filter form-inline mrgn-bttm-md" data-bind-to="bc-geomap">
+		<div class="form-group">
+			<label for="zoomBCArea">B.C. Regions:</label>
+			<select class="form-control" data-filter="aoi" id="zoomBCArea">
+				<option value="63, -110, 50, -128" selected>Select a region</option>
```
GormFrank, line 1314: This value uses commas between each points, but others don't; please make it consistent. 

Whether it should be with or without commas though is a decision that should be based on what's currently used for other configs of Geomap and also the Action manager.

GormFrank (changes requested):
I have not built locally to test; only made a review looking at the source code itself.

@Garneauma Merci :)

src/plugins/deps/geomap-lib.js
Garneauma: Since "extent" seemed to already be used and, since I don't know the plugin throughout, I didn't want to create confusion or conflict between with this new property. However, I am very open to any other name. TBH, I am not a fan of this property name either, but I couldn't come up with a better name myself.

GormFrank: @Garneauma Ok then, even though I am not a fan of the current name either, the rationale makes sense so let's keep `mapExtent`.

GormFrank:
Pre-approved upon successful local testing

src/plugins/geomap/geomap-en.hbs
```diff
@@ -1303,3 +1303,53 @@ var wet_boew_geomap = {
 		</details>
 	</section>
 </section>
+
+<section id="aoiZoomOnLoad">
+	<h2>Zoom to AOI on page load (extent)</h2>
```
GormFrank, line 1308: Peux-tu ajouter l'exemple dans la table des matières svp?

src/plugins/geomap/geomap-fr.hbs
```diff
@@ -1291,3 +1291,53 @@ var wet_boew_geomap = {
 		</details>
 	</section>
 </section>
+
+<section id="aoiZoomOnLoad">
+	<h2>Zoom à AOI au chargement de la page (étendue)</h2>
```
GormFrank, line 1296: Peux-tu ajouter l'exemple dans la table des matières svp? Et aussi, prendre cet opportunité pour régler l'erreur d'orthographe "cartre" dans la page.

GormFrank (changes requested):
Petits changements demandés, mais devrait être tout beau après et prêt à être merged!

@Garneauma Merci :)

GormFrank (approved):
Approved upon successful build

duboisp:
Thanks @GormFrank for the review and @Garneauma for the changes

Issue #9682, opened 2023-10-12
Including server side errors in form validation summary


sgdowney:
I'm using the following code to include a server side validation error message in the form based on the documentation for including server validation errors in the Form Validation plugin.
```
<fieldset class="form-inline chkbxrdio-grp">
    <legend class="required">
        <span class="field-name">Did you attend school or a training course during Week 2 of this report?</span>
        <strong class="required">(required)</strong>
        <span class="label label-danger wb-server-error" id="QuestionModel_SelectedResponse_validationMessage">You have declared that you attended school or training during the period of this report but you also declared that you did not attend school or training during either the first or second week.  You must indicate in which week(s) you attended training.  Review your answers and make corrections where necessary.</span>
    </legend>
    <label class="radio-inline">
        <input checked="checked" class="input-validation-error" data-msg-required="You must provide a yes or no answer." data-rule-required="true" id="QuestionModel_SelectedResponse" name="QuestionModel.SelectedResponse" type="radio" value="No" />
        No
    </label>
    <label for="TrainingYes" class="radio-inline">
        <input class="input-validation-error" id="TrainingYes" name="QuestionModel.SelectedResponse" type="radio" value="Yes" />
        Yes
    </label>
</fieldset>
```

When the page loads in the browser, the validation summary briefly appears and the server validation message is prefixed with **Error 1:**.  Then the validation summary disappears and the prefix is removed from the message.  There are no client side validation errors on this page, only the server validation error.  What do I need to add to the HTML to get the validation summary to remain on the page?

sgdowney:
This appear to be specific to fieldsets.  When I apply the wb-server-error class to a select or input type='text' the error summary is generated when the page loads.

sgdowney:
It looks like this code from formvalid.js does not take into account inline radio or checkbox inputs.
```
//Trigger validation on wb-server-error
$form.find( ".wb-server-error" ).filter( ":not( :hidden )" ).parent().each( function() {
	if ( this.attributes.for && this.attributes.for.value.length > 0 ) {
		$( "form" ).validate().element( $( "[id =" + this.attributes.for.value + "]" ) );
	} else if ( $( this ).find( "input" )[ 0 ] ) {
		$( "form" ).validate().element( $( this ).find( "input" )[ 0 ] );
	} else if ( $( this ).next( ".radio, .checkbox" ).children( "label" ).children( "input" )[ 0 ] ) {
[...]

PR #9686, opened 2023-10-19
Update formvalid.js

### LINKED ISSUE #9684: [BUG] Form Validation - Inline radio and checkbox inputs with wb-server-error class not recognized as errors to display
## Required information

**Describe the bug**
Using the markup for including server side errors is not recognized on inline radio and checkbox inputs.  This prevents the Error Summary from being rendered when the page loads.

**To Reproduce**
Steps to reproduce the behaviour:
See issue #9682  for an example and possible fix.

**Expected behaviour**
When the page is rendered, the Form Validation Error Summary should be displayed. The error message from the server side validation should be updated with the prefix and error counter **Error #;**.

**Who shall do the work?**
I am asking for Principal publisher to please do the work.  I don't have the tools to properly code and test a solution.

**Version of WET-BOEW/GCWeb you are using**
CDTS version 4.1.0

sgdowney:
Update to Form Validation plug-in file formvalid.js
Add conditions for inline radio and checkbox elements marked with wb-server-error class.  Triggers display of error summary on page load when server side errors have been added to the page.

Closes #9684, #9682


sgdowney:
I'm looking at the formvalid-server-en.hbs and formvalid-server-fr.hbs files.  Maybe I'm not fully understanding the purpose of these files as they already have the full markup for the server errors as if the plug-in has already run.  My understanding of how the plug-in should work is that you add the error message from the server validation without providing the error prefix (**Error X:**) and allow the plug-in to add the prefixes and generate the error summary.  The hbs files have the error prefix already embedded in the server side error messages.  What would be the correct way to add these examples to the file?  With the prefix already added or without?

duboisp:
@GormFrank for your review

sgdowney:
I'm going to clean this up a bit better then submit a new pull request to apply the fix.

PR #9675, opened 2023-09-27
Twitter embedded timeline: Add "verified" account example

EricDunsworth:
Timeline widgets currently work differently between standard and "verified" Twitter accounts:
* Standard accounts: Shows a "Nothing to see here - yet" placeholder message
* "Verified" accounts: Shows top 100 most-liked tweets (ignores ``data-tweet-limit`` attribute)

This adds a second timeline for @Canada ("verified" government account) to demonstrate the difference between both "types" of account timelines.

Other changes:
* Use a ``data-height="fb-page"`` attribute for the "verified" account example (via #9691)
* Adjust nearby content to be more presentable:
  * Revise example heading (change to plural, update timeline's official name, use sentence case)
  * Revamp Twitter docs link into an expanded section
* Update French translations:
  * Translate main page content
  * Update outdated page description translation

PR #9691, opened 2023-11-03
Twitter embedded timeline: Add data-* presets

EricDunsworth:
Specifically:
* ``data-height="500"``:
  * WET sets it when:
    * Value is explicitly set to "fb-page"
    * ``data-tweet-limit`` is used without ``data-height``
  * Needed to prevent older timeline widget implementations from growing extremely tall and to set a reasonable "generic" height going forward
  * Starting July 21, 2023... timeline widgets stopped honouring tweet limits and began showing 100 tweets at a time ("verified" accounts only)
  * 500 pixels matches the default height of the Facebook page plugin (aka Facebook embedded pages)
* ``data-lang="zh-cn"``:
  * WET sets it when the in-page language is "zh-Hans"
  * Twitter only supports "zh-cn" for Chinese (Simplified)
* ``data-dnt="true"``:
  * Not used if ``data-lang="false"`` is already hardcoded

Related changes:
* Plugin now checks every Twitter link in every plugin container (not just the first link)
* Adjust the loading icon's logic accordingly:
  * Move it into the ``data-*`` logic's ``foreach`` loop
  * Make its initial ``if`` condition check for ``eventTarget.firstElementChild`` (i.e. only add a loading icon for the first link in the container)
* Documentation:
  * Add specific links to Twitter timelines/languages/properties docs
  * Add notable configuration options (ones affected by this PR's logic changes)

Other:
* Decided against rolling these changes into #9675:
  * These changes go beyond adding support logic for #9675's new example (presetting ``data-lang`` and ``data-dnt`` have nothing to do with "fixing" that example's height)
  * #9675 serves a separate purpose (i.e. demonstrate both non-functional and functional widgets... nothing more)

### DIFF
--- a/src/plugins/twitter/twitter.js
+++ b/src/plugins/twitter/twitter.js
@@ -31,7 +31,7 @@ var componentName = "wb-twitter",
 			protocol = wb.pageUrlParts.protocol;
 
 		if ( eventTarget ) {
-			const twitterLink = eventTarget.firstElementChild;
+			const twitterLinks = eventTarget.querySelectorAll( "a.twitter-timeline" );
 
 			// Ignore IE11
 			// Note: Twitter's widget no longer supports it...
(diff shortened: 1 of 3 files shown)


src/plugins/twitter/twitter.js
```diff
@@ -40,39 +40,78 @@ var componentName = "wb-twitter",
...
+						if ( twitterLinkLang.toLowerCase() === "zh-hans" ) {
+							twitterLinkLang = "zh-cn";
+						}
 
-						// Deal only with removed HTML nodes
-						mutation.removedNodes.forEach( function( removedNode ) {
+						// Set the widget to the link's language
+						twitterLink.dataset.lang = twitterLinkLang;
+					}
+				}
```
EricDunsworth, line 64: Turns out that most of this logic is pointless. Twitter's widget script is already capable of "re-using" the closest parent element's language on its own.

The English widget UI in the [French demo page](https://wet-boew.github.io/v4.0-ci/demos/twitter/twitter-fr.html)'s example threw me off :P. Most of that page's content is wrapped within ``<div lang="en">``... which caused Twitter's script to go with English. I mistook it as Twitter arbitrarily defaulting to English unless ``data-lang`` is explicitly-set.

My next force-push will only keep the Chinese (Simplified) replacement logic at this spot.

EricDunsworth:
This should be good to review now. Btw I'd recommend merging it prior to #9675.

duboisp (changes requested):
We will need to have the accompanying working example for local testing.

We will at least one demo (a demo that combine all) per new added feature/configuration.

EricDunsworth:
Superseded by #9704. Check it for follow-ups to the review comment.

PR #9694, opened 2023-11-09
Formvalid: adding working examples for radios and checkboxes

Garneauma:
Adding a few working examples for the stacked and inline checkboxes and radios because it seems people were not using the correct HTML as some examples were missing.


duboisp:
Pre-approved upon review + local testing.

duboisp (approved):
Reviewed and tested locally. Work as expected.

This is a content change.

PR #9701, opened 2023-11-17
Footnotes: Demonstrate reference link in heading

EricDunsworth:
Using footnote reference links on headings is a valid use case, so let's demonstrate it.

This achieves it by moving footnote 4's second reference link onto its parent section's heading.


duboisp:
Pre-approved upon review + local testing.

This is a content change

duboisp (approved):
Review and tested locally, Work as expected

I confirm this is a content change.

PR #9703, opened 2023-11-20
Formvalid: Fixing dependency load issue

Garneauma:
Queuing dependency load for Formvalid. additional-methods.js is now only loaded after jquery.formvalid.js has done loading. This prevents a racing condition when the page is heavy.

PS: Not sure why Github lists than many changes. Maybe it has to do with the indentation. Only thing that changed is lines 69 to 80 and 394 to 400.

### DIFF
--- a/src/plugins/formvalid/formvalid.js
+++ b/src/plugins/formvalid/formvalid.js
@@ -66,333 +66,338 @@ var componentName = "wb-frmvld",
 			}
 
 			Modernizr.load( {
-
-				// For loading multiple dependencies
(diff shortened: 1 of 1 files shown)


duboisp:
Pre-approved upon review + local testing

duboisp (dismissed):
I reviewed and tested locally and it do work as expected.

We will be able to merge it once the conflict is resolved.

duboisp (approved):
Quickly re-tested and still work as expected.

This is a Patch change

PR #9704, opened 2023-11-20
Twitter embedded timeline: Add data-* presets, "verified" example, skip links, iframe title fix

EricDunsworth:
Combined PR for #9691 / #9675's changes and more:
* Closes #9691
* Closes #9675
* Supporting changes

## Twitter embedded timeline: Add data-* presets

Specifically:
* ``data-height="500"``:
  * WET sets it when:
    * Value is explicitly set to "fb-page"
    * ``data-tweet-limit`` is used without ``data-height``
  * Needed to prevent older timeline widget implementations from growing extremely tall and to set a reasonable "generic" height going forward
  * Starting July 21, 2023... timeline widgets stopped honouring tweet limits and began showing 100 tweets at a time ("verified" accounts only)
  * 500 pixels matches the default height of the Facebook page plugin (aka Facebook embedded pages)
* ``data-lang="zh-cn"``:
  * WET sets it when the in-page language is "zh-Hans"
  * Twitter only supports "zh-cn" for Chinese (Simplified)
* ``data-dnt="true"``:
  * Not used if ``data-dnt="false"`` is already hardcoded

Related changes:
* Documentation:
  * Add specific links to Twitter timelines/languages/properties docs
  * Add notable configuration options (ones affected by this PR's logic changes)

## Twitter embedded timeline: Add "verified" account example

Timeline widgets currently work differently between standard and "verified" Twitter accounts:
* Standard accounts: Shows a "Nothing to see here - yet" placeholder message
* "Verified" accounts: Shows top 100 most-liked tweets (Twitter ignores ``data-tweet-limit`` attributes)

This adds a second timeline for ``@Canada`` ("verified" government account) to demonstrate the difference between both "types" of account timelines.

Other changes:
* Use a ``data-height="fb-page"`` attribute for the "verified" account example
* Adjust nearby content to be more presentable:
  * Revise example heading (change to plural, update timeline's official name, use sentence case)
  * Revamp Twitter docs link into an expanded section
* Update French translations:
  * Translate main page content
  * Update outdated page description translation

## Supporting changes

* Add skip links to the start and end of the timeline:
  * Spares users from needing to tab through several hundred interactive elements to get past "verified" timeline widgets
* Replace the timeline ``iframe``'s English-only ``title`` attribute with an i18n version written in sentence case
* Prevent the timeline's ``iframe`` from getting wider than its parent container (letting it happen causes skip links/notices to look misaligned)
* Allow the timeline to scroll horizontally (in case its ``iframe`` gets narrower than 224px)
* Drop partial support for multiple Twitter links in the same plugin container:
  * Didn't look very presentable
  * Didn't play well with skip links/notices
  * No known implementations in the wild


EricDunsworth:
@duboisp Regarding #9691's review comments...

> We will need to have the accompanying working example for local testing.

Resolved now that both PRs have been combined.

> We will at least one demo (a demo that combine all) per new added feature/configuration.

I can go down that path if needed, but should note the following:
* The logic changes don't introduce any brand new ``data-*`` attributes. Most of the logic changes just cause WET to preset the values of some of Twitter's ``data-*`` attributes (**except** for ``data-height="fb-page"`` - it's a brand new value demonstrated in the new "verified" account example).
* The plugin's old documentation basically said "check Twitter's docs to learn about their config options". What I did with the new config options table was to cover the ``data-*`` attributes WET's plugin is now playing around with. It doesn't cover all of Twitter's config options.
* I'd recommend against adding too many new Twitter examples since it'd risk adding bloat and increased load times (each "verified" timeline loads 100 tweets apiece and the new example already loads kind of slowly as-is...).

duboisp (commented):
I tried locally and I can't not see any difference for the verified account.

I also tried when I was logged in to twitter (In a private browser session) but it resulted to the same experience where there is no difference for verified account.

Did X changed their embed API recently? @EricDunsworth do you experience the same?

EricDunsworth:
@duboisp I just tried locally and the "verified" ``@Canada`` example isn't working for me either :S. Not sure if Twitter deliberately changed anything on their end or if it's just running into technical difficulties atm.

I did notice a few other things just now:
* "Verified" government (e.g. [``@Canada``](https://twitter.com/Canada),[``@USAGov``](https://twitter.com/USAGov), [``@ottawacity``](https://twitter.com/ottawacity)) and personal (e.g. [``@StephenKing``](https://twitter.com/StephenKing)) accounts are currently acting like standard "unverified" accounts:
  * Their profile pages are stuck behind login walls for guests
* "Verified" business accounts (e.g. [``@NintendoCanada``](https://twitter.com/NintendoCanada)) and affiliates of Twitter (e.g. [``@elonmusk``](https://twitter.com/elonmusk), [``XDevelopers``](https://twitter.com/XDevelopers)) work better, but not as well as before:
  * Their profile pages:
    * Still work for guests (no login walls)
    * Only show pinned tweets to guests... normal tweets aren't showing anymore
      * Previous behaviour was to show the top 100 most-liked tweets to guests
* Embedded timelines for all "verified" account types are now showing "Nothing to see here - yet" for guests

> Did X changed their embed API recently?

That's a funny question :D... They've been randomly tinkering with the timeline API since late July, but haven't announced any of their changes and aren't updating their API docs anymore. Keeping up with all the undocumented changes seems to be turning into a game a whack-a-mole.

For all I know, maybe embedded timelines are gone for good now. But it's hard to tell for sure since their docs are still up and I haven't seen any announcement about it anywhere. I haven't even found any forum/community/etc posts about it so far - so what we noticed was probably a _very_ recent change.

duboisp:
Then, I will try again tomorrow and I may probably suggest to add a notice near by our working example regarding the instability of the behaviour caused by unknown. 
Thanks for researching @EricDunsworth

EricDunsworth:
@duboisp The ``@Canada`` embedded timeline example seems to be working again for me... for now :P. But it'd probably for the best to hold off for at least another week and see what happens.

Other things I observed yesterday are all over the place atm :S:
* ``@Canada``'s profile page now works again and is showing 100 tweets
* ``@USAGov`` and ``@NintendoCanada``'s profile pages work but only show one pinned tweet apiece
* ``@ottawacity`` (verified government account) and ``@StephenKing`` (vertified personal account)'s embedded timelines still don't work and their profile pages are still behind login walls for whatever reason

I based the demo page updates on the presumption that embedded timelines work for "verified" accounts and don't work with standard accounts. I also tested my presumptions while prepping the PR.

But I can't make sense of what's happening now :S. Only _some_ "verified" accounts currently have working embedded timelines. Even accounts sharing the same type of "verified" status (like ``@Canada`` and ``@ottawacity`` both being marked as "verified" government accounts) have different embedded timeline behaviour.

duboisp:
I tired again and it is now working for me too. May be they had an service outage yesterday.

I will complete my review and local testing this week and we will merge if it is still working next Monday.

EricDunsworth:
Here's an outline of what I'm currently seeing compared to my [last comment](https://github.com/wet-boew/wet-boew/pull/9704#issuecomment-1830244610):
* All timeline widgets currently show "Nothing to see here - yet" for guests :S
* [@Canada](https://twitter.com/Canada), [@ottawacity](https://twitter.com/) and [@StephenKing](https://twitter.com/StephenKing)'s profile pages are currently stuck behind login walls
* [@USAGov](https://twitter.com/USAGov), [@NintendoCanada](https://twitter.com/NintendoCanada), [@elonmusk](https://twitter.com/elonmusk) and [XDevelopers](https://twitter.com/XDevelopers)'s profile pages work for guests, only show one pinned tweet


> I may probably suggest to add a notice near by our working example regarding the instability of the behaviour caused by unknown.

I'm feeling tempted to go in that direction :)...

duboisp (changes requested):
Tested and review locally. Everything seem to work as expected, but we need

* [x] 1 additional working example where:
  * Multiple timeline is contained inside the same wb-twitter plugin. (2 or more)
  * Set a custom numeric value for the height of the feeds

Once that additional working example is added, we should be good to merge this one.

duboisp (dismissed):
I tested the keyboard navigation and we are feeling trapped when we start to navigate a twitter feed. We will need to apply a work around, something like:
* A skip link before to enter in the twitter feed
* Be able to escape the twitter plugin by pressing "esc" key. It could bring the user just on or before the skip link.

Also I made a PR with the additional working example. https://github.com/EricDunsworth/wet-boew/pull/2

EricDunsworth:
@duboisp I agree that the status quo isn't very usable. But IMO that goes beyond the scope of this PR and WCAG conformance.

The ``@Canada`` example would likely require at least 800 ``Tab`` key presses to get through the widget (assuming 100 tweets * 8 tabs per tweet) - which is completely ridiculous 🤯. But at the same time, it still technically conforms to [Success Criterion 2.1.2 (No Keyboard Trap)](https://www.w3.org/WAI/WCAG22/Understanding/no-keyboard-trap.html) since the plugin doesn't "trap" users inside its ``iframe``. Users are still able to "enter" and "leave" the ``iframe`` solely via keyboard-tabbing. Neither the SC nor its techniques treat excessive amounts of tabbing as constituting a "trap". Btw just for comparison's sake - the [Facebook widget demo](https://wet-boew.github.io/v4.0-ci/demos/facebook/facebook-en.html) only needs 63 ``Tab`` key presses to get through.

Just as an aside, IMO the widget does fail WCAG 2.x in other ways. For example, no visual focus indicator of any kind appears when tabbing to "i" icons in the bottom-right of the widget's tweets. I also noticed strange behaviour when focusing through the first tweet's group of 3 images (each image link can be focused separately... but all of them use an "all-in-one" focus outline that covers all 3 images together... which makes it look like focus has gotten "stuck" when tabbing between the images). But unfortunately WET can't do anything on its end to deal with those ☹.

I can try to implement a skip link of sorts in a separate PR. But it won't be possible to modify the plugin to intercept ``Esc`` key presses from inside the Twitter widget's ``iframe``. Just to be clear - I'd love to be able to do that since it'd be super user-friendly to easily "jump out" of the widget at any time. But browsers don't allow parent pages to intercept child ``iframe``s for security reasons 🐱‍👤.

**PS:**
IMO from Twitter's end, the widget is probably dead in the water and only still partially-working due to a coincidence and/or shared code. It won't surprise me if it eventually ceases to work completely. For example, the widget currently contains a few "X" logos... but is still chock-full of Twitter branding (i.e. "Tweets from ``@???``" [X renamed tweets to "posts"], "Follow on Twitter" and "View on Twitter"). Ditto for its documentation. If any real attention was still being paid to the widget, it'd of probably been fully rebranded to X months ago.

duboisp:
Pre-approved after the multiple feed working example is added and a skip link is added before each iframe.

EricDunsworth:
FWIW I won't be able to get this across the finish line until mid-late January. Sorry :(. I'll change it back into a draft for now.

I recently came up with some experimental logic for skip links locally. But it's not ready for prime time yet. It's currently designed to seek out whichever element comes directly after each timeline ``iframe`` (or its parents), traverse down to the sibling's lowest first child element, set an ID on it and insert a skip link just before the widget.

But it's not able to distinguish between "categories" of elements. So for example, if a timeline's plugin container was followed by a paragraph containing formatted text (spans, italics, etc)... my current logic would triage to the deepest possible first child element... It wouldn't be be smart-enough to know that the paragraph itself is [palpable content](https://html.spec.whatwg.org/multipage/dom.html#palpable-content-2) and stick with that. Since there isn't really a means of programmatically detecting whether an element is palpable content vs flow/phrasing content, I might have to refrain from triaging into deeper elements. Or only look for notable kinds of palpable elements (like headings) and otherwise stick to higher-level elements.

Here are todos I'd like to get to before merging this PR:
* Refine my skip link logic (i.e. improve triaging, account for nodes in general instead of just element nodes, etc...)
* Add "reverse" skip links
* Change the "verified" example to use ``@XDevelopers`` (the timelines/profile pages for all "verified" accounts I know of started working again when I last checked on December 8th).
* Determine how to move forward with EricDunsworth/wet-boew#2

EricDunsworth:
Just wanted to provide a status update on this.

I pushed a few newer commits yesterday.

Today I rebased and made further revisions... but was unable to force-push just now due to proxy issues :(. I'll try again on Monday and hope for the best. I'll mark it as ready for review afterwards.

EricDunsworth:
Finally got around to rebasing and enhancing this :)!

Check out the [OP](https://github.com/wet-boew/wet-boew/pull/9704#issue-2002657208)'s "Supporting changes" section and newer commits for more info on what's changed.

@duboisp Questions:
1. Are you on board with how I've implemented the skip links?

   Here's an explanation of how they currently work:
   * Activating the skip to end link focuses onto an end notice paragraph
   * The end notice paragraph:
     * Is set to ``display: none;`` by default
     * Turns into invisible content if the skip to end link gets activated (to make it focusable)
     * Visually-appears upon receiving focus
     * Reverts to ``display: none;`` upon losing focus (to prevent it from lingering on as invisible content)
   * Activating the skip to start link focuses onto a start notice paragraph and adds a "fake" outline to the timeline
     * The start notice works similarly to the end notice - but never visually-appears
     * IMO this provides the best experience for everyone:
       * Sighted users will perceive focus as having landed back onto the timeline
       * Screen reader users will get an announcement about the start notice (if "real" focus had landed onto the timeline itself, the *entire* timeline would've been automatically announced :O...)
3. How would you like to proceed with EricDunsworth/wet-boew#2? I'd prefer closing it to avoid bloating the demos. But if you feel strongly about it, I could go ahead and revise + merge it.

**Heads up to reviewers:**
* In NVDA 2023.3.3 + Firefox/Chromium, navigating to the inside of the timeline's ``iframe`` causes NVDA's narrator to go completely silent for ~30-40 seconds... then announce all tweets in one shot afterwards :O.
  * Probably related to ``role="region"`` and the enormous size of the timeline. My guess is that maybe NVDA is trying to concatenate all of the region's labels together before announcing anything - and taking ages to do it...
* Firefox handles ``iframe`` elements kind of strangely:
  * [``iframe`` elements get treated as interactive content](https://bugzilla.mozilla.org/show_bug.cgi?id=1483828) (i.e. the element itself gains focus when keyboard-tabbing... which goes against the HTML spec and deviates from every other browser)
  * [Focused ``iframe`` elements get a dotted focus indicator instead of a focus ring](https://bugzilla.mozilla.org/show_bug.cgi?id=1863525)... it's almost imperceptible when tabbing into the Twitter timeline

duboisp:
We will do a code review and local testing.

EricDunsworth:
Noticed that [Twitter Publish's ``@XDevelopers`` timeline ``iframe`` example](https://publish.twitter.com/?query=%40XDevelopers&widget=Timeline) doesn't have a long period of silence in NVDA... then went digging around to figure out why.

It appears to be related to the ``data-height`` attribute and a buggy Firefox polyfill...
* When ``data-height`` is present on a verified account's timeline:
  * Twitter's external widget script adds a ``maxHeight=500px`` param to the ``iframe``'s ``src`` attribute (makes sense)
  * ``maxHeight=500px`` causes a polyfill in the timeline widget to make the tweets' parent ``div`` keyboard-focusable in Firefox (but not Chromium - it doesn't use a polyfill)...
    * When the parent ``div`` gains focus:
      * A ``data-focusvisible-polyfill="true"`` attribute gets added to the ``div``
      * A partial focus outline appears on the ``div`` (looks like an outline covering the bottom of the "Tweets from @" bar...)
      * Since the ``div`` contains all 100 tweets... screen readers attempt to announce everything (not ideal)
      * NVDA goes completely silent and stops responding for ~30 seconds upon encountering the ``div``... probably because NVDA is trying to parse through countless elements before its narrator announces anything.
	  * NVDA's stall doesn't seem to be caused by Firefox... I can still tab forward/backward properly while NVDA is silent. NVDA also stalls when using Chromium if I manually add ``tabindex="0"`` to the ``div`` and tab to it.
* When ``data-height`` **isn't** present:
  * The ``div`` doesn't become focusable in Firefox - which bypasses NVDA's stalling issue

As for how best to deal with this situation...
1. I'll probably open an NVDA issue about that screen reader's silence/stalling behaviour
2. I don't believe there's any way to bring the widget's flaw to X/Twitter's attention ([their accessibility teams were laid-off in late 2022](https://liamodell.com/2022/11/05/twitter-blue-tick-layoffs-redundancies-revenue-accessibility-elon-musk-fires-sacks/) and [``@XA11y``](https://twitter.com/XA11y) has been inactive since)
3. On WET's side of things, we can pick between the following options:
   * **a)** Keep things as-is (it's technically Twitter's fault)
   * **b)** Revise the timeline plugin to handle ``data-height``/``data-width`` in a completely different way:
     * Don't pass on the attributes to Twitter's widget
     * Make WET use the attributes to adjust the timeline ``iframe`` and/or its parent ``div``'s dimensions (won't look as nice since the "Tweets by @" bar will become scrollable instead of "fixed" - unless we instruct the widget to exclude the bar and create a "fake" bar before the ``iframe``)
   * **c)** Scrap the timeline plugin altogether

@duboisp Thoughts?

EricDunsworth:
Just reported the NVDA issue in nvaccess/nvda#16233.

Garneauma:

Garneauma:
The skip links work well in Chrome. However, I noticed something on Safari. The timeline simply is not showing up. The loader is there on page load, then the content is updated, but nothing shows up. Here's a screenshot:
Screen Shot 2024-03-04 at 11 09 23 AM

EricDunsworth:
@Garneauma Are you running into any JS console errors in Safari?

Does the issue exist in the [pre-existing demo page](https://wet-boew.github.io/v4.0-ci/unmin/demos/twitter/twitter-en.html)? If so, what about [Twitter Publish's ``@XDevelopers`` sample](https://publish.twitter.com/?query=%40XDevelopers&widget=Timeline)?

Garneauma:
@EricDunsworth The timeline shows up on Twitter's example.

EricDunsworth:
@Garneauma Thanks for the heads up :)! I kind of wish I had Safari to test against :P.

Based on your screenshot, it looks like the ``iframe``s are loading-in. None of the WET plugin's logic attempts to hide the ``iframe`` or set minimum heights (apart from 100x100 pixels for the loading icon). Some SCSS does temporarily set the ``iframe``'s ``min-width`` property to 0 in a ``:has()`` selector though...

**Quick questions:**
1. Could you try letting this PR's demo page sit there for like 1-2 minutes in case it's loading really slowly?
2. Is ``min-width: 0;`` still set on the ``iframe`` after the loading icon disappears and the giant empty space appears?
3. Is the ``iframe``'s DOM populated? If not, could you check to see if maybe Safari is blocking it in some way (maybe it's getting treated as a third party tracker of sorts when loaded in WET's site)?

Garneauma:
The DOM is in fact populated. I couldn't figure out why it was blank. I will try and let it sit for a while to see if it loads up at some point.

src/plugins/twitter/twitter.js
```diff
@@ -40,37 +41,93 @@ var componentName = "wb-twitter",
...
+							mutation.removedNodes.forEach( function( removedNode ) {
+
+								// If the removed node was a Twitter link, remove its adjacent loading icon, add skip links and stop observing
+								// Note: Twitter's widget script removes "a.twitter-timeline" upon displaying the timeline iframe's content... at which point the loading icon is no longer useful
+								if ( removedNode === twitterLink && mutation.nextSibling === loadingDiv ) {
+									const iframeContainer = loadingDiv.previousElementSibling;
+
+									loadingDiv.remove();
+									addSkipLinks( iframeContainer );
+									observer.disconnect();
```
Garneauma, line 118: Add the following lines to force Safari to display the iframe when it is updated:

```suggestion
									

// The following 2 lines were added as a workaround in Safari where the iFrame is not displayed
$( ".wb-twitter" ).css( "opacity", 1 );
$( ".wb-twitter" ).css( "opacity", "" );
									
observer.disconnect();
```

Garneauma (changes requested):

EricDunsworth:
@Garneauma Just did the following:
* Slightly revised your Safari workaround commit (fixed tabs/trailing spaces, revised commit message, added your committer info)
* Polished your workaround to use pure JS (instead of jQuery) and double-quotes (instead of single quotes) in order to better blend in with the rest of this PR

Garneauma (dismissed):

Garneauma (dismissed):

Garneauma:
@EricDunsworth Please add working examples for all of the plugin options available.

EricDunsworth:
@Garneauma @duboisp Just snuck in an extra examples page... should cover all other scenarios related to this PR's logic changes.

EricDunsworth:
Just snuck in a few more updates to...
* Fix the extra examples page's linting issues
* Change one of the extra page's examples to ``@XDevelopers`` to avoid having redundant ``@Canada`` timelines/skip links
* Bump modified dates to today

Garneauma (approved):

PR #9708, opened 2023-12-04
Form validation: Translate maxWords demo's French label

EricDunsworth:
#6952 previously translated the code sample's label, but missed the demo's counterpart.

This copies the code sample's label content over to the demo.


duboisp (approved):
Review and this is a content change which fix a translation isssue

Issue #8500, opened 2018-10-25
Tables: Filtering on an exact match


RobJohnston:
I'm following the [DataTables filtering example](http://wet-boew.github.io/v4.0-ci/demos/tables/tables-en.html), but in my case, a select list used as a filter has a numeric for a value and a string for the text, as is typical for database work.  I'll try and extend the working example to make my issue clear. 

Image a filter has values like this for column 7:
```html
<select class="form-control" id="dt_subject" name="dt_subject" data-column="7">
<option value="">All Subjects</option>
...
<option value="6">Agriculture</option>
<option value="16">Transport</option>
<option value="26">Military</option>
...
</select>
```
And the table is defined like this, with the ID at position 7 and hidden, the text at position 8 and visible:
```html

<table class="wb-tables table table-striped table-hover" id="dataset-filter" aria-live="polite" data-wb-tables='{
--
  | "bDeferRender": true,
  | "ajaxSource": "ajax/datasource.json",
  | "order": [5, "desc"],
  | "columns": [
  | { "data": "TITLE", "className": "nws-tbl-ttl h4" },
  | { "data": "PUBDATE", "className": "nws-tbl-date" },
  | { "data": "DEPT", "className": "nws-tbl-dept" },
  | { "data": "TYPE", "className": "nws-tbl-type" },
  | { "data": "TEASER",  "className": "nws-tbl-desc" },
  | { "data": "LOCATION",  "visible": false },
  | { "data": "AUDIENCE",  "visible": false },
  | { "data": "SUBJECT_ID", "visible": false },
  | { "data": "SUBJECT", "visible": true },
  | { "data": "MINISTER", "visible": false }
  | ]}'>


```

Currently, when I do a search, for "Agriculture", which has an ID of 6, it shows results that have a "6" anywhere in the ID, so 6, 16, 26, etc.  What I want to to match exactly "6".

Is there an easy way to do this now?  I thought that the answer would be `"search": { "smart": false, "regex": false },`, and then setting the `value` of the select list to strings like "^6$", but that had no effect.  It's what worked for me on the DataTables [Search API (regular expressions)](https://datatables.net/examples/api/regex.html) demo.

Looking at the WET-BOEW code for [tables.js, starting on line 175](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/tables/tables.js#L175), I see the following:

```js
// Lets reset the search;
$datatable.search( "" ).columns().search( "" );

// Lets loop throug all options
[...]

PR #9735, opened 2024-02-19
Tables: adding option to filter with exact match

Garneauma:
Adding exact match option for filters so that the search within a column does not include rows outside the exact value of the filter.

Changes related to WET-452

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -434,8 +434,12 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 
 			// Verifies if regex was preset, if not preset use 'contains value' as default
 			if ( !$regex ) {
-				$value = $value.replace( /\s/g, "\\s*" );
-				$regex = "(" + $value + ")";
+				if ( $elm[ 0 ].getAttribute( "data-exact" ) ) {
+					$regex = "^" + $value + "$";
+				} else {
+					$value = $value.replace( /\s/g, "\\s*" );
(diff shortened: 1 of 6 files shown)


src/plugins/tables/tables.js
```diff
@@ -434,8 +434,12 @@ $document.on( "submit", ".wb-tables-filter", function( event ) {
 
 			// Verifies if regex was preset, if not preset use 'contains value' as default
 			if ( !$regex ) {
-				$value = $value.replace( /\s/g, "\\s*" );
-				$regex = "(" + $value + ")";
+				if ( $elm[ 0 ].getAttribute( "data-exact" ) === "true" ) {
```
duboisp, line 437: Considering the value set on the data attribute don't really matter here because it is simply a flag, let's use[ `hasAttribute( "data-exact" )`](https://dom.spec.whatwg.org/#ref-for-idl-boolean%E2%91%A3%E2%91%A1) instead. If later, the value is somehow use, then it will make senses to check what value the author has specified.

```suggestion
				if ( $elm[ 0 ].hasAttribute( "data-exact" ) ) {
```

duboisp (changes requested):
One little code optimization change.

Tested and do work as expected.

Change:
* Minor configuration change: Add a configuration for making exact filter.

RobJohnston:
Addresses #8500.

duboisp (approved):
Reviewed and tested. do work as expected.

This is a minor change.

PR #9739, opened 2024-02-23
node-sass update

Ricokola:
Update of node-sass to support Node 18 and 20.


package.json
```diff
@@ -87,11 +87,11 @@
     "grunt-sri": "^0.2.0",
     "grunt-string-replace": "^1.3.1",
     "grunt-stylelint": "^0.18.0",
-    "grunt-wet-boew-postbuild": "^0.1.3",
-    "grunt-wget": "~0.1.3",
+    "grunt-wet-boew-postbuild": "git+https://github.com/wet-boew/grunt-wet-boew-postbuild.git#npm20",
+    "grunt-wget": "git+https://github.com/jakejuby/grunt-wget.git#master",
```
duboisp, line 91: Note: We will need to clone it under wet-boew org and use our own adapted copy to ensure stability of our build script

duboisp (changes requested):
See the few inline change request.

Did you successfully built the wet-boew by using NodeJS 18 and NodeJS 20?

PR #9744, opened 2024-03-07
Proximity CSS: Disallow column breaks inside colcount list items

EricDunsworth:
In CSS columns, content "breaks" across columns by default. It helps keep the columns looking balanced. But it looks very strange when multi-line list items break across columns.

This adds a new class (``colcount-no-break``) to prevent it from happening. When used, each list item's content will always "stick together". The list items themselves will still be spread across multiple columns.

Also added a list columns section to the utilities page:
* Used the style guide's content as a starting point
  * Shortened the explanation
* Revised the unordered list example to:
  * Deliberately show multi-line list items splitting across columns
  * Demonstrate combinations of breakpoint classes
* Added a similar description list example
  * Uses ``div`` elements out of necessity (``dt`` and ``dd`` need a common parent to prevent them from getting "separated" across columns)

**CC:**
* @cfarquharson (added you as a co-committer since I borrowed [some of your old content/examples from the WET style guide](https://wet-boew.github.io/wet-boew-styleguide/design/lists-en.html#columns))
* @andrewnordlund (FYI)

### DIFF
--- a/src/base/proximity/_base.scss
+++ b/src/base/proximity/_base.scss
@@ -255,6 +255,31 @@
 	}
 }
 
+/* Disallow column breaks inside list column items */
+%proximity-colcount-no-break {
+	break-inside: avoid-column;
+}
+
+.colcount-no-break {
+	> {
+		dd,
+		dt,
+		li {
+			@extend %proximity-colcount-no-break;
+		}
+	}
+}
+
+dl {
+	&.colcount-no-break {
+		> {
+			div {
+				@extend %proximity-colcount-no-break;
+			}
+		}
+	}
+}
+
 .colcount-xxs-2 {
 	column-count: 2;
 }
--- a/src/other/utility/utility-en.hbs
+++ b/src/other/utility/utility-en.hbs
@@ -7,7 +7,7 @@
 	"tag": "utility",
 	"parentdir": "utility",
 	"altLangPrefix": "utility",
-	"dateModified": "2023-08-15"
+	"dateModified": "2024-03-25"
 }
 ---
 
@@ -25,6 +25,7 @@
 	<li><a href="#spacing">Spacing</a></li>
 	<li><a href="#sizing">Sizing</a></li>
 	<li><a href="#flexbox">Flexbox</a></li>
+	<li><a href="#lists">Lists</a></li>
 	<li><a href="#positioning">Positioning</a></li>
 	<li><a href="#tables">Tables</a></li>
 	<li><a href="#miscellaneous">Miscellaneous</a></li>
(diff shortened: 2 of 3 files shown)


EricDunsworth:
Based on further discussions:
* List item content only splits across columns by default in Safari (WebKit) and Firefox (Gecko):
  * WebKit/Gecko's rendering behaviour has been like that since ~2010 according to this old blog post ([Deal-breaker problems with CSS3 multi-columns](https://web.archive.org/web/20200205032832/http://zomigi.com/blog/deal-breaker-problems-with-css3-multi-columns/) - "Breaking content across columns" section)
  * Chromium (Blink) **_doesn't_** split list items across columns by default... so it seems to be the outlier
  * Need to figure out why Blink behaves differently and what the spec says about it (if anything)
* Unsure if there are any browser differences with `dt`/`dd` elements getting "separated" across columns
* [GCWeb's lists demo page](https://wet-boew.github.io/GCWeb/common/list/lists-en.html) already contains some ``colcount-*-*`` examples derived from the style guide... although I think this PR's WET utilities section would still be worthwhile to add since it has some goodies going for it (e.g. explanation, more fleshed-out examples and a description list example)

EricDunsworth:
After further testing, it looks like Blink actually does behave similarly to Gecko/WebKit for this.

Blink **_does_** break list item content across columns by default, but not at exactly the same spots as Gecko (not sure about WebKit). For example, adding one extra line to a list item at the end of a column might be enough to trigger a break in Gecko, but adding several _will_ ultimately do it in Blink too.

I think the reason for all the confusion in the technical review meeting was because when @Garneauma initially added extra lines to induce a break in [GCWeb's lists page](https://wet-boew.github.io/GCWeb/common/list/lists-en.html), he only added 1-2 lines... which was enough for Gecko, but not for Blink - leaving everyone with the wrong impression 😆.

Also didn't help that I only tested this PR in Firefox before sending it out (the new unordered list example only breaks in Gecko by default without ``break-inside: avoid-column;``).

Anyway, I wasn't expecting this PR to be so contentious. IMO, as a user, list items breaking across columns looks really strange and arguably impedes usability. I can't think of a scenario where anyone would actually find that behaviour to be desirable. But since the change turned out to be so contentious, I guess I'll revise the PR to use an extra class to trigger ``break-inside: avoid-column;``.

EricDunsworth:
Just force-pushed a handful of updates:
* Revised the unordered list examples to show off list items "breaking" in both Blink and Gecko
* Diverted the ``break-inside: avoid-column;`` property into a new class (``colcount-break-inside-avoid-column``)
* Added a section and examples for the new class
* Polished-up French content slightly:
  * Renamed "columns" to "colonnes"
  * Added non-breaking spaces before colons
  * Renamed "item" to "élément")
* Updated modified dates

src/base/proximity/_base.scss
```diff
@@ -255,6 +255,31 @@
 	}
 }
 
+/* Disallow column breaks inside list column items */
+%proximity-colcount-break-inside-avoid-column {
+	break-inside: avoid-column;
+}
+
+.colcount-break-inside-avoid-column {
```
EricDunsworth, line 263: I'd be open to better class name suggestions... :).

Couldn't think of anything short and sweet... so I just rehashed the property's name for now.

duboisp: As discussed during the technical review, our consensus is

`colcount-no-break`

duboisp (dismissed):
As discussed during the technical review, our consensus for the new class name is:

`colcount-no-break`

---

Once this change is completed, this is pre-approved upon the completion of a final review and local testing. 

This is a Minor change. and we will need to update our list working example in GCWeb.

EricDunsworth:
@duboisp This should be ready for a final review now :).

* Revised the class name from ``colcount-break-inside-avoid-column`` to ``colcount-no-break``
* Changed modified dates to today's date
* Rebased

src/other/utility/utility-en.hbs
```diff
@@ -413,6 +414,189 @@
...
+		&lt;dd&gt;Description&lt;br&gt;Line 2&lt;/dd&gt;
+	&lt;/div&gt;
+	&lt;div&gt;
+		&lt;dt&gt;Term 6&lt;/dt&gt;
+		&lt;dd&gt;Description&lt;/dd&gt;
+	&lt;/div&gt;
+&lt;/dl&gt;</code></pre>
+
+<h3><code>colcount-no-break</code></h3>
+<p>Prevents list items from splitting across multople list columns.</p>
```
Garneauma, line 520:
```suggestion
<p>Prevents list items from splitting across multiple list columns.</p>
```
Small typo.

Garneauma (changes requested):
@EricDunsworth There is a small typo. Outside of that, everything looks good.

src/base/proximity/_base.scss
EricDunsworth: Marking this as resolved since the class was renamed as per this suggestion.

Garneauma (approved):

PR #9745, opened 2024-03-19
bgimg: Adding option to have no background image in responsive format

Garneauma:
Adds the option to define no background image for a specific breakpoint which is rather useful when you want no background image on smaller screens.


duboisp (changes requested):
Tested and it do work as expected.

But instead of using/creating a special word "none" let use an URI that identify there is no image. We can use vocab repository in wet-boew for that. We will only need to create a base namespace along with the accessibility conformance report vocabulary

duboisp (approved):
Reviewed and tested locally, everything work as expected

This is a minor change.

duboisp:
The srcset are conform to : https://html.spec.whatwg.org/multipage/images.html#image-candidate-string

Issue #9293, opened 2022-02-28
Creating tables dynamically using jQuery does not work in WET v4.0.44+


adam2:
Hi,

I've found a bug in WET v4.0.44+.  I initially reported this bug to the CDTS group, but I can reproduce it using standard WET files as well.
https://github.com/wet-boew/cdts-sgdc/issues/434

The issue is that when using jQuery to dynamically create an HTML table, it doesn't work in WET 4.0.44+. Some "tr" and "td" elements you would expect to be added are not being added. It's as if the jQuery append function isn't working properly, or something is stripping those elements from the DOM after they are added.  It is difficult to explain, so I made a demo page which make the problem more obvious. Please see the attached HTML file.  The page is based on the WET documentation index page which is using v4.0.47.2.  This bug did not exist in WET 4.0.43.

Please let me know if you have any questions.

Thanks,
Adam
[wet-jquery-tables-bug.html.txt](https://github.com/wet-boew/wet-boew/files/8154557/wet-jquery-tables-bug.html.txt)

duboisp:
Yes and that is intentional.

As documented in the [release note of v4.0.44](https://github.com/wet-boew/wet-boew/releases/tag/v4.0.44.5), in the section "Special notes": `<tr>`, `<td>` and other tags need to be inserted by using the javascript DOM interface. And you can read in the subsequent bullet that all jQuery DOM manipulation are sanitized with [DOMPurify](https://github.com/cure53/DOMPurify).

Since v4.0.44 we patched directly jQuery which has improved the security not just of wet-boew but also to any custom javascript code that leverage jQuery.

If you want to discuss further, please join us at one of our weekly live support from 13h to 16h Estern time every Tuesday afternoon. You will find the meeting link here: https://github.com/wet-boew/wet-boew/wiki/WET-BOEW-Code-sprint

Cheers

adam2:
So you've intentionally broken jQuery and labeled it as "wontfix"?  Thanks for your help.

Now we are forced to fix thousands of lines of code or not upgrade WET to 4.0.44.

duboisp:
@adam2 it was a security patch to prevent XSS attack.

If you want, we can discuss more about your use case during one of our weekly Tuesday afternoon live support. Or alternatively you can find my contact information in GCDirectory and we will coordinate a meeting to discuss further.

duboisp:
@adam2 I didn't saw you at our weekly office hours 😞.

[...]

PR #9746, opened 2024-03-19
Update jquery-fix.js for DataTable allowed tags

mercury64:
DataTable allowed Tag array has html tags that are non-existant.

This updates the datatable tags to what they should be.

Specifically allows dynamically adding rows to a DataTable.

Current broken jquery-fix.js file:
```
...
	dataTableAllowedTag = [
		"<tbody/>",
		"<tr/>",
		"<td />",
		"<td/>"
	],
...
```

Proposed fix:
```
	dataTableAllowedTag = [
		"<tbody>", "<tr>", "<td>", "<td>",
		"</tbody>", "</tr>", "</td>", "</td>"
	],
```

https://github.com/wet-boew/wet-boew/issues/9293
https://github.com/wet-boew/wet-boew/issues/9610

Original commit that caused this issue:
https://github.com/wet-boew/wet-boew/commit/581e1652b8f9130f27a4d8a32fd8d47d651d5a32

### DIFF
--- a/src/core/dep/jquery-fix.js
+++ b/src/core/dep/jquery-fix.js
@@ -171,10 +171,8 @@ var localParseHTML = jQuery.parseHTML,
 	replaceWith = jQuery.fn.replaceWith,
 	jqInit = jQuery.fn.init,
 	dataTableAllowedTag = [
-		"<tbody/>",
-		"<tr/>",
-		"<td />",
-		"<td/>"
+		"<tbody>", "<tr>", "<td>", "<td>",
+		"</tbody>", "</tr>", "</td>", "</td>"
 	],
 	sanitize = function( html ) {
 


pearl-nguyen:
This proposed fix is working for me. Please review. Thanks

pearl-nguyen:
Can we also update jquery-fix to fix the following case:

$("#tbody").append('<tr><td>a</td><td>a</td><td>a</td><td>a</td><td>a</td></tr>')

Change to dataTableAllowedTag does not help to fix this case.

This proposed solution https://github.com/wet-boew/wet-boew/issues/9293#issuecomment-1139236566 fixes this issue

`<script>
(function(){
	const jQueryOriginalFunctions = {
		append:      jQuery.fn.append, 
		prepend:     jQuery.fn.prepend, 
		before:      jQuery.fn.before, 
		after:       jQuery.fn.after, 
		replaceWith: jQuery.fn.replaceWith, 
		init:        jQuery.fn.init,
		html:        jQuery.html
	};
	// Prevent WET-BOEW for overriding these functions without causing a TypeError in the WET-BOEW code
	Object.defineProperty(jQuery.fn, "append", {get:function(){return jQueryOriginalFunctions.append;}, set:function(){}});
	Object.defineProperty(jQuery.fn, "prepend", {get:function(){return jQueryOriginalFunctions.prepend;}, set:function(){}});
	Object.defineProperty(jQuery.fn, "before", {get:function(){return jQueryOriginalFunctions.before;}, set:function(){}});
	Object.defineProperty(jQuery.fn, "after", {get:function(){return jQueryOriginalFunctions.after;}, set:function(){}});
	Object.defineProperty(jQuery.fn, "replaceWith", {get:function(){return jQueryOriginalFunctions.replaceWith;}, set:function(){}});
	Object.defineProperty(jQuery, "html", {get:function(){return jQueryOriginalFunctions.html;}, set:function(){}});
})();
</script>`

duboisp:
@Garneauma can you review, reproduce the issue and test is change request are fixing it. Thanks

Garneauma:
@pearl-nguyen I have just tested your fix and unfortunately it doesn't seem to fix the issue.

Question: what is the intention behind adding that row? Is it simply to add a row to your datatable table? If it is, datatable has a simple way to do so: https://datatables.net/reference/api/row.add()

If that is not the case, could you please explain further what is your intention with that append.

Unfortunately, according to this issue: https://github.com/cure53/DOMPurify/issues/324, it seems that the issue with the append is that the browser does not allow rows to be inserted outside of a table element. It is not DOMPurify itself that does not allow it.

Garneauma:
@pearl-nguyen Do you have any update on this?

pearl-nguyen:
@Garneauma, we are in the process of rewriting our app. Therefore, we do not want to spend resources on fixing existing applications. Recent version of wet broke the javascript in our applications that require changes in many places. The propsed soluton help to fix the issue. thanks

PR #9757, opened 2024-04-29
Core: edit DOMPurify sanitization to allow external links

Garneauma:
External links are safe to include in web pages given that they also have a `rel` attribute containing the value `"noreferrer"`. See the following link for a more detailed explanation: https://owasp.org/www-community/attacks/Reverse_Tabnabbing

### DIFF
--- a/src/core/dep/jquery-fix.js
+++ b/src/core/dep/jquery-fix.js
@@ -163,6 +163,33 @@ jQuery.htmlPrefilter = function( html ) {
  * This implementation leverage DOMPurify for filtering every string prior DOM manipulation by jQuery
  *
  */
+
+// START: add hooks to DOMPurify to allow external links when they meet certain conditions as defined here: https://owasp.org/www-community/attacks/Reverse_Tabnabbing
+DOMPurify.addHook( "beforeSanitizeAttributes", function( node ) {
+
+	// Add "data-wb-external-link" to all <a> with a target="_blank" and rel="noreferrer"
+	if (
+		node.tagName === "A" &&
+		node.getAttribute( "target" ) &&
+		node.getAttribute( "target" ) === "_blank" &&
+		node.getAttribute( "rel" ) &&
+		node.relList.contains( "noreferrer" )
+	) {
+		node.setAttribute( "data-wb-external-link", "true" );
+	}
+} );
+
+DOMPurify.addHook( "afterSanitizeAttributes", function( node ) {
+
+	// Put back the target="_blank" to all <a> with attribute "data-wb-external-link"
+	if ( node.tagName === "A" && node.getAttribute( "data-wb-external-link" ) ) {
+		node.setAttribute( "target", "_blank" );
+		node.removeAttribute( "data-wb-external-link" );
+	}
+} );
+
+// END
+
 var localParseHTML = jQuery.parseHTML,
 	append = jQuery.fn.append,
 	prepend = jQuery.fn.prepend,
--- a/src/core/test.js
+++ b/src/core/test.js
@@ -96,6 +96,36 @@ describe( "wb.core helpers test suite", function() {
 	} );
 
 	/*
+	 * Test external links after DOMPurify sanitization
+	 */
+	describe( "External links", function() {
+		before( function() {
+			$( "body" ).append( `<div class='external-links-test'>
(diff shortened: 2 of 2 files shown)


Garneauma:

src/core/dep/jquery-fix.js
```diff
@@ -163,6 +163,33 @@ jQuery.htmlPrefilter = function( html ) {
...
+// START: add hooks to DOMPurify to allow external links when they meet certain conditions as defined here: https://owasp.org/www-community/attacks/Reverse_Tabnabbing
+DOMPurify.addHook( "beforeSanitizeAttributes", function( node ) {
+
+	// Add "data-external-link" to all <a> with a target="_blank" and rel="noreferrer"
+	if (
+		node.tagName === "A" &&
+		node.getAttribute( "target" ) &&
+		node.getAttribute( "target" ) === "_blank" &&
+		node.getAttribute( "rel" ) &&
+		node.getAttribute( "rel" ).includes( "noreferrer" )
```
duboisp, line 176: Using `.includes()` function is problematic because the following value in the `rel` attribute return true:
* `noreferrers`
* `anoreferrer`
* `snoreferrers` (Add this example as a unit test)

I suggest to do instead and you would not require to check if that attribute is set or not.
```suggestion
		node.relList.contains( "noreferrer" )
```

src/core/dep/jquery-fix.js
```diff
@@ -163,6 +163,33 @@ jQuery.htmlPrefilter = function( html ) {
...
+	) {
+		node.setAttribute( "data-external-link", "true" );
+	}
+} );
+
+DOMPurify.addHook( "afterSanitizeAttributes", function( node ) {
+
+	// Put back the target="_blank" to all <a> with attribute "data-external-link"
+	if ( node.getAttribute( "data-external-link" ) ) {
```
duboisp, line 185: I suggest to pre-check the tagName as this is only applicable to anchor element. Calling a function which need to iterate over each attribute should take more computation time compared to check a static property. That way we reduce the number iteration on non-anchor element.

```suggestion
	if ( node.tagName === "A" && node.getAttribute( "data-wb-external-link" ) ) {
```

src/core/test.js
```diff
@@ -95,6 +95,31 @@ describe( "wb.core helpers test suite", function() {
...
+	/*
+	 * Test external links after DOMPurify sanitization
+	 */
+	describe( "External links", function() {
+		before( function() {
+			$( "body" ).append( `<div class='external-links-test'>
+				<a id='externalLinkTest1' href='http://www.canada.ca' target='_blank' rel='noreferrer'>Valid link 1</a>
+				<a id='externalLinkTest2' href='http://www.canada.ca' target='_blank' rel='noreferrer noopener'>Valid link 2</a>
+				<a id='externalLinkTest3' href='http://www.canada.ca' target='_blank'>Invalid link 1</a>
+			</div>` );
```
duboisp, line 107: Additional unit test case where the result should be undefined

```suggestion
				<a id='externalLinkTest3' href='http://www.canada.ca' target='_blank'>Invalid link 1</a>
				<a id='externalLinkTest4' href='http://www.canada.ca' target='_blank' rel="hellonoreferrerworld>Invalid link 2</a>
				<!-- This one should be valid according to spec because the token are case insensitive but this is a limitation of relList DOM anchor function implemented in browser -->
				<!-- No need to test, but let's keep it here as a documentation item for future reference -->
				<!-- <a id='externalLinkTest5' href='http://www.canada.ca' target='_blank' rel='external noReferrer nofollow'>Valid link 3</a>-->
			</div>` );
```

duboisp (changes requested):
There is an error where it can allow invalid keywork, see the propose change.

There is also a few performance improvement suggestion.

src/core/dep/jquery-fix.js
Garneauma: Merci pour la suggestion!

duboisp (approved):
Reviewed and tested locally. Everything look fine.

Also, I noticed a little performance improvement of the data-ajax working example page compared to my yesterday test. I don't know if that is caused by your latest update or if that is caused by something else running on my test computer.

duboisp:
Note: 
This is a minor change about core and content sanitization feature which allow the possibility to open external links in new windows.

PR #9767, opened 2024-05-22
Twitter embedded timeline: Improve handling of missing i18n vars

EricDunsworth:
#9704's changes had the potential of going into an infinite loop if one of its i18n variables was ``undefined``.

The infinite loop tied into an ``if`` condition in the MutationObserver:
* Its purpose is to replace the Twitter widget's English-only ``iframe`` ``title`` with an i18n version from WET
* It checks for attribute mutations on ``iframe`` elements and whether the ``iframe``'s ``title`` attribute matches the ``i18nText.timelineTitle`` variable
* If the current ``iframe`` ``title`` isn't strictly equal to the i18n variable, the former gets set to the latter (triggering another attribute mutation)

Normally, the ``if`` condition should become truthy in the subsequent attribute mutation.

When the i18n variable wasn't set, this never happened. Why? Because missing i18n variables are treated as ``undefined``. So the ``iframe`` ``title`` got set to "undefined" (implicit conversion of ``undefined`` into a string)... Then the next mutation's ``if`` condition didn't match ("undefined" !== ``undefined``). So the ``if`` condition attempted to set the ``iframe`` ``title`` to "undefined" once again - triggering yet another attribute mutation. This sequence of events repeated forever.

This fixes it by checking whether the i18n variable is a string before changing the ``iframe``'s ``title`` attribute.

Related changes:
* Show a console warning if the plugin's i18n variables appear to be missing
* Don't add skip links if one of its i18n variables isn't a string (prevents console errors, no point in adding the links without content)

### DIFF
--- a/src/plugins/twitter/twitter.js
+++ b/src/plugins/twitter/twitter.js
@@ -58,6 +58,11 @@ var componentName = "wb-twitter",
 					};
 				}
 
+				// Show a warning if the timelineTitle variable isn't a string
+				if ( typeof i18nText.startNotice !== "string" ) {
+					console.warn( componentName + ": i18n text is missing. Iframe title override and skip links will be disabled." );
+				}
+
 				// Set Chinese (Simplfified)'s language code to "zh-cn"
 				// If the link doesn't specify a widget language and its "in-page" language code is "zh-Hans"...
 				// Notes:
@@ -97,8 +102,10 @@ var componentName = "wb-twitter",
 							const mutationTarget = mutation.target;
 
 							// Override the timeline iframe's title right after Twitter's widget script adds it
-							// Note: The timeline's iframe title is English-only, uses "Twitter" and is written in title case ("Twitter Timeline")... This replaces it with an i18n version that uses "X" and is written in sentence case.
-							if ( mutationTarget.nodeName === "IFRAME" && mutationTarget.title !== i18nText.timelineTitle ) {
+							// Notes:
+							// -The timeline's iframe title is English-only, uses "Twitter" and is written in title case ("Twitter Timeline")... This replaces it with an i18n version that uses "X" and is written in sentence case.
+							// -Only proceed if the i18n variable is a string... otherwise this'll trigger an infinite loop of attribute mutations
+							if ( mutationTarget.nodeName === "IFRAME" && mutationTarget.title !== i18nText.timelineTitle && typeof i18nText.timelineTitle === "string" ) {
 								mutationTarget.title = i18nText.timelineTitle;
 							}
 							break;
@@ -161,9 +168,8 @@ var componentName = "wb-twitter",
 		let skipToEndLink;
 		let skipToStartLink;
 
-		// Abort if Twitter username is falsy
-		// Note: Unlikely to happen unless the username doesn't exist... in which case Twitter's third party widget script will have already failed and triggered an exception by this point
-		if ( !username ) {
+		// Abort if Twitter username is falsy or i18n variables aren't strings
+		if ( !username || typeof i18nText.timelineTitle !== "string" ) {
 			return;
 		}
 


src/plugins/twitter/twitter.js
duboisp, line 1: You can verify and fix the value or simply prevent the plugin execution.

```
	i18nText = {
		startNotice: i18n( "twitter-start-notice" ),
		endNotice: i18n( "twitter-end-notice" ),
		skipEnd: i18n( "twitter-skip-end" ),
		skipStart: i18n( "twitter-skip-start" ),
		timelineTitle: i18n( "twitter-timeline-title" ) || "X"
	};

```

Or

```
	i18nText = {
		startNotice: i18n( "twitter-start-notice" ),
		endNotice: i18n( "twitter-end-notice" ),
		skipEnd: i18n( "twitter-skip-end" ),
		skipStart: i18n( "twitter-skip-start" ),
		timelineTitle: i18n( "twitter-timeline-title" )
	};
	if ( typeof i18nText.timelineTitle !== "string" ) {
		console.warn( "Tweeter embed timeline can not start, i18n text is missing" );
		wb.ready( $( eventTarget ), componentName );
		return;
	}
}
```

duboisp (dismissed):
See the suggestion

duboisp:
This is a patch change.

Garneauma:
Pre-approved upon application of change request and successful testing.

src/plugins/twitter/twitter.js
EricDunsworth: I tried revising this PR's logic, but feel like neither of those suggestions would be very ideal in practice.

Here's why:
1. ```js
   timelineTitle: i18n( "twitter-timeline-title" ) || "X"
   ```
   The ``timelineTitle`` variable's purpose is to override a flawed ``iframe`` ``title`` attribute set by Twitter's widget. So if any kind of fallback string were to be needed, I think it'd be best to stick with whatever Twitter itself defined.
   
   But I can't do ``|| (get iframe title from Twitter's widget)`` super early on since the ``iframe`` doesn't exist yet. The mutation that was causing the infinite loop is the earliest point in time at which the ``iframe``'s ``title`` attribute becomes known.

2. ```js
   if ( typeof i18nText.timelineTitle !== "string" ) {
   	console.warn( "Tweeter embed timeline can not start, i18n text is missing" );
   	wb.ready( $( eventTarget ), componentName );
   	return;
   }
   ```
   I don't mind doing something along those lines to quickly show a warning about missing i18n text. I also prefer the way you wrote your code sample's warning compared to what I had for ``TypeError``.
   
   But I don't think it'd be appropriate to abort plugin initialization solely because of missing i18n variables. The i18n text is only used for "enhancements" WET makes to Twitter's widget. The text isn't required for the widget to work in practice.

I'm thinking of going in another direction:
* Immediately show a console warning about missing i18n text (like your second suggestion - without aborting init)
* Keep the i18n ``typeof`` check inside the MutationObserver to avoid the infinite loop (same as what I already have - without ``else``)
* Add an i18n ``typeof`` check before calling ``addSkipLinks()`` to avoid console errors (no point in adding skip links if i18n text is unavailable)

@duboisp Thoughts? Would you be on board with that?

duboisp: @EricDunsworth ideally there is no interference. I let you to apply the best solution. At least now you have a few more options.

> I also prefer the way you wrote your code sample's warning compared to what I had for TypeError.

👍

EricDunsworth: Just implemented what I pitched in the last comment :).

Gonna mark this as resolved now. But feel free to reopen it if you still have any concerns.

@duboisp Btw thanks for challenging what I initially had lol. I feel like the PR better off now vs where it started out 💯.

Garneauma (commented):
@duboisp I have tested the solution and it works well. I will let you also review the code.

@EricDunsworth Thank you!

Garneauma (approved):

PR #9775, opened 2024-05-30
improve lightbox accessibility of close button

### LINKED ISSUE #9412: Lightbox: Modal accessibility issue with close button and screen readers
NVDA and JAWS are currently reading the X close button as "Times, close's overlay (escapte key)" and will repeat "Close's overlay (escape key). I believe that we could fix this problem either by wrapping letter **X** in a "span aria-hidden="true"" or may be by changing the button title attribute, by an aria-label attribute?

radmorecameron:
When using NVDA with the lightbox, it seems to be pretty verbose (eg: the voice over says "Close Close Overlay Close Overlay"). This PR will change it so it only says "Close Overlay". 

I've tested these changes with NVDA running and it seems to be working well

**General checklist**
Make your own list for the purpose of your Pull request.

- [x] Build and test PR's code /// Rouler le script de compilation et tester le code de la PR
- [x] Validate changes against WCAG for accessibility

**Related issues**
closes #9412


Garneauma:

src/plugins/lightbox/lightbox.js
```diff
@@ -21,6 +21,11 @@ var componentName = "wb-lbx",
 	modalHideSelector = "#wb-tphp, body > header, body > main, body > footer",
 	$document = wb.doc,
 	callbacks, i18n, i18nText,
+	defaults = {
+
+		// exclude 'times' from screen reader with aria-hidden span
+		closeMarkup: "<button type='button' class='mfp-close' title='%title%'><span class='mfp-close' aria-hidden='true'>&times;</span></button>"
```
duboisp, line 27: This markup for that button should be revised. Title element should be avoided, use `.wb-inv` as a design pattern like how we do for other similar feature.

src/plugins/lightbox/lightbox.js
```diff
@@ -296,7 +301,7 @@ var componentName = "wb-lbx",
 
 				overlayCloseFtr = "<button type='button' class='btn btn-sm btn-primary pull-left " + closeClassFtr +
 					"' title='" + spanTextFtr + "'>" +
-					closeTextFtr +
+					"<span aria-hidden='true'>" + closeTextFtr + "</span>" +
```
duboisp, line 304:
```suggestion
					"<span class="wb-inv">" + closeTextFtr + "</span>" +
```

src/plugins/lightbox/lightbox.js
```diff
@@ -272,7 +277,7 @@ var componentName = "wb-lbx",
 			complete: function() {
 
 				// Set the dependency i18nText only once
-				$.extend( true, $.magnificPopup.defaults, i18nText );
+				$.extend( true, $.magnificPopup.defaults, i18nText, defaults );
```
duboisp, line 280: Maybe we should allow the author to extend those setting, like how we do for the data table plugin.

duboisp (changes requested):
@Garneauma can you perform a review. Thanks

duboisp:
Note: I will review later this week

src/plugins/lightbox/lightbox.js
duboisp: Hummm,

After looking how those magnificPopup global default setting is used and after a little local test to override the close button setting, this is not the right place as you pointed out. Thanks for this.

So, I recommend to not change this line, but instead update the following one during the magnificPopup plugin initialization here:

https://github.com/wet-boew/wet-boew/blob/e01bc308c1cc123dbe7948cd8dccc14829ed8e70/src/plugins/lightbox/lightbox.js#L93-L102

Where you can set the default like:

```javascript
 // Extend the settings with window[ "wb-lbx" ] then data-wb-lbx 
 settings = $.extend( 
 	true, 
 	defaults, 
 	settings, 
 	window[ componentName ], 
 ); 
 $elm.magnificPopup( 
 	settings 
 ).data( "wbLbxFilter", settings.filter ); 
```

---

For your information, I tried to override the setting by configuring the ligthbox button like:

```html
<a class="wb-lbx" href="ajax/ajax4-en.html" data-wb-lbx='{"filter": ".filter-selector", "closeMarkup": "<button type=\u0027button\u0027 class=\u0027mfp-close\u0027><span class=\u0027mfp-close\u0027 aria-hidden=\u0027true\u0027>&times;</span><span class=wb-inv>%title%</span></button>"}'>AJAX - Content filtered by selector (v4.0.12+)</a>
```

src/plugins/lightbox/lightbox.js
duboisp: Let improve it considering you are currently revising that code. Currently that button fallback on the very last way to provide a label according to the Accessible name and description computation algorithm - https://www.w3.org/TR/accname-1.2/#mapping_additional_nd_name

I propose the following which will do similar as you suggested and still leverage `wb-inv` CSS class:
```suggestion
		closeMarkup: "<button type='button' class='mfp-close'><span class='mfp-close' aria-hidden='true'>&times;</span><span class='wb-inv'>%title%</span></button>"
```

src/plugins/lightbox/lightbox.js
duboisp: Thanks for pointing this @radmorecameron , I didn't realized that it was the close button at the bottom of the overlay. I though it was the top right "X" button.

To make the screen reader less verbose, I recommend to remove the title attribute and to replace the i18n text from "Close overlay" for "overlay" which could end up with this code instead:

```javascript
				overlayCloseFtr = "<button type='button' class='btn btn-sm btn-primary pull-left " + closeClassFtr +
					"'>" +
					closeTextFtr +
					"<span class='wb-inv'>" + spanTextFtr + "</span></button>";
```

Where the i18n text value for `overlay-close` would be modified in this CSV file: https://github.com/wet-boew/wet-boew/blob/master/src/i18n/i18n.csv#L25

Note: If you want to test locally the i18n text change, you will need to modify the following file in your test environment and probably replicate those change for the minified corresponding file:
* https://github.com/wet-boew/wet-boew-dist/blob/v4.0-dist/wet-boew/js/i18n/en.js#L61
* https://github.com/wet-boew/wet-boew-dist/blob/v4.0-dist/wet-boew/js/i18n/fr.js#L61

duboisp (changes requested):
@radmorecameron see the propose change and discussion.

---

cc @Garneauma

duboisp:
I will review the propose change.

src/plugins/lightbox/lightbox.js
duboisp: I see, I retested and I can observe that the close button is not added for the "Gallery" lightbox style.

Having a way to configure that was more a nice to have.

src/i18n/base.js
```diff
@@ -58,7 +58,7 @@ wb.i18nDict = {
 	"srch-menus": "@srch-menus@",
 	email: "@email@",
 	"menu-close": "@menu-close@",
-	"overlay-close": "@overlay-close@",
+	"overlay": "@overlay@",
```
duboisp, line 61: To avoid a breaking change, I suggest to keep the same key.

```suggestion
	"overlay-close": "@overlay-close@",
```

src/i18n/i18n.csv
```diff
@@ -22,7 +22,7 @@ General (i18n),all,All,Tous,Alle,Të gjithë,جميع,Բոլորը,Всички,
...
+,overlay,overlay,la fenêtre superposée,Maak die oortrek,Mbylle mbulesë,إغلاق تراكب,Փակեք կափարիչ,Затворете наслагването,關閉疊加,关闭叠加,Zavřete se šablonou,Sluit het dialoogvenster,Sulge pop-up,Schließen Sie das Pop-up,Κλείστε το αναδυόμενο παράθυρο,पॉपअप बंद,Zárja be az overlay,Loka yfirlagið,Tutup hamparan,Chiudere la sovrapposizione,ᖃᓕᕇᑦ ᒪᑐᓕ,オーバーレイを閉じる,오버레이에게 닫습니다,Aizveriet pārklāt,Uždaryti perdangos,Zamknij nakładkę,Feche a sobreposição,Feche a sobreposição,Закройте вспомогательных линий,Zavrite so šablónou,Cierre la superposición,ปิดการแสดงข้อมูล,Bindirme kapatmak,Закрийте допоміжних ліній,Đóng lớp phủ
```
duboisp, line 25: There is no issue with the content change. 

However, let keep the same i18n key

```suggestion
,overlay-close,overlay,la fenêtre superposée,Maak die oortrek,Mbylle mbulesë,إغلاق تراكب,Փակեք կափարիչ,Затворете наслагването,關閉疊加,关闭叠加,Zavřete se šablonou,Sluit het dialoogvenster,Sulge pop-up,Schließen Sie das Pop-up,Κλείστε το αναδυόμενο παράθυρο,पॉपअप बंद,Zárja be az overlay,Loka yfirlagið,Tutup hamparan,Chiudere la sovrapposizione,ᖃᓕᕇᑦ ᒪᑐᓕ,オーバーレイを閉じる,오버레이에게 닫습니다,Aizveriet pārklāt,Uždaryti perdangos,Zamknij nakładkę,Feche a sobreposição,Feche a sobreposição,Закройте вспомогательных линий,Zavrite so šablónou,Cierre la superposición,ปิดการแสดงข้อมูล,Bindirme kapatmak,Закрийте допоміжних ліній,Đóng lớp phủ
```

Note: I didn't reviewed the other translation.

src/plugins/lightbox/lightbox.js
```diff
@@ -121,8 +126,8 @@ var componentName = "wb-lbx",
 			i18n = wb.i18n;
 			i18nText = {
 				close: i18n( "close" ),
-				oClose: i18n( "overlay-close" ),
-				tClose: i18n( "overlay-close" ) + i18n( "space" ) + i18n( "esc-key" ),
+				oClose: i18n( "overlay" ),
+				tClose: i18n( "close" ) + i18n( "space" ) + i18n( "overlay" ) + i18n( "space" ) + i18n( "esc-key" ),
```
duboisp, line 130:
```suggestion
				tClose: i18n( "close" ) + i18n( "space" ) + i18n( "overlay-close" ) + i18n( "space" ) + i18n( "esc-key" ),
```

duboisp (dismissed):
I reviewed and tested. This does work as expected. Everything is good except about i18n change.

We will just need to adjust the i18n change in a way that it don't create a breaking change for the situation where the i18n files are not fully in sync with the plugin js. That would result in having the lightbox plugin to miss some important text (or error) and that would create a more serious accessibility issue.

To prevent that, I propose to keep the same i18n key `overlay-close`. Even if the word "Close" is removed from its assigned text value. IMHO, its purpose does remain the same, which is being part of the text content for the close button in a lightbox dialog.

src/plugins/lightbox/lightbox.js
```diff
@@ -121,8 +126,8 @@ var componentName = "wb-lbx",
 			i18n = wb.i18n;
 			i18nText = {
 				close: i18n( "close" ),
-				oClose: i18n( "overlay-close" ),
-				tClose: i18n( "overlay-close" ) + i18n( "space" ) + i18n( "esc-key" ),
+				oClose: i18n( "overlay" ),
```
radmorecameron, line 129:
```suggestion
				oClose: i18n( "overlay-close" ),
```

radmorecameron:
Updated and rebased @duboisp

duboisp (commented):
Looks good, we will be able to approve and merge this in our next release once we re-confirmed if everything work as expected considering the last change. @Garneauma can you do that last functional test. Thanks.

Garneauma (approved):

PR #9776, opened 2024-06-04
Core: Ensure wb.getId() always returns unique IDs

EricDunsworth:
WET core's ``getId()`` method was initially designed to increment a seed number in conjunction with a WET-specific auto ID naming scheme ("wb-auto-#").

That's fine in theory, but doesn't guarantee that it'll return a unique ID. It can clash with hardcoded IDs that use the same naming scheme - such as pages whose authors mistakenly-hardcode "post-JS" HTML markup generated by plugins.

For example, if a "botched" plugin implementation contains a hardcoded "wb-auto-5" ID... ``wb.getId()`` might independently generate the same ID and return it to a "good" plugin somewhere else in the page. End result is that the "good" plugin will insert a duplicate ID into the page and possibly even break.

This resolves the potential for clashes by revising ``getId()`` to:
* Check for conflicting hardcoded IDs the first time it's called
* Increment its initial seed number if needed (to surpass the highest conflicting ID's number)
* Show console errors for any hardcoded IDs starting with WET's auto ID prefix (``wb-auto-``)

Also adds a WET core working examples page to demonstrate an ID conflict scenario in action.

Thanks to @fsnoddy for initially spotting this and providing insight.

### DIFF
--- a/src/core/wb.js
+++ b/src/core/wb.js
@@ -188,7 +188,32 @@ var getUrlParts = function( url ) {
 		},
 
 		getId: function() {
-			return "wb-auto-" + ( seed += 1 );
+			var idPrefix = "wb-auto-",
+				ids,
+				numberCandidate,
+				numbers = [];
+
+			// Check for conflicting hardcoded IDs the first time an ID is requested
+			if ( !seed ) {
+				ids = document.querySelectorAll( "[id^='" + idPrefix + "']" );
+
+				// Loop through elements whose IDs begin with the prefix
+				ids.forEach( function( currentElm ) {
+					numberCandidate = currentElm.id.substring( idPrefix.length );
+
+					// Verify whether the ID ends with a conflicting number and add it to an array
+					if ( numberCandidate.search( /^\d+$/ ) !== -1 ) {
+						numbers.push( numberCandidate );
+					}
+
+					console.error( "wb.getId: ID '" + currentElm.id + "' isn't supposed to be hardcoded in the page. Please remove it or change its prefix to something different than '" + idPrefix + "'." );
+				} );
+
+				// Set the seed to the array's highest number (will be incremented later)
+				seed = numbers.length ? Math.max.apply( null, numbers ) : seed;
+			}
+
+			return idPrefix + ( seed += 1 );
 		},
 
 		init: function( event, componentName, selector, noAutoId ) {
--- a/site/pages/docs/ref/core/core-en.hbs
+++ b/site/pages/docs/ref/core/core-en.hbs
@@ -6,7 +6,7 @@
 	"categoryfile": "other",
 	"description": "Core WET object.",
 	"altLangPrefix": "core",
-	"dateModified": "2023-03-28"
+	"dateModified": "2024-07-02"
 }
 ---
 <div class="wb-prettify all-pre hide"></div>
@@ -19,6 +19,14 @@
 <section>
 	<h2>Purpose</h2>
 	<p>Core WET object.</p>
+</section>
+
+<section>
+	<h2>Working example</h2>
+		<ul>
+			<li><a href="../../../demos/core/core-en.html">English examples</a></li>
+			<li><a href="../../../demos/core/core-fr.html">French examples</a></li>
+		</ul>
 </section>
 
 <section>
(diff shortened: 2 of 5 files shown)


src/core/wb.js
```diff
@@ -188,7 +188,20 @@ var getUrlParts = function( url ) {
 		},
 
 		getId: function() {
-			return "wb-auto-" + ( seed += 1 );
+			var initialSeed = seed,
+				candidateId;
```
EricDunsworth, line 192: Wanted to use ``const`` and ``let`` here... but fell back on ``var`` to lessen the risk of issues in really old browsers.

``getId()`` is called in basic HTML mode and obsolete versions of IE are forced into it...

EricDunsworth:
Example demonstrating hardcoded ``wb-auto-id`` IDs preventing the tables (DataTables) plugin from running:
```html
<div id="wb-auto-1"></div>
<div id="wb-auto-2"></div>
<div id="wb-auto-3"></div>
<div id="wb-auto-4"></div>
<div id="wb-auto-5"></div>

<table class="wb-tables table">
	<thead>
		<tr>
			<th>Rendering engine</th>
			<th>Browser</th>
		</tr>
	</thead>
	<tbody>
		<tr>
			<td>Trident</td>
			<td>Internet Explorer 4.0</td>
		</tr>
	</tbody>
</table>
```

Garneauma:
Pre-approved upon successful review and internal discussions.

Garneauma:
As discussed, please adjust the PR to check for existing wb ID's only once. Example selector: `document.querySelectorAll( '[id*="wb-auto-"]' )`. If items are returned, log an error in the console like so: `console.error( "Hard coded wet-boew ID for the following element and it must be removed" );`, and log the element in question.

duboisp (changes requested):
Je propose une approche différente où qu'on devrait faire seulement une seule vérification lorsque la fonction est appelé.

Par example, tu peux voir tout les ID problématique avec `var elmWithBadIDs = document.querySelectorAll( '[id*="wb-auto-"]' )`

Après logger une erreur: console.error( "Duplicate ID probable - Hard coded wet-boew ID for the following element and it must be removed" );
Logger les éléments problématiques: console.log( elmWithBadIDs );

EricDunsworth:
Just force-pushed a new approach based on the review comments :).

It'll now check the document ``object`` only once, the very first time ``wb.getId()`` gets called (based on whether ``seed`` is truthy). If ``querySelectorAll()`` finds any hits for ``[id^='wb-auto-']``, it'll figure out which of the IDs use numbers (i.e. conflict candidates), add them into a ``numbers`` array and complain in the console. Then it'll ultimately set the ``seed`` to the highest conflicting ID's number and increment it. That way, WET will increase the baseline number of its auto IDs to avoid conflicts.

Btw this logic should be able to work in IE9+ (although IE6+ are forced into basic HTML mode...).

Also, I went with a console warning instead of an error for the following reasons:
* My logic can overcome auto ID conflicts... so the conflicts aren't a show-stopper (nothing will literally break if a conflict is encountered).
* None of WET's documentation currently forbids hardcoding IDs prefixed with ``wb-auto-``... I didn't feel like it'd be appropriate to treat a violation of an undocumented expectation as an error.

As far as the console message go, I felt like it'd be better to spam console warnings for each conflicting ID as opposed to grouping them all together into a single warning. That way, the spam warnings will make a bigger deal out of the issue.

duboisp (changes requested):
The code look good.

Can you create a test page or a unit test?

Like a new page here: https://github.com/wet-boew/wet-boew/tree/master/src/other/utility

EricDunsworth:
* Changed ``console.error()`` to ``console.warn()`` (latter was my real intent)
* Added a WET core working examples page demonstrating an ID conflict scenario
  * @duboisp I was originally going to go with your suggestion to sneak something into the utilities folder... but while prepping it I couldn't help but feel like it just didn't "fit in" no matter how I looked at it :S. Since [WET core already has a docs page](https://wet-boew.github.io/v4.0-ci/unmin/docs/ref/core/core-en.html), I ultimately decided to create a corresponding working examples page and place an ID conflict demo in there. Similar to how the [tag filter plugin's edge cases page](https://wet-boew.github.io/v4.0-ci/unmin/demos/tagfilter/tagfilter-edge-cases-en.html) is setup.

**PS:**
I don't currently have a working local build system. So I haven't been able to verify that the new pages are building/linking-up properly. Everything _looks_ alright to me though.

EricDunsworth:
Just force-pushed a couple more updates:
* Changed ``console.warn()`` back to ``console.error()`` (as requested)

duboisp (approved):
Reviewed and tested. Everything work as expected and it is demoed.

PR #9777, opened 2024-06-17
Tag filter: Fix exception when used with filter plugin

EricDunsworth:
When the filter and tag filter plugins were used together, typing anything into the text filter field caused the tag filter plugin to raise TypeError exceptions. Although it didn't noticeably break anything.

The exceptions were caused by some logic introduced in #9624 to fix an incompatibility between the tag filter and pagination plugins. An ``if`` condition was added to evaluate the source property of the "wb-contentupdated" event's ``data`` parameter whenever filter controls were toggled... except ``data`` doesn't always exist. Only the tag filter and data JSON plugins make use of ``data`` parameters. The filter plugin doesn't.

This fixes the issue by adjusting the ``if`` condition to ensure that the ``data`` parameter is truthy.

Also revises the edge case examples (to demonstrate that the logic fix doesn't interfere with text filter + pagination scenarios):
* Adds a text filter to example 1
* Adds a second example (based on the normal demo page's second example) with a text filter and pagination

### DIFF
--- a/src/plugins/tagfilter/tagfilter-edge-cases-fr.hbs
+++ b/src/plugins/tagfilter/tagfilter-edge-cases-fr.hbs
@@ -6,7 +6,7 @@
 "tag": "filter",
 "parentdir": "tagfilter",
 "altLangPrefix": "tagfilter-edge-cases",
-"dateModified": "2023-07-10"
+"dateModified": "2024-07-03"
 }
 ---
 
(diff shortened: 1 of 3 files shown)


Garneauma:

src/plugins/tagfilter/tagfilter-edge-cases-fr.hbs
```diff
@@ -222,20 +249,576 @@
...
+			&lt;div class="wb-tagfilter-noresult"&gt;
+				&lt;p&gt;Aucun item ne correspond à cette combinaison de filtres.&lt;/p&gt;
+			&lt;/div&gt;
+			&lt;div id="filteredEventsListPagination"&gt;&lt;/div&gt;
+		&lt;/div&gt;
+	&lt;/div&gt;
+&lt;/section&gt;</code></pre>
+</details>
+
+FR EXAMPLE 2 UPDATED:
```
Garneauma, line 270: @EricDunsworth Is there a need to have this?

Garneauma (dismissed):
Only one small text edit.

src/plugins/tagfilter/tagfilter-edge-cases-fr.hbs
EricDunsworth: Nope, will remove it in a moment. Btw thanks for catching it :)!

It was probably some filler content I added while prepping the new example locally... only to overlook it later on :P.

EricDunsworth:
Just force-pushed a few minor updates:
* Removed "FR EXAMPLE 2 UPDATED:" from the new French edge case's source code
* Made the new edge case's English/French source code more consistent (EN didn't match its demo, FR wasn't truncated at the end at the same spot as EN)
  * **Side note:** Probably botched these originally since I rushed-in the new edge cases just before a technical review meeting a couple of weeks ago and lacked time for self-QA

duboisp (approved):
Reviewed and tested.

I was able to reproduce the issue and I do confirm this fix it.

PR #9782, opened 2024-06-21
Background image: adding half-width utilities

Garneauma:
Adds:
- `bg-half-left`: background image takes the left half width of the element
- `bg-half-right`: background image takes the right half width of the element

### DIFF
--- a/src/base/_component.scss
+++ b/src/base/_component.scss
@@ -18,6 +18,17 @@
 .bg-norepeat {
 	background-repeat: no-repeat;
 }
+.bg-half-right,
+.bg-half-left {
+	background-repeat: no-repeat;
+	background-size: 50%;
+}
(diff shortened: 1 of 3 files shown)


src/base/_component.scss
```diff
@@ -18,6 +18,17 @@
 .bg-norepeat {
 	background-repeat: no-repeat;
 }
+.bg-half-width-right,
+.bg-half-width-left {
```
duboisp, line 22: I propose the shorter name

```suggestion
.bg-half-right,
.bg-half-left {
```

duboisp (changes requested):

Garneauma:

duboisp:
This change has been cancelled by DTO on June 2, see comment in request WET-464 for more details

PR #9791, opened 2024-08-02
Focus: Assign focus more resiliently

EricDunsworth:
If this plugin is used to focus onto a non-interactive element, it adds a ``tabindex="-1"`` attribute (to make the destination focusable) before proceeding.

Since adding ``tabindex="-1"`` isn't always necessary, the plugin exempts elements that already have a ``tabindex`` attribute, as well as "known" interactive elements.

However...
* The plugin removed interactive elements it wasn't aware of (like ``summary``) from sequential keyboard navigation:
  * End result was WCAG 2.x failures: SC 2.1.1 and 2.4.3 (via F44)
* Tying exemptions to a list of "known" interactive elements is a bad practice:
  * Isn't future-proof
  * Keeping it up-to-date is a maintenance burden
  * [Interactive content as defined by the HTML spec](https://html.spec.whatwg.org/multipage/dom.html#interactive-content) isn't suitable for that purpose (e.g. ``details`` and ``label`` are technically interactive... but aren't keyboard-tabbable)
  * Doesn't account for different behaviours across browser rendering engines (e.g. ``video`` element without the ``controls`` attribute is tabbable in Gecko, but not in Blink)

This alleviates the need for an exemptions list by revamping how focusing works:
1. The plugin will first try to focus directly onto the element
2. If the element hasn't gained focus, the plugin will add a ``tabindex="-1"`` attribute and retry
3. If the element still hasn't gained focus... the plugin will remove ``tabindex="-1"`` and show a console error

Credit goes to @fsnoddy for initially spotting this issue (by using the toggle plugin's ``{"print": "on"}`` configuration option on a ``summary`` element). Also added "boring" print state examples to the toggle plugin's demo page to show it in action.

### DIFF
--- a/src/plugins/wb-focus/focus.js
+++ b/src/plugins/wb-focus/focus.js
@@ -28,7 +28,8 @@ var $document = wb.doc,
 // Bind the setfocus event
 $document.on( setFocusEvent, function( event ) {
 	if ( event.namespace === "wb" ) {
-		var $elm = $( event.target ),
+		var elm = event.target,
+			$elm = $( elm ),
 			$closedParents = $elm.not( "summary" ).parents( "details, [role='tabpanel']" ),
 			$closedPanels, $closedPanel, len, i;
 
@@ -48,20 +49,41 @@ $document.on( setFocusEvent, function( event ) {
 			}
 		}
 
-		// Set the tabindex to -1 (as needed) to ensure the element is focusable
-		$elm
-			.filter( ":not([tabindex], a[href], button, input, textarea, select)" )
-			.attr( "tabindex", "-1" );
-
 		// Assigns focus to an element (delay allows for revealing of hidden content)
 		setTimeout( function() {
+			let addedTabIndexAttr = false;
+
 			$elm.trigger( "focus" );
 
-			var $topBar = $( ".wb-bar-t[aria-hidden=false]" );
+			// If the element hasn't gained focus, retry with a tabindex attribute...
+			if ( elm !== document.activeElement && elm.getAttribute( "tabindex" ) === null ) {
 
-			// Ensure the top bar overlay does not conceal the focus target
-			if ( $topBar.length !== 0 ) {
-				document.documentElement.scrollTop -= $topBar.outerHeight();
+				// Add a tabindex="-1" attribute and keep track of it for later
+				elm.setAttribute( "tabindex", "-1" );
+				addedTabIndexAttr = true;
+
+				// Try re-focusing
+				$elm.trigger( "focus" );
+			}
+
+			// If the element has gained focus...
+			if ( elm === document.activeElement ) {
+				const $topBar = $( ".wb-bar-t[aria-hidden=false]" );
+
+				// Ensure the top bar overlay does not conceal the focus target
+				if ( $topBar.length !== 0 ) {
+					document.documentElement.scrollTop -= $topBar.outerHeight();
+				}
+			} else {
+
+				// Remove the element's tabindex attribute (no point in keeping it if it didn't help)
+				if ( addedTabIndexAttr ) {
+					elm.removeAttribute( "tabindex" );
+				}
+
+				// Show a console error
+				console.error( setFocusEvent + ": Unable to focus onto the destination element... maybe it's hidden?" );
+				console.error( elm );
 			}
 
 			return $elm;
--- a/src/plugins/toggle/toggle-en.hbs
+++ b/src/plugins/toggle/toggle-en.hbs
@@ -8,7 +8,7 @@
 	"parentdir": "toggle",
 	"altLangPrefix": "toggle",
 	"css": ["demo/toggle"],
-	"dateModified": "2020-03-10"
+	"dateModified": "2024-08-14"
 }
 ---
 <span class="wb-prettify all-pre linenums"></span>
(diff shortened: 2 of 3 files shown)


src/plugins/wb-focus/focus.js
EricDunsworth, line 1: Would breaking support for IE11 and possibly old Edge be a show-stopper?

[jQuery's ``focus`` event documentation](https://api.jquery.com/focus/) mentions a few quirks with IE's native ``focus`` event:

**1.**
> Attempting to set focus to a hidden element causes an error in Internet Explorer. Take care to only use ``.trigger( "focus" )`` on elements that are visible. To run an element's focus event handlers without setting focus to the element, use ``.triggerHandler( "focus" )`` instead of ``.trigger( "focus" )``.

That behaviour might pre-empt the console errors this PR's logic would ultimately produce in that scenario.

Btw it isn't clear from jQuery's explanation whether IE would treat _any_ unfocusable element as "hidden". If it does, it'd probably error-out before this PR's logic tries re-focusing with ``tabindex="-1"`` (which might've otherwise been successful had ``tabindex="-1"`` been in play on the first try).

**2.**
> The native ``focus`` event is asynchronous in all versions of IE, contrary to other browsers. To avoid issues related to this discrepancy, as of jQuery 3.7.0, jQuery uses ``focusin`` as the native backing event for ``focus`` in IE.

That behaviour might intefere with the premise of this PR.

Since this PR's logic works by immediately checking where focus went before deciding how to proceed... there's a risk that IE's goofy behaviour might run into unexpected outcomes :S.

For example:
* Adding ``tabindex="-1"`` onto an interactive element that's supposed to be tabbable.
* Skipping scroll-offsetting in the [overlay plugin](https://wet-boew.github.io/v4.0-ci/unmin/demos/overlay/overlay-en.html)'s top bar example.

EricDunsworth:
Here's a simplified example of what @fsnoddy was trying to do:
```html
<details>
	<summary class="wb-toggle" data-toggle='{"print": "on"}'>My summary</summary>
	<p>My description</p>
</details>
```

His goal was to use the toggle plugin's ``{"print": "on"}`` option to auto-expand ``details`` elements when printing. But he noticed that activating the ``summary`` element caused ``tabindex="-1"`` to get added to it... which took  it out of the keyboard-tabbing sequence from that point onwards 💀.

I subsequently determined that it was caused by the toggle plugin's reliance on the focus plugin. I could've fixed it just by adding ``summary`` to the [focus plugin's ``tabindex="-1"`` exclusions](https://github.com/wet-boew/wet-boew/blob/e01bc308c1cc123dbe7948cd8dccc14829ed8e70/src/plugins/wb-focus/focus.js#L51-L54)... but really disliked the exclusions list. So I completely reworked it to use a simpler, browser-agnostic approach.

Btw whatwg/html#4607's [focus testing page](https://boom-bath.glitch.me/tabindex.html) demonstrates plenty of focusable element examples and how different browser engines handle them differently.

For example:
* ``object`` is programmatically-focusable in Blink, but not Gecko
* ``embed`` is focusable in Gecko, but not Blink
* "text1 text2" (``div`` with overflowing content and scrollbars) is focusable in Gecko, but not Blink
* ``video`` is focusable in Gecko, but not Blink
* ``video[src]`` is focusable in Gecko, but not Blink
* ``area`` with no ``href`` is focusable in Gecko, but not Blink

All in all, it isn't realistic to be using exclusion lists to figure out whether it's appropriate to add ``tabindex="-1"``. There's way too much variation between browser implementations for that to be practical.

Garneauma:
Pre-approved upon successful testing and review. This will need a thorough testing of all plugins that use this utility.

duboisp:
Step to reproduce:
1. Insert the code sample
2. Tab until the summary get the focus
3. The first time it get the focus and open (activate) the detail/summary.
4. Continue the tabbing on the next element
5. Then get back on the summary element previously open (shift+tab)
6. Result:
  * Pass: We can focus and operate the summary element
  * Fail: The focus on the summary are skip or it is inoperable. [This change is to fix this bug]

Todo:
* [ ] Create a working example to reproduce the bug
* [ ] Test all plugins (currently active with GCWeb) that are leveraging the wb-focus
* [ ] Test with a few browser: Chrome, Safari, Edge

EricDunsworth:
@duboisp Just added "boring" ``details`` examples to the toggle plugin's demo page to demonstrate @fsnoddy's scenario (which ultimately led to this PR).

As far as other tests go...
* This successfully focuses on the first try:
  * **Example:** [``<input id="fname1">`` (Form validation)](https://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-en.html#fname1)
* This fails on the first try, then successfully retries with ``tabindex="-1"``.
  * **Example:** [``<h1 id="wb-cont">`` (Accessibility Conformance Report Methodology)](https://wet-boew.github.io/v4.0-ci/unmin/demos/acr/acr-en.html#wb-cont)
* This tries twice, fails and produces a console error since the ``<details>`` element is hopeless:
  * **Example:** [``<details id="json-ld-acr">`` (Accessibility Conformance Report Methodology)](https://wet-boew.github.io/v4.0-ci/unmin/demos/acr/acr-en.html#json-ld-acr)

Garneauma (approved):
The suggested fix works to resolve the mentioned issue. Thank you @EricDunsworth for creating a working example.

Tested all plugins using the `setfocus.wb` event in Chrome, Safari and Edge.

PR #9792, opened 2024-08-05
JSON manager docs: Remove redundant event description

EricDunsworth:
The events section covered ``wb-init.wb-jsonmanager`` twice in a row (two identical copies of the same content).

This removes the second copy of the content.


duboisp (approved):
Reviewed and approved. This is only a content change.

PR #9800, opened 2024-08-21
Build - Continuous deployment - Run deployment with only one NodeJS version

duboisp:
To fix a racing condition when deploying and creating new tags for wet-boew via our continuous deployment script.


Ricokola (dismissed):
Seems all good

duboisp:
I updated the version of wet-boew for refreshing all the releases files

Ricokola (approved):

PR #9807, opened 2024-09-06
Update package.json to support newer versions of npm

Garneauma:


duboisp (approved):
Reviewed and it looks good. Didn't tested as it only affecting the build which seems to work according to our CI.

PR #9814, opened 2024-09-25
Updating i18n

Garneauma:


duboisp (approved):
Reviewed and this PR do revert, as expected, the undesirable changes applied during the google spreadsheet sync which contained incomplete translation.

Goal: Revert the change made in this PR to the i18n.csv : https://github.com/wet-boew/wet-boew/pull/9813/files#diff-962f09b704d01f39c99b2a9fab246f880e03ac2f5fdc87adc978c59a233ba509

Proof: Compare with the previous version, I can see that only the overlay, as expected, is updated. And we can see that some double quote are removed for the twitter embedded timeline (which was not in our master file previously). 
https://github.com/wet-boew/wet-boew/compare/v4.0.81.4...Garneauma:v4.0.81.4:master#diff-962f09b704d01f39c99b2a9fab246f880e03ac2f5fdc87adc978c59a233ba509

Thanks @Garneauma for fixing our master translation file and for this patch PR.

PR #9830, opened 2024-11-19
Form Validation:  Only allow display of error summary on form submit.

### LINKED ISSUE #9827: Form Validation: Error summary being displayed even when the form is not submitted.
During a recent Accessibility audit of our application, an issue was raised regarding the error summary.  The summary was being displayed with the message **The form could not be submitted** even though the form was not submitted.  The assessor felt this could cause confusion for clients with cognitive issues.  

To recreate the issue, go to the Form Validation example page and select a value for the Title drop down, then change the drop down back to Select a title.  This will trigger the error summary to be displayed.

I've tried configuring jQuery Form Validation to disable validation on onfocusout=false, onkeyup=false and onclick=false.

Is there a way to prevent the error summary from being generated if the page was not submitted?

sgdowney:
This pull request introduces the following behaviours to the Form Validation plugin:

- Only display the error summary when a form is submitted with validation errors or server-side validation errors exist.
- When a field becomes invalid and the field still has focus, do not update the `arialive` region to prevent the error from being announced twice by screen readers.  
- When a field becomes invalid and does not have focus, update the `arialive` region to prompt the screen reader to announce the error.
- When a validation error is triggered, only update the error summary if it is displayed.

Closes #9827


Garneauma:
This will require further investigation whether it fails WCAG or it is more user preference.

sgdowney:
This was flagged under WCAG 3.3.1 Error Identification (Level A).  The assessor provided the following in the report:
Impact on Users: Users with cognitive variations could be confused by the error message that says the form was submitted without selecting the "Continue" button.  The error message "The form could not be submitted" should only be shown when we try to submit the form, otherwise it's misleading and can be confusing.

Garneauma:
@sgdowney Thank you for your response! We will proceed with a technical review.

Garneauma (approved):

PR #9834, opened 2024-11-27
Facebook embedded pages: Add overflow-x scrollbar

EricDunsworth:
Facebook's page widget isn't fully responsive:
* Default width is 340px
* Officially supports widths between 180-500px (doesn't support percentages)
* Automatically shrinks to fit within its container when inited (via ``data-adapt-container-width="true"``, default behaviour)
* Can't shrink or expand after initing (can't use CSS to resize its iframe either... width is set inside the iframe's ``body``)

As a result, if the widget initially renders itself within a wide container, resizing the viewport won't cause it to scale... So if the viewport width shrinks and the widget overflows beyond its container, the entire page will become horizontally-scrollable.

This limits the damage by adding an ``overflow-x`` property to the widget's container (to make only the widget itself scroll horizontally).

### DIFF
--- a/src/base/_wet-boew.scss
+++ b/src/base/_wet-boew.scss
@@ -100,6 +100,7 @@
 @import "../plugins/dismissable-content/base";
 @import "../plugins/eqht-css/base";
 @import "../plugins/equalheight/base";
+@import "../plugins/facebook/base";
 @import "../plugins/filter/base";
 @import "../plugins/footnotes/base";
 @import "../plugins/formvalid/base";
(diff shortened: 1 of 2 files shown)


EricDunsworth:
**Reproduction steps:**
1. Visit the [Facebook embedded pages demo](https://wet-boew.github.io/v4.0-ci/demos/facebook/facebook-en.html) in a full-screen desktop browser window.
2. Open your browser's mobile/responsive design emulator.
3. Decrease the viewport width to 320px.
4. Scroll-down to the widget.

**Current behaviour (before this PR):**
Entire page is horizontally-scrollable.

**Expected behaviour (after this PR):**
Only the widget is horizontally-scrollable (*not* the whole page).

**PS:**
Another way of dealing with this scenario would be to reload the widget if the viewport is resized. IMO it'd do a better job of adhering to [WCAG 2.x's reflow (1.4.10) success criterion](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html)... but come with other drawbacks due to the refresh (e.g. mini change of context, losing the ``iframe``'s previous focus selection and/or scroll position). IMO the lesser of two evils is an horizontal scrollbar. Plus this scenario isn't very common in any case.

Garneauma:

Garneauma (approved):

PR #9836, opened 2024-12-04
Core: adding native DOM dispatch event on component ready

Garneauma:


duboisp (changes requested):
Can you document this addition here: https://github.com/wet-boew/wet-boew/tree/master/src/other/core
With one sentence on how someone can use it.

And if you want add this code sample
```
<script>

        document.addEventListener( "wet-boew-ready", function(e) {
                console.log( "DOM event - WET-BOEW is now ready" );
        });

</script>
```

I tested and it does work as expected. Once the documentation completed, we will be able to merge this one

Considering this is required for tracking by our AA to track contextual feature that are display randomly on our home page, this is essential and will allow us to implement a stable solution for them to track. Also this solution will simplify our AA configuration when they need to wait for WET to be ready.

thanks

duboisp (approved):
Reviewed and tested locally

Everything work as expected.

This is a minor change

PR #9840, opened 2024-12-11
Build: Replace Node Sass with Dart Sass

EricDunsworth:
Minimum changes needed to get it to successfully build.

Contains countless deprecation warnings, but those can be dealt with in the future.

Haven't thoroughly compared old (Node Sass) vs new (Dart Sass) compiled CSS, but haven't spotted any show-stopper differences so far.


EricDunsworth:
Btw GCWeb is able to successfully build for me locally (node.js v20.18.0 on Windows 10) when replacing GCWeb's wet-boew dependency with this PR's branch... even when using Node Sass for GCWeb itself :D.

EricDunsworth:
Here's a quick overview of some of the differences I noticed while trying to diff Dart Sass vs Node Sass' compiled CSS:
* Dart places comma-separated selector rules in a different order than Node
* Dart places closing curly brackets (``}``) at the start of the next line after a selector's final property (Node used a space and closing curly bracket on the same line as each selector's final property)
* Dart places ``0``s in front of property values containing decimal numbers (Node didn't)
* Dart doesn't retain multiple indents in front of properties unless nested in a media query (Node sometimes retained multiple indents)
* Dart adds indents to multi-line comments (Node didn't)
* Dart sometimes writes colour values as rgb() values (in scenarios where Node sometimes used colour names or HEX values)
* Dart omits 0 suffixes from the end of decimal numbers
* Dart adds a space character at the end of IE8 hack values ending with \9 (Node didn't)
* Dart omits final newlines

I haven't been able to do a comprehensive comparison between them since what they output isn't extremely similar... so it's not easy to compare them head-to-head. But I haven't spotted any differences in my spot checks that'd give me any reason to think anything would break if WET moved over to Dart Sass.

As a result, I'll shift this PR into a ready to review state.


**PS:**
Merging this, rolling out a new WET release and updating GCWeb's WET dependency should be enough to make GCWeb buildable again in node.js. In fact, wet-boew/GCWeb#2457 isn't even technically needed for it (although it'd be kind of strange for it to be using both Dart Sass for its WET dependency and Node Sass for itself :D).

Garneauma:

duboisp:
I compiled and I can now build wet-boew on my armbian device. That is great.

I compared both CSS (old and this new one) and I do confirm the difference explained by Eric. The transformation that is interesting is that a few hex color are transformed into RBG with decimal value. After research that should not impact as those value are more precise compared to hex color value.

In term of functional testing, we will need at least to check the following plugin for style changes:
* Form validation plugin
* Static input form that leverage the CSS class `.input-group-sm` with inner input that are type: date 
* Multimedia player to check the content value with this CSS condition: `.wb-mltmd.video:not(.playing):not(.waiting) .display:after`
  * It may currently broken and this Dart sass may fix it
* Glyphicon
* Color. For example HTML that are matching the following CSS selector
  * `a.bg-primary:hover`
  * `a.text-danger:hover`

@Garneauma can you do that functional testing? 

Once the functional testing completed and successful, we can approve this PR. This is **Patch change** that impact **all components**.

Now I am going to compile a version of GCWeb with this fix.

Garneauma (approved):
I have completed functional testing and everything looks the same as the live version.

PR #9842, opened 2024-12-27
JSON manager - Example of use multi data source to render a data table and racing condition edge case fix

duboisp:
Example of use multi data source to render a data table and racing condition edge case fix


Garneauma:

Garneauma (changes requested):
In Chrome, when I try a working example with a lot of initializations on the page, I get the following browser alert: "DataTables warning: table id=wb-auto-42 - Cannot reinitialise DataTable. For more information about this error, please see https://datatables.net/tn/3".

Garneauma (approved):

PR #9845, opened 2025-01-08
Build: downgrade uglify-js version to fix jsonmanager issue

Garneauma:
Downgraded `uglify-js` to point to version `3.17.4` inside the `grunt-contrib-uglify` project (now forked from the main project) to fix JSON Manager "too much recursion issue".

Related to WET-507

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "grunt-contrib-copy": "^1.0.0",
         "grunt-contrib-cssmin": "^3.0.0",
         "grunt-contrib-htmlmin": "^3.1.0",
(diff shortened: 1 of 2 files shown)


duboisp (approved):
There is some inconsistency in our local copy of grunt-contrib-uglify in regards of matching the package.json along with the package-lock.json, but that is not a stopper.

I was able to reproduce and confirm the issue, then I ran a "npm install" and tested again and it seem to work as expected.

This is a great interim fix for now.
We should investigate what has changed after uglifyjs version "3.17.4", see what change has been applied since to the library and try to fix our code base to avoid the issue.

We can merge this at our next technical review meeting.

PR #9850, opened 2025-01-16
Update _base.scss

Christopher-O:
Change was to remove the legacy CSS code for details/summary for IE11 (which did not support this element) as its no longer required.  The removal was for the display:none and visibility:  visible/hidden styles.

This fixes two issues: 
1. It prevents the focus of the keyboard tab from moving to a hidden tab in desktop if it contains an details element thus violating WCAG SC 2.4.7 (Level AA) Focus Visible.  
2. It allows the user to use Ctrl+F5 to search for a word/phrase inside of a details element and also opens the element if the word/phrase exists.  

Impact:
The only impact this will have may be for anybody with an unsupported browser that does not support detail elements (Internet Explorer), or very early browser versions.  If there is a desire to support this, a script could be written or a fallback stylesheet could be created.  Otherwise the issue should be minimal.

No documentation needed to be added or updated.

- [X] Build and test PR's code /// Rouler le script de compilation et tester le code de la PR
- [X] Validate changes against WCAG for accessibility /// Valider les changements avec WCAG pour l'accessibilité

None

Pure keyboard and search functionality

### DIFF
--- a/src/polyfills/details/_base.scss
+++ b/src/polyfills/details/_base.scss
@@ -49,14 +49,3 @@ details {
 		}
 	}
 }
-
-// Prevent FOUC when polyfill is disabled
-.wb-disable {
-	details {
-		& > {
-			*:not(summary) {
-				display: block !important;
-			}
-		}
-	}
-}
--- a/src/polyfills/details/_base.scss
+++ b/src/polyfills/details/_base.scss
@@ -8,16 +8,8 @@ summary {
 	display: list-item !important;
 	list-style-type: none; // Prevent list bullets from appearing in browsers that lack native details/summary support (e.g. IE/Edge)
 	list-style-type: disclosure-closed; // Show summary arrows in Firefox 49.0+ and future versions of Blink/Webkit
-	visibility: visible !important;
 }
 
-%details-nested-in-hidden-details {
-	details {
-		summary {
-			display: none !important;
-		}
-	}
-}
 
 details {
 	margin-bottom: .25em;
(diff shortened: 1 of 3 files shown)


duboisp:
@Christopher-O on which browser you had issue? Did you tested with IE11 or another browser? I will need that information for test and review. Thanks

Christopher-O:
> @Christopher-O on which browser you had issue? Did you tested with IE11 or another browser? I will need that information for test and review. Thanks

This is an issue for all Chromium browsers as well as Firefox.  I have not yet tested this on Safari.

Garneauma:

Garneauma (approved):

src/polyfills/details/_base.scss
```diff
@@ -56,48 +48,12 @@ details {
...
-		details {
-			&:not([open]) {
-				visibility: visible;
-			}
-		}
-	}
 }
 
 // Prevent FOUC when polyfill is disabled
 .wb-disable {
```
Garneauma, line 54: @Christopher-O Could you also remove this whole rule? I just tested in basic mode and doesn't seem to help with anything. It actually causes issues for table elements and other elements that are not by default block elements.

Garneauma (changes requested):

duboisp:
@Christopher-O sorry for my incomprehension in previous comment, I was under the perception that CSS inside the polyfill was dedicated to the details polyfill only when it is active. But thanks to @Garneauma that have explained to me that the CSS inside that polyfill are actually styling the baseline for all details/summary element regardless if the polyfill are active or not, which goes beyond that polyfill scope. In short, that CSS file are definitely miss placed, we might move it the future.

src/polyfills/details/_base.scss
Christopher-O: @Garneauma  just to confirm, you're suggesting to remove the entire rule for .wb-disable?

Garneauma: @Christopher-O Yes, exactly. Thank you.

Garneauma (approved):

PR #9855, opened 2025-01-27
Postback: fix submitter not being sent

Garneauma:
Submitter object was not being sent correctly. This PR fixes it.

WET-514


Garneauma:

duboisp (approved):
Review and tested locally, it does work as expected

PR #9857, opened 2025-01-31
PII scrub: fix support for Postback and data on submitter

Garneauma:
Fixing known issue when using the PII scrub in conjunction with the Postback plugin. This PR also fixes a previously unknown issue where the data available on the submit button is not being sent when using the PII scrub plugin.


duboisp (approved):
I tested each example with and without PII info and it does work as expected,

Change:
* Minor (was Patch) - PII scrub
  * I recommend to increase the impact to Minor, event if the change is rated as Patch because that are going to be the minimal version where the PFT component will be able to work along with the PII scrub plugin.
* Patch - Postback

PR #9864, opened 2025-02-03
chore: Update Stylelint and add Stylelint for Sass files

nschonni:
_(Pick the language of your choice to fill out the following information. / Choisissez la langue de votre choix pour remplir l'information ci-dessous.)_

## What does this pull request (PR) do? / Que fait cette demande « pull » (PR)?
Migrate to Stylelint v16 and removed depreciated rules.
Second commit adds linting and some cleanup for the SCSS files. This could be split out. Not sure if the `sass-lint` part is still needed

## Additional information (optional) / Information additionnelle (optionel)

**General checklist / Liste de contrôle générale**
Make your own list for the purpose of your Pull request. /// Faites votre propre liste en fonction des besoins de votre demande « pull ».

- [ ] Create/update documentation /// Créer/mettre à jour la documentation
- [ ] Build and test PR's code /// Rouler le script de compilation et tester le code de la PR
- [ ] Validate changes against WCAG for accessibility /// Valider les changements avec WCAG pour l'accessibilité
- [ ] Ensure documentation is bilingual /// S'assurer que la documentation soit bilingue

PR #9866, opened 2025-02-05
Forms: Add asterisk disable class for required fields

EricDunsworth:
Required form fields are visually-prefixed with an asterisk (*) by default (via the "required" class). But asterisks aren't always desirable...

For instance:
* They aren't meaningful in and of themselves
* WET implements them as visual decorations (via a CSS pseudo-element and the ``content`` property)
* They might be perceived as visual bloat, particularly in the following scenarios:
  * Form consisting solely of one required field
  * Form consisting entirely of required fields
  * Experimental plugins like steps form (WET) and steps quiz (GCWeb) that are likely to show only one question at a time
* They're not necessary to achieve WCAG 2.x compliance:
  * [Success criterion 3.3.2's technique H90](https://www.w3.org/WAI/WCAG22/Techniques/html/H90) doesn't obligate field labels/legends to use asterisks to indicate their requiredness (although it's allowable so long as the asterisk's meaning is defined in content beforehand)

Why not just omit the "required" class when asterisks aren't desirable? Because that class is needed to make the bolded "(required)" text red (unsure if due to a bug or by design). So up to this point, it was almost a foregone conclusion that all of WET's required fields always used asterisks.

This makes it possible to disable asterisks by:
* Adding a new class ("required-no-asterisk") and setting it on any required field's parent element
* Demonstrating the class in the form validation plugin's demo page (via the "Examples using required fields without asterisks" section)
* Describing the class' purpose in the form validation plugin's documentation (in the "How to implement" section)

### DIFF
--- a/src/base/forms/_base.scss
+++ b/src/base/forms/_base.scss
@@ -12,12 +12,16 @@ label,
 legend,
 .checkbox {
 	&.required {
-		&:before {
-			@extend %forms-color-red-bold;
+		/* Set required asterisk styles... unless any parent element uses a disable class */
+		/* Credit: Stack Overflow answer (https://stackoverflow.com/a/73773370) by magda and Zach Jensz (zach-jensz) */
+		&:not(.required-no-asterisk .required) {
+			&:before {
+				@extend %forms-color-red-bold;
 
-			content: "* ";
-			margin-left: -$wb-forms-asterisk-width;
-			vertical-align: top;
+				content: "* ";
+				margin-left: -$wb-forms-asterisk-width;
+				vertical-align: top;
+			}
 		}
 
 		strong {
--- a/src/plugins/formvalid/formvalid-en.hbs
+++ b/src/plugins/formvalid/formvalid-en.hbs
@@ -7,7 +7,7 @@
 	"tag": "formvalid",
 	"parentdir": "formvalid",
 	"altLangPrefix": "formvalid",
-	"dateModified": "2022-04-13"
+	"dateModified": "2025-02-24"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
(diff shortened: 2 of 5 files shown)


src/plugins/formvalid/formvalid-en.hbs
```diff
@@ -357,6 +357,22 @@
 				</details>
 			</fieldset>
 
+			<div class="form-group">
+				<label for="cname1" class="required required-no-asterisk"><span class="field-name">Country name</span> (no asterisk) <strong class="required">(required)</strong></label>
```
EricDunsworth, line 361: Btw I could add ``fieldset``/``legend`` and standalone ``checkbox`` examples if needed.

Didn't do it from the outset since I felt like they'd amount to bloat. Especially since the "required-no-asterisk" class doesn't have its own properties (it's nothing more than a ``:not()`` pseudo-class).

EricDunsworth:
* Revised this PR's OP + commit message
* Revised the ``required-no-asterisk`` class' selector to apply only to parent elements (was trickier than I expected :D... credited [this Stack Overflow answer](https://stackoverflow.com/a/73773370) in a comment)
* Revamped the country name field demo I initially added into a dedicated section demonstrating a group of required fields without asterisks
* Revised the item I initially added to the form validation plugin docs' "How to implement" section
* Updated modified dates
* Rebased to fix conflicts with #9864 and #9867's changes

src/base/forms/_base.scss
```diff
@@ -12,12 +12,16 @@ label,
 legend,
 .checkbox {
 	&.required {
-		&:before {
-			@extend %forms-color-red-bold;
+		/* Set required asterisk styles... unless any parent element uses a disable class */
+		/* Credit: Stack Overflow answer (https://stackoverflow.com/a/73773370) by magda and Zach Jensz (zach-jensz) */
+		&:not(.required-no-asterisk .required) {
```
EricDunsworth, line 17: Should a ``:where()`` pseudo-class be leveraged to reduce this selector's specificity?

What I had in mind was backwards-compatibility with any pre-existing WET implementations that might've tried overriding the required asterisk styles (would be very surprised if anyone's bothered to try :P).

src/plugins/formvalid/formvalid-en.hbs
```diff
@@ -456,6 +456,91 @@
 	</div>
 </section>
 <hr>
+<section>
+	<h2>Examples using required fields without asterisks</h2>
```
EricDunsworth, line 460: @Garneauma @duboisp Should a paragraph be added to this section (and perhaps even the docs) to explicitly state that the ``required-no-asterisk`` class is an alternate way of styling required fields? Or is the status quo good enough to get the point across?

Garneauma: Yes, please.

Garneauma:
Pre-approved upon successful review and addition of paragraph added.

src/base/forms/_base.scss
EricDunsworth: Was determined that it'd be best not to proceed with this.

src/plugins/formvalid/formvalid-en.hbs
EricDunsworth: Done :)!

EricDunsworth:
Just force-pushed a few more content revisions:
* Added intro content to the demo page's no asterisks section
* Revised the title of the French demo page's no asterisks section (renamed "requis" to "obligatoires")
* Slightly-revised the documentation item's wording and copied-over the demo page's recommendation against mixing and matching asterisk styles
* Updated modified dates

@Garneauma Should be fully ready to go now :).

src/plugins/formvalid/formvalid-fr.hbs
```diff
@@ -456,6 +456,93 @@
 	</div>
 </section>
 <hr>
+<section>
+	<h2>Exemples utilisant des champs obligatoires sans astérisques</h2>
+	<p>Les champs obligatoires peuvent être alternativement stylés sans astérisques en ajoutant la classe <code>required-no-asterisk</code> à l’un de leurs éléments parents (comme <code>&lt;div class="wb-frmvld"&gt;</code> ou <code>&lt;form&gt;</code>).</p>
+	<p>Que les astérisques sont utilisés, c’est recommandé qu’ils soient implémentés d’une façon uniforme dans le même formulaire. S’il vous plaît, ne mélangez pas des styles d’astérisques différents dans le même formulaire.</p>
```
Garneauma, line 462:
```suggestion
	<p>Que les astérisques soient utilisés ou non, il est recommandé qu’ils soient implémentés de façon uniforme dans un même formulaire. Veuillez ne pas mélanger les styles d’astérisques différents dans un même formulaire.</p>
```

Garneauma (changes requested):
@EricDunsworth Only a minor text change. Other than that, it works as expected. Thanks !

Garneauma (approved):

Issue #9806, opened 2024-09-05
[BUG] Non-breaking Space Missing in French Translations


raviraj-mangukiya:
## Required information

**Describe the bug**
In French typography, it is customary to include a non-breaking space before certain punctuation marks, such as colons, semicolons, question marks, and exclamation points. This practice is part of French typographic standards and is important for maintaining clarity and readability.

Our project uses both wet-boew and gctheme, and we’ve identified that the non-breaking space issue is present due to how these dependencies handle French translations. This issue is also observed in the current version of [Canada.ca](https://www.canada.ca) and NVC https://nvc-cnv.canada.ca/fr/tradename-immunizing-agents
also in All demo where we have table ex. https://wet-boew.github.io/v4.0-ci/demos/index-fr.html

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to '[...](https://www.canada.ca/fr/immigration-refugies-citoyennete/services/travailler-canada/employeurs-non-conformes.html)'
2. Open NVDA download screen reader if its not installed  '[....](https://www.nvaccess.org/download/)'
3. Scroll down on page until you see table header
4. Hear speech where it reads &#160; while reading title See error screenshot attached

**Expected behaviour**
it should read title without &$160;

**Who shall do the work?**
I am asking for Principal publisher to please do the work

## Additional information (optional)

**Version of WET-BOEW/GCWeb you are using**
All versions including latest v4.0.81.3

**Desktop/Smartphone (please complete the following information)**
 - OS or Device: Windows, Mac
 - Browser [Edge, chrome, safari]
 - Version 

**Screenshots**
![image](https://github.com/user-attachments/assets/bb67e71a-1446-42b9-a620-5a57da2335b6)

**Additional context**
The issue can be resolved once the dependencies (wet-boew and gctheme) are updated to handle non-breaking spaces correctly in French translations by editing wet-boew code: [wet-boew/tables.js](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/tables/tables.js#L246).

mercury64:
I think the problem is with the NVDA, as it shouldn't read/interpret the numeric character reference literally rather than as a non-breaking space. I'd file a bug report with them.

Unfortunately NVDA is not available for Mac OS or Safari.

I have tested on Mac OS with Safari and the default OS screen reader interprets the &#160; correctly.

I'd report the issue over here: https://github.com/nvaccess/nvda/issues

PR #9899, opened 2025-02-20
fix: Replacing HTML whitespace character so that it doesn't render in table header tooltip

WildRyc:

The i18n csv that contains the additional information for table headers includes a leading space before the colon. This is entered as a `#&160;` which preserves the space that would otherwise be discarded. However, when the page is refreshed, as in sorting the table, this whitespace character is not rendered correctly.

The addition of a replace on the character to a `" "` should cover this. I'm not sure if there are downstream effects, or if we should instead use `&nbsp`.

## Additional information (optional) / Information additionnelle (optionnel)

Related to #9806.

***Before***
Screenshot 2025-02-20 170937
***After***
Screenshot 2025-02-20 170633

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -195,7 +195,7 @@ $document.on( "draw.dt", selector, function( event, settings ) {
 			$btn = $th.find( "button" );
 		if ( order && order.length && order[ 0 ][ 0 ] === index ) {
 			var label = ( order[ 0 ][ 1 ] === "desc" ) ? i18nText.aria.sortAscending : i18nText.aria.sortDescending;
-			label = $btn.text() + label.replace(/&#160;/g, " ");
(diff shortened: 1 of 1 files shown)


Garneauma:
@WildRyc Could you use an non breaking space (&nbsp;) instead?

Garneauma:
@WildRyc I just tried locally and indeed `&nbsp;` does not work either, so I approve your solution. Please fix the issue brought up by the CI build and then it should be good to go. Thanks !

Garneauma (changes requested):

duboisp (approved):
Reviewed and tested locally, it does fix the issue and work as expected

duboisp:
This is a **Patch** change for the **data table** plugin

Garneauma (approved):

PR #9901, opened 2025-02-26
Build: fix stylelint CI/CD issue

Garneauma:


duboisp (approved):
Look good, we can merge this change as it only impact the build

PR #9902, opened 2025-02-26
Forms: Fix misc asterisk disable class issues

### LINKED ISSUE #9900: Form Validation plugin required fields
I was trying out the new option to remove the asterisk from required fields.  Using the demo page, I could only remove the asterisk if I applied the new class at the parent level and at the label level at the same time.  Using one or the other did not hide the asterisk.  Only having to apply the class at the parent level would be preferable to having to update each label.  Is there an additional mark-up change required for this to work?
@EricDunsworth

EricDunsworth:
This is a follow-up to #9866 that fixes flaws identified in #9900.

Specifically:
* Build:
  * Bumped grunt-contrib-cssmin from v3.0.0 to v5.0.0 to make the "required-no-asterisk" class functional in minified CSS (via clean-css v5.1.2):
    * The older version stripped spaces out of selectors situated in pseudo-classes... which in turn caused ":not(.required-no-asterisk .required):before" to be transformed into ":not(.required-no-asterisk.required):before"
* Base forms SCSS:
  * Fixed right-to-left (RTL) support... prevents red "(required)" text from overlapping asterisks
* Form validation plugin demo page ("Examples using required fields without asterisks" section):
  * Removed redundant/non-functional "required-no-asterisk" class from ``label`` elements in the nickname/country name fields' code samples

Fixes #9900.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,7 @@
         "grunt-contrib-concat": "^1.0.1",
         "grunt-contrib-connect": "^3.0.0",
         "grunt-contrib-copy": "^1.0.0",
-        "grunt-contrib-cssmin": "^3.0.0",
+        "grunt-contrib-cssmin": "^5.0.0",
         "grunt-contrib-htmlmin": "^3.1.0",
         "grunt-contrib-uglify": "github:wet-boew/grunt-contrib-uglify#v1.0.0",
         "grunt-contrib-watch": "^1.1.0",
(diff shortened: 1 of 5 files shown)


EricDunsworth:
@Garneauma Here's an overview of changes I noticed while diffing WET's minified CSS in ``grunt-contrib-cssmin`` v5.0.0... Didn't spot any differences I'd consider to be show-stoppers.

**Differences between v3.0.0 and v5.0.0:**
* ``-webkit-box-shadow``, ``-webkit-box-sizing``, ``-webkit-transition`` and ``-webkit-transform`` got removed (leaving just the unprefixed property with the same value... both the vendor-prefixed and non-prefixed versions were previously used side-by-side)
* ``:nth-child(even)`` changed to ``:nth-child(2n)``
* Duplicate ``@charset`` declarations + comments from the header blocks of certain CSS files were consolidated
* Spaces in``-webkit-gradient()``, ``color-stop()`` and ``rgba()`` were removed
* Quotes were added to ``url()``
* Spaces between properties/values were removed in ``@supports ((...) or (...))``
* Spaces are retained in pseudo-classes like ``:not()`` and ``:has()``
* Redundant ``transition`` properties were consolidated
* ``color-stop(0, rgba(0,0,0,.5))`` was replaced with ``from(rgba(0,0,0,.5))`` in ``-webkit-gradient()``
* ``word-spacing:.16em!important`` changed to ``word-spacing:0.16em!important``
  * **Note:** Can't see how this benefits minification, might be a minor regression. Or maybe there's an advantage to literally writing 0 that I'm overlooking?

nschonni:
> * `-webkit-box-shadow`, `-webkit-box-sizing`, `-webkit-transition` and `-webkit-transform` got removed (leaving just the unprefixed property with the same value... both the vendor-prefixed and non-prefixed versions were previously used side-by-side)

Looked at a few of those when enabling https://github.com/wet-boew/wet-boew/blob/0aac4c512b3fd7ca1458b3452788f65f3b7182d0/stylelint.config.js#L100. It seems pretty safe, as `autoprefixer` adds back and required vendor prefixes based on the `browserlist` criteria. Hacky way as to `git add dist/**/*.css --force` then look at the diff after changing the SCSS and rebuilding the css

Garneauma (approved):
Tested locally and looked at the minified CSS files differences and I agree that nothing stands out as being relevant (outside of the fixes for the :not() pseudo-selector).

PR #9920, opened 2025-03-10
build: use NPM for grunt-contrib-uglify

nschonni:

It seems like the fork only pins to this older version. Since NPM 8, there is a different way of doing this https://docs.npmjs.com/cli/v11/configuring-npm/package-json#overrides
The upstream repo looks off, but this is a simpler way of adjusting this dependency. EX: If you change the override to `3.18.0` and re-install, you'll see the change in the lock file. 

## Additional information (optional) / Information additionnelle (optionnel)

- [ ] Create/update documentation /// Créer/mettre à jour la documentation
- [ ] Build and test PR's code /// Rouler le script de compilation et tester le code de la PR
- [ ] Validate changes against WCAG for accessibility /// Valider les changements avec WCAG pour l'accessibilité
- [ ] Ensure documentation is bilingual /// S'assurer que la documentation soit bilingue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "grunt-contrib-copy": "^1.0.0",
         "grunt-contrib-cssmin": "^5.0.0",
         "grunt-contrib-htmlmin": "^3.1.0",
-        "grunt-contrib-uglify": "github:wet-boew/grunt-contrib-uglify#v1.0.0",
+        "grunt-contrib-uglify": "^5.2.2",
         "grunt-contrib-watch": "^1.1.0",
         "grunt-eslint": "^24.3.0",
         "grunt-gh-pages": "^4.0.0",
@@ -68,7 +68,7 @@
         "time-grunt": "^2.0.0"
       },
       "engines": {
-        "npm": "^10.7.0"
+        "npm": ">=10.7.0"
       }
     },
     "node_modules/@babel/code-frame": {
@@ -7217,13 +7217,14 @@
     },
     "node_modules/grunt-contrib-uglify": {
       "version": "5.2.2",
-      "resolved": "git+ssh://git@github.com/wet-boew/grunt-contrib-uglify.git#df7c9e2c5cd479c322488c807c4a682ff7ff35a4",
+      "resolved": "https://registry.npmjs.org/grunt-contrib-uglify/-/grunt-contrib-uglify-5.2.2.tgz",
+      "integrity": "sha512-ITxiWxrjjP+RZu/aJ5GLvdele+sxlznh+6fK9Qckio5ma8f7Iv8woZjRkGfafvpuygxNefOJNc+hfjjBayRn2Q==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
         "chalk": "^4.1.2",
         "maxmin": "^3.0.0",
-        "uglify-js": "3.17.4",
+        "uglify-js": "^3.16.1",
         "uri-path": "^1.0.0"
       },
       "engines": {
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "grunt-contrib-copy": "^1.0.0",
     "grunt-contrib-cssmin": "^5.0.0",
     "grunt-contrib-htmlmin": "^3.1.0",
-    "grunt-contrib-uglify": "github:wet-boew/grunt-contrib-uglify#v1.0.0",
+    "grunt-contrib-uglify": "^5.2.2",
     "grunt-contrib-watch": "^1.1.0",
     "grunt-eslint": "^24.3.0",
     "grunt-gh-pages": "^4.0.0",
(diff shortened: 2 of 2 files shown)


Garneauma:
Note: test Data-JSON and JSON Manager to make sure they still work with this update as the previous update to uglify broke those plugins.

Garneauma:
@nschonni @EricDunsworth I just tested locally (and in AEM) the latest version of grunt-contrib-uglify (5.2.2) (without the override) and there doesn't seem to be any issue with JSON Manager and Data JSON. Maybe the recent changes done to our JS files for ESLint fixes the issue we were having. Still investigating...

nschonni:
OK, let me know and I can remove the `overrides` and bump the `uglify` version in this PR if you want.

EricDunsworth:
Just tried cloning this PR's latest commit, running ``npm install``, ``grunt dist`` and copying its generated wet-boew.min.js file into my local GCWeb clone's /dist/ folder (which is fully up-to-date vs upstream master). Tried a few local test pages derived from the [What to do when someone dies](https://www.canada.ca/en/services/death/checklist.html) and [Welcoming a child](https://www.canada.ca/en/services/child/questionnaire-birth.html) checklist pages. Also tried my WET clone's JSON manager and data JSON demo pages.

Test results:
* **Firefox:** Still getting an "Uncaught InternalError: too much recursion" console error in the checklist test pages and the [data JSON conditional template demo](https://wet-boew.github.io/wet-boew/demos/wb-data-json/conditional-en.html). The unminified version of wet-boew.js doesn't produce any errors.
* **Chrome/Edge:** Not producing any console errors.

Here are ``grunt-contrib-uglify``'s "resolved" dependencies in my wet-boew clone (via the ``nschonni:overrides-grunt-contrib-uglify`` branch):
> **npm list --depth=1**
> ....
> +-- grunt-contrib-uglify@5.2.2
> | +-- chalk@4.1.2
> | +-- maxmin@3.0.0 deduped
> | +-- uglify-js@3.19.3
> | `-- uri-path@1.0.0
> ...

Based on all that, #9922 will probably still ultimately be needed. Or revising the affected JSON plugin to target [strict JS mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) (which is what the ``modules`` option toggles).

IMO it'd probably be best to tackle both issues. I don't think improving the problematic plugin would be enough in and of itself. Other plugins might potentially have their own issues in strict mode... Plus it seems kind of silly for a minifier to make assumptions that risk breaking what it's supposed to be minifying :P.

nschonni:
I could go back to pinning with `overrides`, but I had removed that and force pushed this after openning

Garneauma:
@EricDunsworth Thank you very much for testing this on your environment! It is much appreciated.

@nschonni Please add back the override. Thank you!

nschonni:
I added back the `overrides` if you want to retest

EricDunsworth:
Just retried without any luck (got onto the very latest commit, ran ``npm install`` and ``grunt dist``, retested affected pages).


Here are my "resolved" dependencies for ``grunt-contrib-uglify`` this time around:
> **npm list --depth=1**
> ....
+-- grunt-contrib-uglify@5.2.2 overridden
| +-- chalk@4.1.2
| +-- maxmin@3.0.0 deduped
| +-- uglify-js@3.18.0 overridden
| `-- uri-path@1.0.0
> ...

@nschonni Maybe try changing the ``uglify-js`` override to target [3.17.4](https://github.com/mishoo/UglifyJS/releases/tag/v3.17.4) instead of [3.18.0](https://github.com/mishoo/UglifyJS/releases/tag/v3.18.0)? That'd be in line with what [WET's fork of ``grunt-contrib-uglify`` was doing](https://github.com/gruntjs/grunt-contrib-uglify/compare/main...wet-boew:grunt-contrib-uglify:main).

3.18.0 is the first version that enabled the ``module`` option by default. So IMO it makes sense that the status quo is still producing console errors atm.

nschonni:
@EricDunsworth sorry about that, I was half paying attention and used that number from the starting comment without actually parsing that it was an example. PS: I usually use `npm ci`, since it then wipes out the node_modules and makes a clean install after

EricDunsworth:
@nschonni Thanks :)! Btw thanks for the heads up about ``npm ci``.

My [What to do when someone dies](https://www.canada.ca/en/services/death/checklist.html) and [Welcoming a child](https://www.canada.ca/en/services/child/questionnaire-birth.html) checklist test pages are no longer producing any console errors :)! My cloned version of the [data JSON conditional template demo](https://wet-boew.github.io/wet-boew/demos/wb-data-json/conditional-en.html) is also currently free of console errors (but it's not producing console errors at all today :S - not even with "problematic" versions of wet-boew.min.js).

Here are my "resolved" dependencies for ``grunt-contrib-uglify`` at this point:
> **npm list --depth=1**
> ...
> +-- grunt-contrib-uglify@5.2.2 overridden
> | +-- chalk@4.1.2
> | +-- maxmin@3.0.0 deduped
> | +-- uglify-js@3.17.4 overridden
> | `-- uri-path@1.0.0
> ...

Garneauma (approved):

PR #9934, opened 2025-03-13
Multimedia: add support to define frame size for YouTube videos

Garneauma:
Plugin now handles the "width" and "height" defined on the video element to resize the YouTube frame accordingly.

Also removed some deprecated/removed YouTube API options and added the language setting so that the player's interface matches the page's language.

Changes related to WET-526

### DIFF
--- a/src/plugins/multimedia/multimedia-youtube-en.hbs
+++ b/src/plugins/multimedia/multimedia-youtube-en.hbs
@@ -133,3 +133,29 @@
 		</details>
 	</section>
 </div>
+
+<div class="row">
+	<section class="col-md-6">
+		<h2>Video with defined video frame size</h2>
+		<figure class="wb-mltmd">
+			<video title="Suspect" width="720" height="1280">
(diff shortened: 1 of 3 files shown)


Garneauma:

duboisp (changes requested):
* [x] Add a working example of a portrait video instead of a square yt video. Like this video URL: 
  * EN: https://www.youtube.com/watch?v=4pe_UoNdj5s
  * FR: https://www.youtube.com/watch?v=YS6zbBaoSz0
  * Video player size: ~ `width="720" height="1280"`
* [ ] Can we set a max-height on the video that is bellow the height of the view port? That will avoid to have the top/bottom to be off screen for portrait video and have enough space to display the player command when those player command don't take more than 20% of the view port height. ("20%" value here is full arbitrary, please update as needed)
* [x] Can you check if it is possible to remove the "More videos" panel added by youtube when we pause the video? I don't see that behaviour on landscape video.

duboisp (approved):
Reviewed and I tested locally Work as expected.

This is a Minor change for the multimedia player. 
impact: Add support for youtube portrait video and refactoring of the video resizing code

Issue #9923, opened 2025-03-10
Referencing or forking for SideBySideImproved Flot plug-in


nschonni:
This is the last thing that in remaining till Bower can be removed from the build.

From https://github.com/wet-boew/wet-boew/pull/9890#discussion_r1972375138
> Noticing that there is a package.json in the dev branch https://github.com/pkoltermann/SideBySideImproved/tree/dev. Possible that it might work, but a fork to the WET org + adding the package.json is probably "safer"

@Garneauma / @duboisp  I think I still have permission to fork into this org, but I didn't really think that was a good idea. Do you have a preference on approach?

1. Do nothing, keep Bower. 
    - Might work for some time still, but Bower isn't receiving any support anymore
    - Upstream repo could be deleted by the maintainer, and then 
2. Fork the repo, add the package.json, and update the `dependencies` in NPM, drop Bower in CI + build. 
    - This is the "safest", but adds another repo in the org to maintain.
    - Quick search shows that maybe there might be some things were a fork might still be work having though like https://github.com/wet-boew/wet-boew/issues/8132.
    - This could be published to NPM or not
3. Reference the `dev` branch from upstream, and update the `dependencies` in NPM, drop Bower in CI + build.
    - This has the risk that the branch could be removed in the future and builds would start failing.

nschonni:
Looks like #9965 did Option 2

PR #9965, opened 2025-04-30
Charts: edit prepare task and remove Bower

Garneauma:
- Adding prepare Grunt task
- Updating prepare action
- Remove bower
- Moved SidebySideImproved to npm dependencies

This is a patch change for the Charts plugin.


.github/workflows/ci.yml
```diff
@@ -36,3 +36,4 @@ jobs:
         run: npm ci
 
       - run: npm test
+      - run: grunt deploy-build
```
duboisp, line 39: Can we avoid to re-build the item (JS/CSS/Assets) that was already built via the "npm run prepare" script?

duboisp (changes requested):
See inline comment.

Note: This is partial review.

.github/workflows/ci.yml
Garneauma: @duboisp Done. I have removed the tasks that are already executed in prepare from "deploy-build" since this task is only run in our CI/CD and is always prepended by the prepare task.

duboisp (approved):
Reviewed and tested,

Work as expected.

This is only a build change, no impact

package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "test": "grunt pre-mocha && node test-runner",
     "update-i18n": "node script/i18n.mjs",
-    "prepare": "bower update && grunt deploy-build"
+    "prepare": "grunt prepare"
```
duboisp, line 12: This break chart and graph regarding the use of the side-by-side-improve used to generate the bar chart

duboisp (changes requested):

duboisp (approved):
Review and tested locally,

Work as expected. The Bar Chart also work as expected.

This impact Patch: Charts and graph

nschonni:
If you're dropping the Bower part, I think this should get moved from `prepare` to `prepublishOnly` https://docs.npmjs.com/cli/v6/using-npm/scripts.
Building all the assets doesn't need to be done everytime `npm install` is run.

nschonni:
Closes #9942

duboisp:
We are dropping bower because it causing us some challenge because of our work computer restriction when we are building wet-boew on it.

nschonni:
Yup, I agree with dropping Bower, I also opened up #9923 and a draft PR #9942. What I'm flagging is, now that you're dropping Bower, you should reconsider the asset building you're doing in the `prepare` task, rather than expanding it's use.

Garneauma:
@nschonni We need the assets as we're fetching them from the `node_modules` directory when in GCWeb. From what I can see, the `prepublishOnly` is not run on npm install, so the assets would not be generated. Let me know if I am missing something!

PR #9966, opened 2025-05-02
Patch: Geomap - Fix build error about PhantomJS path not found

duboisp:

Fix a local build error about PhantomJs path not found. That error was caused by our pre-compiled version of Openlayer which tried to re-compile. Openlayer is used by Geomap. I also make it part of our list of dependencies as that libraries is used directly from our distributions.

This indirectly impact Geomap (no code change) as we need to ensure the expected 3rd party libraries are still copied over­ as expected and confirming the build of overlay-dist was not necessary.


Garneauma (approved):
Tested locally and works the same as before. This is a patch change.

PR #9987, opened 2025-05-29
PII Scrub: escaping HTML before generating modal

Garneauma:
This PR adds escaping of any HTML entered by the user in the text box so that it is not rendered in the generated modal.


duboisp (changes requested):
It works but the redacted text image should not be escaped.
I should not see this:

```
<span role='img' aria-label='redacted'>
```

Garneauma:
@duboisp Fixed the issue about the span wrapping the scrubbed data being escaped.

duboisp (approved):
Reviewed and tested,

Work as expected.

This is a Patch change to the PII Scrub

PR #10006, opened 2025-07-29
Filter: adding support for items using shadowDOM

Garneauma:
Adding support for GCDS elements using ShadowDOM in the Filter plugin.

Working example will be added in GCWeb.

### DIFF
--- a/src/plugins/filter/filter-en.hbs
+++ b/src/plugins/filter/filter-en.hbs
@@ -7,7 +7,8 @@
 	"tag": "filter",
 	"parentdir": "filter",
 	"altLangPrefix": "filter",
-	"dateModified": "2018-05-23"
+	"dateModified": "2025-08-07",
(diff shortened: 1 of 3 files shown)


duboisp (changes requested):
I tested and it don't work as expected.

The filter plugin show no entries and I think it is because of line 210 (`$items = $elm.find( secSelector + settings.selector ),`) because it get sanitized via our DOM Purify config.

What I did was to use the gcds-card working example on GCWeb, I added the filter plugin by wrapping it with a div and I added this configuration `data-wb-filter='{"selector":"gcds-card"}'`. I suggest to create a standalone working example in wet-boew that add a GCDS card and use the filter. It will make it easier to test too.

@tristandasilva for your info because we will need this feature to support the GCWeb home page redesign that you are working on.

Garneauma:
@duboisp Added working example. It does work as expected.

duboisp (approved):
Reviewed and tested locally

Work as expected when the filter is configured directly on the gcds-grid container, this is a minor change.

PR #10008, opened 2025-07-30
Steps form: Remove GCWeb classes from quiz examples

EricDunsworth:
This plugin's quiz feature started out as a méli-mélo project in GCWeb that was ultimately ported upstream (into WET).

However, some of the classes its examples originally used in GCWeb don't exist in WET. Therefore, there's no value in keeping them in WET's quiz examples.

Specifically:
* ``lst-spcd-2``
  * Note: The standard ``lst-spcd`` class exists in WET. [GCWeb's ``lst-spcd-2`` class styles also contain a note mulling about moving it over to WET](https://github.com/wet-boew/GCWeb/blob/c0fdaa7df0c7fc99efd3bf92ba721cbd8a06886f/common/list/_lists.scss#L115-L117) (unclear why it wasn't initially introduced in WET in the first place).
* ``gc-chckbxrdio``
*   * Note: This class overrides the behaviour of Bootstrap's ``checkbox`` and ``radio`` classes. Bootstrap's classes expect ``input``s to be nested within ``label``s, whereas GCWeb's ``gc-chckbxrdio`` class expects them to be adjacent. Using GCWeb's structure in WET causes large spaces to appear between ``input``s and ``label``s.

This fixes it revising WET's quiz examples to use "stacked" checkbox/radio buttons with ``li`` elements. It's the closest WET equivalent to what GCWeb was originally doing.

### DIFF
--- a/src/plugins/stepsform/stepsform-en.hbs
+++ b/src/plugins/stepsform/stepsform-en.hbs
@@ -7,7 +7,7 @@
 	"tag": "stepsform",
 	"parentdir": "stepsform",
 	"altLangPrefix": "stepsform",
-	"dateModified": "2025-06-02"
+	"dateModified": "2025-08-04"
 }
 ---
 <section>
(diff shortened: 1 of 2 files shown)


Garneauma:
Pre-approved upon successful review and adjustment of the examples.

EricDunsworth:
@Garneauma Just adjusted the examples, should be good to go now.

Garneauma (approved):

PR #10010, opened 2025-08-01
Steps form: fixing broken conditional

Garneauma:
This conditional always returned true, which caused and issue when the latest version of WET-BOEW and the Nahanni compilation in GCWeb were both on the same page.


duboisp (approved):
Manually tested via my GCWeb testing environment. I was able to reproduce the issue and I do confirm this fix the double progress bar.

This change can be merged.

PR #10011, opened 2025-08-01
PII Scrub: fix issue when closing modal with Escape button

Garneauma:
An issue was introduced with PR #9957 where when a user would close the PII modal using the Escape key, the form would be submitted since the confirm button had the class "popup-modal-dismiss". The class has been removed and instead thje modal is closed via the available method.

A temporary workaround has also been added to leave authors some time to implement this fix.

### DIFF
--- a/site/pages/docs/ref/pii-scrub/pii-scrub-fr.hbs
+++ b/site/pages/docs/ref/pii-scrub/pii-scrub-fr.hbs
@@ -80,7 +80,7 @@
 					<ul>
 						<li>L'espace réservé pour les champs IPI&nbsp;: <code>&lt;div data-scrub-modal-fields>&lt;/div></code>. Cet espace réservé sera rempli avec la liste des champs IPI et leur libellé associé.</li>
 						<li>Un bouton Annuler. Ce bouton doit avoir la classe CSS <code>popup-modal-dismiss</code>. Par exemple&nbsp;: <code>&lt;button type="button" class="btn btn-link btn-block <strong>popup-modal-dismiss</strong>">Revenir en arrière et modifier les champs&lt;/button></code></li>
-						<li>Un bouton Soumettre. Ce bouton doit avoir l'attribut <code>data-scrub-submit</code> et la classe CSS <code>popup-modal-dismiss</code>. Par exemple&nbsp;: <code>&lt;button type="button" class="btn btn-primary btn-block <strong>popup-modal-dismiss</strong>" <strong>data-scrub-submit</strong>>Envoyer le commentaire&lt;/button></code></li>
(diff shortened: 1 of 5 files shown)


site/pages/docs/ref/pii-scrub/pii-scrub-fr.hbs
```diff
@@ -80,7 +80,7 @@
 					<ul>
 						<li>L'espace réservé pour les champs IPI&nbsp;: <code>&lt;div data-scrub-modal-fields>&lt;/div></code>. Cet espace réservé sera rempli avec la liste des champs IPI et leur libellé associé.</li>
 						<li>Un bouton Annuler. Ce bouton doit avoir la classe CSS <code>popup-modal-dismiss</code>. Par exemple&nbsp;: <code>&lt;button type="button" class="btn btn-link btn-block <strong>popup-modal-dismiss</strong>">Revenir en arrière et modifier les champs&lt;/button></code></li>
-						<li>Un bouton Soumettre. Ce bouton doit avoir l'attribut <code>data-scrub-submit</code> et la classe CSS <code>popup-modal-dismiss</code>. Par exemple&nbsp;: <code>&lt;button type="button" class="btn btn-primary btn-block <strong>popup-modal-dismiss</strong>" <strong>data-scrub-submit</strong>>Envoyer le commentaire&lt;/button></code></li>
+						<li>Un bouton Soumettre. Ce bouton doit avoir l'attribut <code>data-scrub-submit</code> et la classe CSS <code>popup-modal-dismiss</code>. Par exemple&nbsp;: <code>&lt;button type="button" class="btn btn-primary btn-block" <strong>data-scrub-submit</strong>>Envoyer le commentaire&lt;/button></code></li>
```
duboisp: To match the English instruction

```suggestion
						<li>Un bouton Soumettre. Ce bouton doit avoir l'attribut <code>data-scrub-submit</code>. Par exemple&nbsp;: <code>&lt;button type="button" class="btn btn-primary btn-block" <strong>data-scrub-submit</strong>>Envoyer le commentaire&lt;/button></code></li>
```

src/plugins/pii-scrub/pii-scrub.js
```diff
@@ -213,6 +213,11 @@ var $document = wb.doc,
 
 		// Add PII fields HTML if using a custom UI template
 		if ( modalTemplate ) {
+
+			// Temporary fix for implementers that added the "popup-modal-dismiss" class to the submit button
+			// Deprecate on version 18.0.0 of GCWeb and Remove on version 19.0.0
```
duboisp: Not sure if we will be able to get this line removed eventually, we are in wet-boew not GCWeb. There is no plan to apply any breaking change to wet-boew.

I think this comment need to be reviewed consider that.

I suggest to keep that workaround to avoid a breaking change but don't mention any removal version.

duboisp (changes requested):
Reviewed and tested. the code work as expected and it does fix the issue when closing the dialog with the escape keyboard key.

I manually tested the workaround and it does work as expected. But can you create a working example for that in a different page?

Currently as is, this is a Patch change.

Although, some edit are required in the documentation.

duboisp (approved):
Reviewed and tested, it works as expected, the documentation was updated and a working example was added.

This is a Patch

PR #10012, opened 2025-08-01
Steps form (quiz): Remove "phantom" spacing below legends

EricDunsworth:
To disable the "magic" layout of ``fieldset`` ``legend``s, WET floats ``legend``s to the left by default.

The steps form plugin's ``quiz`` option previously disabled ``legend`` floating to prevent layout issues with nested ``fieldset``s. But that in turn re-activated the ``legend``'s "magic" layout and caused an extra 10px of "phantom" space to appear below the current step's ``legend``.

This resolves it by simplifying the layout fix for nested ``fieldset``s:
* Remove the ``legend``'s ``float: none;`` and ``border/padding-top`` overrides
* Add ``clear: left;`` to any elements adjacent to ``legend``s to suppress unusual floating behaviour

### DIFF
--- a/src/plugins/stepsform/_base.scss
+++ b/src/plugins/stepsform/_base.scss
@@ -46,18 +46,10 @@
 
 
 		fieldset {
-			&:not(:first-child) legend {
-				border-top: 1px solid #e5e5e5;
-				padding-top: 10px;
-			}
-
-			&:not(:first-child):not(:has(legend)) {
-				border-top: 1px solid #e5e5e5;
-				padding-top: 10px;
-			}
-
 			legend {
-				float: none;
+				& + * {
+					clear: left;
+				}
 			}
 		}
 


EricDunsworth:
@Garneauma Would it be better to take an "all-in" or "hands-off" approach?

WET's form validation plugin and "normal" steps form implementations are both affected by this issue. But this PR and the old fix it replaces only target steps form quizzes :S.

The issue only seems to affect WET due to a [global override used to float ``legend``s to the left](https://github.com/wet-boew/wet-boew/blob/127aa1d85746a3d13cedf3fb1f3d226bdd355df2/src/base/bootstrap-overrides/_core-forms.scss#L73-L85). It doesn't seem to be an issue in Bootstrap. It can be manually-fixed by adding ``<div class="clearfix"></div>`` in-between elements that float strangely. But another way of dealing with it would be to make this PR's fix work globally...

Thoughts?

Garneauma:

Garneauma (approved):

PR #10020, opened 2025-08-13
jQuery 4 - Fix flot charts and Replace $.trim with native String.prototype.trim.

joejoseph00:

We have upgraded a project to Drupal 11 using jQuery 4, and wet-boew `8.0.85`. After the upgrade, the client's flot charts are now broken. There were two issues, one was wet-boew was using `$.trim` which is no longer available in jQuery4. Another issue is that flot was using `$.trim` as well, and that project seems to be unsupported now.

So @smulvih2 has made the following changes in a local patch to my project and it fixes the charts; they now render using jQuery 4.

    * Replaced all instances of `$.trim` with native `.trim()` to remove jQuery dependency
 
    * Applied patch to flot v0.8.3 using patch-package to support jQuery 4
 
    * Added postinstall script to ensure patch is applied on install


olstjos:
@Garneauma There's absolutely nothing that can be done for the flot patch lint since we cannot change it or the patch will cease to apply correctly. Do you have a skip list or ignore list for the linting?  We should add the flot patch to an ignore.

Garneauma (changes requested):
Remove the spaces at the beginning of lines 6, 7, 8, 12, and 13. This will fix the linting issues.

Please re-run npm install to make sure all the changes are present in the package-lock.json. When I ran it, it added `"hasInstallScript": true`.

Outside of that, I'm ok with the changes.

joejoseph00:
@Garneauma  all checks have now passed.  Please review!

Garneauma (approved):

PR #10030, opened 2025-08-27
Details close: stabilizing component

Garneauma:


src/plugins/details-close/details-close.js
```diff
@@ -30,46 +29,24 @@ var componentName = "wb-details-close",
...
-			// Get the plugin JSON configuration set on attribute data-wb-details-close
-			// Will define one set settings for all .wb-details-close on the page
-			breakpoint = $elm.data( "breakpoint" ) || "sm";
-
-			// reset breakpoint if config is passed
-			if ( views.length === viewsClass.length ) {
-				i = views.indexOf( breakpoint );
-				viewsClass = viewsClass.slice( 0, i + 1 );
+			let $elm = $( elm ),
+				breakpoint = elm.dataset.breakpoint || "sm", // Get the target breakpoint from data attribute or default to "sm"
```
duboisp, line 36: Should we support an array of breakpoint? like by default xxs, xs and sm?

duboisp (changes requested):
Question: Should we have the plugin to be executed during screen resize if the impacted details don't contain the focus neither contain the active element? Or also when it is after the current view port? What do you think?

duboisp:
I did a quick test and I see it select the current up to the larger view. I think we would probably want the reverse because I assume that plugin is going to be use to save space space in content for small device.

Let's discuss but I see it like when the medium config is set, it will be closed for "xxs, xs, sm md" not be closed for "md, lg". 
Also it don't change on screen resize, but what about to only trigger a change when the view port change from 2 view port distant. For example switching from lg to md wont trigger a change, but changing from lg to sm (or lower) will trigger a change. what do you think?

duboisp (approved):
Looks good, I was confused this morning.

It is actually working as expected. So if someone configure "medium" that detail will be close from x-tra small device up to medium (inclusivly) which is the expected behaviour.

I also reviewed all the code that goes beyond this PR which relate to the stabilisation of the details/close component.

Cheers

PR #10037, opened 2025-09-12
Allow leading and trailing whitespaces during postal code validation

tristandasilva:
Jira Ticket [WET-561](https://jtickets.atlassian.net/browse/WET-561)


Garneauma (changes requested):
As discussed, hook on the postalCodeCA rule instead of the field itself.

tristandasilva:
@Garneauma I was able to override the default validation method, should work as expected now :)

Garneauma (approved):

BeraJosh (commented):
Approved

PR #10055, opened 2025-10-03
Tabbed interface: Use flexbox for tab markers

EricDunsworth:
Previously, in medium view and over, if any tab markers wrapped across multiple lines, "short" tab markers behaved strangely compared to "tall" ones.

Specifically:
* Short tab markers didn't use 100% height and looked bottom-aligned
* If the active tab used a short tab marker:
  * The marker's thick top grey border wasn't top-aligned (vs tall tab markers)
  * A thin grey border appeared above the marker
* The "linked area" of inactive short tab markers was misleading (looked as if marker links used 100% height, but didn't really):
  * Caused by hover/focus effects (like background colours and ``cursor: pointer;``) being set on ``li`` elements instead of ``a`` elements.

This resolves it by:
* Adding a second line to the third tab marker ("Example 3"/"Exemple 3") in the tabbed interface demo's first example
* Adjusting tab markers to use flexbox (to make short tab markers match the height/alignment of tall ones)
* Adjusting margins where necessary
* Tying hover/focus styles solely to links (not their parent elements)
* Removing irrelevant hover/focus styles (e.g. cursor properties, placeholder selector for a hidden element)

Notes:
* Flexbox changed carousel spacing slightly (for the better):
  * Carousel style 1:
    * Removed 10px space above the carousel
    * Reduced spacing between the next and play/pause buttons from 10px to 5px
  * Carousel style 2:
    * Added 10px space between previous button and tab counter (matches space between counter and next button) (deliberate change)
    * Fully-aligned play/pause button to the right (previously had extra space on the far right side)
    * Added hover background colour to "show-thumbs" class' thumbnails (byproduct of hover/focus style changes)

**Screenshots (Firefox):**
* **Tabbed interface - Active short tab marker**
  * **Before:** tabs-active-before
  * **After:** tabs-active-after
* **Tabbed interface - Inactive short tab marker (tab 3 active, hovering over tab 2's marker, dashed red borders illustrate linked areas)**
  * **Before:** tabs-inactive-before
  * **After:** tabs-inactive-after
* **Carousel style 1**
  * **Before:** carousel-style1-before
  * **After:** carousel-style1-after
* **Carousel style 2**
  * **Before:** carousel-style2-before
  * **After:** carousel-style2-after

### DIFF
--- a/src/plugins/tabs/_base.scss
+++ b/src/plugins/tabs/_base.scss
@@ -141,11 +141,6 @@ $tab-margin-width: 10px;
 	box-shadow: none;
 }
 
-%tabs-carousel-border-top-none-padding-top-1 {
-	border-top: 0;
-	padding-top: 10px;
-}
-
 %tabs-display-inline-block {
 	display: inline-block;
 }
@@ -208,7 +203,8 @@ $tab-margin-width: 10px;
 
 	[role="tablist"] {
 		border-spacing: $tab-margin-width 0;
-		display: table;
+		display: flex;
+		flex-wrap: wrap;
 		list-style: none;
 		margin: 0;
 		padding: 0;
@@ -220,9 +216,7 @@ $tab-margin-width: 10px;
 
 				background: $tablist-bg-color;
 				color: $carousel-tablist-tabcount-color;
-				cursor: pointer;
-				display: table-cell;
-				left: -$tab-margin-width;
+				display: flex;
 				position: relative;
 				text-align: center;
 				border: {
@@ -239,31 +233,30 @@ $tab-margin-width: 10px;
 					// changed padding and margin from em to px to fix web-kit layout issue
 					padding: 10px;
 					text-decoration: none;
-				}
 
-				&:focus,
-				&:hover {
-					background: $tablist-hover-bg-color;
-					background: color.adjust($tablist-hover-bg-color, $alpha: -.1);
+					&:focus,
+					&:hover {
+						background: $tablist-hover-bg-color;
+						background: color.adjust($tablist-hover-bg-color, $alpha: -.1);
+					}
 				}
 
 				&.active {
-					@if $tablist-active-bg-color {
-						background: $tablist-active-bg-color;
-					} @else if $main-bg {
-						background: $main-bg
-					} @else if $content-bg {
-						background: $content-bg
-					} @else {
-						background: $body-bg;
-					}
-
 					border-bottom: 0;
-					cursor: default;
 					padding-bottom: 1px;
 					z-index: 2;
 
 					a {
+						@if $tablist-active-bg-color {
+							background: $tablist-active-bg-color;
+						} @else if $main-bg {
+							background: $main-bg
+						} @else if $content-bg {
+							background: $content-bg
+						} @else {
+							background: $body-bg;
+						}
+
 						border: {
 							color: $tablist-active-link-border-color;
 							style: $tablist-active-link-border-style;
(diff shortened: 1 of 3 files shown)


src/plugins/tabs/_base.scss
```diff
@@ -141,11 +141,6 @@ $tab-margin-width: 10px;
 	box-shadow: none;
 }
 
-%tabs-carousel-border-top-none-padding-top-1 {
-	border-top: 0;
-	padding-top: 10px;
-}
-
```
EricDunsworth, line 148: This placeholder selector's properties appear to have been meant for carousel style 1's tab markers... which are always set to ``display: none;``. So the placeholder selector was pointless.

Only noticed this because it was tied to a suspicious-looking hover/focus selector (wasn't used on an interactive element).

src/plugins/tabs/_base.scss
```diff
@@ -348,30 +339,11 @@ $tab-margin-width: 10px;
...
-							cursor: default;
-						}
-					}
-
-					&.active,
-					&:focus,
-					&:hover {
-						a {
-							@extend %tabs-carousel-border-top-none-padding-top-1;
-						}
```
EricDunsworth, line 374: Here's where that sketchy placeholder selector was previously being used.

EricDunsworth:
* Fixed a regression that caused hover styles to appear on the active tab marker (hover effects aren't supposed to apply to it)
* Replaced ``10px`` in the ``&.tab-count`` selector with a reference to the ``$tab-margin-width`` SCSS variable
* Adjusted carousel style 2's notes at the end of the commit message/OP
* Added screenshots to the OP to clearly visualize what's changed
* Adjusted demo modified dates

As far as this PR's impacts on GCWeb go...
* I haven't attempted to test it out, but don't expect it to break anything
* [GCWeb contains a handful of tabbed interface/carousel overrides](https://github.com/wet-boew/GCWeb/tree/master/components/tabs), but none appear likely to clash with this PR in practice
* I have concerns about potential clashes with [components/tabs/_screen-xxs-max.scss](https://github.com/wet-boew/GCWeb/blob/master/components/tabs/_screen-xxs-max.scss)'s ``.plypause`` and ``.tab-count`` overrides... but those selectors appear to be used in a deprecated use case, are very targeted and broken atm:
  * They're tied to a ``prm-flpr`` parent class ([only sign of its carousel implementation is in a deprecated code sample in the news template](https://wet-boew.github.io/GCWeb/templates/news/news-doc-en.html#:~:text=prm-flpr))
  * Selector rules are malformed (they don't take effect even when the parent class is present... caused by [this commit](https://github.com/wet-boew/GCWeb/commit/a343ead7ec9196b56cfe08ade5b142d9ab729b4d#diff-ec2814ee54254957f3fa8f0e18e44901f6d0ca9acc81a742f97936e02bf939d8L8-R16) via wet-boew/GCWeb#675).
  * @Garneauma Would there be any value in trying to fix those selector rules and make them compatible with this PR's changes? Or should the selectors be removed entirely?

EricDunsworth:
Just force-pushed a minor update:
* Added ``flex-wrap: wrap;`` as a precaution
* Removed ``cursor: default;`` property from an ``&.active`` selector (which would map to ``li.active`` in practice...):
  * That property is no longer useful given that hover/focus selectors are now solely tied to links
  * Kept it in place for links situated _inside_ ``&.active`` (aka current tab marker)

Garneauma (approved):

PR #10056, opened 2025-10-07
Deprecation of Twitter plugin

tristandasilva:
Deprecation of Twitter plugin + fallback mechanism to verify if the issue still exists.

Jira Ticket [WET-579](https://jtickets.atlassian.net/browse/WET-579)

Related #10047

### DIFF
--- a/src/plugins/twitter/twitter.js
+++ b/src/plugins/twitter/twitter.js
@@ -43,7 +43,6 @@ var componentName = "wb-twitter",
 
 			// Process the Twitter link
 			if ( twitterLink ) {
-				const loadingDiv = document.createElement( "div" );
 				let observer;
 
 				// Only initialize the i18nText once
@@ -88,10 +87,6 @@ var componentName = "wb-twitter",
 					twitterLink.dataset.dnt = "true";
 				}
 
-				// Add a loading icon below the link
-				loadingDiv.className = "twitter-timeline-loading";
-				twitterLink.after( loadingDiv );
-
 				// Observe DOM mutations
 				observer = new MutationObserver( function( mutations ) {
 					mutations.forEach( function( mutation ) {
@@ -115,12 +110,11 @@ var componentName = "wb-twitter",
 							case "childList": {
 								mutation.removedNodes.forEach( function( removedNode ) {
 
-									// If the removed node was a Twitter link, remove its adjacent loading icon, add skip links and stop observing
-									// Note: Twitter's widget script removes "a.twitter-timeline" upon displaying the timeline iframe's content... at which point the loading icon is no longer useful
-									if ( removedNode === twitterLink && mutation.nextSibling === loadingDiv ) {
-										const iframeContainer = loadingDiv.previousElementSibling;
+									// If the removed node was a Twitter link add skip links and stop observing
+									// Note: Twitter's widget script removes "a.twitter-timeline" upon displaying the timeline iframe's content
+									if ( removedNode === twitterLink ) {
+										const iframeContainer = eventTarget.querySelector( "div.twitter-timeline" );
 
-										loadingDiv.remove();
 										addSkipLinks( iframeContainer );
 
 										// The following 2 lines were added as a workaround in Safari where the iFrame is not displayed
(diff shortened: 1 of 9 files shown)


BeraJosh (approved):
In the src/plugins/twitter/twitter.js on line 163 I see it says 5000. Just want to ensure this is correct as above was changed from 5000 to 6000 in src/plugins/twitter/test.js.

src/plugins/twitter/twitter.js
```diff
@@ -143,14 +143,42 @@ var componentName = "wb-twitter",
...
+
+				// The script will allow 5 seconds for the timeline to load, after which it will fallback to a link to the Twitter page
+				const timeout = setTimeout( function() {
+					if ( !timelineLoaded ) {
+						const fallbackLink = document.createElement( "a" );
+						const fallbackLinkText = eventTarget.querySelector( "a.twitter-timeline" ).textContent || i18nText.timelineTitle;
+						fallbackLink.href = twitterLink.href;
+						fallbackLink.textContent = fallbackLinkText;
+						eventTarget.innerHTML = "";
+						eventTarget.appendChild( fallbackLink );
```
EricDunsworth: It isn't really necessary to programmatically recreate the ``@`` link here.

Twitter/X's timeline script will only remove the hardcoded ``@`` link if it decides to "fill-in" and reveal the the timeline ``iframe``. That'll never happen if X's rate limit gets exceeded.

src/plugins/twitter/twitter.js
```diff
@@ -143,14 +143,42 @@ var componentName = "wb-twitter",
...
+						wb.ready( $( eventTarget ), componentName );
+					}
+				}, 5000 );
```
EricDunsworth: A 5 second timeout might be too aggressive.

AFAIK X's timeline widgets can get populated with up to 100 tweets - which can take a while to load if they're packed with media. IIRC before X started imposing rate limits, I believe certain timelines could potentially take way more than 5 seconds to load. Plus there's always a risk of slower connections taking longer.

src/plugins/twitter/twitter.js
EricDunsworth, line 1: FWIW all these logic changes can be avoided by scrapping the ``loadingDiv`` variable.

The ``loadingDiv`` variable is tied to a ``div`` with a made-up class name. The ``twitter-timeline-loading`` class doesn't come from X's widget script - it's a class name I improvised (because I couldn't get away with ``::before/::after`` :D).

If you try renaming the class to something else on line 92 in the plugin's pre-existing logic, the endless loading icons will go away. ``wb.ready`` will still work too since Modernizr only checks if X's third party script loads - but doesn't care if it fails to accomplish its purpose :P.

EricDunsworth:
@tristandasilva

> Thanks for all the info :) ! Taking into consideration what you mentioned about the timeline requiring a longer duration to attempt loading, I allowed 1 minute before resorting to a fallback message.

Awesome :)!

> I was never able to get the widget working on either the corporate or personal connection, so if you have any more insight apart from what you’ve already mentioned regarding that please feel free to share!

Try these steps:
1. Open your browser's network inspector (F12 - Network)
2. Visit https://wet-boew.github.io/v4.0-ci/demos/twitter/twitter-en.html
3. Enter ``https://syndication.twitter.com/srv/timeline-profile/screen-name/`` in your network panel's search filter... should only get 2 hits
4. Look for ``x-rate-limit-reset`` in the HTTP response headers. It represents a time (in UTC epoch seconds) for when X's rate limiter will reset. Copy/paste it into a site like https://www.epochconverter.com/ to convert it to your local time.
5. If you ran into endless loading icons and got HTTP 429 errors in step 3's network requests... wait until around a minute before the reset time, then start refreshing the demo page over and over and hope for the best. The goal is to get the feeds to load before anyone else in your corporate network, _right_ after the rate limit resets :P...
6. If the timelines loaded, check-out the ``x-rate-limit-limit`` and ``x-rate-limit-remaining`` headers to see how close you got to surpassing the limit :P.

> As for the rate limiting, from what I can see there doesn’t seem to be a solid solution for embedded content, the only potential solution I could see would be to use server-side caching and rendering with the twitter API, but this would require an infrastructure change to the plugin, twitter API access (likely requiring a subscription for our needs) and would be a bit more complex to maintain...but again, **if you have any more insights to resolve the rate limiting issue please feel free :)**

My suggestions would be:
* Don't expect the widget to work in corporate networks 99% of the time :P
* Hope X relaxes or eliminates the timeline plugin's rate limit in the future :P (it didn't used to be subject to rate limits at all)
* This isn't really a suggestion, but even when the timelines appear correctly, X seems to treat verified and non-verified accounts differently.
  
  For example:
  * ``@WebExpToolkit`` (unverified)'s timeline shows a misleading filler message ("**Nothing to see here - yet.** When they post, their posts will show up here."). Its profile page (https://twitter.com/WebExpToolkit) also claims it hasn't tweeted anything. However, [XCancel](https://news.ycombinator.com/item?id=41546640)'s mirror of it (https://xcancel.com/WebExpToolkit) shows many older tweets.
  * ``@Canada`` (verified)'s timeline and profile (https://twitter.com/Canada) both show recent tweets.

src/plugins/twitter/twitter.js
```diff
@@ -141,16 +135,26 @@ var componentName = "wb-twitter",
...
-			} );
+			}
+
+			// Load the Twitter widget script
+			const script = document.createElement( "script" );
+			script.src = ( protocol.indexOf( "http" ) === -1 ? "http:" : protocol ) + "//platform.twitter.com/widgets.js";
+			script.onload = function() {
+				wb.ready( $( eventTarget ), componentName );
+			};
+			document.head.appendChild( script );
```
Garneauma: This logic seems to add a the script to the head everytime the plugin is called. Please wrap this in a condition whether the script has already been loaded or not.

Garneauma (changes requested):

Garneauma (approved):

Garneauma:
@BeraJosh For your review.

BeraJosh (approved):
No issues observed from change version.

PR #10069, opened 2025-10-28
Multimedia Player: added default cc functionality

BrahimMahadi:
Added closed captions when video is played functionality to the multimedia player component as well as added working examples and updated documentation. Reference JIRA ticket WET-592.

### DIFF
--- a/src/plugins/multimedia/_base.scss
+++ b/src/plugins/multimedia/_base.scss
@@ -117,10 +117,12 @@
 	}
 
 	&.cc_on {
-		.wb-mm-cc {
-			display: table;
-			height: calc(#{$line-height-base * $mm-cc-lines}em + #{$mm-cc-padding * 2});
-			padding: $mm-cc-padding;
+		&.played:not(.youtube) {
+			.wb-mm-cc {
+				display: table;
+				height: calc(#{$line-height-base * $mm-cc-lines}em + #{$mm-cc-padding * 2});
+				padding: $mm-cc-padding;
+			}
 		}
 
 		&:not(.errmsg) {
--- a/src/plugins/multimedia/multimedia.js
+++ b/src/plugins/multimedia/multimedia.js
@@ -27,6 +27,8 @@ var componentName = "wb-mltmd",
 	templateLoadedEvent = "templateloaded" + selector,
 	cuepointEvent = "cuepoint" + selector,
 	captionClass = "cc_on",
+	firstPlayClass = "played",
+	defaultCaptions = false,
 	multimediaEvents = [
 		"durationchange",
 		"playing",
@@ -581,7 +583,9 @@ var componentName = "wb-mltmd",
 			},
 			$mltmPlayerElm,
 			mltmPlayerElm,
-			isMuted;
+			isMuted,
+			settings,
+			$ccButton;
 
 		switch ( event.data ) {
 			case null: // init
@@ -591,11 +595,29 @@ var componentName = "wb-mltmd",
 
 				// Put video on mute if the video is muted on init, run once
 				$mltmPlayerElm = $media.parentsUntil( selector ).parent();
+				mltmPlayerElm = $mltmPlayerElm.get( 0 );
+
 
 				// Mute the player, GUI
 				if ( $mltmPlayerElm.data( "putMutedOnInit" ) ) {
 					youTubeApi.call( $mltmPlayerElm.get( 0 ), "setMuted", true );
 					$mltmPlayerElm.data( "putMutedOnInit", false );
+				}
+
+				// Check if closed captions should be enabled by default
+				settings = wb.getData( mltmPlayerElm, componentName );
+
+				if ( settings !== undef ) {
+					if ( settings.closedCaptions !== null && settings.closedCaptions === true ) {
+						youTubeApi.call( mltmPlayerElm, "setCaptionsVisible", true );
+
+						$ccButton = $( mltmPlayerElm ).find( ".cc" );
+
+						$ccButton.attr( {
+							"title": i18nText.cc_off,
+							"aria-pressed": true
+						} );
+					}
 				}
 				break;
 			case -1:
@@ -706,6 +728,10 @@ $document.on( initializedEvent, selector, function( event ) {
 		if ( settings !== undef ) {
 			data.shareUrl = settings.shareUrl;
 			data.fullscreen = settings.fullscreenBtn || false;
+			data.closedCaptions = settings.closedCaptions || false;
+			if ( data.closedCaptions !== undef ) {
+				defaultCaptions = data.closedCaptions;
+			}
 		}
 
 		$this.addClass( type );
@@ -868,7 +894,9 @@ $document.on( renderUIEvent, selector, function( event, type, data ) {
 			captionsUrl = wb.getUrlParts( data.captions ),
 			currentUrl = wb.getUrlParts( window.location.href ),
 			$media = data.media,
-			$eventReceiver;
+			$eventReceiver,
+			$button,
+			$buttonSpan;
 
 		$media
 			.after( tmpl( template, data ) )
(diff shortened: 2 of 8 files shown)


Garneauma (changes requested):
It works well for regular videos, however it does not work for YouTube videos. Please make sure it works for both.

src/plugins/multimedia/multimedia.js
```diff
@@ -581,7 +582,8 @@ var componentName = "wb-mltmd",
 			},
 			$mltmPlayerElm,
 			mltmPlayerElm,
-			isMuted;
+			isMuted,
```
Garneauma, line 586: You should get the settings using the `wb.getData()`. See around line 700 how it is used in the initializedEvent.

src/plugins/multimedia/multimedia.js
```diff
@@ -612,6 +614,18 @@ var componentName = "wb-mltmd",
...
+				captionStatus = mltmPlayerElm.getAttribute( "data-wb-mltmd" );
+
+				if ( captionStatus !== null ) {
+
+					captionStatus = captionStatus.replace( /"/g, "" );
+
+					if ( captionStatus.includes( "{closedCaptions: true}" ) ) {
+						youTubeApi.call( mltmPlayerElm, "setCaptionsVisible", true );
+					}
+				}
```
Garneauma: After getting the settings through the `wb.getData()` function, you should be able to simply check the value like `settings.closedCaptions == true`.

src/plugins/multimedia/multimedia.js
```diff
@@ -1064,14 +1086,31 @@ $document.on( multimediaEvents, selector, function( event, simulated ) {
...
 				if ( eventType === "ended" ) {
 					this.loading = clearTimeout( this.loading );
 				}
 				$this.removeClass( "playing" );
 			}
 			$button
-				.attr( "title", buttonData )
+				.attr( {
+					title: buttonData,
+					"aria-pressed": isPlay
```
Garneauma: Why was this added? This would make the "Pause" button as "pressed" when the video is playing.

src/plugins/multimedia/multimedia.js
```diff
@@ -1064,14 +1086,31 @@ $document.on( multimediaEvents, selector, function( event, simulated ) {
 			if ( isPlay ) {
 				$this.addClass( "playing" );
 				$this.find( ".progress" ).addClass( "active" );
+
```
Garneauma: The attribute "aria-pressed" should be set to true on page load. Not only when the video starts. The `<figure>` element should also have the class "cc_on" so the styles of the button are applied.

src/plugins/multimedia/multimedia.js
```diff
@@ -1064,14 +1086,31 @@ $document.on( multimediaEvents, selector, function( event, simulated ) {
...
+				if ( $this.data( "wbMltmd" ) !== undef ) {
+					captionStatus = $this.data( "wbMltmd" ).closedCaptions;
+				}
+
+				if ( captionStatus !== undef && defaultCaptions !== undef  && !defaultCaptions ) {
+					$this.addClass( captionClass );
+
+					// Trigger caption visibility change event only on the first time the video plays
+					defaultCaptions = true;
+				}
```
Garneauma: We should add the class and attribute aria-pressed on page load instead of on first play. So, please move this to the "renderUIEvent" function. Also, this code does not work if, for example, a user adds "closedCaptions: false".

For the black bar for regular videos, you could add a CSS class the first time the video is played and show the CC when only this class is added and the class "cc_on" is also there. For example: "played". That way, the CC bar would not be visible on page load, but would not hide when clicking pause.

Garneauma (changes requested):

src/plugins/multimedia/multimedia.js
BrahimMahadi: Fixed in amended commit.

src/plugins/multimedia/multimedia.js
BrahimMahadi: Fixed in amended commit.

src/plugins/multimedia/multimedia.js
BrahimMahadi: Fixed in amended commit.

src/plugins/multimedia/multimedia.js
BrahimMahadi: This was meant for the closed captions button, it has been removed from the play/pause button.

Garneauma (changes requested):
@BrahimMahadi Please test the Youtube version. Let me know if everything works as expected.

Garneauma (changes requested):
2 things are still not handled correctly:
- For Youtube, the button is not selected by default (aria-pressed=true).
- For both Youtube and not Youtube, the title of the button needs to be changed to "Hide closed captioning" on page load since it is selected by default.

src/plugins/multimedia/_base.scss
```diff
@@ -117,10 +117,14 @@
...
-			padding: $mm-cc-padding;
+		&.played {
+			&:not(.youtube) {
+				.wb-mm-cc {
+				display: table;
+				height: calc(#{$line-height-base * $mm-cc-lines}em + #{$mm-cc-padding * 2});
+				padding: $mm-cc-padding;
+			}
+			}
 		}
```
Garneauma: Merge these two.
```suggestion
		&.played:not(.youtube) {
			.wb-mm-cc {
				display: table;
				height: calc(#{$line-height-base * $mm-cc-lines}em + #{$mm-cc-padding * 2});
				padding: $mm-cc-padding;
			}
		}
```

src/plugins/multimedia/multimedia.js
```diff
@@ -591,12 +595,28 @@ var componentName = "wb-mltmd",
 
 				// Put video on mute if the video is muted on init, run once
 				$mltmPlayerElm = $media.parentsUntil( selector ).parent();
+				mltmPlayerElm = $mltmPlayerElm.get( 0 );
```
Garneauma, line 598: What's the difference between these two?

src/plugins/multimedia/multimedia.js
```diff
@@ -591,12 +595,28 @@ var componentName = "wb-mltmd",
...
+
 
 				// Mute the player, GUI
 				if ( $mltmPlayerElm.data( "putMutedOnInit" ) ) {
 					youTubeApi.call( $mltmPlayerElm.get( 0 ), "setMuted", true );
 					$mltmPlayerElm.data( "putMutedOnInit", false );
 				}
+
+				// Check if closed captions should be enabled by default
+				settings = wb.getData( mltmPlayerElm, componentName );
```
Garneauma, line 608: I get a JS error in the console because of this line. Please investigate.

`Uncaught TypeError: Cannot read properties of undefined (reading 'closedCaptions')`

src/plugins/multimedia/multimedia.js
```diff
@@ -926,6 +952,32 @@ $document.on( renderUIEvent, selector, function( event, type, data ) {
...
+			$this.addClass( captionClass );
+
+			// Trigger caption visibility change event only on the first time the video plays
+			defaultCaptions = true;
+
+			$button = $this.find( ".cc" );
+
+			$button
+				.attr( {
+					"title": "Hide closed captioning",
```
Garneauma: What about in French?

src/plugins/multimedia/multimedia.js
```diff
@@ -591,12 +595,28 @@ var componentName = "wb-mltmd",
...
+				// Check if closed captions should be enabled by default
+				settings = wb.getData( mltmPlayerElm, componentName );
+
+				if ( settings.closedCaptions !== null && settings.closedCaptions === true ) {
+					youTubeApi.call( mltmPlayerElm, "setCaptionsVisible", true );
+
+					$ccButton = $( mltmPlayerElm ).find( ".cc" );
+
+					$ccButton.attr( {
+						"title": "Hide closed captioning",
```
Garneauma: What about French?

src/plugins/multimedia/multimedia.js
```diff
@@ -926,6 +952,32 @@ $document.on( renderUIEvent, selector, function( event, type, data ) {
...
+			$button = $this.find( ".cc" );
+
+			$button
+				.attr( {
+					"title": "Hide closed captioning",
+					"aria-pressed": true
+				} );
+
+			$buttonSpan = $button.find( ".wb-inv" );
+			$buttonSpan.text( "Hide closed captioning" );
```
Garneauma: What about French?

Garneauma (changes requested):
Although it seems to be working well, I get a JS error on the Youtube page. Also, some items are only in English, please use the i18n dictionary.

src/plugins/multimedia/_base.scss
BrahimMahadi: Fixed in amended commit.

src/plugins/multimedia/multimedia.js
BrahimMahadi: The line at 598 is necessary to send the component properly to the Youtube Api.

src/plugins/multimedia/multimedia.js
BrahimMahadi: Fixed in amended commit.

src/plugins/multimedia/multimedia.js
BrahimMahadi: Fixed in amended commit.

src/plugins/multimedia/multimedia.js
BrahimMahadi: Fixed in amended commit.

src/plugins/multimedia/multimedia.js
BrahimMahadi: Fixed in amended commit.

Garneauma (approved):

BeraJosh (approved):
Thank you for the lint checks! All clear

PR #10073, opened 2025-11-12
Geomap: adding gc-checkbox support and fix mobile scroll

Garneauma:
This PR does the following:
- Add support for GC Checkbox and radio buttons
- Fix horizontal scrollbar issue in mobile view using the "table-responsive" class

Changes related to WET-622

### DIFF
--- a/src/plugins/deps/geomap-lib.js
+++ b/src/plugins/deps/geomap-lib.js
@@ -268,7 +268,7 @@ var componentName = "wb-geomap",
 		if ( this.settings.accessible ) {
 			this.map.layerDiv.append( "<div class='panel panel-default'><div class='panel-heading'><div class='panel-title' role='heading'>" +
 					this.settings.title + "</div></div><div class='panel-body'><div data-layer='" +
-					this.id + "' class='geomap-table-wrapper' style='display:none;'></div></div></div>" );
+					this.id + "' class='geomap-table-wrapper table-responsive' style='display:none;'></div></div></div>" );
 		}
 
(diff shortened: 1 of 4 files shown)


src/plugins/deps/geomap-lib.js
```diff
@@ -2422,10 +2431,18 @@ MapLayer.prototype.addToLegend = function() {
 	$label = $( "<label>", {
 		"for": "cb_" + this.id,
 		text: this.settings.title
-	} ).prepend( $chkBox );
+	} );
 
-	$li = $( "<li class='checkbox geomap-lgnd-layer'>" )
-		.append( $label, "<div id='sb_" + this.id + "'></div>" );
+	// Display large checkboxes if class is present
+	if ( $ul.closest( ".wb-geomap" ).hasClass( "large-checkboxes" ) ) {
```
duboisp: I think you can access directly to the element like `this.map.legendDiv` like you did later.


```suggestion
	if ( this.map.mapDiv.context.classList.contains( "large-checkboxes" ) ) {
```

duboisp (changes requested):
I completed a full code review and a complete functional testing by leveraging the working example in the GCWeb PR.

It does work as expected and produce the expected outcome.

The following change request are only for consistency purpose to help code readability.

duboisp (approved):
Reviewed and tested locally via GCWeb working example, it works as expected.

This is a minor change

Yves-Denis (commented):
Security Review completed on 2025-11-19 - passed.

PR #10074, opened 2025-11-13
Keep tabs in a single row instead of wrapping to next line

tristandasilva:
Jira Ticket Reference: [WET-614](https://jtickets.atlassian.net/browse/WET-614)

### DIFF
--- a/src/plugins/tabs/_base.scss
+++ b/src/plugins/tabs/_base.scss
@@ -204,9 +204,10 @@ $tab-margin-width: 10px;
 	[role="tablist"] {
 		border-spacing: $tab-margin-width 0;
 		display: flex;
-		flex-wrap: wrap;
 		list-style: none;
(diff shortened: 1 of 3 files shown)


Garneauma (dismissed):
Great fix.

This is a Patch change.

tristandasilva:
A working example has been added to show that when there are many tabs, they are managed by a horizonal scroll bar. The previous iteration would wrap the tabs to the next line, which you can see [here](https://www.canada.ca/en/environment-climate-change/services/canadian-environmental-protection-act-registry.html) and [here](https://www.canada.ca/en/environment-climate-change/services/species-risk-public-registry.html) as examples :)

@duboisp @Garneauma

Garneauma (approved):

BeraJosh (approved):
Great, thanks!

PR #10076, opened 2025-11-17
Tabbed interface: remove border under active tab

Garneauma:
A previous pull request introduced a regression issue for the active tab where the border-bottom was visible. This change fixes that issue.

### DIFF
--- a/src/plugins/tabs/_base.scss
+++ b/src/plugins/tabs/_base.scss
@@ -205,7 +205,7 @@ $tab-margin-width: 10px;
 		border-spacing: $tab-margin-width 0;
 		display: flex;
 		list-style: none;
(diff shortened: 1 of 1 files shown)


duboisp (approved):
I reviewed and tested locally, it does remove the underline for the selected tab, this is a bug fix.

Change impact: **Patch - Tab interface - Fix: remove the bottom border for the active tab.**

Note, I also observed this focus highlight behavious, but that focus indicator are managed through the browser, so no change are required.

We did have this before (v4.0.92)

Image

And with this fix, the bottom outline are impacted, it is now 1px instead of 2px like before. However because this focus indicator is determined by the user agent, that seems to be an exception according to SC 2.4.13

Image

PR #10080, opened 2025-11-27
Stabilize the tag filter component and adjust for screen reader compa…

tristandasilva:
Related to [WET-468](https://jtickets.atlassian.net/browse/WET-468)

### DIFF
--- a/src/plugins/filter/filter.js
+++ b/src/plugins/filter/filter.js
@@ -24,7 +24,7 @@ var componentName = "wb-filter",
 			section: ">tbody"
 		},
 		tblgrp: {
-			selector: " th:not([scope])" + notFilterClassSel,
+			selector: "th:not([scope])",
 			hdnparentuntil: "tbody",
 			section: ">tbody"
 		}
@@ -67,7 +67,8 @@ var componentName = "wb-filter",
 				i18n = wb.i18n;
 				i18nText = {
 					filter_label: i18n( "fltr-lbl" ),
-					fltr_info: i18n( "fltr-info" )
+					fltr_info: i18n( "fltr-info" ),
+					itemsFound: i18n( "items-found" )
 				};
 			}
 
@@ -89,7 +90,6 @@ var componentName = "wb-filter",
 					if ( uiInfo ) {
 						uiInfoID = uiInfo.id || uiInfoID;
 						uiInfo.id = uiInfoID;
-						uiInfo.setAttribute( "role", "status" );
 					}
 				} else {
 					console.error( componentName + ": " + "an <input type=\"search\"> is required in your UI template." );
@@ -104,7 +104,7 @@ var componentName = "wb-filter",
 					"<label for=\"" + inptId + "\" class=\"input-group-addon\"><span class=\"glyphicon glyphicon-filter\" aria-hidden=\"true\"></span> " + i18nText.filter_label + "</label>" +
 					"<input id=\"" + inptId + "\" class=\"form-control " + inputClass + "\" data-" + dtNameFltrArea + "=\"" + elm.id + "\" aria-controls=\"" + elm.id + "\" type=\"search\">" +
 					"</div>" +
-					"<p role=\"status\" id=\"" + uiInfoID + "\">" + i18nText.fltr_info + "</p>" );
+					"<p id=\"" + uiInfoID + "\">" + i18nText.fltr_info + "</p>" );
 
 				if ( settings.source ) {
 					$( settings.source ).prepend( filterUI );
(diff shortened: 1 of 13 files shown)


src/plugins/tagfilter/tagfilter-en.hbs
```diff
@@ -1124,19 +1114,19 @@
 						<ul class="list-unstyled">
 							<li class="radio">
 								<label>
-									<input type="radio" name="language" class="wb-tagfilter-ctrl" value=""
+									<input type="radio" name="langue" class="wb-tagfilter-ctrl" value=""
```
Garneauma: Why was this changed? If it is because it clashed with another name, maybe something like "filter_language" would be better. "Langue" is French and this is the English page.

Garneauma (changes requested):
@tristandasilva This is great work!

Can you try to fix the status message for the Filter plugin? It seems to not say the number of items correctly, which impacts the accessibility of the tagfilter plugin when they are implemented together.

Also, can you add some descriptions of the tests you performed when doing your accessibility assessment? Only for the most relevant success criteria like Info and relationship, Meaningful sequence, Keyboard, Headings and labels, Status messages.

src/plugins/tagfilter/tagfilter-en.hbs
tristandasilva: Yes! That is exactly why I changed it, I'll switch it to your proposed "filter_language" that's a lot better

src/plugins/filter/filter.js
```diff
@@ -263,7 +287,8 @@ $document.on( "keyup", selectorInput, function( event ) {
 	if ( wait ) {
 		clearTimeout( wait );
 	}
-	wait = setTimeout( filter.bind( this, $input, $elm, $elm.data() ), 250 );
+
+	wait = setTimeout( filter.bind( this, $input, $elm, $elm.data() ), 900 );
```
tristandasilva: Note that here I increased the time because it seems to fix the issue where VoiceOver cuts itself off when reading the status message. However if the delay is too long let me know and I can check for an alternative solution!

src/plugins/filter/filter.js
```diff
@@ -226,8 +238,8 @@ var componentName = "wb-filter",
 			}
 
 			if ( !searchFilterRegularExp.test( text ) ) {
-				if ( hndParentSelector ) {
-					$item = $item.parentsUntil( hndParentSelector );
+				if ( hdnParentSelector ) {
+					$item.parentsUntil( hdnParentSelector ).addClass( filterClass );
```
Garneauma, line 242: Was something broken here? It seems that, before, the class was only added to the parent element, now the class is added to the item and the parent?

src/plugins/filter/filter.js
Garneauma: Let's bring that back to 250.

src/plugins/tagfilter/tagfilter.js
```diff
@@ -135,45 +148,35 @@ const componentName = "wb-tagfilter",
...
-									filterGroupActiveFilters.push( filterItem.value );
-								}
-								break;
-							}
+					{
+						let selectedItem = filterGroup.find( item => item.isChecked );
+						if ( selectedItem && selectedItem.value !== "" ) {
+							selectedFilters.push( selectedItem.value );
+						} else {
+							console.warn( componentName + ": Radio button groups must have a default selected value. If you want to display all items, add an option called \"All\" with an empty value." );
```
Garneauma: This doesn't seem to work. I get the warning for most of the examples now. The "if" statement seems wrong. The "All" item must have a value of "", so this wouldn't work.

src/plugins/tagfilter/tagfilter.js
```diff
@@ -212,50 +214,67 @@ const componentName = "wb-tagfilter",
...
-			instance.activeFilters.forEach( function( filterGroup ) {
-				if ( filterGroup.length === 0 ) {
-					matchCount++;
-				} else {
-					let itemIncludesFilter = filterGroup.filter( function( f ) {
-						return item.tags.includes( f );
-					} ).length;
-
-					if ( itemIncludesFilter ) {
+			let matchCount = 0;
```
Garneauma: For code consistency, we try to keep all variable declarations within a block together. Please move back to where it was.

tristandasilva:
@Garneauma I reverted the timeout duration for the "keyup" event to 250, and added a timeout when updating the status message (line 278 in filter.js) as you suggested and it works well!

src/plugins/tagfilter/tagfilter-en.hbs
```diff
@@ -131,15 +121,15 @@
 				</template>
 			</ul>
 			<div class="wb-tagfilter-noresult">
-				<p>No item match this combination of filters.</p>
+				<p>No item matches this combination of filters.</p>
```
Garneauma, line 124: Can you make that change to all examples in the edge cases pages? Thanks!

Garneauma (changes requested):
Accessibility messages seem to be working very well! Only a few small changes.

src/plugins/filter/filter.js
tristandasilva: @Garneauma Yes exactly! Remember the issue I showed you where the item counter for the "Table filtering" example wasn't working (i.e. the number of items found was not updating)? Well this change seems to fix the issue. I think it's because we were only hiding the parent row, so the actual item was still being counted therefore the numbers never changed.

Garneauma (approved):

Yves-Denis (approved):
Security Assessment Completed - Passed, No concerns identified

PR #10089, opened 2026-01-26
Implement fade transition for .wb-disable class

Christopher-O:
Add fade transition styles for .wb-disable class

This allows the content that is hidden on load (.fade.out) to be visible if page template is set to simple HTML mode.   This will add in accessibility (content not hidden when wet4 js files are blocked)

### DIFF
--- a/src/base/transitions/_base.scss
+++ b/src/base/transitions/_base.scss
@@ -136,6 +136,14 @@
 	}
 }
 
+.wb-disable {
+	&.fade {
+		&.out {
+			@extend %visible;
+		}
+	}
+}
+
 @keyframes fadein {
 	0% {
 		@include invisible;
--- a/src/base/transitions/_base.scss
+++ b/src/base/transitions/_base.scss
@@ -138,9 +138,7 @@
 
 .wb-disable {
 	.fade {
-		&.out {
-			@extend %visible;
-		}
+		opacity: 1;
 	}
 }
 
--- a/src/base/transitions/_base.scss
+++ b/src/base/transitions/_base.scss
@@ -137,7 +137,7 @@
 }
 
 .wb-disable {
-	&.fade {
+	.fade {
 		&.out {
 			@extend %visible;
 		}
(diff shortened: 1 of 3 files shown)


Garneauma (changes requested):
@Christopher-O Please add a working example in the transitions page.

FYI, the basic HTML mode will no longer be available when we transition to the GCDS header. Also, the noscript is very much a thing of the past. When we transition to GCDS, If anyone uses their page without scripts, they would no longer have any header or footer on their page.

Garneauma (changes requested):
@Christopher-O This solution does not seem to fix the issue. On basic HTML mode, the elements remain hidden because de "out" class is added in JS by the plugin on page load. You should remove the "&.out" part.

You can view a working example on the Data inview plugin.

src/base/transitions/_base.scss
```diff
@@ -136,6 +136,14 @@
 	}
 }
 
+.wb-disable {
+	.fade {
+		&.out {
+			@extend %visible;
+		}
```
Garneauma:
```suggestion
		opacity: 1;
```

src/plugins/data-inview/_noscript.scss
```diff
@@ -0,0 +1,11 @@
+/*
+ * Web Experience Toolkit (WET) / Boîte à outils de l'expérience Web (BOEW)
+ * wet-boew.github.io/wet-boew/License-en.html / wet-boew.github.io/wet-boew/Licence-fr.html
+ */
+
+.fade {
+	&.out {
+		opacity: 1;
+		visibility: visible;
+	}
```
Garneauma:
```suggestion
	opacity: 1;
```

Garneauma (changes requested):
@Christopher-O It seems that Bootstrap has some styles that adds `opacity: 0` to the `.fade` class which is causing the issue. To fix that in Basic HTML mode, simply put `opacity: 1` in `.wb-disable`.

Garneauma (approved):

BeraJosh (approved):
No issues

PR #10091, opened 2026-02-03
Build: Restore parseargs to package-lock.json

EricDunsworth:
It was initially added via #9935, but somehow got removed in #9938.

Running ``npm install`` in node.js 24.13.0 (which uses npm 11.6.2) restores it.

**PS:**
Node.js 24.13.0 also adds ``"peer": true`` all over the place... but those have been deliberately excluded since they're caused by a bug in npm 11.6.2 (see npm/cli#8690... fixed in 11.6.3, should be bundled in a future version of node.js 24.x).

CC @nschonni @Garneauma

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1131,6 +1131,17 @@
         "url": "https://opencollective.com/parcel"
       }
     },
+    "node_modules/@pkgjs/parseargs": {
+      "version": "0.11.0",
+      "resolved": "https://registry.npmjs.org/@pkgjs/parseargs/-/parseargs-0.11.0.tgz",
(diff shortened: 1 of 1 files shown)


EricDunsworth:
Btw it doesn't look like ``parseargs`` currently exists in [GCWeb's package-lock.json file](https://github.com/wet-boew/GCWeb/blob/50d821d5ddc4452572ca230d22a0b5cc9fd1e069/package-lock.json).

Running ``npm install`` in node.js 24.13.0 didn't make it appear for me in that repo. Not sure what to make of it... :S.

Maybe that dependency is too "deep" in WET for GCWeb's package-lock.json file to bother trying to keep track of it? Although wouldn't that defeat the whole purpose of package-lock.json?

Garneauma (approved):

BeraJosh (approved):
Approved

PR #10119, opened 2026-03-18
Flot: removing patch, include wet-boew/flot with patch instead

Garneauma:
The patch caused issues in GCWeb when removing `node_modules` and `package-lock.json`, and running `npm install`.

Applied the patch to a fork hosted in the wet-boew organization and changed the dependency to that fork.


duboisp (approved):
Reviewed and tested with the replaced dependency. It still work and render as expected

Impact:
* **Patch** - Chart and graph - Build, dependency location updated
 
(medium)

BeraJosh (approved):
Changed trusted

PR #10140, opened 2026-05-04
Fixes screen readers not announcing feedback messages

tristandasilva:
Screen readers were not reliably announcing the feedback messages after a submission. For example, after a user clicks "Submit" screen readers weren't announcing "Thank you!".

Instead of using a status message which was proving to be inconsistent in its announcements, moving the focus to the feedback message after submission guarantees screen readers will consistently announce the text. 

This PR addresses the failed criteria (4.1.3 Status Messages) of the accessibility assessment of the Feedback area in GCWeb found in [GCWeb PR #2781](https://github.com/wet-boew/GCWeb/pull/2781)

### DIFF
--- a/src/plugins/wb-postback/wb-postback.js
+++ b/src/plugins/wb-postback/wb-postback.js
@@ -86,7 +86,7 @@ var $document = wb.doc,
 						if ( settings.successURL && !settings.success ) {
 							window.location.href = settings.successURL;
 						} else {
-							$selectorSuccess.removeClass( classToggle );
+							$selectorSuccess.removeClass( classToggle ).trigger( "setfocus.wb" );
 						}
 						$elm.trigger( successEvent );
 					} )
@@ -96,7 +96,7 @@ var $document = wb.doc,
 						if ( settings.failureURL && !settings.failure ) {
 							window.location.href = settings.failureURL;
 						} else {
-							$selectorFailure.removeClass( classToggle );
+							$selectorFailure.removeClass( classToggle ).trigger( "setfocus.wb" );
 						}
 						$elm.trigger( failEvent, response );
 					} )


src/plugins/wb-postback/wb-postback.js
```diff
@@ -86,7 +86,7 @@ var $document = wb.doc,
 						if ( settings.successURL && !settings.success ) {
 							window.location.href = settings.successURL;
 						} else {
-							$selectorSuccess.removeClass( classToggle );
+							$selectorSuccess.removeClass( classToggle ).attr( "tabindex", "-1" ).focus();
```
EricDunsworth: Would triggering [WET's built-in focus plugin](https://github.com/wet-boew/wet-boew/blob/master/src/plugins/wb-focus/focus.js) be a better bet for this?

Example of it in action (via the footnotes plugin):
https://github.com/wet-boew/wet-boew/blob/f48bd7c249c9b25492c74e70fea6d7f8cc70cee1/src/plugins/footnotes/footnotes.js#L20
https://github.com/wet-boew/wet-boew/blob/f48bd7c249c9b25492c74e70fea6d7f8cc70cee1/src/plugins/footnotes/footnotes.js#L67-L68

Garneauma: As @EricDunsworth mentioned, using the "setfocus.wb" would also work and be more concise. Another option would be to add the role status to the message. That's what we did for the GC Feedback. One might argue that the success/failure block will not always be a status message though. 

@EricDunsworth What do you think?

```suggestion
							$selectorSuccess.removeClass( classToggle ).trigger( "setfocus.wb" );
```

Garneauma (changes requested):

tristandasilva:
@EricDunsworth @Garneauma I applied your suggested changes using 'setfocus' :)

Garneauma:
@tristandasilva It seems the GC Feedback was announcing the "No" content when using Edge+NVDA on Windows. However, it was not announcing it using Chrome+VoiceOver on Mac. This PR fixes the latter.

As for the rest of Postback examples, this change does fix all examples.

Garneauma (approved):
API changes:
- Patch - Postback: fix screen readers not announcing feedback messages after submit

Yves-Denis (approved):
Security Assessment Completed - Passed, no issues identified

PR #7177, opened 2015-11-03
Remove modernizr build for now.

andre-gh:
Fixes #7176 

@pjackson28

Issue #7176, opened 2015-11-03
details/summary bugs in all browsers


LaurentGoderre:
In browsers using the polyfill, the arrows indicating the state of the the details/summary is not showing up. In Browsers with native support for details/summary, WET breaks the existing support

pjackson28:
The icons aren't showing up because no-details is not being added to the html element any more. Modernizr was doing that but none of the no-* classes are being added any more. Did someone mess with Modernizr?

LaurentGoderre:
I believe the version of Modernizr was updated

pjackson28:
Oddly enough the polyfills seem to be loading, it is just that Modernizr is no longer adding classes to the html element which some polyfills (including details/summary polyfill) was dependent upon. When I manually add no-details to the HTML element then the polyfill works fine. Was that a documented change for the new version of Modernizr? If Modernizr is no longer updating the HTML element then we'll need to review and update all plugins and polyfills as this is a breaking change.

Suggest reverting back the Modernizr version until we can update safely.

pjackson28:
Also suggest we improve our testing to ensure version bumps of our dependencies don't break functionality like what happened here.

andre-gh:
The Modernizr bump was done due to a build failure on new installs (404 downloading community something something) and thus happened without much testing.

andre-gh:
If we want to revert, we will have to be careful not to revert to a version which no longer builds.

pjackson28:
So I see a few possible approaches then.

1. Roll back to a previous versions that still adds the classes to the HTML element but also builds.

2. Test and fix all plugins and polyfills (and core) to remove the dependency upon the Modernizr add classes (or to at least add what is no longer being added).

Either way, something needs to be done quickly as there could be multiple things broken now. Note also that because details/summary doesn't work, the mobile panel will be broken and there is content that will seem unavailable to users (since they won't know that the details elements can be opened).

@masterbee What should be the way forward to address this issue?

andre-gh:
@pjackson28 O.o does grunt-modernizr just build the master version of Modernizr?

andre-gh:
@LaurentGoderre Am I reading their code right?  Do we/does it just follow the Modernizr master regardless of the grunt-modernizr version.

pjackson28:
@andre-gh Yes it looks like it is doing a custom build based upon pass parameters. Here is the build URL in modernizr-custom.js:
```
[...]

PR #10143, opened 2026-05-20
Removed Modernizr MathML test

tristandasilva:
_Related to [WET-665](https://jtickets.atlassian.net/browse/WET-665)_

- Fixes the issue described in the ticket where on Canada.ca pages (including WET and GCWeb), there is extra text that appears at the bottom of the last page in a print preview (when you go to print a page).

- The issue seemed to have been related to the removal of MathJax from WET, because when I removed the MathML test in modernizr-custom.js, the bug was resolved.

- Tested in Chrome, Edge, Firefox and Safari by making the change in WET, building the prod files and overriding wet-boew.min.js in the browser on Canada.ca pages.

- From the diff in Github it doesn't look like the change is very clear, but essentially I just removed this block of code from modernizr-custom.js (see attached screenshot)

block-removed


Garneauma (dismissed):
This does fix the issue. Thank you @tristandasilva !

BeraJosh:
Maryanne deserves all the credit for this security review (excellent and thorough!). 
See below:

Looking at Tristan's PR, I can confirm that the changes made are exactly those he mentioned; only the removal of the Modernizr test, editing the one file

I searched for references to the removed test, and found the following CSS:
src > base > views > _screen.scss > wb-dtmd

Line 39: Hide MathML check * /
This could be removed to improve readability of the file since the Modernizr MathML test was removed. Its purpose was to hide the Modernizr MathML check after the check was completed.

Would you like this removed before the approval to keep everything concise?
@Garneauma @EricDunsworth

Garneauma:
@BeraJosh YEs, that can be removed. Thanks @mmoya064 .

@tristandasilva Please remove the mentioned code, thanks.

src/core/dep/modernizr-custom.js
```diff
@@ -1,4 +1,4 @@
 /* Modernizr (Custom Build) | MIT & BSD
  * Build: https://modernizr.com/download/#-elem_details-elem_progress_meter-mathml-cors-load-mq-css3-input-inputtypes-svg-cssclasses-csstransitions-fontface-backgroundsize-borderimage-teststyles-testprops-testallprops-hasevents-prefixes-domprefixes
```
EricDunsworth, line 8: I'd recommend updating the build URL to reflect the updated Modernizr settings (MathML test removal):

```diff
- https://modernizr.com/download/#-elem_details-elem_progress_meter-mathml-cors-load-mq-css3-input-inputtypes-svg-cssclasses-csstransitions-fontface-backgroundsize-borderimage-teststyles-testprops-testallprops-hasevents-prefixes-domprefixes
+ https://modernizr.com/download/?backgroundsize-borderimage-cors-csstransitions-details-fontface-input-inputtypes-progressbar_meter-svg-domprefixes-mq-prefixes-setclasses-testallprops-teststyles
```
(Note: Looks like option names have shifted a bit in the updated URL - which I obtained by visiting the old URL and de-selecting the "MathML" option)

From what I can tell, the old and revised minified JS code exactly match what the Modernizr site generates for "Build" with/without the MathML option.

tristandasilva:
- Removed the 'Hide MathML check' CSS in `src/base/views/_screen.scss`
- Updated the Modernizr build URL, removing MathML from the selected options. I can confirm that the newly generated minified code that Modernizr generated matches my updated minified code, as @EricDunsworth mentioned

Let me know if there's anything else :) 

@Garneauma

Garneauma (dismissed):

EricDunsworth:
Uh oh - turns out that the Modernizr site generates completely different minified JS than what we thought :(.

The "Copy to clipboard" link on Modernizr's site is misleading. All it does is open a dropdown with its minified JS highlighted. It doesn't copy anything to the clipboard for real :S. So its generated JS code still needs to be manually copied/pasted :P.

Based on commit timelines, I believe the minified JS that's currently in use for [``src/core/dep/modernizr-custom.js``](https://github.com/wet-boew/wet-boew/blob/master/src/core/dep/modernizr-custom.js) was based on Modernizr 3.2.0:
* c4e4522380e2489143d05fa15cf96fff501290ef was committed on November 2015
* [Modernizr 3.2.0](https://github.com/Modernizr/Modernizr/blob/master/CHANGELOG.md#v320) was the latest release at the time (next release was [3.3.0](https://github.com/Modernizr/Modernizr/blob/master/CHANGELOG.md#v330) in January 2016)

**Potential options (from easiest to hardest):**
* **Option 1:** Keep this PR's updated JS code "as-is" and rework the comment at the start of the file to provide extra context:
  * Maybe keep the old build URL "as-is"
  * State it uses Modernizr 3.2.0
  * State that the MathML logic has been manually-removed
* **Option 2:** Update the minified JS code to match the Modernizr site's build generator:
  * This would bump Modernizr to version 3.6.0
  * More straightforward than option 1... as long as it can be tested
* **Option 3:** Drop Modernizr completely:
  * This would involve many changes... but probably mainly removals, so likely less painful than option 5
  * Biggest challenge with this would probably be to figure out what to do with all the [ancient polyfills](https://github.com/wet-boew/wet-boew/tree/master/src/polyfills) that [depend on WET's Modernizr implementation](https://github.com/wet-boew/wet-boew/blob/master/src/core/wb.js#L560-L621)
* **Option 4:** Switch back to the [``grunt-modernizr``](https://github.com/Modernizr/grunt-modernizr) dependency:
  * WET used it in the past:
    * c4e4522380e2489143d05fa15cf96fff501290ef switched to the site's build generator
    * Merge commit for it is 8e0e4dedca793eebc5abbe0aacf22f95ea0ea580... which references #7177 (doesn't exist anymore?)
    * #7176 contains background info for what led to it:
      * Looks like only "half" of a conversation (@andre-gh's comments appear to have all been deleted)
      * Sounds like something caused Modernizr to stop adding feature class names to the ``<html>`` element at the time - which broke WET's polyfills... and the site's outdated build generator was seen as the solution?
    * Modernizr/grunt-modernizr#126 has additional context (seems like ``grunt-modernizr`` started out as an "unofficial" project, then got transferred to the Moderznir team and got overhauled by them at the time)
  * Modernizr itself and ``grunt-modernizr`` don't follow the same versioning schemes:
    * [Modernizr's official changelog](https://github.com/Modernizr/Modernizr/blob/master/CHANGELOG.md) covers versions up to v4.0.0-alpha ([its site's news section](https://modernizr.com/news/) contains copies of older changelog entries)
    * [``grunt-modernizr``'s changelog](https://github.com/Modernizr/grunt-modernizr/blob/main/CHANGELOG.md) goes all the way up to 5.0.3... but those versions are "detached" from the Modernizr versions it implements

@tristandasilva @Garneauma Thoughts?

src/core/dep/modernizr-custom.js
```diff
@@ -1,4 +1,4 @@
 /* Modernizr (Custom Build) | MIT & BSD
- * Build: https://modernizr.com/download/#-elem_details-elem_progress_meter-mathml-cors-load-mq-css3-input-inputtypes-svg-cssclasses-csstransitions-fontface-backgroundsize-borderimage-teststyles-testprops-testallprops-hasevents-prefixes-domprefixes
+ * Build: https://modernizr.com/download/?-backgroundsize-borderimage-cors-csstransitions-details-fontface-input-inputtypes-progressbar_meter-svg-domprefixes-mq-prefixes-setclasses-testallprops-teststyles
```
EricDunsworth: Remove dash before URL's question mark
```suggestion
 * Build: https://modernizr.com/download/?backgroundsize-borderimage-cors-csstransitions-details-fontface-input-inputtypes-progressbar_meter-svg-domprefixes-mq-prefixes-setclasses-testallprops-teststyles
```

EricDunsworth: Btw this suggestion might be irrelevant now given what's covered in https://github.com/wet-boew/wet-boew/pull/10143#issuecomment-4547517527.

Garneauma:
@EricDunsworth @tristandasilva Let's go with option 1 above. Thanks !

Garneauma (approved):

BeraJosh (approved):
Approved, thank you!

PR #10152, opened 2026-06-05
Core: Update focusable/tabbable helper methods

EricDunsworth:
WET used altered copies of those methods from jQuery UI... but they were buggy and became outdated overtime. They were susceptible to selecting elements that should've been out of scope - such as hidden ones (e.g. ``<a href="#" hidden>``, ``<input type="hidden">``, etc). That in turn could result in plugins trying to focus onto them in vain.

This resolves it by updating WET's copies of those methods to jQuery UI's latest versions. Retains an extra optional parameter for WET's ``discoverable`` method.

Unlike the old logic (which used an extra ``visibility`` param that was passed-in by both the regular ``focusable``/``tabbable`` methods)... the new logic uses an ``isVisible`` param that's set to ``true`` by default. So the burden is now on the ``discoverable`` method to pass-in a ``false`` parameter to ``focusable``.

Related changes:
* Added inline MIT license block from jQuery UI (slightly-amended to combine "Focusable" and "Tabbable" on the same line)
* Fixed "tabbable" typo in comment
* Revised comment block to clarify WET's deliberate changes and exactly which code was borrowed from jQuery UI
* Removed ``visible`` method (old ``focusable`` method used it, no longer exists in latest jQuery UI logic)
* Removed ``data`` method (unused code from an old version of jQuery UI)
* ESLint: Changed ``!= null`` to ``!== undefined`` to adhere to "no-eq-null" rule

Related demo change:
* Overlay plugin: Add an alternate middle screen example that begins with a hidden link:
  * Forward-tabbing beyond the top-right X button previously failed in that scenario (since the old logic mistakenly-believed the hidden link was tabbable)

### DIFF
--- a/src/plugins/overlay/overlay-fr.hbs
+++ b/src/plugins/overlay/overlay-fr.hbs
@@ -9,7 +9,7 @@
 	"altLangPrefix": "overlay",
 	"js": ["demo/overlay"],
 	"css": ["demo/overlay"],
-	"dateModified": "2014-07-22"
+	"dateModified": "2026-06-08"
 }
 ---
 <span class="wb-prettify all-pre hide"></span>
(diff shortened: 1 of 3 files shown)


EricDunsworth:
Net-new changes (compared to the draft):
* Revised OP + commit message
* Added jQuery UI's MIT license block
* Fixed ":tabbable" typo in comment
* Revised comment block to clarify WET's deliberate changes and exactly which code was borrowed from jQuery UI
* Removed data method (appears to be old unused code copied from jQuery UI)
* Fixed ESLint issues (removed commented-out return statements, changed ``!= null`` to ``!== undefined``)
* Polished-up the new example in the overlay demo (simplified/renamed it, moved it, created a French variant of it)

EricDunsworth:
@nschonni @Garneauma Just rebased since I was getting [failed CI jobs from my branch on GitHub](https://github.com/EricDunsworth/wet-boew/actions/workflows/ci.yml) (probably related to #10150 having been merged in-between the time I sent this as a draft PR and now).

src/plugins/overlay/overlay-fr.hbs
```diff
@@ -414,6 +431,30 @@ wb.doc.on( "click", "#overlay-open-btn", function( event ) {
...
+					<li>liste ordonnée (<code>ol</code>) deuxième niveau - apparence par défaut
+						<ol>
+							<li>liste ordonnée (<code>ol</code>) troisième niveau - apparence par défaut</li>
+							<li>liste ordonnée (<code>ol</code>) troisième niveau - apparence par défaut</li>
+						</ol>
+					</li>
+				</ol>
+			</li>
+		</ol>
+		<p><a href="#">Lien - apparence par défaut</a></p>
```
Garneauma, line 454: Not a deal breaker, but you don't have this extra link in the EN version. Do you want to add it in the EN or remove it here?

Garneauma (commented):
Just a small comment, otherwise it works well!

src/plugins/overlay/overlay-fr.hbs
EricDunsworth: I noticed it too while testing. It seems to be on all of the French demo's middle/full-screen examples, so I kept it in place in the new example too.

I think the best course of action would be to remove those link examples (or add them to English) in a separate PR, after this one gets merged. That way, they can all be dealt with consistently in one shot.

Garneauma (approved):

BeraJosh (approved):
Approved @Garneauma

Garneauma:
API changes:
- Patch - Core: Update focusable/tabbable helper methods
- Patch - Overlay: fix focus trap issue
- Patch - Lightbox: fix focus trap issue

PR #10154, opened 2026-06-09
Multimedia: fixing captions not appearing on full screen mode

mmoya064:
Related to WET-658.

Fixing issue with captions not appearing in fullscreen.

### DIFF
--- a/src/plugins/multimedia/_base.scss
+++ b/src/plugins/multimedia/_base.scss
@@ -47,6 +47,11 @@
 	.display {
 		&:before {
 			text-align: center;
+			z-index: 1;
+
+			&:fullscreen {
+				padding-bottom: 10em;
+			}
 		}
 
 		&:after {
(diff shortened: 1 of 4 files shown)


BrahimMahadi (approved):

Yves-Denis (dismissed):

Ricokola:
@Garneauma initial tech review has been made and approved by @BrahimMahadi and security assessment has been made by @Yves-Denis. It's now ready for your final review in order to be merged.

Garneauma (changes requested):
@BrahimMahadi @mmoya064 Just a few things I noticed:
- The larger the screen is, the bigger the space for the CC get. Needs to be fixed.
- Please do not add new CSS classes or ID's if it's not really needed. I believe everything can be targeted using context (this, @this, or elm) and the right CSS selector. Only files that should be changed are the CSS and JS files.

Garneauma (changes requested):
The behaviour is still a bit weird. When in full screen, the video does not behave as a regular full screen video. The bottom control bar is still visible, which is a lesser issue, but the transcript portion is also there, which it shouldn't be. The video itself should be full screen and centered on the screen, but it is now moved to the top. I think this can all be adjusted using CSS.

PR #10155, opened 2026-06-09
Menu: Move aria-expanded, scrap aria-hidden attributes

EricDunsworth:
They were previously being used on the mega menu (menubar pattern) and mobile menu (menu pattern)'s submenu ``<ul>`` elements... which wasn't very ideal.

Issues with that setup:
* Didn't have any effect in practice (compared to if they hadn't been used... at least not in NVDA)
* The lists were already completely hidden via CSS when collapsed (``display: none;`` in the mega menu, ``content-visibility: hidden;`` via ``::details-content`` in the mobile menu)
* Setting the ``aria-expanded`` attribute on "child" content is an unusual convention (both the APG and Bootstrap's menu pattern examples place it on trigger links directly)

This resolves it by:
* Moving ``aria-expanded`` attributes to trigger links
* Removing ``aria-hidden`` attributes and revising logic that depended on it (now primarily relies on the ``open`` attribute or "open" class... along with some dead code)

Notes:
* Setting ARIA roles on ``summary`` elements violates the ARIA in HTML spec and fails HTML validation... fixing it would necessitate major changes to the menu plugin
* Decided against removing ``aria-expanded`` attributes for the following reasons:
  * It's a supported state for the ``menuitem`` role in [ARIA 1.2 and up](https://www.w3.org/TR/wai-aria-1.2/#menuitem)
  * Had concerns about how assistive technology might handle a combination of a ``summary`` with a ``menuitem`` role combined with its parent ``details`` element's intrinsic expanded/collapsed state... figured that having an explicit ``aria-expanded`` attribute would be a safer bet
* Screen reader handling of ``role="menuitem"`` and ``aria-expanded`` appears to be spotty:
  * NVDA deliberately doesn't announce the current collapsed/expanded states of menu items, unless the user is in the act of expanding (i.e. announcing "expanded" right after an expansion event occurs). Refer to https://github.com/nvaccess/nvda/issues/9096#issuecomment-1497085780 for more details.
  * Unsure about other screen readers... some might behave differently given that ARIA itself allows for that combination of roles/states

### DIFF
--- a/src/plugins/menu/menu.js
+++ b/src/plugins/menu/menu.js
@@ -93,11 +93,9 @@ var componentName = "wb-menu",
 			// if there is a submenu lets put in the aria for it
 			if ( $subMenu.length !== 0 ) {
 
-				$elm.attr( "aria-haspopup", "true" );
-
-				$subMenu.attr( {
-					"aria-expanded": "false",
-					"aria-hidden": "true"
+				$elm.attr( {
+					"aria-haspopup": "true",
+					"aria-expanded": "false"
 				} );
 
 				// recurse into submenu
@@ -120,8 +118,8 @@ var componentName = "wb-menu",
 			sectionHtml = "<li><details>" + "<summary class='mb-item" +
 				( $section.hasClass( "wb-navcurr" ) || $section.children( ".wb-navcurr" ).length !== 0 ? " wb-navcurr'" : "'" ) +
 				menuitem + sectionsLength + posinset + ( sectionIndex + 1 ) +
-				"' aria-haspopup='true'>" + $section.text() + "</summary>" +
-				"<ul class='list-unstyled mb-sm' role='menu' aria-expanded='false' aria-hidden='true'>";
+				"' aria-haspopup='true' aria-expanded='false'>" + $section.text() + "</summary>" +
+				"<ul class='list-unstyled mb-sm' role='menu'>";
 
 		// Convert each of the list items into WAI-ARIA menuitems
 		for ( k = 0; k !== itemsLength; k += 1 ) {
@@ -435,23 +433,20 @@ var componentName = "wb-menu",
 	 * @param {boolean} removeActive Whether or not to keep the active class
 	 */
 	menuClose = function( $elm, removeActive ) {
+
+		// Adjust top-level menu item's aria-expanded attribute
+		$elm
+			.find( "[aria-haspopup=true]" )
+			.attr( "aria-expanded", "false" );
+
 		$elm
 			.removeClass( "sm-open" )
 			.children( ".open" )
 			.removeClass( "open" )
-			.attr( {
-				"aria-hidden": "true",
-				"aria-expanded": "false"
-			} )
 
 		// Close nested submenus
 			.find( "details" )
-			.removeAttr( "open" )
-			.children( "ul" )
-			.attr( {
-				"aria-hidden": "true",
-				"aria-expanded": "false"
-			} );
+			.removeAttr( "open" );
 
 		if ( removeActive ) {
 			$elm.removeClass( "active" );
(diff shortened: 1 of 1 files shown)


src/plugins/menu/menu.js
```diff
@@ -96,7 +96,7 @@ var componentName = "wb-menu",
 				$elm.attr( "aria-haspopup", "true" );
 
 				$subMenu.attr( {
-					"aria-expanded": "false",
+					//"aria-expanded": "false",
 					"aria-hidden": "true"
```
EricDunsworth: Might be worthwhile to scrap ``aria-hidden`` attributes too if it doesn't do any harm...

IIRC ``display: none;`` is in play in all the same scenarios as ``aria-hidden`` - which _should_ technically have the same effect in practice.

EricDunsworth: Proceeding with this.

Btw I was wrong about ``display: none;``, but from what I checked, browser accessibility APIs don't expose non-``summary`` children of collapsed ``details`` elements to ATs. So IMO ``aria-hidden`` is still unlikely to be useful in practice in WET's menus.

EricDunsworth:
Should be effectively ready now, will need some final testing tomorrow. Then I'll bring it into a single commit and mark this PR as ready.

**Edit:** Needs linting fixes too.

src/plugins/menu/menu.js
```diff
@@ -748,13 +739,6 @@ $document.on( "keydown", selector + " [role=menuitem]", function( event ) {
 
 					// Close any other open menus
 					if ( !isOpen ) {
-						$( parent )
-							.closest( "[role^='menu']" )
-							.find( "[aria-hidden=false]" )
-							.parent()
-							.find( "[aria-haspopup=true]" )
-							.not( menuItem )
-							.trigger( "click" );
```
EricDunsworth, line 757: Scrapped this since it never ran in practice and wouldn't achieve anything even if it were to be fixed.

This logic is copied from a click event handler somewhere else in the plugin (basically meant to be a keyboard equivalent of it):
https://github.com/wet-boew/wet-boew/blob/5f1bdefbdb450e482a74c956d43fceac3e20792c/src/plugins/menu/menu.js#L590-L598

Unlike the click event's logic, the "vanilla" ``parent`` variable that's getting passed to jQuery is an undeclared variable in this context. But due to its name... it's being treated like an alias for [``Window.parent``](https://developer.mozilla.org/en-US/docs/Web/API/Window/parent), which in turn causes it to always return the ``Window`` object. So nothing ever actually happens here.

I tried changing ``$( parent )`` to ``$parent`` and updated the selectors... but it was useless in practice. The click event's fake-click logic already covers all the same scenarios as the keyboard event's fake-click.

EricDunsworth:
@duboisp @Garneauma This PR should be ready to go now :)... even though it might not ultimately be merged.

Btw my goal with this was solely to adjust ARIA semantics in the hopes of improving screen reader behaviour (which unfortunately didn't pan out for NVDA - although it'd be interesting to see how JAWS/VoiceOver behave now).

My line of thinking was that this could serve as a "step 1" of sorts and potentially be followed by a "step 2" to revise the menu plugin to use the [disclosure pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/).

Yves-Denis (approved):

EricDunsworth:
Just to add to [my previous comment](https://github.com/wet-boew/wet-boew/pull/10155#issuecomment-4784226940)... I'm currently working on a "step 2" PR to revise WET's menu plugin to use the [disclosure navigation menu pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/). It'll be separate from this one.

**PS:**
From my perspective, there's nothing stopping this PR from being merged if anyone's interested in doing so. If nothing else, it improves semantics... even though it isn't a definitive solution (given that the better semantics didn't change NVDA's behaviour in practice).

EricDunsworth:
The "step 2" PR is #10192. It'll ultimately supersede this one.

Btw the "step 2" PR uses this PR as its starting point (first draft commit is literally a copy of this PR's changes).

PR #10157, opened 2026-06-12
Menu: Don't focus onto mobile menu link when switching to larger views

EricDunsworth:
The menu plugin is designed to close the mobile menu overlay when switching from smaller to larger views. But it previously worked in a flawed manner.

Its old behaviour was to trigger the overlay plugin's "close" event upon switching to larger views. The overlay plugin in turn attempted to focus onto the mobile menu link... which just so happens to be hidden in larger views. End result was that ``setfocus.wb`` eventually triggered console errors.

This resolves the errors by adding a ``noFocus: true`` property to the mobile menu plugin's "close" event (derived from how it's done in the data-inview plugin).


EricDunsworth:
Btw this could potentially be improved upon by focusing at a predetermined spot (such as the mobile menu overlay's closest _visible_ parent container, ``#wb-bnr`` or maybe even ``#wb-tphp``). IMO the closest equivalent to the menu overlay's starting content would be the language toggle links.

Garneauma (approved):

BeraJosh (approved):
Approved, thank you

PR #10169, opened 2026-07-02
Menu: Use focusEvent consistently

EricDunsworth:
The plugin previously called a mix of ``focusEvent`` (an alias for "setfocus.wb") and "setfocus.wb" directly - which goes against the point of having a ``focusEvent`` variable.

This replaces all the direct references to "setfocus.wb" to ``focusEvent`` for the sake of consistency.

Having only one way of calling the focus plugin might also reduce the potential for confusion.


Garneauma (approved):

Garneauma:

BeraJosh (approved):
Approved

PR #10171, opened 2026-07-07
Tables: trigger plugins init when table is drawn (pagination, search, order, etc.)

Garneauma:
When datables includes wet-boew plugins, it only initializes them if the content is already on the page on the first load. This PR trigger the initialization of any plugin that was not previously initialized within the table.

To test, in tables-fltrs-en/fr.html, search for "Biomass" and click on the "Biomass Program Continues to Support Transition to Renewable Energy in Manitoba" link.


BeraJosh (approved):
Thank you!

duboisp (approved):
I reviewed the code, look good. I also did a functional check and it is working as expected.

This is **Minor change for our data table plugin** - Enable in paging data, wet-boew plugin inside data cell content.

PR #10172, opened 2026-07-07
Menu: Fix lower-case i typo in menuItemSelector var

EricDunsworth:
The plugin's keyboard event handlers previously used a ``menuitemSelector`` variable (with a lower-case "i" in "item") instead of ``menuItemSelector`` (upper-case "I" in "Item"). Some logic even literally assigned ``menuitemSelector`` to ``menuItemSelector``'s value ("> a, > details > summary") to prevent it from being treated as ``undefined``.

This resolves it by removing the ``menuitemSelector`` variable's declaration and replacing all references to it with ``menuItemSelector``.

Note:
Won't change any behaviour in practice. The ``menuitemSelector`` variable was never ``undefined`` in any of the logic that used it.


Garneauma (approved):

Garneauma:

Yves-Denis (approved):

PR #10186, opened 2026-07-28
add support for custom tags and attributes in dom purify

tristandasilva:
Adds the GC Combobox to the allowed tags during DOMPurify sanitation to support using it with other plugins (e.g. Fieldflow,  using ajax to fetch and render)

### DIFF
--- a/src/core/dep/jquery-fix.js
+++ b/src/core/dep/jquery-fix.js
@@ -199,6 +199,15 @@ var localParseHTML = jQuery.parseHTML,
 		"<td />",
 		"<td/>"
 	],
+	allowedTags = [
+		"gc-combobox"
+	],
+
+	// Declare the allowed attributes for custom tags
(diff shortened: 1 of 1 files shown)


Garneauma (changes requested):
I realized it still strips out the custom attributes from the combobox as they are not valid HTML attributes. Please add support for the combobox custom attributes. Thanks!

Garneauma (approved):
Looks good!

Garneauma:
@wet-boew/sc-pull-request-approvers This PR adds the combobox tag and its attributes to the allowlist of DOMPurify so that it is possible for the combobox to be ajaxed in.

Yves-Denis (approved):
Security Assessment - Passed:
Additional Sanitization enhances security vulnerabilities with better HTML parsing and reduces risks of XSS attacks.

PR #10197, opened 2026-08-15
Remove invalid aria-pressed from Data Table pagination

### LINKED ISSUE #10196: [BUG] Data Table accessibility
Data Table does not pass accessibility testing

Image
From playwright accessibility report https://playwright.dev/docs/accessibility-testing

aria-pressed causes error
https://cms-sgc.esdc-edsc.canada.ca/eng/service_catalogues/video/index.shtml?site=iservice-prv

opensource-joe:

Removes `aria-pressed` from the Data Table pagination buttons, which is the `aria-allowed-attr` error reported in #10196.

`aria-pressed` is only valid on a button role. DataTables renders these buttons as anchors with an explicit `role="link"`, so the attribute is not allowed on them and axe flags every page button as a critical violation.

It is also redundant. The comment above the code says "Should be pushed upstream to DataTables", and that has since happened: DataTables 1.13 sets `aria-current="page"` on the active button, which is the correct pattern for pagination. Nothing in WET touches `aria-current`, so it already reaches the page intact. Removing `aria-pressed` loses no information. It leaves the current page marked one correct way instead of two ways, one of them invalid.

The rendered markup after the change, unchanged apart from the removed attribute:

```html
<a class="paginate_button previous disabled" role="link">
<a class="paginate_button current"          role="link" aria-current="page">
<a class="paginate_button"                  role="link">
```

The invisible "Page" prefix on each number is untouched.

## Verification

Against `dist/unmin/demos/tables/tables-en.html` built from this branch, in Chrome, using axe-core:

| | violations | passes |
|---|---|---|
| before | **16** | 40 |
| after | **0** | 56 |

Rule set: `aria-allowed-attr`, `aria-allowed-role`, `aria-required-attr`, `aria-valid-attr-value`. The 16 violating nodes move into the pass column, and `aria-current="page"` is confirmed still present on the current button afterwards.

Build and existing tests are clean on the branch:

```
grunt                exit 0
grunt pre-mocha      exit 0
node test-runner     exit 0   118 suites, 241 tests, 241 passes, 0 failures
```

**Disclosed honestly:** I did not add a unit test. `src/plugins/tables` has no `test.js`, and standing up a DataTables harness to assert one removed attribute seemed out of proportion to a deletion, as well as a bigger call than a first contribution should make on its own. Happy to add one if you would like it, and equally happy to be told the right shape for it.

## Related issues (optional) / Requêtes associées (optionnel)

Fixes #10196

## Also noticed, not changed here

`src/plugins/paginate/paginate.js` uses `aria-current="true"` on its page buttons. That is valid, but `aria-current="page"` is the more precise token for pagination and would match what DataTables does in the tables plugin. Deliberately left out of this PR since it is a different plugin and not part of the reported bug. Glad to send it separately if wanted.

### DIFF
--- a/src/plugins/tables/tables.js
+++ b/src/plugins/tables/tables.js
@@ -152,8 +152,15 @@ var componentName = "wb-tables",
 		$pagination.empty();
 		$pagination.append( ol );
 
-		// Update the aria-pressed properties on the pagination buttons
-		// Should be pushed upstream to DataTables
+		// Prefix each page number with an invisible "Page" so the link reads as
+		// "Page 2" rather than a bare "2".
+		//
+		// The current page is not marked here. DataTables sets aria-current="page"
+		// on it, which is the correct pattern for pagination. This used to add
+		// aria-pressed as well, from a time when DataTables did not, but these are
+		// anchors carrying an explicit role="link" and aria-pressed is not allowed
+		// on that role, so it was an axe aria-allowed-attr error on every page
+		// button.
 		$pagination.find( ".paginate_button" )
 			.attr( {
 				"href": "#" + navFocusOnId
@@ -167,12 +174,9 @@ var componentName = "wb-tables",
 			} )
 
 			.not( ".previous, .next" )
-			.attr( "aria-pressed", "false" )
 			.html( function( index, oldHtml ) {
 				return "<span class='wb-inv'>" + i18nText.paginate.page + " </span>" + oldHtml;
-			} )
-			.filter( ".current" )
-			.attr( "aria-pressed", "true" );
+			} );
 	};
 
 // Bind the init event of the plugin


opensource-joe:
@Garneauma Would you be able to approve the workflow runs on this PR?

As a first-time contributor here my workflows need a maintainer to release them, so all three (`CI`, `JavaScript Linting`, `Space Linting`) are sitting at `action_required` and none have executed. Nothing reports in the checks list at all, so it reads as though CI simply has not started rather than as something waiting on a person.

I am not asking for a review yet, only for CI to be allowed to run, so that any failures are mine to fix before I take up reviewer time.

For what it is worth, on this branch locally `grunt` and `grunt pre-mocha` both exit 0, and `node test-runner` gives 118 suites / 241 tests / 241 passes / 0 failures. The change itself is a six-line deletion in one file, so there should not be much for CI to disagree with, but I would rather have that confirmed than assumed.

Garneauma (approved):
The rationale behind the use of `true` instead of `page` in the pagination plugin, was that the MDN documentation considers a "page" as the document, rather than a group of visible elements. Therefore, using `true` made more sense as it is a bit more broad.

> true
Represents the current item within a set.

Garneauma:

opensource-joe:
Thanks for the review and for the escalation, @Garneauma.

On the `true` versus `page` reasoning, I think we are talking about two different plugins, and it is worth pinning down so nobody reconciles them later on the assumption they should match.

Your rationale describes **`src/plugins/paginate/paginate.js`**, which sets `aria-current="true"` in three places, at lines 131, 184 and 189. That is WET's own markup and your reasoning for it holds up: that widget paginates a group of elements inside one document rather than a set of documents, so `true`, "the current item within a set", is the better fit. I would not change it.

This PR touches **`src/plugins/tables/tables.js`**, where the value is not WET's to choose. The pagination markup comes from DataTables, pinned at `~1.13.11` in `package.json`, and it sets the attribute itself:

```js
// jquery.dataTables.js, 1.13.11, line 14946
'aria-current': btnClass === classes.sPageButtonActive ? 'page' : null,
```

The same object literal sets `'role': 'link'` on the line above, which is the reason `aria-pressed` had to go: it is only valid on a button role, so axe flagged every page button under `aria-allowed-attr`.

So after this change the current page is still marked, by DataTables, as `aria-current="page"`. Nothing is lost, and there is no WET code left that would need to pick a value. Overriding it to `true` would mean rewriting an attribute the vendored library already sets correctly for its own widget.

Short version: the two plugins disagree on purpose and both are right for their own context. Happy to add a code comment noting that, if you think a future reader would otherwise try to make them consistent.

Garneauma:
@opensource-joe Yes, I was just responding to your comment below for the rationale behind true vs page. I approved your PR.

> ### Also noticed, not changed here
> src/plugins/paginate/paginate.js uses aria-current="true" on its page buttons. That is valid, but aria-current="page" is the more precise token for pagination and would match what DataTables does in the tables plugin. Deliberately left out of this PR since it is a different plugin and not part of the reported bug. Glad to send it separately if wanted.

Garneauma:
@wet-boew/sc-pull-request-approvers Reminder to review this PR. Thanks!

Yves-Denis (approved):

PR #10207, opened 2026-09-02
Formvalid: email field validation without domain

mmoya064:
Related to WET-699.
Fixed email address validation so that emails without top-level domains (.com, .ca, .net., etc) are not accepted.

### DIFF
--- a/src/plugins/formvalid/formvalid.js
+++ b/src/plugins/formvalid/formvalid.js
@@ -384,6 +384,11 @@ var componentName = "wb-frmvld",
 								$form.validate().element( this );
 							} );
 
+							// Ensuring email validation isn't done without the top-level domain (.com, .ca, .net, etc)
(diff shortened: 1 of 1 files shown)


Yves-Denis (approved):
Security Assessment Completed - Passed, new code does not introduce any security issues.

duboisp (changes requested):
This need to be implemented as a new validation method, the current one does conform to HTML 5 email validation algorithm

Ricokola:
@Garneauma This PR will be updated according to @duboisp  comment above. The ticket type and business requirements are currently been re-written.

Ricokola:
It has been confirmed that this is not an issue and the email validation that the requester is looking for will be available in GCDS so our recommendation and as per client response, this can wait until full migration. An email has been sent to the requester so this PR can now be closed.

@Garneauma

PR #10216, opened 2026-09-11
add new combo box attribute to dompurify allow list

tristandasilva:
Adds support for the `required-message` attribute that was newly added to the combo box (related to [WET-708](https://jtickets.atlassian.net/browse/WET-708)).

Note: the `required` attribute was also added to the combo box, but since this is a native attribute it's not necessary to have it in the DOMPurify allow list.


Yves-Denis (approved):
Security Assessment Completed - Passed, no security issues introduced or identified (Only tag attributes are affected).

Issue #10189, opened 2026-07-30
[BUG] Video class has missing aria-label


pmirowski:
Error occurs with playwright accessibility testing: 
https://playwright.dev/docs/accessibility-testing

volume input results in an error.
https://wet-boew.github.io/v4.0-ci/demos/multimedia/multimedia-en.html

   Object {
     "description": "Ensure that every form element has a visible label and is not solely labeled using hidden labels, or the title or aria-describedby attributes",
     "help": "Form elements should have a visible label",
     "helpUrl": "https://dequeuniversity.com/rules/axe/4.11/label-title-only?application=playwright",
     "id": "label-title-only",
     "impact": "serious",
     "nodes": Array [
       Object {
         "all": Array [],
         "any": Array [],
         "failureSummary": "Fix all of the following:
   Only title used to generate label for form element",
         "html": "<input type=\"range\" class=\"volume\" aria-controls=\"wb-auto-3-md\" aria-labelledby=\"wb-auto-3-md\" aria-describedby=\"wb-auto-3-md-vlm-lbl\" title=\"Volume\" min=\"0\" max=\"100\" value=\"100\" step=\"5\">",
         "impact": "serious",
         "none": Array [
           Object {
             "data": null,
             "id": "title-only",
             "impact": "serious",
             "message": "Only title used to generate label for form element",
             "relatedNodes": Array [],
           },
         ],
         "target": Array [
           ".volume",
         ],
       },
     ],
     "tags": Array [
       "cat.forms",
       "best-practice",
     ],
   },

Please look into this

PR #10219, opened 2026-09-15
Patch - Multimedia player: Add visible volume label

Pablo-seixas:
The multimedia volume range currently relies on its title attribute while the descriptive text is hidden. This causes the axe label-title-only accessibility rule to report a serious violation.

This change adds a visible label associated with the volume range through for and a unique per-player id. The existing localized volume text is preserved, and aria-controls continues to identify the controlled media element.

Related issue: https://github.com/wet-boew/wet-boew/issues/10189

Validation completed: grunt test; grunt prepare demos; npm test with 241 of 241 tests passing; grunt css; axe-core label-title-only comparison; HTML validation of all 10 multimedia pages; responsive Chromium checks in English and French at 320 and 1280 pixels.


Garneauma (changes requested):
@Pablo-seixas Please link the exact WCAG success criterion the multimedia player volume slider fails. As far as I know, the default <video> HTML tag does not even have a visible volume label. Are you saying the default tag fails WCAG?

Furthermore, your solution breaks the layout of the multimedia component entirely. This cannot be accepted as is.

Pablo-seixas:
Thank you for the review. You are correct: axe's `label-title-only` rule is categorized as a Deque best practice, not as a failure of a specific WCAG success criterion. The existing `title` provides an accessible name, and W3C technique H65 recognizes this approach when a visible label cannot be used. I should not have described the axe finding as a WCAG failure.

I have reverted the visible-label change because it altered the multimedia player's layout. Since the proposed change is no longer justified and the branch now restores the original component, I am closing this PR.

